Issue #16, opened 2026-04-02
[MEDIUM] Length checks are byte-based not char-based


alphaonedev:
All validate.rs length checks use .len() (bytes) not .chars().count(). Unicode titles allow fewer characters than intended.

**Status:** Accepted — byte-based limits are intentional for storage constraints. Documented behavior.

Issue #15, opened 2026-04-02
[MEDIUM] Silent ROLLBACK failure in touch() and consolidate()


alphaonedev:
ROLLBACK errors silently swallowed with `let _`. Transaction state corruption hidden.

**Fix:** Added tracing::error! logging for ROLLBACK failures in both functions.

**Commit:** ca799c0

PR #409, opened 2026-04-26
fix(cli): AI_MEMORY_AGENT_ID env no longer leaks into list/search filter

alphaonedev:
## Summary

The top-level `--agent-id` flag was declared with both `global = true` and `env = \"AI_MEMORY_AGENT_ID\"`. clap's `global = true` propagated the env-derived value into every subcommand's `--agent-id` field — including the per-command `agent_id` **filter** fields on `list` / `search`. As a result, exporting `AI_MEMORY_AGENT_ID` (e.g. a campaign harness, a CI runner, or a personal shell profile) silently scoped read results to the caller's identity and hid every memory whose `metadata.agent_id` did not match.

```bash
$ env AI_MEMORY_AGENT_ID=charlie ai-memory --json list
{\"count\":0,\"memories\":[]}                   # was: silently filtered
{\"count\":2,\"memories\":[{...alice...},{...bob...}]}  # now: returns all rows
```

## Charter motivation

Phase 1 in-scope bug fix per the v0.6.3 grand-slam charter §\"In scope for this campaign\" (\"Bug fixes that surface during the above\"). Surfaced repeatedly in the campaign-runner harness over iters #15 / #16, where ~10 integration tests under `test_agentid_*` returned 0 rows on plain `list` because the harness exports `AI_MEMORY_AGENT_ID=campaign-runner:ai-memory-v063` for write-path attribution.

## Fix

- **`src/main.rs`** — drop `env = \"AI_MEMORY_AGENT_ID\"` from the top-level `--agent-id` flag. `global = true` stays so `--agent-id X` can still appear anywhere in the command line.
- **`src/identity.rs`** — refresh comments around `ENV_AGENT_ID` and `resolve_agent_id` step 2. The env-var fallback for **write** paths (CLI store/delete/update/sync, MCP tool calls) is unchanged — `resolve_agent_id` already reads `AI_MEMORY_AGENT_ID` at step 2 of the documented precedence chain. Removing it from the clap flag just stops clap from auto-merging the env into the per-subcommand **filter** fields.

## Test plan

- [x] `cargo fmt --check`
- [x] `cargo clippy -- -D warnings -D clippy::all -D clippy::pedantic`
- [x] `AI_MEMORY_NO_CONFIG=1 cargo test` — 186 passed (was 185 passing + 1 failing in the campaign env pre-fix)
- [x] New regression test `test_agentid_env_does_not_leak_to_list_filter` covers both `list` and `search`, plus a sanity check that an explicit `--agent-id` flag still filters when the env is set. Verified to fail on un-fixed code (`assertion left == right failed: env should not filter list results; got: {\"count\":0,\"memories\":[]}`) and pass after.
- [x] `test_mcp_store_invalid_metadata_defaults_to_empty` now `.env_remove(\"AI_MEMORY_AGENT_ID\")`s so its NHI-default assertion holds under a runner-exported env override (pre-existing test rigidity, in the same neighborhood).
- [ ] CI green

## Backwards compatibility

- **Writes**: identical behavior. `AI_MEMORY_AGENT_ID=alice ai-memory store …` still tags the memory with `agent_id=alice` because `identity::resolve_agent_id` reads the env at step 2. CLI `--help` still documents the env var.
- **Reads**: `AI_MEMORY_AGENT_ID=alice ai-memory list` no longer silently filters. To filter, pass `--agent-id alice` explicitly. This restores the documented precedence chain (\"agent_id is per-write provenance; reads see all rows by default\").

## AI involvement

- **Author**: Claude Opus 4.7 (1M context), authority class **Standard** per `docs/AI_DEVELOPER_GOVERNANCE.md`. Two-line clap edit + comment refresh + one regression test on a release-track branch under operator supervision.
- **Reviewer**: Operator merge required (release-track guard).
- Generated as iter 17 of the `ai-memory-v063` campaign run.

### DIFF
--- a/src/identity.rs
+++ b/src/identity.rs
@@ -37,8 +37,13 @@ use anyhow::Result;
 
 use crate::validate;
 
-/// Environment variable override for `agent_id` (used by CLI via clap's
-/// `env = "AI_MEMORY_AGENT_ID"`; read directly for MCP fallback).
+/// Environment variable override for `agent_id`. Read directly here as
+/// step 2 of the precedence chain for *write* paths (CLI store/delete/
+/// update/sync, MCP tool calls). Intentionally NOT bound to clap's `env`
+/// attribute on the `--agent-id` flag because `global = true` would
+/// otherwise propagate the env value into the per-subcommand `--agent-id`
+/// *filter* fields on `list` / `search`, silently hiding memories from
+/// the caller.
 const ENV_AGENT_ID: &str = "AI_MEMORY_AGENT_ID";
 
 /// Environment variable opt-out for the hostname-revealing default (#198).
@@ -114,7 +119,7 @@ fn sanitize_component(input: &str) -> String {
 /// See module docs for precedence. Returned id is always valid per
 /// [`validate::validate_agent_id`].
 pub fn resolve_agent_id(explicit: Option<&str>, mcp_client: Option<&str>) -> Result<String> {
-    // 1. Explicit caller value (already env-merged by clap for CLI)
+    // 1. Explicit caller value (CLI `--agent-id` flag, MCP tool param)
     if let Some(id) = explicit
         && !id.is_empty()
     {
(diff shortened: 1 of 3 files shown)


alphaonedev:
Closing: the v0.6.3 release line is superseded by the shipped v0.6.4 release and the active release/v0.7.0 development line. This v0.6.3-campaign PR is abandoned per operator direction (current release = v0.6.4). Reopenable if any individual change still has standalone value against a live branch.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

PR #592, opened 2026-05-05
feat: v0.7-b3 — pre-computed family-descriptor embeddings

alphaonedev:
## Summary
- Adds an `Arc<Vec<(Family, Vec<f32>)>>` cache to `AppState`, populated at HTTP boot from eight 1-2 sentence descriptors (one per `Family` variant) so B2's `memory_smart_load(intent)` can match an intent string against pre-computed embeddings with one embed call + eight cosine comparisons.
- Fault-tolerant: when the embedder is `None` (keyword tier or model-load failure) the cache is empty and `AppState::best_family_match` returns `None`, letting B2 degrade gracefully.
- No new MCP tools, no schema or `HookEvent` changes — pure backend addition. Tool count stays at 45.

## Test plan
- [x] `cargo build --tests` clean
- [x] `cargo test --lib` — 2164 passed, 0 failed
- [x] `cargo test --lib b3_` — 4 new tests pass (descriptor coverage, no-embedder→empty, local-embedder→8, empty-cache→None)
- [x] Local-embedder test actually loaded the model and produced 8 vectors at the embedder's native dim
- [x] `cargo clippy --bin ai-memory -- -D warnings` clean (the authoritative gate per `src/lib.rs:11`); pre-existing `--all-targets` errors in `tests/proptest_embeddings.rs`, `tests/integration.rs:11667`, and `src/llm.rs::build_client` are not from this change


alphaonedev:
## CI fix pushed (Option A)

The synchronous `AppState::precompute_family_embeddings(embedder.as_ref())` call inside `bootstrap_serve` was blocking HTTP `/health` past the integration suite's 5 s `wait_for_health` budget on CI runners without a pre-warmed `hf-hub` model cache. The embedder loaded fine; it was the **first `embed()` call** that triggered model download and hung. ~30 integration tests including `http_capabilities_returns_json_with_version` were failing at `tests/integration.rs:8924` on Ubuntu, macOS, and Windows.

### Fix

Moved the precompute to a detached `tokio::spawn` task launched from `bootstrap_serve`:

- `AppState::family_embeddings` type changes from `Arc<Vec<(Family, Vec<f32>)>>` to `Arc<tokio::sync::RwLock<Option<Vec<(Family, Vec<f32>)>>>>`. `None` = "not yet populated"; `Some(empty)` = "embedder unavailable, never will populate".
- The spawned task wraps the (sync) embedding work in `tokio::task::spawn_blocking` and writes the result to the cell when ready.
- The `JoinHandle` parks on `task_handles` so daemon shutdown aborts the precompute if it's still in flight.
- `best_family_match` uses `try_read()` and returns `None` until the cache populates - matches B2's existing degrade-to-non-embedding-path contract.

### Regression test

`tests/b3_precompute_doesnt_block_serve.rs::b3_precompute_does_not_block_serve_health` boots the binary the same way `DaemonGuard::spawn` does and asserts `/api/v1/health` responds within **2 s** (well under the 5 s integration budget). If the precompute ever creeps back onto the startup path, this test fails first.

### Verified locally

- `cargo fmt` clean
- `cargo clippy --lib -- -D warnings -D clippy::all -D clippy::pedantic` clean
- `cargo test --lib` - 2171 passed (the `task_handles.len()` assertion bumped 4 to 5 for the new precompute task)
- `cargo test --test integration http_capabilities_returns_json_with_version` - passes
- new regression test passes in 1.4 s

### Cascade safety

Pure boot-path fix. No changes to tool count, schema, webhook events, or `HookEvent` variants.

alphaonedev:
Root cause: under heavy parallel-test load, the B3 detached precompute spawns 8 back-to-back `embed()` calls that serialise on `Embedder::Local`'s `std::sync::Mutex<BertModel>` against every request-path embed (notify content, `sync_push` peer-side embedding refresh, single-row create_memory), pushing federation-quorum acks past their 5 s window — manifesting as 503 `quorum_not_met` and reqwest `TimedOut` on multi-peer fanout tests. The prior B3-fix's `spawn_blocking` detach solved the startup-path block but not the steady-state mutex contention. Fix: gate the precompute behind `AI_MEMORY_PRECOMPUTE_FAMILY_EMBEDDINGS=1` (default OFF) since `best_family_match` is dead code in production until B2 lands. Verified by reproducing 3/3 FAIL on this branch under `--test-threads=16`, then 5/5 PASS after the gate (full integration suite at `--test-threads=8` also green: 211 passed, 21 s). Pushed as b5f92f2.

PR #602, opened 2026-05-06
feat: v0.7-g10 — pre_recall daemon-mode hook

alphaonedev:
## Summary

- Adds **`HookEvent::PreRecallExpand`** (the 21st lifecycle variant) for synchronous query-expansion hooks on the recall hot path. Wire-name `pre_recall_expand`, payload `RecallExpandQuery { query, namespace, k }`.
- Introduces **`EventClass::HotPath`** with a **50ms** class deadline (v0.6.3 recall p95 budget). G6's chain runner enforces the ceiling on the whole chain; operators MUST configure these hooks in `mode = "daemon"` — fork+exec alone exceeds the budget.
- Wires the fire site into a new `handle_recall_with_pre_recall_hook` wrapper around `handle_recall` (src/mcp.rs). The wrapper applies `Modify`-side rewrites to the in-flight `(query, namespace, k)` triple before delegating, and short-circuits on `Deny` with a `meta.diagnostic.pre_recall_denied` block.
- Helper `crate::hooks::recall::apply_pre_recall_expand` marshals the payload through `HookChain::fire` and projects the chain result back to a `PreRecallOutcome { Allow | Modified | Denied }`.

`Modify` overloads three `MemoryDelta` fields: `content` -> `query`, `namespace` -> `namespace`, `priority` -> `k` (cast `i32 -> u32`, non-positive falls back). The overload is documented inline rather than forking `MemoryDelta` per-event.

## Cascade safety

- HookEvent variant count: **20 -> 21** (verified via the existing `hook_event_all_variants_round_trip` test before edit).
- Tool count, MCP schema, webhook events: unchanged.
- `event_class` mapping covers the new variant; the compiler's exhaustiveness check pins the table.
- `is_pre_event` updated.

## Test plan

- [x] `cargo test --lib` -> `2179 passed; 0 failed` (was 2173; +6 new lib tests across recall.rs / events.rs / decision.rs / timeouts.rs)
- [x] `cargo test --test hooks_pre_recall` -> `6 passed; 0 failed`
- [x] `cargo test --test hooks_executor_test --test hooks_timeout_budget --test hooks_hot_reload` -> all green (no regression on existing G3/G6 surface)
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --lib --test hooks_pre_recall -- -D warnings` clean
- Integration test (`tests/hooks_pre_recall.rs`) covers: type/class mapping, in-process Allow / Modify / Deny / empty-chain via mock executor, and a real daemon-mode subprocess round-trip pinning the NDJSON wire shape.

## AI involvement

Implemented end-to-end by Claude Opus 4.7 (1M context) in auto mode. Author retains review responsibility per AI_DEVELOPER_GOVERNANCE.

### DIFF
--- a/src/hooks/decision.rs
+++ b/src/hooks/decision.rs
@@ -379,6 +379,9 @@ pub fn is_pre_event(event: HookEvent) -> bool {
             | HookEvent::PreGovernanceDecision
             | HookEvent::PreArchive
             | HookEvent::PreTranscriptStore
+            // G10: hot-path query expansion fires before the recall
+            // call — Modify decisions rewrite the in-flight query.
+            | HookEvent::PreRecallExpand
     )
 }
 
@@ -652,7 +655,7 @@ mod tests {
 
     #[test]
     fn is_pre_event_classifies_all_variants() {
-        // Pre- variants
+        // Pre- variants (G10 added PreRecallExpand)
         for ev in [
             HookEvent::PreStore,
             HookEvent::PreRecall,
@@ -664,6 +667,7 @@ mod tests {
             HookEvent::PreGovernanceDecision,
             HookEvent::PreArchive,
             HookEvent::PreTranscriptStore,
+            HookEvent::PreRecallExpand,
         ] {
             assert!(is_pre_event(ev), "expected {ev:?} to be a pre- event");
         }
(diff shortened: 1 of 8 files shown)


alphaonedev:
Rebased onto main. K7 (#603) is still open, so the harness_integration CORE_DESCRIBE_OPENING fix (5 -> 6 memory tools, post-B1) was applied directly here as a follow-up commit.

Issue #2024, opened 2026-07-14
[v1.x][skills] Operator-authorized lifecycle management for registered skills


catan271:
## Problem

Registered Agent Skills cannot currently be removed or retired. The skill surfaces expose `register`, `list`, `get`, `resource`, `export`, `promote`, and `compose`, but no lifecycle operation for a registered skill.

This leaves obsolete compatibility skills discoverable indefinitely. For example, the registered `ai-memory` and `ai-memory-skills` compatibility skills have been superseded by `ai-memory-base`, `ai-memory-operations`, and `ai-memory-skill-registry`, but there is no supported way to remove them from discovery. The generic `memory_delete` endpoint only accepts memory IDs and rejects skill IDs.

## Proposed solution

Add an operator-authorized skill lifecycle surface that can retire or delete a registered skill by ID, with parity across MCP, CLI, and HTTP where those surfaces support skills.

Suggested initial behavior:

- Target a specific skill ID/version, not only a name.
- Default to a reversible `retired` state that is omitted from normal `skill list` discovery.
- Preserve provenance/audit history and make retirement explicit in `skill get`.
- Provide a separate, deliberately explicit hard-delete/purge operation only if it is compatible with the skill-version and audit model.
- Require the same authorization/governance checks used for destructive lifecycle operations.
- Add tests for discovery filtering, direct retrieval behavior, authorization failures, and CLI/MCP/HTTP parity.

## Alternatives considered

- Direct SQLite/database edits: unsupported and risks breaking registry invariants or audit history.
- Re-registering a replacement: leaves the deprecated skill discoverable.
- Restricting the MCP startup profile: changes which tools are exposed, not which skills the registry advertises.

## User impact

This blocks operators from removing deprecated or mistakenly registered skills without manually mutating storage. It also means every agent still sees compatibility skills that should no longer be selected.

alphaonedev:
Thanks for the detailed, well-scoped report @catan271 — this is a **legitimate gap** and we're implementing it for v1.0.0. We put it through a 2×5 adversarial design review (two waves of five reviewers against the actual code); the verdict and plan follow, plus two small factual corrections for the record.

### Confirmed
[...]

PR #2099, opened 2026-07-16
infra(#1803): IronClaw lan-parity compose peer-key auto-provisioning

### LINKED ISSUE #1803: [v1.x] Test-infra: IronClaw lan-parity compose peer-key auto-provisioning
## Context
The 2026-06-24 live IronClaw A2A campaign (docs/v0.8.0/test-campaign-2026-06-24-ironclaw-a2a/track-b-a2a-results.md) brought up the `infra/lan-parity-test` stack (alice↔bob over shared pg-age, HEAD 3e57ec3c, x-ai/grok-4.3 brain) and validated the v0.8.0 secure-default deployment posture with zero defects.

## Finding (test-infra, NOT a product defect)
Post-#1789 (peer-enrollment fail-closed = v0.8 secure default), a bare `docker compose -f infra/lan-parity-test/docker-compose.yml up -d` does **not** pre-provision peer keys, so the **enrolled-federation happy path cannot be smoke-tested live** out of the box:
- alice's `/sync/push` to bob → `401 peer_not_enrolled` (correct secure-default).
- Manually enrolling `host:ic-parity-alice` (key-dir `.pub`) was confirmed effective at the gate (an unsigned probe then returns `x_memory_sig_missing`), but the daemon's **own** signed push still 401'd — its presented broadcast `sender_agent_id` + the FED-P3a CA-credential path differ from a manual `.pub` enrollment, so manual enrollment alone doesn't unblock the live mesh.

## Proposed fix (v0.9, ~test-infra only)
Add a peer-key provisioning step to the IronClaw compose (an init container or entrypoint hook) that cross-enrolls alice↔bob via the real operator workflow (`ai-memory identity import` / credential issuance) keyed by each daemon's **actual** resolved federation `sender_agent_id`, so a bare `compose up` yields a working enrolled mesh for live A2A smoke (federation roundtrip + #1464 content attestation + #1709 coordination cross-daemon). Document the resolved sender_agent_id (fed-identity) ↔ enrollment-key relationship.

## Scope
v0.9 test-infra. NOT a v0.8.0 ship blocker — the enrolled happy-path data flows are covered by the 8-green CI suite (a2a_campaign_round1, federation_identity_e2e, federation_signing, federation_sync_state_merge_1709, #1464 unit + live-PG parity). The live campaign validated the deployment SECURITY POSTURE (all gates/secure-defaults firing correctly).

_Filed by AI NHI from the live IronClaw A2A campaign (operator-authorized, mem b4ce5017)._

alphaonedev:
## Summary

Closes #1803. A bare `docker compose -f infra/lan-parity-test/docker-compose.yml up -d` now yields a **working enrolled 2-daemon mesh** with no manual peer-key provisioning step — closing the gap the 2026-06-24 live IronClaw A2A campaign hit, where alice's signed `/sync/push` to bob `401`'d even after a manual `.pub` copy.

## Root cause (source-pinned; test-infra, not a product defect)

Each daemon's OUTBOUND federation `/sync/push` signing key is loaded by `governance::audit::load_daemon_signing_key(&sender_agent_id)` (`src/federation/peer.rs`), keyed by the **resolved federation identity** (`AI_MEMORY_FED_IDENTITY` env > `host:<hostname>`; `src/federation/identity/resolver.rs`) — a *different* on-disk file from the fixed `DAEMON_KEYPAIR_LABEL = "daemon"` link/audit keypair the entrypoint already mints. `ai-memory serve` has **no auto-generate fallback** for that identity-labeled key, so federation pushes went out UNSIGNED and any peer enforcing the v1.0.0 secure defaults (`AI_MEMORY_FED_REQUIRE_WRITE_SIG` / `_SIGNAL_SIG`) rejected them. A manually-copied `.pub` had nothing to verify against because the authoring side never held the matching signing key under that name — exactly the "manual `.pub` enrollment doesn't unblock the live mesh" symptom in the issue.

## Changes

- **`entrypoint.plan-c.sh`** — on first boot each daemon also mints a keypair under its resolved federation identity in `ai-memory serve`'s own key directory (`identity::keypair::default_key_dir`), idempotent across restarts.
- **`infra/lan-parity-test/docker-compose.yml`** — pin `AI_MEMORY_FED_IDENTITY` explicitly on alice/bob, plus a new one-shot `ic-parity-peer-key-provisioner` service (gated `depends_on: service_healthy`) that runs `provision-peer-keys.sh` to cross-copy each side's real `<fed_identity>.pub` into the other's key volume under the exact name `identity::verify::lookup_peer_public_key` (and the `peer_not_enrolled` gate) reads. Only `.pub` material ever crosses sides.
- **`infra/lan-parity-test/provision-peer-keys.sh`** — new one-shot cross-enrollment script with the full root-cause writeup inline.
- **`tests/lan_parity_provision_peer_keys.sh`** — standalone assertion harness (happy-path cross-enrollment, timeout fail-closed, missing-env guard), mirroring the `tests/plan_c_preflight.sh` precedent.
- **`CHANGELOG.md`** — `### Added` entry.

## Verification

No Rust source touched (infra/config + shell only). Verified end-to-end against the freshly-built release binary (docker is unavailable to the autonomous process, so the mesh itself can't boot here — but the load-bearing logic was exercised directly):

- `ai-memory identity generate --agent-id host:ic-parity-alice` mints `host:ic-parity-alice.{pub,priv}` exactly as the entrypoint expects.
- Running `provision-peer-keys.sh` against those real keys makes `ai-memory identity list` show each peer's key enrolled on the *other* side — the exact lookup the `peer_not_enrolled` gate performs — with byte-matching `.pub` copies and **no private key leaked**.
- `tests/lan_parity_provision_peer_keys.sh` passes all three assertions.
- `docker compose config` parses the new service cleanly; both shell scripts pass `bash -n` / `sh -n`; `cargo fmt --check` clean.

## AI involvement

Implemented autonomously by Claude (Opus 4.8) under the v1.0.0 campaign. Test-infra scope, no core `src/` change.


alphaonedev:
### CI note — `token-budget gates` failure is pre-existing on base, unrelated to this PR

The one red check (`token-budget gates`) is **not** caused by this change:

- This PR touches **zero Rust and zero MCP tool schemas** — `git diff --stat origin/release/v1.0.0...HEAD -- src/` is empty. It is infra/compose + shell + CHANGELOG only.
- The failure is `full_profile_tools_list_under_post_859_ceiling`: the MCP `tools/list` wire payload is **6563 cl100k_base tokens vs the 6500 ceiling** — a function of the advertised MCP tool schemas, which this PR does not alter.
- The identical check is **already failing on the base `release/v1.0.0` branch** (`gh api repos/.../commits/release/v1.0.0/check-runs` → `token-budget gates: failure`). It was pushed over the ceiling by the merged #2024 skill retire/delete tools (advertised tool count 101 → 103). That is a separate, pre-existing item to resolve on the base branch (trim `wire_compact_descriptions` per the panic hint), not something an infra PR can or should fix.

The relevant always-on gate for this change, `tool-count grep gate`, **passes**. The heavier Rust build/test/clippy workflows are path-filtered and correctly did not trigger for an infra/shell-only diff.

Local verification (docker is unavailable to the autonomous process, so the load-bearing logic was exercised directly against the freshly-built release binary):
- `ai-memory identity generate --agent-id host:ic-parity-alice` mints `host:ic-parity-alice.{pub,priv}` as the entrypoint expects.
- `provision-peer-keys.sh` cross-copies so `ai-memory identity list` shows each peer enrolled on the other side (the exact `peer_not_enrolled`-gate lookup), byte-matching `.pub`, no private key leaked.
- `tests/lan_parity_provision_peer_keys.sh` passes all three assertions; `docker compose config` parses; `bash -n`/`sh -n` + `cargo fmt --check` clean.

Issue #2127, opened 2026-07-16
[#2086 Fable audit] #2115 residual: cli/rules.rs + recover/transcript_paths.rs HOME-mutating tests still on module-local locks, race the shared test_env_lock cohort


alphaonedev:
Fable pre-merge audit of PR #2098 (final gate over the Opus #2114/#2115 fix round).

## Finding (LOW-MEDIUM — test hygiene; residual of #2115)

The #2115 fix correctly introduces the crate-canonical process-global env lock `crate::config::test_env_lock()` and moves the `$HOME`-mutating tests in `src/embeddings.rs`, `src/reranker.rs` (and `src/cli/commands/config.rs`, `src/config.rs` were already on it) onto it. The specific embeddings-vs-reranker race #2115 flagged is fixed.

However, two OTHER `$HOME`-mutating in-process test modules remain on **module-local locks** and therefore still race the `test_env_lock` cohort (including PR #2098's new reranker tests) under parallel `cargo test`:

- `src/cli/rules.rs` (~line 2234): `static HOME_ENV_LOCK: Mutex<()>` — module-local; `set_var("HOME", ...)` at ~2268/2316.
- `src/recover/transcript_paths.rs` (~line 210): `static LOCK: OnceLock<Mutex<()>>` — module-local; `set_var("HOME", ...)` at ~226/361.

A `transcript_paths` or `rules` test holding only its own lock can swap `$HOME` out from under a reranker/embeddings test that holds `test_env_lock` mid-assertion — the exact cross-module flake mechanism #2115 described ("Share ONE process-global serialization mutex across every $HOME-mutating test in the crate"), just with different counterparties.

## Fix

Route both modules' env guards through `crate::config::test_env_lock()` (one-line change per module, same pattern as `src/cli/commands/config.rs::env_lock()`), and grep-gate future `set_var("HOME"` additions to require the shared lock.

## Severity

LOW-MEDIUM — no production impact; intermittent-CI-flake vector for the #2019 hermetic coverage runs this PR chain is meant to unblock.

_Filed by the Fable pre-merge audit of PR #2098; follow-up to #2115 (which is resolved for the pair it named). Non-blocking for #2098 itself — the residual lives in files the PR does not touch._

alphaonedev:
[...]

PR #2142, opened 2026-07-16
test(#2127): migrate cli/rules.rs + recover/transcript_paths.rs HOME tests to shared test_env_lock

alphaonedev:
## Summary

- Fixes #2127 — the `#2115` residual flagged by the Fable pre-merge audit of #2098: `src/cli/rules.rs` and `src/recover/transcript_paths.rs` still serialized their `$HOME`-mutating tests on **module-local** locks (`static HOME_ENV_LOCK` / a private `OnceLock<Mutex<()>>`), which race the **shared** `crate::config::test_env_lock()` cohort (`src/embeddings.rs`, `src/reranker.rs`, `src/config.rs`, `src/cli/commands/config.rs`) under parallel `cargo test` — the exact cross-module `$HOME`-swap flake mechanism #2115 fixed for the embeddings/reranker pair, just with different counterparties.
- Both modules now delegate to the crate-canonical `crate::config::test_env_lock()`, mirroring the established `src/cli/commands/config.rs::tests::env_lock()` thin-wrapper pattern (`fn env_lock() -> ... { crate::config::test_env_lock() }`).
- Test-only mechanical change. No production code touched; no new behavior.

## Applied rust-skills / rust-microsoft rules

- `test-fixture-raii` (rust-skills) — the returned `MutexGuard` is bound to a `let _guard = ...` local for the duration of the critical section, never dropped early.
- `err-no-unwrap-prod` (rust-skills) — `test_env_lock()` recovers from mutex poisoning (`unwrap_or_else(PoisonError::into_inner)`) so one panicking test doesn't cascade-poison the shared lock for the rest of the suite; preserved by delegating rather than reimplementing.
- `unsafe-safety-comment` (rust-skills) — the `unsafe { std::env::set_var(...) }` blocks in `cli/rules.rs` keep (and correct) their `// SAFETY:` comments to point at the new shared guard.
- `M-AVOID-STATICS` (rust-microsoft) — removes the module-local `static HOME_ENV_LOCK: Mutex<()>` and the module-local `static LOCK: OnceLock<Mutex<()>>`, consolidating on the one process-global static already owned by `crate::config`.
- `M-DESIGN-FOR-AI` (rust-microsoft) — doc comments at both call sites now name the shared-lock rationale and the sibling modules that share it, so a future migration follows the same trail #2115 → #2127 did.

## Tests migrated

- `src/cli/rules.rs::tests::run_rules_sign_seed_neither_layout_falls_through_to_legacy_path_and_errors`
[...]

Issue #2115, opened 2026-07-16
[#2086 audit] Reranker offline tests mutate process-wide $HOME and can race the embedder module's HOME-mutating tests under parallel cargo test


alphaonedev:
Independent higher-tier audit of PR #2098.

## Finding (MEDIUM — test hygiene / potential flake)

The three new tests in `src/reranker.rs` (`cross_encoder_fallback_bails_loud_when_uncached_2086`, `cross_encoder_fallback_succeeds_when_prestaged_2086`, `cross_encoder_resolve_files_routes_through_fallback_when_offline_2086`) call `std::env::set_var("HOME", ...)` (and `AI_MEMORY_EMBED_OFFLINE`) guarded by a **module-local** `static LOCK: Mutex<()>`. The test comments themselves admit the lock *"only covers this module's own tests."*

`src/embeddings.rs` ALSO has HOME-mutating tests (the `load_from_fallback` family). `cargo test` runs tests from different modules concurrently on multiple worker threads, and the two modules use **different** `static LOCK`s — so a reranker test can have `$HOME` swapped out from under it by an embeddings test mid-assertion. Because process env is global, `cross_encoder_fallback_succeeds_when_prestaged_2086` can intermittently observe a foreign `$HOME` and fail (or, worse, a bail test spuriously pass for the wrong reason).

Additionally, `std::env::set_var` is documented-unsound in a multithreaded process (it is `unsafe` as of edition 2024 for exactly this reason).

## Fix
- Route the fallback resolvers through an explicit base-dir parameter (or a `#[cfg(test)]` override seam) so tests pass a temp dir instead of mutating `$HOME`; OR
- Share ONE process-global serialization mutex across every `$HOME`-mutating test in the crate (embeddings + reranker), and/or gate them behind `--test-threads=1` via a serial-test guard.

## Severity
MEDIUM — no production impact, but a real cross-module data race that can flake CI (and the #2019 coverage-CI hermeticity this PR is meant to unblock).

_Filed by Opus 4.8 audit of Sonnet-authored PR #2098._

alphaonedev:
[...]

Issue #1998, opened 2026-07-13
[ci][flaky] config::tests intermittently fail under multi-threaded runner (AI_MEMORY_* env-var race)


alphaonedev:
**Symptom.** `cargo test --lib --features sal config::tests` (default multi-threaded runner) intermittently reports a failure; the SAME command with `-- --test-threads=1` passes 214/0 deterministically (verified twice). Surfaced during the #1953 merged-tree battery.

**Root cause.** Multiple `config::tests` set/remove process-global `AI_MEMORY_*` env vars (each with a `// SAFETY: single test owns this var` comment) but rely on the runner NOT interleaving them. Under the default multi-threaded test runner two such tests race on the shared env, so an assertion reads a var another test set/cleared. It is a test-isolation defect, not a production bug.

**Impact.** CI's coverage job serializes (`-- --test-threads=1`) so it is unaffected there, but the per-OS `Check` jobs (impact-aware `cargo test`) can run multi-threaded and could flake. Pre-existing (not introduced by any specific PR).

**Proposed fix.** Serialize the env-touching config tests with a shared `Mutex` guard (the `key_dir_env_lock()` / `store_url_env_lock()` precedent already used elsewhere in the crate) so they cannot interleave, OR gate them behind a `#[serial]` attribute. ~small.

Auto-filed-by-agent during #1953 integration. Verified per prime directive pm-v3 (memory cd8ede94): reproduced-then-serialized-passes (twice), confirmed pre-existing.

PR #2149, opened 2026-07-16
test-infra(#2146): add grep-gate for $HOME set_var without the shared test_env_lock guard

### LINKED ISSUE #2146: [#2127 Fable audit] grep-gate half of #2127's fix not delivered by PR #2142 — future set_var("HOME") test additions can silently regress off the shared test_env_lock cohort
Fable pre-merge audit of PR #2142 (fix/2127-test-env-lock).

## Finding (LOW — test-infra guard gap; the migration itself is verified correct)

Issue #2127's Fix section specified TWO deliverables:

> Route both modules' env guards through `crate::config::test_env_lock()` (one-line change per module, same pattern as `src/cli/commands/config.rs::env_lock()`), **and grep-gate future `set_var("HOME"` additions to require the shared lock.**

PR #2142 delivers the first half completely and correctly (audited: every `set_var("HOME"`/`remove_var("HOME")` site in the lib crate — `src/embeddings.rs`, `src/reranker.rs`, `src/config.rs`, `src/cli/commands/config.rs`, `src/cli/rules.rs`, `src/recover/transcript_paths.rs` — now serializes on the ONE canonical `crate::config::test_env_lock()`; poison handling is the single canonical `unwrap_or_else(PoisonError::into_inner)`; guard/HomeGuard drop order is correct; no new lock-ordering edges — the migrated tests hold no second lock, so no deadlock cycle is possible; `tests/form_7_agent_external_wiring.rs` is a separate test BINARY/process and documents its own `--test-threads=1` serialization, so it is out of the in-process cohort).

The second half — the mechanical gate — is absent from the PR and unmentioned. Since the PR carries `Closes #2127`, the gate deliverable silently drops. This class has now recurred across three audit generations (#1998 -> #2115 -> #2127), each time because a new module-local lock was hand-rolled instead of the shared one; the grep gate is what actually closes the CLASS rather than the current instances.

## Proposed fix (~30-line script or an extension of an existing gate)

A `scripts/check-test-env-lock.sh` (or an arm in an existing script gate) that:
- flags any `set_var("HOME"` / `remove_var("HOME"` under `src/` whose enclosing test module does not reference `crate::config::test_env_lock` (file-scope grep pairing is sufficient at this codebase's scale: every current file with a HOME mutation also contains the `test_env_lock` token);
- flags any NEW `static ... Mutex<()>`-style module-local lock adjacent to a `set_var("HOME"` site;
- `--self-test` plants a violating fixture and confirms non-zero exit (pm-v3.2 load-bearing-gate discipline).

## Severity

LOW — no live race remains at PR #2142 HEAD; this is regression-guard coverage for the recurrent class.

_Filed by the Fable pre-merge audit of PR #2142. Non-blocking for #2142 itself._

alphaonedev:
## Summary

Issue #2146 (Fable audit of PR #2142 / `fix/2127-test-env-lock`) found that #2142 delivered the migration half of issue #2127's Fix section — routing every `$HOME`-mutating test onto the crate-canonical `crate::config::test_env_lock()` — but not the mechanical gate the issue also specified: *"grep-gate future `set_var("HOME"` additions to require the shared lock."* This defect class has recurred **three times**: #1998 → #2115 → #2127, each time because a new module hand-rolled its own module-local `Mutex<()>` instead of reusing the shared lock, so two different modules' `$HOME`-mutating tests could still race cross-module even though each was individually serialized within its own file.

`scripts/check-test-env-lock.sh` HARD-BLOCKS any `src/**/*.rs` or `tests/**/*.rs` file that contains a `std::env::set_var("HOME"` / `remove_var("HOME"` call without also referencing `test_env_lock` somewhere in the same file.

## Design: file-scoped, not function-scoped (deliberate)

The issue's proposed fix suggested structural detection ("`set_var("HOME")` without a `test_env_lock()` acquisition in the enclosing test fn/scope"). A brace-balanced **per-function** scan — the naively more precise option — produces a real **false positive** against this exact codebase: `src/config.rs`'s test module defines a one-line delegate

```rust
fn env_var_lock() -> std::sync::MutexGuard<'static, ()> {
    super::test_env_lock()
}
```

and several `$HOME`-mutating tests in that file call `env_var_lock()`, never the literal `test_env_lock` token — a per-fn scan would flag those as violations even though they ARE correctly serialized (transitively, through the delegate).

**File-scope presence-pairing has no such false positive**: the delegate's own body cites `test_env_lock`, so the FILE satisfies the pairing even though an individual call site's tokens do not. This mirrors the Fable audit's own proposed fallback verbatim: *"file-scope grep pairing is sufficient at this codebase's scale: every current file with a HOME mutation also contains the test_env_lock token"* — verified true against every current site.

**Known residual gap, documented honestly (not overclaimed) in the script header:** a NEW test fn added to a file that is ALREADY compliant (contains `test_env_lock` for its other tests) could theoretically hand-roll its own unserialized `$HOME` mutation without tripping this file-scoped gate. Closing that fully needs fn-scoped tracking of every `test_env_lock`-delegating wrapper fn, which is a real but narrower follow-up. Today: every one of the six production sites is correctly gated, and any wholesale-new file that mutates `$HOME` without ever mentioning `test_env_lock` — the shape the #1998/#2115/#2127 recurrence history actually exhibited each time — is caught unconditionally.

## Allowlist

`tests/form_7_agent_external_wiring.rs` — a separately compiled test **binary** that documents its own `--test-threads=1` serialization inline and structurally cannot reach the `pub(crate)` `test_env_lock` across the test-binary boundary.

## Self-test (load-bearing)

Three fixtures: a plain omission (no lock reference at all), a hand-rolled-local-lock (the recurrent #1998/#2115/#2127 shape — defines its own `static Mutex<()>` and never references the shared guard), and a compliant control. Verifies the gate catches both violators and spares the compliant fixture:

```
Test-env-lock gate self-test: PASS (caught both contrived violations, spared the compliant fixture; exit=1)
```

## CI wiring

New job `test-env-lock-gate` in `.github/workflows/c8-precheck.yml`, alongside the existing `test-stdin-gate` job (same pattern: main check step + self-test step).

## Verification

- `bash scripts/check-test-env-lock.sh --self-test` → PASS.
- `bash scripts/check-test-env-lock.sh` on the clean `release/v1.0.0` tree → PASS (all six production `$HOME` sites — `embeddings.rs`, `reranker.rs`, `config.rs`, `cli/commands/config.rs`, `cli/rules.rs`, `recover/transcript_paths.rs` — correctly recognized compliant; `tests/form_7_agent_external_wiring.rs` correctly allowlisted).
- `bash scripts/check-vendor-literals.sh` → PASS.
- `bash scripts/check-hardcoded-literals.sh` → PASS.
- `bash scripts/check-docs-vs-ssot.sh` → PASS.
- `bash -n scripts/check-test-env-lock.sh` → syntax clean.
- `python3 -c "import yaml; yaml.safe_load(...)"` on the modified workflow → valid YAML.

## Test plan
- [x] `--self-test` catches both fixture violations, spares the compliant control
- [x] Main gate passes clean on `release/v1.0.0` HEAD, all 6 real sites + the 1 allowlisted binary
- [x] Vendor-literal / hardcoded-literal / docs-vs-ssot gates unaffected

Closes #2146.

### DIFF
--- a/.github/workflows/c8-precheck.yml
+++ b/.github/workflows/c8-precheck.yml
@@ -176,3 +176,31 @@ jobs:
         # --self-test stages a tests/ file that blocking-reads the real
         # process stdin in a #[test]; the gate must reject it.
         run: bash scripts/check-test-stdin-reads.sh --self-test
+
+  test-env-lock-gate:
+    # Issue #2146 — the grep-gate half of issue #2127's Fix section,
+    # not delivered by PR #2142 (Fable pre-merge audit of #2142).
+    # HARD-BLOCKS any src/**/*.rs or tests/**/*.rs file that mutates
+    # the process-global $HOME environment variable
+    # (`std::env::set_var("HOME"` / `std::env::remove_var("HOME"`)
+    # without also referencing the crate-canonical
+    # `crate::config::test_env_lock()` serialization guard somewhere in
+    # the same file. `std::env::set_var` is unsound under concurrent
+    # access; this defect class recurred THREE times (#1998 -> #2115
+    # -> #2127) because a new module kept hand-rolling its own
+    # module-local lock instead of reusing the shared one. This gate
+    # keeps a new module from silently re-introducing that class.
+    name: Test-env $HOME-lock gate (#2146)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run test-env-lock gate
+        run: bash scripts/check-test-env-lock.sh
+      - name: Self-test the gate (regression evidence)
+        # --self-test stages a $HOME-mutating fixture with no
+        # test_env_lock reference (the plain-omission shape) and a
+        # second fixture that hand-rolls its own local lock instead of
+        # the shared one (the #1998/#2115/#2127 recurrent shape); the
+        # gate must reject both and spare a compliant control fixture.
+        run: bash scripts/check-test-env-lock.sh --self-test
--- a/scripts/check-test-env-lock.sh
+++ b/scripts/check-test-env-lock.sh
@@ -1,0 +1,282 @@
+#!/usr/bin/env bash
+# Copyright 2026 AlphaOne LLC
+# SPDX-License-Identifier: Apache-2.0
+#
+# v1.0.0 (issue #2146 -- the grep-gate half of issue #2127's Fix section,
+# not delivered by PR #2142; filed by the Fable pre-merge audit of #2142).
+#
+# HARD-BLOCK any src/**/*.rs or tests/**/*.rs file that mutates the
+# process-global $HOME environment variable (`std::env::set_var("HOME"`
+# / `std::env::remove_var("HOME"`) without also referencing the
+# crate-canonical `crate::config::test_env_lock()` serialization guard
+# somewhere in the SAME file.
+#
+# WHY THIS GATE EXISTS. `std::env::set_var` is unsound under concurrent
+# access -- two cargo-test threads mutating $HOME at the same time is a
+# real data race, not merely a style nit. This defect class has
+# recurred THREE times: #1998 -> #2115 -> #2127, each time because a
+# new module hand-rolled its own module-local `static LOCK: Mutex<()>`
+# instead of reusing the ONE canonical `crate::config::test_env_lock()`
+# -- so two DIFFERENT modules' $HOME-mutating tests could still race
+# each other cross-module even though each was individually serialized
+# within its own file. PR #2142 (fix/2127-test-env-lock) migrated every
+# then-existing $HOME site onto the shared lock but shipped WITHOUT the
+# mechanical gate issue #2127's Fix section also specified. Issue #2146
+# (the Fable audit of #2142) is that residual; this script closes it.
+#
+# SCOPE: FILE-scoped presence-pairing, deliberately NOT function-scoped.
+# A brace-balanced per-fn scan (the naively "more precise" structural
+# option) produces a real FALSE POSITIVE against this exact codebase:
+# src/config.rs's test module defines a local one-line delegate
+#
+#     fn env_var_lock() -> std::sync::MutexGuard<'static, ()> {
+#         super::test_env_lock()
+#     }
+#
+# and several $HOME-mutating tests in that same file call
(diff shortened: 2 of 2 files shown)


alphaonedev:
**Fable pre-merge audit: FINDINGS-NON-BLOCKING — merge OK.**

Audited cold by executing the gate. Scope clean: `scripts/check-test-env-lock.sh` (new) + `.github/workflows/c8-precheck.yml` (new `test-env-lock-gate` job, main + `--self-test` steps) — no product code. The gate is WIRED and green on this PR's own CI run.

**Verified:**
- Main gate PASS on the clean tree; all 6 known-legit `$HOME`-mutating files pass, including the `transcript_paths.rs` `home_lock()` delegate (file-scope pairing recognizes it — its body cites `test_env_lock`).
- Structural design is FILE-scope presence-pairing, not a fragile enclosing-scope bash heuristic and not a count ratchet — so helper-fn / closure / far-above-lock shapes are all correctly admitted, and the `==`-vs-`<=` ratchet hazard does not arise. The config.rs `env_var_lock()` delegate false-positive the header documents is real (confirmed the delegate exists) and file-scoping is the right call.
- A NEW file with `set_var("HOME")` and zero lock reference FAILS the gate (verified with a scratch fixture, removed) — the entire #1998→#2115→#2127 recurrence class is closed.
- Allowlist is exact-path; a lookalike filename does NOT ride the `form_7_agent_external_wiring.rs` exemption (verified).
- Self-test is load-bearing: both violator probes assert CAUGHT (filename greps), the compliant probe asserts SPARED, non-zero exit required.
- Failure message is actionable (names the guard line + the established embeddings.rs/reranker.rs pattern).

**Finding (LOW, non-blocking):**
- #2153 — the pairing token-match is comment-insensitive (a file hand-rolling a local lock while merely MENTIONING `test_env_lock` in a comment passes), and the #2146-proposed module-local-lock detector arm (`static ... Mutex<()>` adjacent to a `$HOME` mutation) was not delivered. A const-indirected `set_var(HOME_KEY, ...)` also evades but is a deliberate-evasion spelling with no repo precedent; not filed.

CI: all completed checks green at audit time, incl. the new `Test-env $HOME-lock gate (#2146)` job.

_Fable pre-merge audit, v1.0.0 iteration-2._

Issue #2130, opened 2026-07-16
recover_from_transcript_store (SAL twin) ignores RecoverOpts.bypass_fast_path — latent #2126 tail-loss on the postgres path


alphaonedev:
## Finding (LOW — latent correctness inconsistency; surfaced by the #2100 / #2126 re-audit)

The #2126 fix (commit `63fc424a`, PR #2100) added `RecoverOpts.bypass_fast_path` (default `false`). The **sqlite** recovery entry point honours it:

```rust
// src/recover/mod.rs — recover_from_transcript()
if !opts.bypass_fast_path {
    if let (Some(mtime), Some(watermark_iso)) = (mtime, watermark.as_deref()) {
        ... // step-2 watermark short-circuit
    }
}
```

The **SAL/postgres twin** `recover_from_transcript_store` (`src/recover/mod.rs:536`) does **not**:

```rust
// src/recover/mod.rs — recover_from_transcript_store()  (~line 536)
if let (Some(mtime), Some(w)) = (mtime, watermark.as_deref()) {   // <-- no `if !opts.bypass_fast_path` guard
    if let Ok(wdt) = chrono::DateTime::parse_from_rfc3339(w) {
        ...
        report.fast_path_hit = true;
        return Ok(report);
    }
}
```

So a `RecoverOpts { bypass_fast_path: true }` routed through the store twin would **silently still fast-path** — the exact tail-loss #2126 fixes on the sqlite path.

## Severity: LOW (latent, not live)

Verified during the #2100 re-audit — this cannot be hit today:

- The L3 poll watcher (the only setter of `bypass_fast_path = true`, `src/recover/watcher.rs:416`) calls **only** `recover_from_transcript` (sqlite) — `src/recover/watcher.rs:423`. `WatchArgs` has **no** `--store-url` field and `Command::Watch` (`src/daemon_runtime.rs:1880`) dispatches with `db_path` only. The watcher is structurally sqlite-only.
- The **only** production caller of `recover_from_transcript_store` is `run_store` (`src/cli/commands/recover_previous_session.rs:161`), which builds opts via `build_opts` → `bypass_fast_path: false`. So no live caller ever pairs `true` with the store twin.

The gap becomes **live** the moment anyone (a) gives `ai-memory watch` a `--store-url` / postgres path, or (b) constructs a bypassing `RecoverOpts` for the SAL path. At that point #2126's tail-loss re-opens on postgres.

## Fix

[...]

Issue #2126, opened 2026-07-16
[#1978 Fable audit] watch pending_drain (#2117 fix) is cleared by the L2 watermark fast-path — un-drained limit-tail still droppable on a static transcript


alphaonedev:
Fable pre-merge audit of PR #2100 (final gate over the Opus #2116-#2119 fix round).

## Finding (MEDIUM — correctness; residual hole in the #2117 fix)

The #2117 fix adds `HostPollState.pending_drain` (`src/recover/watcher.rs`): when a tick's recovery reports `lines_skipped_limit > 0`, the next tick re-runs recovery on an otherwise-unchanged file, and clears the flag from the NEXT report:

```rust
entry.pending_drain = report.lines_skipped_limit > 0;
```

But `recover_from_transcript` (`src/recover/mod.rs`) has a **watermark fast-path** that runs BEFORE any parse: if `MAX(created_at) WHERE agent_id_idx = ?` (across ALL of the agent's memories) is >= the transcript's mtime, it returns early with `fast_path_hit = true` and `lines_skipped_limit = 0` — reporting "nothing skipped" while the un-drained tail was never even looked at.

Recovered rows are stamped with their **turn timestamps** (`created_at: turn.timestamp_iso`, mod.rs), so under pure watcher operation the watermark stays behind the file mtime and draining progresses. The hole opens when any **same-agent wall-clock write** (a normal L1 `memory_store` under the same resolved agent id — the common single-agent setup) lands between drain ticks:

1. Tick 1: backlogged transcript (N > limit new lines, e.g. the watcher's first tick against a long pre-existing transcript) → recovers `limit`, `pending_drain = true`.
2. Agent stores any L1 memory → watermark jumps to wall-clock-now >= file mtime.
3. File goes static (session ends). Tick 2: `pending_drain` correctly forces `PollDecision::Recover` → `recover_from_transcript` **fast-paths** → `lines_skipped_limit == 0` → `pending_drain = false`.
4. Every later tick: `Unchanged`. The `N - limit` tail is silently never captured — exactly the #2117 loss scenario, resurfacing through the fast-path interaction.

## Secondary facet (LOW — `--dry-run` busy re-parse)

[...]

PR #2100, opened 2026-07-16
feat(recover): L3 substrate poll-based filesystem-watcher daemon (#1978)

alphaonedev:
Closes #1978

## Summary

Issue #1978 was previously dispositioned **OPERATOR-BLOCKED** because the canonical L3 design (policy memory `f62cb182`, #1389 layered-capture architecture) named the `notify` crate (inotify/FSEvents/ReadDirectoryChangesW) as the mechanism, and adding a new external dependency is gated under the sole-authority no-external-injection rule — an AI NHI agent cannot add a dependency.

This PR resolves the blocker with a different, dependency-free design that satisfies the SAME architectural contract:

- **`src/recover/watcher.rs`** — a **std-only bounded poll loop**. Every `--interval-secs` tick (default 5s, clamped `[1, 3600]`), it diffs `std::fs::metadata` mtime/size per watched host transcript (`claude-code` / `codex` / `gemini`, tracked independently rather than the `Auto` most-recent-wins union so concurrent sessions across hosts are all captured). On a detected change it feeds the change straight into the **existing** shared L2 pipeline (`crate::recover::recover_from_transcript`) — same `transcript_line_dedup` idempotency, same per-host graceful degradation, **zero new parsing code**.
- **`src/cli/watch.rs`** — the opt-in CLI surface: `ai-memory watch --once [--json]` runs a single poll tick and reports; `ai-memory watch --daemon [--host <h>]... [--namespace <ns>] [--limit <n>] [--dry-run]` loops until SIGINT, mirroring `curator --daemon`'s `Notify`→`AtomicBool`→`spawn_blocking` bridge exactly.
- **`src/daemon_runtime.rs`** — new `Command::Watch(WatchArgs)` variant + dispatch arm + `run_watch_daemon_with_primitives` (the same bridge pattern as `run_curator_daemon_with_primitives`).
- The watcher never runs unless explicitly invoked — no implicit activation from `serve`/`mcp`.

## SSOT / doc sync

- `EXPECTED_CLI_SUBCOMMANDS_DEFAULT` 88→89, `EXPECTED_CLI_SUBCOMMANDS_SAL` 90→91 (`src/lib.rs`), mechanically re-verified by `tests/cli_subcommand_count_invariant.rs`.
[...]

Issue #2156, opened 2026-07-17
[v1.0][security][#2129 Fable audit round-3] BLOCKING: #2140 reflect body-binding is unconditional — regresses #1317 header-optional contract + breaks inert-by-default (2 CI tests red)


alphaonedev:
## Summary (BLOCKING — required CI red)

Commit `cfc157d7` (PR #2129, #2140 reflect fix) **regresses the shipped #1317 HTTP-reflect wire contract** and **violates its own "inert out of the box" guarantee**. Two required tests fail in the **Postgres feature gate** job (run 29542740360):

```
tests/issue_1317_http_reflect_wire_layer_preserves_caller_metadata.rs
  http_reflect_empty_metadata_preserves_canonical_shape ... FAILED  (:364)
  http_reflect_preserves_caller_supplied_entity_id_and_probe ... FAILED  (:288)

#1317: HTTP reflect with caller metadata must 200 OK; got 400 Bad Request
{"error":"invalid agent_id: agent_id_body_header_mismatch: body-supplied agent_id
 \"test-agent-1317-http\" disagrees with authenticated header-resolved id
 \"anonymous:req-01660e78\""}
```

## Root cause

`cfc157d7` added the #2140 body-agent_id binding to `handle_reflect_http` (`src/handlers/route_1111.rs` ~L190):

```rust
let body_agent = body.get("agent_id").and_then(Value::as_str);
let caller = match crate::handlers::parity::resolve_caller_agent_id(body_agent, &headers, None) {
    Ok(id) => id,
    Err(e) => return err_response(e),
};
let mut body = body;
if let Some(obj) = body.as_object_mut() {
    obj.insert("agent_id".to_string(), Value::String(caller));
}
```

This block runs **UNCONDITIONALLY** — there is no `enrolled_agent_keys.is_empty()` / `mode != Off` guard, unlike the `enforce_idor_identity` gate directly above it (which IS inert when no keys are enrolled). `resolve_caller_agent_id` treats the header as authoritative: with **no `X-Agent-Id` header** and a **non-empty body `agent_id`**, it synthesizes `anonymous:req-<uuid8>`, the body claim mismatches, and it returns `agent_id_body_header_mismatch` → `400`.

[...]

PR #2158, opened 2026-07-17
fix(recover/#2130): SAL/pg twin honors bypass_fast_path — cross-backend parity for the #2126 tail-loss fix

alphaonedev:
## Summary
Closes **#2130** — the SAL/postgres twin `recover_from_transcript_store` ignored `RecoverOpts.bypass_fast_path`, a latent #2126-class tail-loss on the postgres path. The sqlite path gained the `if !opts.bypass_fast_path` guard on the step-2 watermark fast-path via the merged #2126 fix (PR #2100); this ports that guard to the SAL/pg twin so both backends honor `bypass_fast_path` identically and the L3 poll watcher never fast-path-skips an un-drained `pending_drain` tail on postgres.

## Parity
Mirrors the sqlite guard verbatim at the single-source fast-path in `recover_from_transcript_store` (backend-blind). No other pg/sqlite recover divergence found in the fast-path handling.

## Verification
Live-postgres behavior is exercised by CI's **Postgres feature gate** (authoritative live-PG signal); the tail-preserved-under-bypass assertion mirrors the sqlite twin's test structure.

## Gate compliance
- rust-skills + rust-microsoft rule IDs cited in commit body.
- No hardcoded-literal duplication; no new tool/route/CLI/schema (no SSOT churn).

Fable pre-merge audit + terminal CI (Postgres + Per-Module Coverage) gate the merge.


alphaonedev:
**Fable pre-merge audit: AUDIT-CLEAN** (adversarial, @ 38c9ad7d)

**Scope**: verified the core two-sites claim, guard equivalence, the #2126 lane on the pg path, test validity via mutation, divergence sweep, compile matrix, and CI to terminal.

- **Two fast-path sites confirmed** (the key question): the sqlite twin `recover_from_transcript` carries its own inline watermark short-circuit, guarded by `if !opts.bypass_fast_path` at `src/recover/mod.rs:373` since PR #2100; the SAL store twin `recover_from_transcript_store` carried a second, textually separate copy of the same `mtime <= MAX(created_at)` short-circuit that was **unguarded** pre-fix (verified against pre-fix source). This PR guards that second site at `mod.rs:554` — the fix is real, correctly placed, and not a no-op. The commit's "single-source" claim is accurate *within* the store twin: the skip decision lives once in the backend-blind function; only `agent_max_created_at` differs per backend (`store/sqlite.rs:212` via indexed `agent_id_idx`, `store/postgres.rs:14118` via `metadata->>'agent_id'`), so the one guard covers both backends.
- **Verbatim-equivalence**: the ported guard is semantically identical to the sqlite #2126 guard — same `!opts.bypass_fast_path` gate, same RFC3339 parse, same `mtime_dt <= watermark` comparison, same effect (`fast_path_hit` + early return). Only local variable names differ (`w`/`wdt` vs `watermark_iso`/`watermark_dt`). Both twins symmetrically still run the watermark query under bypass (parity preserved).
- **pg tail-loss lane**: pre-fix, a `bypass_fast_path = true` RecoverOpts routed through the store path would still short-circuit once a same-agent wall-clock write pushed the watermark past the static file's mtime — returning `lines_skipped_limit = 0`, which a watcher-style caller uses to clear `pending_drain`, silently stranding the tail (the exact #2126 loss). Post-fix the parse runs, the limit is respected, and `lines_skipped_limit` is truthful. Latent-not-live confirmed: the only `bypass_fast_path = true` setter (watcher.rs:502) calls the sqlite twin; the store twin's only production caller (`run_store`) passes `false`.
- **Test validity (mutation-verified)**: `store_path_honors_bypass_fast_path_2130` passes at HEAD (1 passed, 0.17s). Reverting the guard to the pre-fix shape makes it FAIL on the exact `"bypass_fast_path must skip the watermark short-circuit (#2130)"` assertion — it genuinely catches the regression, and the control run proves the seed is an effective fast-path trigger (not a vacuous pass). Note (non-blocking): the test drives `SqliteStore`; that is the exact fixed backend-blind path, and the pg-specific piece (`agent_max_created_at` SQL) is unchanged by this PR and pre-existing under the CI Postgres gate.
- **Divergence sweep**: no other pg/sqlite fast-path divergence — each twin has exactly one watermark short-circuit, both now guarded; `since` derives only from `opts.since_iso` in both; no other early return keyed off the watermark.
- **Local**: `cargo check --features sal-postgres` clean; qual_10 + qual_6_7 ceilings pass; worktree clean at 38c9ad7d.
- **CI terminal-green, all checks pass** — including **Postgres feature gate** and **Per-Module Coverage Thresholds** (the #2156-catchers), Check ×3 OS, SAL-only gate, MSRV, Lint, Docker build. No failures, no cancelled dups.

Rules: M-STRONG-TYPES (fast-path contract single-sited per twin, backend-blind in the SAL lane); anti-empty-catch (no silent tail drop); test-descriptive-names / test-arrange-act-assert (control + bypass runs). No blocking or non-blocking findings — merge-ready.

PR #2147, opened 2026-07-16
test-infra(#2143,#2144,#2145): close three check-test-stdin-reads.sh residual vectors

alphaonedev:
## Summary

Fable audit of PR #2139 (fix/2120-stdin-gate) filed three residual false-pass vectors against `scripts/check-test-stdin-reads.sh` (the #1989 full-lib-suite-wedge regression guard), each with a verified repro and a concrete proposed fix. This PR implements all three:

- **#2143 (MEDIUM)** — `cfg_test_ranges` only triggered on the literal `#[cfg(test)]` attribute spelling; a composed predicate (`#[cfg(all(test, ...))]`, `#[cfg(any(test, ...))]`) placed above a file's `mod tests {}` evaded it even though it is exactly as test-reachable. The repo carries live `#[cfg(all(test, feature = "sal"))]` / `#[cfg(all(test, unix))]` / `#[cfg(any(test, debug_assertions))]` sites today (curator/mod.rs, curator/reflection_pass.rs, handlers/postgres_gate.rs, handlers/federation_signing_check.rs, hooks/executor.rs, hooks/post_reflect/auto_export.rs, hooks/timeouts.rs). Widened the awk trigger to admit any cfg predicate containing the bare `test` token per the auditor's proposed pattern, while deliberately excluding `#[cfg(not(test))]` (never test-reachable).
- **#2144 (LOW-MED)** — `cfg_test_ranges` brace-balanced with a plain `gsub` count, so a `"}"` string literal inside a `#[cfg(test)]` item closed the range early — fail-SILENT (the `helper_ranges` twin has the same lexical limitation but fails LOUD instead, by narrowing its carve-out). Fixed by stripping string literals + line comments before brace-counting.
- **#2145 (LOW)** — the paren-anchored `STDIN_PATTERN` missed two paren-less acquisition forms that both survive `cargo fmt --check`: an item-renaming import (`use std::io::stdin as input;`) and a fn-pointer binding (`let f = std::io::stdin;`). Added a second `STDIN_TOKEN_PATTERN` over the qualified `io::stdin` token (excludes an adjacent `(` so it never double-matches an ordinary call), scanned over the same test-reachable region.

## Self-test (load-bearing)

Extended `--self-test` with one fixture probe per closed vector (probes 5-8), each proven to trip the gate; the sanctioned helper line and the benign `.stdin(` builder call remain unflagged:

```
[...]

Issue #2145, opened 2026-07-16
[#2120 Fable audit] check-test-stdin-reads.sh: paren-anchored STDIN_PATTERN misses paren-less acquisitions that survive rustfmt (alias import, fn-pointer binding)


alphaonedev:
Fable pre-merge audit of PR #2139 (fix/2120-stdin-gate).

## Finding (LOW)

The broadened `STDIN_PATTERN='(^|[^A-Za-z0-9_.])stdin[[:space:]]*\('` requires the call paren adjacent to the `stdin` token. Two acquisition forms that survive `cargo fmt --check` (so CI's fmt gate does NOT normalize them away) still false-pass:

1. **Item-renaming import** — `use std::io::stdin as input;` then `input().read_line(...)` (verified: rustfmt leaves the alias intact; gate PASSES).
2. **Fn-pointer binding** — `let f = std::io::stdin;` then `f().read_line(...)` (verified: gate PASSES).

For completeness: a multiline split (`std::io::stdin` newline `()`) also evades the line-based grep, but rustfmt joins it onto one line, so the CI fmt gate kills that vector before this gate ever sees it — verified with `rustfmt --emit stdout`; NOT counted as a finding. Module aliasing (`use std::io as x; x::stdin()`) IS caught by the broadened bare-`stdin(` pattern.

## Proposed fix (~2 lines)

Both surviving forms necessarily spell the qualified path token `io::stdin` WITHOUT an adjacent `(` (that's what makes them evade). A paren-less `io::stdin` token has essentially no legitimate use in test-reachable code (the sanctioned helper calls it with parens and is carved out anyway), so add a second pattern:

```bash
STDIN_TOKEN_PATTERN='(^|[^A-Za-z0-9_.])io::stdin([^A-Za-z0-9_(]|$)'
```

scanned over the same test-reachable region with the same comment-skip, plus one self-test probe per form.

## Severity

LOW — these are deliberate-looking spellings rather than the accidental patterns the #1989 guard primarily defends against, but the gate's own contract post-#2120 is "the acquisition, however spelled, in test-reachable code", and both forms survive every other CI gate.

_Filed by the Fable pre-merge audit of PR #2139. Non-blocking for #2139 itself._

alphaonedev:
[...]

PR #2161, opened 2026-07-17
test-infra(#2151,#2152,#2153): close three gate-script residual evasion vectors

### LINKED ISSUE #2151: [#2145 Fable audit] check-test-stdin-reads.sh: grouped-alias import `use std::io::{stdin as input, ...}` evades both STDIN patterns (survives rustfmt)
Fable pre-merge audit of PR #2147 (fix/2143-2145-stdin-gate-hardening).

## Finding (LOW — residual of the #2145 class the PR closes)

PR #2147's new `STDIN_TOKEN_PATTERN='(^|[^A-Za-z0-9_.])io::stdin([^A-Za-z0-9_(]|$)'` requires the qualified token `io::stdin` verbatim. A GROUPED item-renaming import spells it `io::{stdin`:

```rust
use std::io::{stdin as input, Write};
```

so neither pattern fires — `STDIN_PATTERN` sees no adjacent `(`, and the token pattern's `io::stdin` never matches across the `{`. The subsequent `input()` call is invisible to both. **Verified evaded at PR #2147 HEAD (2026-07-17)** with fixture `tests/.audit_h_grouped_alias.rs`; `rustfmt --emit stdout` PRESERVES the grouped form (it does not split grouped imports), so the CI fmt gate does not normalize it away — unlike the multiline-attribute vector, which rustfmt joins and which is therefore NOT filed.

The single-import sibling (`use std::io::stdin as input;`) IS caught (self-test probe 7). This is the same class one grouping-brace over.

## Proposed fix (~1 line + 1 self-test probe)

Add a third pattern over the same test-reachable region catching an aliased `stdin` inside a use-group, e.g.:

```bash
STDIN_ALIAS_PATTERN='(^|[{ ,])stdin[[:space:]]+as[[:space:]]'
```

(the `stdin as` spelling is the load-bearing token — it has no legitimate non-import use), plus a probe mirroring probe 7 with the grouped form.

## Severity

LOW — deliberate-looking spelling, same severity tier as parent #2145; the gate's post-#2120 contract is "the acquisition, however spelled, in test-reachable code" and this form survives every other CI gate.

_Filed by the Fable pre-merge audit of PR #2147. Non-blocking for #2147 itself (all three filed vectors #2143/#2144/#2145 verified closed with load-bearing probes)._

### LINKED ISSUE #2152: [#2144 Fable audit] check-test-stdin-reads.sh: per-line string strip cannot see multi-line string state — a lone `}` line inside a multi-line raw string still truncates a cfg_test_ranges range fail-silent
Fable pre-merge audit of PR #2147 (fix/2143-2145-stdin-gate-hardening).

## Finding (LOW — residual of the #2144 class the PR closes)

PR #2147's `gsub(/"[^"]*"/, "", counted)` strip is PER-LINE, so it removes single-line string literals (including single-line raw strings like `r#"}"#` — verified caught) but has no state for a string spanning lines. A multi-line raw-string fixture whose body carries a net-closing-brace line still closes the brace-balanced range early, silently un-gating a stdin read below it in the same `#[cfg(test)]`-family item:

```rust
#[cfg(test)]
mod sneaky_ml {
    const J: &str = r#"
    }
    "#;
    fn reads_stdin() {
        let mut buf = String::new();
        let _ = std::io::stdin().read_line(&mut buf); // <- NOT flagged
    }
}
```

**Verified evaded at PR #2147 HEAD (2026-07-17)** with fixture `src/.audit_l_multiline_rawstring.rs` (no `mod tests` in file). This is distinct from the DOCUMENTED escaped-quote residual ("does not handle escaped quotes") — that one was accepted in #2144's own proposed fix and is not re-filed.

## Realism in this repo's style

- src/ carries 27 multi-line `r#"` fixture openers today, and **6 multi-line raw strings whose body contains a net-closing-brace line, across 4 files** (`src/federation/peer_attestation.rs`, `src/governance/mod.rs`, `src/mine.rs`, `src/synthesis/mod.rs` — JSON-ish fixtures).
- 7 live `#[cfg(all(test, ...))]` above-boundary sites exist (the class #2143 just widened the trigger for), so the preconditions co-occur in current style.

## Proposed fix (small)

Track multi-line raw-string state in the awk (a `in_raw` flag toggled on an unclosed `r#"` opener / a closing `"#`, skipping brace-counting while set) — still no full lexer, but removes the multi-line-fixture class the way the per-line strip removed the single-line class. Document the remaining escaped-quote + nested-hash residuals as the strip comment already does. One self-test probe mirroring probe 6 with a multi-line raw string.

## Severity

LOW — stacked preconditions (cfg-gated item above/without `mod tests` AND a brace-bearing multi-line string above a stdin read), but the failure direction is the same fail-SILENT widen as parent #2144.

_Filed by the Fable pre-merge audit of PR #2147. Non-blocking for #2147 itself._

### LINKED ISSUE #2153: [#2146 Fable audit] check-test-env-lock.sh: file-scope pairing token-match is comment-insensitive, and the proposed module-local-lock detector arm was not delivered
Fable pre-merge audit of PR #2149 (fix/2146-home-setvar-gate).

## Finding (LOW — guard-coverage residual; the primary #1998/#2115/#2127 class IS closed)

The gate passes a file when the literal token `test_env_lock` appears ANYWHERE in it — including a COMMENT. A file that hand-rolls a module-local lock (the exact recurrent defect class) while merely MENTIONING the canonical lock in prose passes the gate:

```rust
// NOTE: unlike test_env_lock, we use our own local mutex here.
static LOCAL: std::sync::Mutex<()> = std::sync::Mutex::new(());
#[test]
fn mutates_home_with_local_lock() {
    let _g = LOCAL.lock().unwrap();
    unsafe { std::env::set_var("HOME", "/nonexistent") }; // <- NOT flagged
}
```

**Verified evaded at PR #2149 HEAD (2026-07-17)** with fixture `src/.audit_p2_comment_token.rs` (gate exit 0). The self-test's hand-rolled probe only covers the no-token-at-all shape.

This pairs with the second bullet of #2146's proposed fix, which PR #2149 did not deliver:

> flags any NEW `static ... Mutex<()>`-style module-local lock adjacent to a `set_var("HOME"` site

That detector arm is exactly what would catch the comment-token shape (and the header-documented "already-compliant file gains a new hand-rolled lock" residual) independent of the token pairing.

## Proposed fix (~10 lines + 1 self-test probe)

Add a second arm: in any non-allowlisted file matching `HOME_PATTERN`, flag `static [A-Z_]+: *(std::sync::)?Mutex<\(\)>` (or `Mutex::new(())`) declarations, regardless of the token pairing — the canonical lock lives in `src/config.rs`, so a fresh module-local `Mutex<()>` co-resident with a $HOME mutation is the recurrent shape by definition. Optionally also restrict the pairing grep to non-comment lines (`grep -v '^[[:space:]]*//'`).

## What was verified working (for the record)

- New file with `set_var("HOME")` and no token: gate FAILS (the entire recurrence-history class — closed).
- All 6 legit sites + the `transcript_paths.rs` `home_lock()` delegate: pass via file-scope pairing.
- Allowlist is exact-path; a lookalike `tests/.audit_p4_form7_lookalike.rs` FAILS correctly.
- Self-test is load-bearing (both violators asserted CAUGHT, compliant fixture asserted spared).
- CI wiring: new `test-env-lock-gate` job in `.github/workflows/c8-precheck.yml` runs main + `--self-test`; green on the PR.

## Severity

LOW — requires an author to name the canonical lock in prose while deliberately not using it; the historically-observed shapes are all caught.

_Filed by the Fable pre-merge audit of PR #2149. Non-blocking for #2149 itself._

alphaonedev:
## Summary

Fable pre-merge audit residuals filed against PR #2147 (stdin gate) and PR #2149 (env-lock gate) on 2026-07-17. This PR closes all three: #2151, #2152, #2153.

### #2151 — grouped-alias import evasion (`check-test-stdin-reads.sh`)

`use std::io::{stdin as input, Write};` evaded both existing STDIN patterns — the `{` breaks the unbroken `io::stdin` substring `STDIN_TOKEN_PATTERN` anchors on, and `rustfmt` does not split/rejoin a grouped `use` item, so the evasion survives the CI fmt gate. **Fixed**: added `STDIN_ALIAS_PATTERN`, anchored on the `stdin as` token itself (covers both the plain single-import form #2145 already closed and the grouped form). Self-test probe 9 added and asserted caught.

### #2152 — multi-line raw-string truncation (`check-test-stdin-reads.sh`)

`cfg_test_ranges`' per-line string strip has no state for a string spanning multiple lines, so a multi-line raw string whose body carries a net-closing-brace line can truncate a `#[cfg(test)]` range one or more braces early, fail-silent. **This is genuinely beyond a single-line bash strip** — per the orchestrator framing, no multi-line-string state machine was added (explicitly out of scope, matching the "don't chase a perfect parser" directive).

What I DID achieve: `cfg_test_ranges` now emits a loud WARN to stderr naming file+line whenever the per-line brace count drives `depth` **negative** (the single-line-evaluable subset of the truncation class — e.g. a stripped line with two net `}` and no matching `{`). The `2>/dev/null` that was silently swallowing this signal on the awk invocation is removed for that function only.

**Honest scope**: the narrower "depth lands exactly at 0" coincidental-close shape (matching the issue's own canonical repro fixture) is NOT distinguishable from a legitimate range close by brace-count alone, and stays an open, explicitly documented residual in the script header. Self-test probe 10 asserts the WARN fires — it does NOT assert the probe's own stdin read gets caught (that would overclaim; the range IS still truncated in that probe).

### #2153 — env-lock gate comment-insensitivity + missing detector arm (`check-test-env-lock.sh`)

Two parts:

- **(a) Comment-insensitive pairing.** The file-scope `test_env_lock` pairing check matched the token anywhere in the file, including inside a `//` comment. Fixed via a `strip_line_comments` helper (mirrors the sibling stdin gate's per-line strip) applied before the pairing grep.
- **(b) Module-local-lock detector arm** (proposed by #2146, not delivered by PR #2149). Added: independent of whole-file pairing, a file that both mutates `$HOME` and declares its own `static ... Mutex<()>`-style lock within a 50-line adjacency window of that mutation, with no `test_env_lock` reference in the same window, is now flagged — closing the "already-compliant file gains a new hand-rolled lock" residual the #2146 header documented as open. The **window** (not a whole-file "any local lock anywhere" check) is what lets this arm coexist with `src/config.rs`, which legitimately declares several unrelated module-local locks (`GATE_LOCK`, `CAP_LOCK`, `test_env_lock()`'s own internal static) thousands of lines from its `$HOME`-mutating tests — verified no false positive against the live tree.

**Two real bugs surfaced and fixed during this change's own verification** (caught by the mandatory clean-tree false-positive check before merge, not left for CI to find):
1. A `set -euo pipefail` interaction where an empty-result `grep | cut` pipeline killed the whole script on the first file with no local-lock declarations.
2. A `grep -q` piped after a live `sed` process that could SIGPIPE the writer under `pipefail`, false-flagging every `$HOME` line in fully-compliant files (`src/reranker.rs`, `src/cli/rules.rs`, `src/config.rs`, `src/embeddings.rs`). Fixed by materializing comment-stripped content once per file and reading it via here-string instead of a live pipe.

## Self-test probes added

- `check-test-stdin-reads.sh`: probe 9 (grouped-alias import, #2151 — caught), probe 10 (multi-line negative-depth, #2152 — WARN asserted).
- `check-test-env-lock.sh`: comment-only-mention probe (#2153a — caught), module-local-lock-adjacency probe (#2153b — caught, with >50-line padding to isolate the arm-b-only shape from arm-a).

## Verification

- `bash -n` clean on both scripts.
- `--self-test` PASS on both scripts.
- Both gates run clean (`PASS`, zero false positives) against this branch's full `src/` + `tests/` tree — including the six legit `test_env_lock` production sites and the real `#[cfg(test)]` multi-line-raw-string fixtures already in `src/`.

Script-only change; no cargo gates apply. No CHANGELOG entry (matches the precedent of the parent commits 61cf7fae / 3456149e, which are dev-tooling/internal, not user-facing).

Closes #2151
Closes #2152
Closes #2153

## Test plan

- [x] `bash -n scripts/check-test-stdin-reads.sh` / `scripts/check-test-env-lock.sh`
- [x] `scripts/check-test-stdin-reads.sh --self-test` → PASS
- [x] `scripts/check-test-env-lock.sh --self-test` → PASS
- [x] `scripts/check-test-stdin-reads.sh` (clean tree) → PASS
- [x] `scripts/check-test-env-lock.sh` (clean tree) → PASS
- [ ] CI green on the PR (c8-precheck.yml gate jobs)

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/scripts/check-test-env-lock.sh
+++ b/scripts/check-test-env-lock.sh
@@ -45,19 +45,56 @@
 # mutation also contains the test_env_lock token") -- verified true
 # against every current site as of this gate's authorship.
 #
-# KNOWN RESIDUAL GAP (documented, not silently overclaimed): a NEW test
-# fn added to a file that is ALREADY compliant (contains test_env_lock
-# somewhere, for its OTHER tests) could theoretically hand-roll its own
-# unserialized $HOME mutation without tripping this file-scoped gate.
-# Closing that gap needs fn-scoped tracking of every test_env_lock
-# DELEGATE wrapper fn (generalizing the config.rs pattern above), which
-# is a real but narrower follow-up. Today: (a) every one of the six
-# production sites (src/embeddings.rs, src/reranker.rs, src/config.rs,
-# src/cli/commands/config.rs, src/cli/rules.rs,
-# src/recover/transcript_paths.rs) is correctly gated, and (b) any
-# wholesale-NEW file that mutates $HOME without EVER mentioning
-# test_env_lock is caught unconditionally -- the class the issue's
-# recurrence history (#1998/#2115/#2127) actually exhibited each time.
+# CLOSED-BY-#2153 RESIDUAL: the file-scope pairing above only proves the
+# literal token `test_env_lock` appears SOMEWHERE in the file -- it does
+# not distinguish a real call from a bare MENTION inside a `//` comment
+# (issue #2153a: a file could hand-roll its own local lock while merely
+# commenting "unlike test_env_lock, we use our own local mutex here" and
+# still pass), nor did it catch the KNOWN RESIDUAL GAP #2146 originally
+# flagged: a NEW test fn added to a file that is ALREADY compliant
+# (contains test_env_lock somewhere, for its OTHER tests) hand-rolling
+# its own unserialized $HOME mutation nearby. Two arms close both:
+#
+#   (a) Comment-insensitive pairing. The token search now runs over
+#       comment-STRIPPED content (`//` to end-of-line removed per line,
+#       mirroring the sibling stdin gate's per-line strip) before the
+#       whole-file `grep -q` -- a comment-only mention of the token no
+#       longer satisfies the pairing. This is a single-line strip: it
+#       does not understand block comments (`/* ... */`) or a `//`
+#       inside a string literal (e.g. a URL) -- both out of scope for
+#       this bash gate, same bound as the stdin gate's twin.
+#
+#   (b) Module-local-lock adjacency arm (the #2146-proposed detector
+#       arm PR #2149 did not deliver). Independent of whole-file
+#       pairing, any `static <NAME>: ...Mutex<()>`-style module-local
+#       lock declaration WITHIN A WINDOW_LINES-line window of a $HOME
+#       mutation, with no (comment-stripped) `test_env_lock` reference
+#       in that SAME window, is flagged -- even in a file that is
+#       otherwise file-scope-compliant for its OTHER tests. The window
+#       (not a whole-file "any local lock anywhere" check) is what lets
+#       this arm coexist with src/config.rs, which legitimately declares
+#       several UNRELATED module-local locks (GATE_LOCK, CAP_LOCK, and
+#       test_env_lock()'s own internal static) thousands of lines away
+#       from its $HOME-mutating tests -- a whole-file check would
+#       false-positive there; a windowed one does not (verified against
+#       the live tree at authorship).
+#
+# Every one of the six production sites (src/embeddings.rs,
+# src/reranker.rs, src/config.rs, src/cli/commands/config.rs,
+# src/cli/rules.rs, src/recover/transcript_paths.rs) is correctly gated
+# under both arms, and any wholesale-NEW file that mutates $HOME without
+# EVER mentioning test_env_lock in real code is caught unconditionally --
+# the class the issue's recurrence history (#1998/#2115/#2127) actually
+# exhibited each time.
+#
+# NARROWER RESIDUAL (documented, not silently overclaimed): a hand-
+# rolled local lock declared MORE than WINDOW_LINES lines from the
+# $HOME mutation it guards (e.g. at the top of a very large file, used
+# far below) would still evade arm (b) while the file stays file-scope
+# "compliant" if it also references test_env_lock elsewhere. This is a
+# real but narrower gap than the one #2153 closes; a fn/block-scoped
+# structural scan (generalizing the config.rs delegate-wrapper pattern)
+# is the follow-up if this shape is ever observed in practice.
 #
 # ALLOWLIST (a structural exception, not a denylist of bad patterns): a
 # SEPARATE TEST BINARY that documents its own out-of-process
@@ -72,14 +109,19 @@
 #   scripts/check-test-env-lock.sh
 #     - exit 0 on clean, exit 1 on any violation
 #   scripts/check-test-env-lock.sh --self-test
-#     - injects a violating fixture ($HOME mutation, zero test_env_lock
+#     - injects: a violating fixture ($HOME mutation, zero test_env_lock
 #       reference anywhere in the file), a compliant fixture (same
-#       shape, but the file also references test_env_lock), and a
+#       shape, but the file also references test_env_lock), a
 #       hand-rolled-local-lock fixture (defines its own static
 #       Mutex<()> and never references test_env_lock -- exercises the
-#       #1998/#2115/#2127 recurrent class directly), verifies the gate
-#       catches both violators and spares the compliant fixture, then
-#       cleans up. Exit 0 on PASS.
+#       #1998/#2115/#2127 recurrent class directly), a comment-only-
+#       mention fixture (issue #2153a -- the token appears ONLY inside
+#       a `//` comment), and a module-local-lock-adjacency fixture
+#       (issue #2153b -- file-scope-compliant for its FIRST test, but a
+#       SECOND test >WINDOW_LINES away hand-rolls its own local lock for
+#       its own $HOME mutation); verifies the gate catches all four
+#       violators and spares the compliant fixture, then cleans up.
+#       Exit 0 on PASS.
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


alphaonedev:
**Fable pre-merge audit: FINDINGS-NON-BLOCKING — merge OK.** All 24 CI checks terminal-green. Both scripts executed adversarially in the PR worktree (`bash -n`, `--self-test`, clean-tree main gate, plus custom evasion fixtures), not just read.

**#2151 (grouped-alias stdin import) — VERIFIED CLOSED.** Adversarial fixtures, all CAUGHT: plain `use std::io::stdin as x;`, grouped alias-not-first `{Write, stdin as inp}`, no-space `{stdin as inp,Write}`, space-indented multiline grouped, nested-path `use std::{io::stdin as inp};`. One theoretical evasion found: a **hard-TAB-indented** multiline grouped import evades `[{ ,]` (no tab in the class) — but `cargo fmt --check` (required CI gate) fails on it and rustfmt rewrites it into the caught single-line form (verified empirically), so it cannot survive CI. Non-blocking → #2162 (with a one-char widening suggestion and a header-comment nit: rustfmt DOES rejoin grouped imports, contra the new comment — inaccuracy is in the safe direction).

**#2152 (multi-line-string truncation) — HONEST PARTIAL, VERIFIED.** (a) WARN fires on the negative-depth fixture naming file+line (`...probe.rs:15`); (b) the threat-model bound is documented in the header ("KNOWN LIMITATION (issue #2152)... the depth==0 residual... stays open"); (c) clean-tree run produces **zero** stderr — no spurious WARNs on the live multi-line raw strings in src/ (non-vacuous: many src files with `r#"` fixtures inside test modules were traversed). The removed `2>/dev/null` is correctly scoped to the one awk invocation, with a NOTE comment.

**#2153a (comment-insensitive pairing) — VERIFIED CLOSED** via self-test probe (comment-only mention flagged). Block-comment mention (`/* test_env_lock */`) still defeats it — explicitly documented out of scope; accepted.

**#2153b (module-local-lock adjacency arm) — VERIFIED, one doc gap.** New-file local `Mutex<()>` + `set_var("HOME")` FAILS (self-test probe); `OnceLock<Mutex<()>>` shape within window also CAUGHT (my fixture). **config.rs no-false-positive confirmed non-vacuously**: 4 lock decls at lines 4196/5961/6097/8895 vs nearest $HOME mutation at 9765 → min gap 870 lines, matching the header's "~870" claim exactly; clean-tree gate exits 0 in 2.2s with reranker.rs / cli/rules.rs / embeddings.rs / config.rs all passing (the SIGPIPE false-flag set is fixed). Doc gap: a **naked unguarded mutation with no local lock at all** in an already-compliant file still evades both arms and is not listed in the NARROWER RESIDUAL paragraph ("Two arms close both" overclaims slightly). Not a regression; non-blocking → #2163.

**Both self-caught bash bugs independently confirmed real and correctly fixed:**
1. Pre-fix empty `grep|cut` command-substitution under `set -euo pipefail` kills the script (reproduced: rc=1 death). `|| true` is correctly scoped — worst masked case is arm-(b) skip for one unreadable file, not a silent global pass.
2. Pre-fix live `sed | grep -q` shape reproduced on src/config.rs: **rc=141 (SIGPIPE) despite a real match** — exactly the false-flag mechanism described. The materialize-then-here-string fix is genuinely SIGPIPE-safe: content is fully materialized before grep runs (no live writer to signal; a redirection is not a pipeline so pipefail can't see it; 545KB config.rs takes bash's temp-file here-string path). **Trailing-newline analysis clean:** command substitution strips only trailing newlines at EOF, which cannot contain the token; `sed 's#//.*$##'` is line-count-preserving, so `grep -n` line numbers over stripped content stay aligned with original-file numbering for the window arithmetic; no brace-balance is computed over the materialized content (the awk counter reads the file directly). No large-file or last-line behavior change.

**Hygiene CONFIRMED:** all 15 self-test fixture paths are `${ROOT}/src|tests`-scoped; exactly two `rm -f` invocations, each listing exact quoted variable names — no globs, no `-rf`, no `/`-rooted or `/tmp` filesystem paths (the `/tmp/...` strings are Rust literals inside heredoc fixture *content*, never script paths). `ROOT` derives from `dirname "$0"/..`. Worktree left clean, no leftover probes.

**#2152 close-or-keep-open recommendation: CLOSE via this PR.** The actionable in-scope work (fail-LOUD WARN + documented bound) shipped honestly; the depth==0 residual is a permanent threat-model bound of a bash gate, not a pending work item — the in-script header is the right home for it, and #2163 now carries the adjacent env-lock residual so nothing is silently dropped. Keeping #2152 perpetually open as documentation would be worse than the header doc.

Findings: #2162, #2163 (both non-blocking).

Issue #1963, opened 2026-07-09
[v1.0][P1][R68/D14] Signed refusals + egress-class gate decisions extended to the inference plane


alphaonedev:
Gate-2. Signed-refusal object + gate-decision records (R68 core the prose dropped); extend default-deny egress to the inference plane (memory content POSTs to embed/LLM vendors at semantic/smart tiers). Forbidden-export TAXONOMY rides #1838.

Filed under the v1.0.0 development epic #1940 (operator grant memory f9a0f397). Gate-item carrier from the 2026-07-09 kickoff audit (run wf_3c42dc07-761). P0/P1 substrate items are T1+T4 5-agent-vote crossroads; Rust skills (/rust-skills + /rust-microsoft) + the no-hardcoded-literals gates are mandatory on all src/ work.

alphaonedev:
Landed on main via `451c72f6`. Inference-plane egress gate + signed refusals (R68/D14): new src/egress.rs — EgressClass (D14 taxonomy), InferenceEgressMode (allow/loopback-only/deny) from AI_MEMORY_INFERENCE_EGRESS, evaluate_inference_egress + emit_inference_egress_refusal (signed egress.inference_refused event). Gate wired at 4 boot chokepoints (build_llm_client + build_embedder API-arm in daemon_runtime.rs; MCP stdio LLM+embedder init in mcp/mod.rs). **Enforced (not advisory):** on refuse the outbound LLM/embedder client is NEVER constructed (None) — enforcement is the ABSENCE of the egress path, not a bypassable per-request check; the local in-process candle/MiniLM embedder never egresses and is never gated; default allow = byte-identical legacy. The signed audit row is best-effort (skipped-with-WARN if no db conn at the boot site; enforcement still holds). 7 egress unit tests + CLAUDE.md env row 131. No schema. Verified on the freed host: full lib suite green (a pre-existing store-url env-pollution flake fixed en route, e31965e8).

PR #2170, opened 2026-07-17
feat(#2166): hot-swappable [llm] config — reload model/provider without serve restart

### LINKED ISSUE #2166: [v1.0.0] Hot-swappable [llm] config — reload the LLM model/provider without restarting the serve process (LLM ONLY)
## Problem

Changing the `[llm]` model/provider in `~/.config/ai-memory/config.toml` does **not** take effect on a running MCP server. Surfaced 2026-07-17 swapping `google/gemma-4-31b-it` → `x-ai/grok-4.5`: the curator daemons picked it up (their `build_curator_llm` at `src/cli/curator.rs:126` calls `AppConfig::load()` fresh each sweep, and a `systemctl --user restart` reloaded them), but the MCP-invoked LLM path did not.

Root cause (verified via codegraph at release/v1.0.0 ff04c9e6): the serve path resolves + builds the LLM client **once at boot** — `run_mcp_server` → `build_llm_client` (`src/daemon_runtime.rs:2893`) → `resolve_llm` → `OllamaClient`, then arms it into the reflect/consolidate hooks threaded through the server. Each `memory_reflect`/`memory_consolidate`/persona/auto_tag/expand_query/contradiction-detect call **reuses that boot-built client**. So the only way to change the MCP LLM model today is a full `ai-memory serve` (Claude Code) restart.

The gap is specifically the **long-lived serve processes** (MCP stdio serve + the HTTP daemon serve). The per-sweep curator path is already effectively hot.

## Proposed approaches (pick one; SIGHUP is the conventional daemon pattern)

1. **SIGHUP config reload.** On SIGHUP: `AppConfig::load()` → validate → rebuild the LLM client → swap it behind an `arc_swap::ArcSwap<Option<OllamaClient>>` (or `RwLock`) so in-flight requests are unaffected. Mirrors the graceful-shutdown signal wiring already in `daemon_runtime::serve`.
2. **Operator-triggered reload verb.** An admin-family MCP tool (`memory_reload_config`) and/or `POST /api/v1/admin/reload` (admin-gated) that re-resolves and hot-swaps. Explicit, auditable.
3. **Lazy generation check.** Cheap `stat` of config.toml mtime on each LLM-op; rebuild only when it changed. Simplest, but adds a stat to the hot path.

## Design considerations (the reason this is non-trivial)

- **Concurrency.** The client is shared across concurrent reflect/consolidate. Use `ArcSwap`/`RwLock`; never tear down a client an in-flight request holds. **The reflect/consolidate hooks currently CAPTURE the client at boot** (`ReflectHooks` armed in `run_mcp_server`) — the hook must resolve the *current* client at call-time instead of capturing, which is the main refactor.
- **Validate-before-swap.** `AppConfig::load()` already falls back to default on parse error; a reload must NOT swap in a broken client — validate, keep the old one on failure, log loudly.
- **Egress gate re-eval.** `build_llm_client` consults the #1963 inference-egress gate; a reload must re-evaluate it (a reload could legitimately disable the client).
- **Provenance.** Log which precedence layer won on reload (the resolver already emits this at boot).
- **Scope: `[llm]` ONLY.** Do NOT hot-swap `[embeddings]` — changing the embedding model mid-corpus corrupts the shared vector space (dim/geometry mismatch). Reranker is borderline; default to `[llm]`-only for this issue.

## Milestone rationale — v1.x, not v1.0.0

Deferred to v1.x on purpose: it's a feature-add that touches the boot/serve wiring, the hook arming, and introduces shared-mutable-client concurrency — **freeze-hostile** for the current GA endgame — while the workaround (restart the serve process) is trivial and the curator path is already hot. If the operator wants it pulled into v1.0.0, re-scope. Filed per the 'every open issue reaches a documented terminal state' directive with this rationale.

## References
- Load path: `src/daemon_runtime.rs:2893` (`build_llm_client`), `src/cli/curator.rs:126` (`build_curator_llm`), `run_mcp_server` hook arming.
- Surfaced by the 2026-07-17 grok-4.5 swap.

alphaonedev:
## Summary

Hot-swappable `[llm]` config (#2166) — the long-lived serve processes now pick up a `config.toml` `[llm]` model/provider change **without a full restart**. Surfaced by the 2026-07-17 `gemma-4-31b` → `grok-4.5` swap (curators were already hot; the MCP + HTTP daemon LLM paths captured the boot client for the process lifetime).

**Scope is `[llm]` ONLY** (operator-locked). `[embeddings]` stays restart-only (`ai-memory reembed`) because hot-swapping the embedding model mid-corpus corrupts the shared vector space; `[reranker]` is out of scope.

## Design

- **`crate::reload::SwappableLlm`** — concurrency-safe holder over the current `[llm]` client that `impl AutonomyLlm` by resolving the CURRENT handle per call (capture-inversion: armed once, never re-armed). Zero-dep `RwLock<Arc<Option<OllamaClient>>>`; reads clone the `Arc` and drop the guard before any `.await` — no lock guard crosses a suspension point. (Operator-named `ArcSwap` deferred to avoid an unauthorized `Cargo.toml` dependency add per the sole-authority rule; the LOE named both as correct.)
- **`AppState.llm` → `Arc<SwappableLlm>`**, **`AppState.resolved_models` → `Arc<Swappable<ResolvedModels>>`**; every HTTP read site resolves current via `app.llm.current()` / `app.resolved_models.current()`, so `memory_capabilities.models.llm` never lies about the active model post-swap.
- **`AppConfig::try_load` / `try_load_from`** — Result-returning loaders that PROPAGATE a parse/secret-validation error instead of swallowing it to `default()` the way `load_from` does. Validate-before-swap: a broken/typo'd config KEEPS the current working client and logs a loud WARN — never swaps in the compiled-default model.
- **Reload triggers**: HTTP daemon **SIGHUP** on the `serve` loop (⚠️ deliberately converts SIGHUP kill→reload, the conventional daemon pattern — a behavior change); MCP stdio a **lazy `config.toml` mtime check between requests** (single-threaded → race-free; no SIGHUP on a stdio child). Both rebuild via `daemon_runtime::build_llm_client` (HTTP, async) / the shared `reload::resolve_and_build_mcp_llm` (MCP, sync), so the #1963 inference-egress gate is **re-evaluated every reload** — a reload can legitimately DISABLE the client (egress `deny`).

## Note / deviation from the Fable LOE

The `auto_persona` `build_post_reflect_hook` the LOE identified as the "boot-captured hook to invert" has **zero production callers** at HEAD (verified via codegraph + grep) — the production reflect handler uses the `auto_export` hook (no LLM). There is no boot-captured hook bundle to re-arm; the real per-request LLM consumers (MCP `handle_request` `llm` param; HTTP `app.llm` read sites) are made swap-aware directly. `SwappableLlm` still carries the `AutonomyLlm` impl so any future hook arms with the swap handle and needs no re-arming.

## SSOT

SIGHUP + the MCP mtime check add **no** MCP tool / HTTP route / CLI subcommand / schema version — no SSOT pin moves (an admin reload verb was deferred to v1.x per the issue). No new dependency. `qual_10` config.rs ceiling bumped 12850→12950 in lockstep (dated). CHANGELOG `[Unreleased]` entry added.

## Tests

`tests/hot_swap_llm_2166.rs` — 4/4 green:
- `swap_succeeds_current_reflects_new_model_2166`
- `validate_before_swap_keeps_old_client_2166`
- `egress_deny_reload_disables_client_2166`
- `concurrent_swap_and_reads_no_panic_2166`

The `AppState.llm`/`resolved_models` field-type change ripples mechanically into ~115 test routers.

## Gates run green

`cargo fmt --all --check`; `cargo check --all-targets` (default) + `--features sal` + `--features sal-postgres`; `cargo clippy --all-targets -- -D warnings` (default) + `--features sal`; `cargo check --examples`; `qual_10_module_size_ceiling`; `qual_6_7_legacy_error_type_ceiling`; vendor-literal / hardcoded-literal / docs-vs-ssot script gates.

## Do NOT merge yet

A mandatory Fable pre-merge security + concurrency audit runs before merge (every consumer routes through the swap point, no guard across `.await`, egress gate inside the rebuild, parse-fail never replaces a working client).

Closes #2166

Rule IDs: own-arc-shared, own-rwlock-readers, async-no-lock-await, anti-lock-across-await, err-no-unwrap-prod, obs-structured-fields, M-SERVICES-CLONE, M-TYPES-SEND, M-STRONG-TYPES-GUARD.


alphaonedev:
# Fable pre-merge audit — **NOT CLEAN: 1 BLOCKING finding (#2172) + CI red**

Audited PR head `fe9d554d` against the #2166 operator spec + Fable LOE, adversarially, across both lanes (HTTP daemon + MCP stdio) with a full codegraph/grep enumeration of every `OllamaClient`/`AutonomyLlm` consumer.

## BLOCKING — #2172: two MCP funnels bypass the swap point

`atomise_handler` (`memory_atomise`) and `ingest_multistep_handler` (`memory_ingest_multistep`) are built ONCE at MCP boot with `llm_client.clone()` (`src/mcp/mod.rs` ~:3800 / ~:3837) and the #2166 reload block reassigns only `llm` + `resolved_models`. Post-swap these two tools silently run the **old model/provider** (the exact bug class this PR closes); a reload that disables the client (egress refusal of the new target) leaves them **still egressing memory content to the old vendor**, contradicting the PR/CHANGELOG "#1963 gate re-evaluated on every reload" claim; and the signed `atomisation_complete` `curator_model` field misattests the live model. Both are loop-locals in the single-threaded stdio loop — rebuilding them inside the reload block is the same race-free swap already used for `llm`. Details + fix shape in #2172.

## Swap-point completeness enumeration (every consumer → resolves current?)

| Consumer | Lane | Resolves current? |
|---|---|---|
| `handle_request` `ctx.llm` (auto_tag, expand_query, detect_contradiction, consolidate, session_start, persona, store synthesis/classifier) | MCP | **YES** — loop-local `llm` reassigned on mtime reload |
| `resolved_models` capability surface | MCP + HTTP | **YES** — refreshed in lockstep (`refresh_llm_model_surface` / `Swappable.current()`) |
| `maybe_auto_tag`, `maybe_detect_conflicts` (`handlers/http.rs`) | HTTP | **YES** — `app.llm.current()` |
| `resolve_consolidate_summary`, `auto_tag_handler`, `expand_query_handler` (`handlers/power_consolidation.rs`) | HTTP | **YES** — `.current()`; `is_none()`→`current()` TOCTOU degrades gracefully (`let Some(..) else` empty-result arm) |
| `atomise_handler` → `LlmCurator` | MCP | **NO — boot-captured (#2172, blocking)** |
| `ingest_multistep_handler` → `OllamaDispatch` | MCP | **NO — boot-captured (#2172, blocking)** |
| Reflect post-hook | both | n/a — production reflect uses the **auto_export** hook (no LLM), built per-request; the LLM-bearing `auto_persona::build_post_reflect_hook` has **zero production callers** (codegraph: 5 callers, all tests) — the PR's LOE-deviation claim is **verified correct** |
| HTTP `memory_atomise` route | HTTP | n/a — `handle_atomise_http` passes `None` handler (no LLM on this route) |
| Curator daemon | — | already hot (per-sweep `AppConfig::load()`), untouched |
| CLI one-shots (expand/persona/atomise/…) | — | n/a — fresh process |
| `inference::CpuBackend` | — | n/a — unwired scaffolding, no production callers |

## Concurrency — CLEAN

`Swappable<T>` = `RwLock<Arc<T>>`; `current()` clones the `Arc` and drops the guard within the expression — no guard ever crosses `.await` at any read site (verified all five HTTP sites bind the `Arc`, then await) — `async-no-lock-await` / `anti-lock-across-await` / `own-rwlock-readers` / `own-arc-shared` / M-SERVICES-CLONE satisfied. No nested lock acquisition anywhere → no deadlock; poison handled via `PoisonError::into_inner` (never panics). Swap-vs-in-flight-read is safe: an in-flight request keeps its snapshot `Arc`, the old client drops on last release. SIGHUPs are serialized by the single handler task. Minor (noted in #2174): the `llm`/`resolved_models` pair swap is two `store()`s, so one concurrent capabilities read can transiently pair new client + old reported model — cosmetic.

## Validate-before-swap — CLEAN, one hardening issue

`try_load`/`try_load_from` propagate parse + `validate_secret_handling` errors (vs `load_from`'s swallow-to-default); both lanes KEEP the working client and WARN on failure — verified by test (b) and by reading both reload paths. Missing/unreadable config at mtime-check time → no reload, keep old (fail-safe). Residual: a torn read of a *non-atomically-written* config can present valid-but-incomplete TOML that parses `Ok` and transiently downgrades the model (self-heals next mtime tick; stateless LLM → degrade, not corruption) — **#2173, non-blocking hardening** (stability/debounce check).

## Egress (#1963) — correct at the swap point, bypassed on the #2172 funnels

Both rebuild paths (`daemon_runtime::build_llm_client` async / `reload::resolve_and_build_mcp_llm` sync) evaluate the gate INSIDE the rebuild that produces the swapped-in value; refusal → `None` swaps in; `None`-current is handled at every current-resolving consumer (trait no-ops, handler early-returns/degrade arms). The #2172 boot-captured handlers are the exception — covered above.

## SIGHUP — safe, intentional, documented

Unix-only, `serve` (HTTP daemon) only; kill→reload conversion is loudly documented in CHANGELOG + code comment; no interaction with the `ctrl_c` graceful-shutdown future (different signals, no select contention); registration failure degrades to no-reload with a WARN; MCP stdio deliberately keeps default HUP semantics (parent-exit). Handles cloned before `app_state` moves into the router — no ownership hazard.

## ~115-file test migration — mechanical, security tests intact

Diffed every test file: outside `tests/hot_swap_llm_2166.rs` the only changes are the two AppState construction hunks (`SwappableLlm::new(None)` / `Swappable::new(ResolvedModels::default())`), the dated `qual_10` config.rs ceiling bump (12850→12950), and `cov_handlers_llm_wired_1660`'s fixture keeping its `Some(client)` wiring (LLM-wired coverage NOT vacuated). Spot-checked `admin_header_trust_1570`, `admin_visibility_bypass_gate_1582`, `agent_attestation_integrity`, `agent_id_spoof_901_regression`, `archive_purge_owner_gate_936`: admin ids, routers, and assertions untouched — no security test weakened.

## North Star

No storage/recall data-path change; the swapped object is a stateless outbound client; a failed reload keeps the old client, a disabling reload degrades. The only integrity blemish is the #2172 `curator_model` signed-event misattestation.

## CI — RED (fix required)

`Lint (fmt + clippy)` and `Postgres feature gate` both fail on **3 `clippy::pedantic` lints in the new `tests/hot_swap_llm_2166.rs`** (reproduced locally with the CI invocation `cargo clippy --all-targets --features sal -- -D warnings -D clippy::all -D clippy::pedantic`): `redundant_closure_for_method_calls` (:28), `items_after_statements` (:68), `single_match_else` (:116). The local-gates list in the PR body ran default-profile clippy on the sal build — CI's sal legs add `-D clippy::pedantic`. Trivial fixes; the 4/4 suite itself passes locally, as do `qual_10` (2/2 with the dated bump).

## Verdict

**BLOCKING — do not merge** until #2172 (rebuild/invert the two MCP handlers in the reload block + a regression test that swaps then drives `memory_atomise`) and the pedantic clippy failures are fixed. #2173/#2174 are non-blocking follow-ups and may land separately. Everything else — concurrency, validate-before-swap, SIGHUP wiring, egress-at-the-swap-point, the LOE-deviation claim, and the test migration — audits clean.

Issue #2167, opened 2026-07-17
[v1.0.0][data-integrity] Silent recall corruption: embedding-space provenance (model+prefix at same dim) is never persisted or checked — config-swap, interrupted reembed, restore, default drift (both backends)


alphaonedev:
## Summary

Stored embedding vectors live in a coordinate space defined by the producing **model (+ its task-prefix scheme)** — but ai-memory persists **no model provenance anywhere**: not in a meta key, not in a column, not in the G13 BLOB header. Every guard in the substrate — write-time G4 namespace-dim invariant, recall-time H7 `CosineComparison`, the pgvector `vector(N)` column type, the `AI_MEMORY_REQUIRE_DIM_MATCH` HNSW strict mode (#114) — checks **dimension only**. Any path that lands a *same-dimension, different-model* vector in the corpus produces silent recall corruption: cosine returns numerically-valid, semantically-meaningless scores; rows rank wrong; nothing errors, warns, or counts.

## Correctness contract (ACCEPTANCE CRITERION — operator hard requirement)

The fix must enforce a **structural invariant**, not a set of case-by-case catches:

> **INVARIANT: Recall NEVER compares two vectors from different embedding spaces, AND an incompatible vector NEVER enters a space unflagged.**

If this holds at every point, corruption is **impossible regardless of cause** — config swap, partial reembed, federation inbound, version-default drift, restore — without enumerating causes. Every enforcement point below is **FAIL-CLOSED**.

**Key principle:** the source TEXT is the durable truth; **vectors are disposable derivatives**. Therefore the worst-case failure mode is **DEGRADED recall** (an incompatible/unverified vector is temporarily unsearchable — excluded from semantic scoring, still keyword-recallable — until `reembed`/backfill regenerates it from text), **NEVER WRONG recall**, and the system **SELF-HEALS** as reembed runs. No data is ever lost. This is the correctness contract the implementation and the Fable audit will verify against (tests T-INV-1..3 below).

### What is persisted today
- **G13 BLOB header** (`src/embeddings.rs:1050-1170`): a **single byte** encoding float endianness only (`0x01` LE f32; `0x02` BE reserved). No model, no dim. Dim is implicit from byte length + the separate `embedding_dim` column.
- **`embedding_dim` column** (both backends; sqlite v0.6.3.1 P2/G4, pg `postgres_schema.sql:83`): per-row **dim**, no model.
[...]

Issue #1598, opened 2026-06-11
API-backed embeddings facility: allow [embeddings].backend=openai-compatible/openrouter so CPU-only nodes need no Ollama (wire support already exists)


alphaonedev:
**Parent:** #1588 (dogfood eval). Filed at operator request (2026-06-11): "we have the OpenRouter backend llm Gemma 4 26B — should we have API facility for the smaller llm also?"

**Current state (v0.7.0 @ 361f6163):**
- **LLM**: fully provider-agnostic post-#1067 (15 API vendor aliases) — a CPU-only node needs no local inference for expand/auto-tag/contradiction/consolidate. ✅
- **Embeddings**: local-only. `EmbeddingModel` enum = `MiniLmL6V2` (in-process candle, 384-dim) | `NomicEmbedV15` (requires Ollama, hardcoded `NOMIC_OLLAMA_MODEL`, `src/embeddings.rs:369-377`); `resolve_embeddings()` accepts any backend string but construction only ever builds Ollama clients (#1143 boot path). ❌ No API option.
- **Reranker**: local CPU only (and the actual CPU pain point — see #1597: 2.8-7.7s/recall; Ollama nomic embedding itself is fine on CPU: measured 80-245ms).

**Key fact — the wire plumbing already exists:** `LlmClient::embed_text_async` (`src/llm.rs:1550-1561`) already speaks BOTH wire shapes: Ollama-native `/api/embed` AND OpenAI-compatible `/embeddings` + Bearer. The gap is purely config + construction.

**Live verification (node f2, existing OpenRouter key, 2026-06-11):**
```
POST openrouter.ai/api/v1/embeddings model=openai/text-embedding-3-small → OK dim=1536
POST ... model=qwen/qwen3-embedding-8b                                   → OK dim=4096
POST ... model=google/gemini-embedding-001                                → OK dim=3072
POST ... model=nomic-ai/nomic-embed-text-v1.5                             → 400 does not exist
```
OpenRouter serves embeddings under the SAME key the node already holds — but NOT nomic-embed-text-v1.5, so switching an existing corpus = full re-embed (this DB: 2,022 rows ≈ cents) and a deliberate dim migration (KNOWN_EMBEDDING_DIMS handles per-row dim-mismatch skip, but mixed-dim recall loss is the #1531 H7 class).

[...]

Issue #1584, opened 2026-06-10
SEC (MED): shipped federation vectors (#1579 B1) validated for dim only — no finite/L2-norm check (recall poisoning)


alphaonedev:
## Severity: MEDIUM
CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:U/C:N/I:H/A:L

## Root cause (train-introduced — #1579 B1)
The B1 embed-once-replicate-vector change ships the sender's stored embedding with each federated row. The receive-path store arm validates **dimension only** — `receiver_dim == Some(se.dim) && se.vector.len() == se.dim` — before calling `set_embedding`/`update_embedding` directly:
- `src/handlers/federation_receive.rs:819` (sqlite path)
- `src/handlers/federation_signing_check.rs:249` (postgres twin)

There is **no `is_finite()` check and no L2-normalization** of the f32 components. Downstream:
- `src/embeddings.rs` `cosine_similarity` has no NaN guard — a NaN component makes scores NaN, and NaN is unordered under `partial_cmp`, silently corrupting candidate ordering.
- `src/hnsw.rs:218` `cosine_distance = 1.0 - dot` assumes both operands are L2-normalized; a high-magnitude non-normalized shipped vector inflates its dot product against every query, ranking itself artificially high across all recalls.

## Attack
An enrolled+signing peer (or any peer in the permissive zero-config posture) ships `content="benign"` paired with a hand-crafted NaN/Inf or high-magnitude vector. The signature attests the **bytes came from the sender**, not that the vector honestly embeds the content. Receiver semantic recall ranking is poisoned invisibly (FTS still keyword-matches the honest content).

## Proposed fix
On the shipped-vector arm (both backends), before `set_embedding`: reject/skip any vector with a non-finite component, and L2-normalize before storing (locally-computed embeddings are already normalized; shipped ones must be). Add a NaN-safe comparator in the cosine paths as defense-in-depth. Document the residual content↔vector trust assumption (signature attests sender, not embedding honesty) — value-domain validation is the fix; do NOT revert B1's perf win by re-embedding. Regression test: a shipped NaN/non-normalized vector is normalized-or-rejected and does not corrupt ranking.

Surfaced by the operator-mandated 10-agent #1579 security audit (release/v0.7.0 @ c7a5af46).
🤖 Generated with [Claude Code](https://claude.com/claude-code)

alphaonedev:
[...]

Issue #1579, opened 2026-06-10
perf: full-spectrum performance audit — v0.7.0 FINAL GATE (measured findings, tiered work breakout, LOE + AI-NHI estimates, ship-vs-fix recommendation)


alphaonedev:
# perf: full-spectrum performance audit — v0.7.0 FINAL GATE (4-agent measured audit; findings + work breakout + LOE + disposition)

**Status: AUDIT ONLY — no remediation applied. All fixes gated on operator decision per directive.**

Audit executed 2026-06-09/10 against `release/v0.7.0` @ 578b8b48 (golden binary 21bd6899 on the live do-1461 fleet). Four parallel audit agents:

- **P1** — local hot paths + SQLite endpoint (built-in bench, MCP stdio per-op at empty/10k/100k, CLI overhead, PRAGMA A/B, EXPLAIN)
- **P2** — HTTP(S) API surface + concurrency model (per-endpoint percentiles, c=1→32 scaling, sqlite-vs-postgres, overhead stack, wire efficiency)
- **P3** — live fleet: the 3 encrypted legs + A2A + PG/AGE/pgvector + federation throughput retrospective (15 droplets, 3 regions)
- **P4** — code-level perf-vector census via CodeGraph (Reviewer-5 F5.x re-verification + missed-vector sweep + exponential ranking)

Raw artifacts: `.local-runs/perf-audit/{p1,p2,p3,p4}/` on the build host. Memories: `perf-audit-v070` ns (efa8356e, e13d1ba1, 0e0abf3c), ai-memory ns (8958e1b7).

---

## 1. Executive summary

The system is **fast at small scale and architecturally sound on the wire** — quorum convergence is excellent (<0.6s globally), mTLS legs behave exactly as modeled (fresh conn ≈ 4.3×RTT, reused ≈ 1×RTT), PG/HNSW per-query primitives are millisecond-class, and the per-request HTTP floor is ~0.1–0.2ms. The performance debt is concentrated in **five structural points**, all measured:

1. **Corpus-scale reads** — recall/search/list grow linearly with corpus (no bounded scan / missing composite index / per-row tsvector recompute). PERFORMANCE.md budgets are blown at 100k rows on both adapters (sqlite recall 269ms p50 vs 50ms budget; PG 306ms at 8k rows → ~4s extrapolated at 100k). The built-in bench gates at 500 rows, so CI never sees this.
[...]

PR #2171, opened 2026-07-17
fix(#2168): fingerprint-gate federation inbound embedding vectors (both backends)

### LINKED ISSUE #2168: [v1.0.0][security][data-integrity] Federation inbound stores same-dim foreign-MODEL shipped embeddings verbatim — ShippedEmbedding.model is on the wire but never compared (both backends)
## Summary

The federation embed-ship receive path (#1566 / #1579 B1) stores a peer-shipped embedding vector **verbatim** into the local corpus when it passes the dimension + finiteness/norm gates — but it **never compares the shipped `model` field against the local embedder's identity**, even though the wire already carries it. A same-dimension vector produced by a *different* embedding model (or the same model under a different task-prefix scheme) lives in a different coordinate space: cosine against the local query embedder returns a numerically-valid but semantically-meaningless score. Recall silently returns wrong matches — no error, no WARN, no counter. This is the **automatic** (no-operator-action) entry vector for cross-model corpus corruption, and it fires identically on BOTH backends.

## Correctness contract (ACCEPTANCE CRITERION — operator hard requirement)

This gate enforces the write-point half of the structural invariant defined in companion issue #2167:

> **INVARIANT: an incompatible vector NEVER enters a space unflagged** (and recall NEVER compares cross-space vectors — the recall half is #2167 step 4).

**FAIL-CLOSED = reject-or-reembed:** a shipped vector whose space fingerprint does not PROVABLY match the receiver's active space is **never stored verbatim** — it is routed to the existing deferred local re-embed path, which regenerates the vector from the shipped TEXT under the receiver's own space. The source text is the durable truth; vectors are disposable derivatives. Worst case is therefore **DEGRADED recall** (the row is keyword-only for the seconds-to-minutes deferred-embed window — exactly the window that already exists today for dim-mismatched/no-vector rows), **NEVER WRONG recall**, and it **SELF-HEALS automatically** with no operator action and no replica divergence (the memory row itself always applies; only the derived vector is regenerated — CRDT-safe).

Framing (design intent): this is NOT a new checker — it **generalizes the existing dim-equality gate at this exact guard point into a vector-space IDENTITY check**. The dim axis is already load-bearing here; the model + prefix-scheme axes are carried/derivable but never compared. A vector-value or schema-type inspector cannot detect this class **by construction** — a same-dim vector from model A is byte-identical in shape to one from model B; the only correct mechanism is recorded provenance compared at the gate.

## Evidence (release/v1.0.0 HEAD ff04c9e6)

**The wire carries the model — and documents it as decorative** (`src/federation/mod.rs:187`):

```rust
pub struct ShippedEmbedding {
    pub memory_id: String,
    /// Human-readable id of the model that produced the vector
    /// (sender's `Embedder::model_description()`). Observability only —
    /// the dim gate is the load-bearing safety check.
    pub model: String,
    pub dim: usize,
    pub vector: Vec<f32>,
}
```

**sqlite receive funnel** (`src/handlers/federation_receive.rs:1512-1541`): acceptance predicate is `receiver_dim == Some(se.dim) && se.vector.len() == se.dim` + `sanitize_shipped_vector` (finite + L2 norm, #1584). `se.model` is destructured **only into the failure WARN string**; on success the foreign vector goes straight into `db::set_embedding` AND the live in-memory HNSW index (`hnsw_updates` → `idx.insert`), so corruption is live for recall before any restart/backfill.

**postgres receive funnel** (`src/handlers/federation_signing_check.rs:364-390`, `sync_push_via_store`): byte-parallel logic — same dim-only predicate, `store.update_embedding(&ctx, &applied_id, Some(&vector))` verbatim. Backend parity of the *gap* is exact.

**Why dim-only is insufficient:** `KNOWN_EMBEDDING_DIMS` (`src/config.rs:6721`) already ships **six distinct 768-dim model families** (nomic-embed-text-v1.5, bge-base-en-v1.5, embedding-001, text-embedding-004, granite-embedding, snowflake-arctic-embed:m), four at 1024, three at 384, two each at 1536/3072. Post-#1598 every node resolves its embedder independently (`AI_MEMORY_EMBED_*` env per node), so a heterogeneous fleet where peer A runs nomic-768 and peer B runs granite-768 is one env var of drift away — and every replicated row then silently poisons the receiver's semantic ranking. The H7 recall guard (`src/embeddings.rs:296` `CosineComparison`) is dimension-aware only, so it cannot catch this after the fact.

**Prefix-scheme axis (#1520):** the nomic family embeds documents under `search_document: ` and queries under `search_query: ` (`src/embeddings.rs:73-74`, `embed_with_role`), gated by `model_requires_nomic_prefix` — a **binary-version-implied** mapping. Two peers on the *same model id* but different binary prefix behavior (pre-/post-#1520, or a future release that starts sending `task_type` for Gemini-family models — today the embed wire is bare `{model, input, truncate}`, `src/llm.rs::embed_text`, so gemini-embedding-2 via OpenRouter is used role-symmetrically) produce different document spaces under an identical model label. The identity compared at the gate must therefore be the **fingerprint `(model_id, dim, prefix_scheme)`**, not model+dim alone.

## Proposed fix (S, no schema change, both funnels)

The deferred-local-re-embed fallback **already exists** for dim-mismatch/no-vector rows — the fix routes every not-provably-matching vector down that same path:

1. **Canonical space fingerprint.** Add `Embedder::space_fingerprint() -> String` = `<canonical_model_id>#<prefix_scheme>` where `prefix_scheme ∈ {"none", "nomic-task-v1", …}` (canonical id via the existing `canonicalise_embedding_model`; `all-MiniLM-L6-v2` for the Local variant — NOT the `model_description()` display prose that currently pollutes `se.model` with `"nomic-embed-text (768-dim, remote)"`). Dim is deliberately NOT folded into the string — it stays on the existing `dim` field/gates so there is one SSOT per axis. Recording the prefix scheme **explicitly** (not model-implied) is the point: it guards cross-version drift of the prefix logic itself.
2. **Wire.** Sender ships `model = space_fingerprint()` (field name kept for wire back-compat). Receiver canonicalises: old-prose senders (`"name (768-dim, remote)"`) parse the leading token for the model axis.
3. **Receive gate (both funnels) — verbatim store requires PROOF of match; everything else re-embeds.** Extended acceptance predicate: `dim matches && fingerprint provably matches`. Any other case — foreign fingerprint, **absent/empty fingerprint (legacy sender)**, unparseable prose — is routed to `deferred_embed` (local re-embed from the shipped title/content under the receiver's own space; the memory row itself is untouched — no replica divergence, CRDT-safe) + a structured WARN (`M-LOG-STRUCTURED`) naming both fingerprints. **No accept-with-WARN posture and no permissive env:** the fallback is cheap, automatic, and correct, so there is no legitimate case for storing an unproven vector — this is what makes the write-point invariant total. (Rolling-upgrade cost: pushes from pre-fix peers are re-embedded locally instead of stored verbatim for the duration of the fleet upgrade — a perf regression to the pre-#1566 embed cost for that window, never a correctness regression, and the #1566 ack-latency win is preserved because the re-embed is already deferred off the ack path.)
4. Once #2167's v84 `embedding_space` column lands, the gate additionally stamps accepted vectors with the verified fingerprint (and locally re-embedded ones with the local fingerprint).
5. Rule IDs: `M-STRONG-TYPES-GUARD` (the fingerprint newtype guards the space invariant; compare canonical ids, not display prose), `M-LOG-STRUCTURED`, `M-DOCUMENTED-MAGIC` (delete the "observability only" sentence; document the comparison contract).

### Files
- `src/embeddings.rs` — `Embedder::space_fingerprint()` + prefix-scheme token (~25 LOC)
- `src/federation/mod.rs` — doc fix + fingerprint canonicalise/compare helper (~30 LOC)
- `src/federation/sync.rs` — sender ships the fingerprint (~5 LOC)
- `src/handlers/federation_receive.rs` — sqlite gate (~15 LOC)
- `src/handlers/federation_signing_check.rs` — postgres gate (~15 LOC)

### Test plan — T-INV-2 (the write/federation-gate invariant proof, per #2167's acceptance criteria)
- `tests/federation_1566_embed_ship.rs`: same-dim foreign-fingerprint shipped vector → memory row applied, vector NOT stored verbatim (assert stored blob is NOT the shipped bytes), row lands on the deferred re-embed path, HNSW never contains the shipped vector (sqlite).
- postgres twin (`#[ignore]` + sal-postgres, run `--include-ignored`): same assertion through `sync_push_via_store` / `update_embedding`.
- **absent-fingerprint legacy sender → re-embed path, never verbatim** (the fail-closed posture).
- same model id + different prefix_scheme → re-embed path (the #1520 axis).
- matching-fingerprint vector still stored verbatim (no perf regression of #1566 for upgraded fleets); prose-form `model` from an old sender that canonicalises to a match on the model axis still requires the full fingerprint to be provable — otherwise re-embed.

~90 production LOC + ~280 test LOC. Opus-implementable under this spec with the standard review gate; the predicate change is small and localized. The Fable audit verifies T-INV-2 against the correctness contract above.

## Severity
Silent data-integrity corruption, automatic, network-triggered, both backends: **high within the data-integrity class**; labeled `security-medium` because it requires an *enrolled* peer (the #1584 analysis already treats an enrolled peer as semi-trusted for vector honesty — but honest heterogeneous-config fleets hit this with zero malice). Milestone **v1.0** per the "no gaps" release bar: the guard is S-sized and the fallback path already exists.

---
*Filed by Fable (hard-coder tier) during the v1.0.0 full-spectrum data-integrity assessment; acceptance bar raised to the structural invariant per operator hard requirement (2026-07-17). Companion: #2167 (embedding-space provenance substrate — per-row column + boot/recall/reembed/restore guards).*

alphaonedev:
Closes #2168

## Problem

The #1566 / #1579 B1 embed-ship receive path stored a peer-shipped embedding vector **verbatim** once it passed the dimension + finiteness/L2-norm (#1584) gates — but it never compared the shipped `model` against the receiver's own embedder identity, even though `ShippedEmbedding.model` already carries it on the wire (its doc said "observability only"). A same-dimension vector produced by a **different** embedding model — or the same model under a different prefix scheme (#1520 nomic asymmetric `search_document:`/`search_query:`) — lives in a different coordinate space: cosine against the local query embedder returns a numerically-valid but semantically-meaningless score, so recall silently returns wrong matches (no error, no WARN, no counter). Post-#1598 every node resolves its embedder independently, so a heterogeneous fleet (peer A nomic-768, peer B granite-768) is one env var of drift from silently poisoning the receiver's vector space — and it fired identically on **both** backends.

## Approach

Migration-free receive-gate fix (independent of the #2167 provenance schema). A new canonical vector-space fingerprint `<canonical_model_id>#<prefix_scheme>` generalises the existing dim-equality gate into a vector-space **identity** check:

- `embeddings::embedding_space_fingerprint(model)` + `Embedder::space_fingerprint()` — the model id is the token before the ` (` prose suffix (prose-tolerant + wire-back-compat: no wire change), lowercased; the prefix scheme derives from that id **exactly** as the local embed path derives it (`Embedder::model_requires_nomic_prefix`, the #1520 predicate). Dimension is **excluded** from the fingerprint — it stays on the separate dim gate so there is one SSOT per axis (M-DOCUMENTED-MAGIC).
- On inbound (`federation_receive::sync_push` sqlite + `federation_signing_check::sync_push_via_store` postgres — **identical** gate, cross-backend parity), the receiver compares the shipped fingerprint (from `se.model`) against its **own** configured embedder fingerprint (resolved from the same `app.embedder`):
  - **MATCH** → the vector is stored verbatim exactly as before (no #1566 perf regression).
  - **MISMATCH** → the foreign vector is **not** stored; the row falls through to the **existing** deferred-local-re-embed fallback (`deferred_embed` → `spawn_deferred_embedding_refresh`), so it still lands (CRDT convergence preserved — the memory is not dropped, only its foreign vector is refused) and is re-embedded under the **local** model, plus a structured WARN (M-LOG-STRUCTURED, `federation::attestation` target) naming both fingerprints + the peer.

**CORE INVARIANT:** a foreign-fingerprint vector never enters the local space (nor the live HNSW index). Distinct model ids never collide, so the gate can produce a false MISMATCH (a safe local re-embed) but never a false MATCH (corruption): **degrade, never corrupt.** M-STRONG-TYPES-GUARD.

## Files changed

- `src/embeddings.rs` — `embedding_space_fingerprint` free fn + `Embedder::space_fingerprint` + prefix-scheme consts + `space_fingerprint_2168_tests` unit module.
- `src/handlers/federation_receive.rs` — sqlite gate (`receiver_fp` + fingerprint check).
- `src/handlers/federation_signing_check.rs` — postgres gate (byte-parallel twin).
- `src/federation/mod.rs` — `ShippedEmbedding.model` doc fix (deleted "observability only"; documented the comparison contract).
- `tests/federation_2168_embed_fingerprint.rs` — new suite.
- `tests/federation_1566_embed_ship.rs` — two verbatim-store cases now ship a matching fingerprint (model is now load-bearing).
- `CHANGELOG.md` — `[Unreleased]` Security entry.

## Tests

- `tests/federation_2168_embed_fingerprint.rs` (4/4): foreign-model → deferred re-embed (foreign vector never stored, invariant asserted); matching-model → verbatim; bare-id prose-tolerant match; #1584 norm still applies under a matching fingerprint.
- `src/embeddings.rs::space_fingerprint_2168_tests` (5/5): prose≡bare-id, same-dim-foreign differs, minilm≠nomic, prefix-scheme axis, case/whitespace normalisation.
- `tests/federation_1566_embed_ship.rs` (9/9): unchanged behaviour after the matching-fingerprint update.
- Gates green: `cargo check --all-targets` (default + `sal` + `sal-postgres`), `clippy --all-targets --features sal -D warnings` + default clippy, `fmt`, `check --examples`, `qual_10_module_size_ceiling`, `qual_6_7_legacy_error_type_ceiling`, hardcoded-literal / vendor-literal / docs-vs-ssot gates. No new tool/route/CLI/migration → no SSOT-count move.

> **MANDATORY Fable security audit before merge** — both funnels, CRDT-convergence preserved, no foreign vector path to HNSW, fingerprint derivation matches the local embed path exactly.


alphaonedev:
## Fable pre-merge audit: AUDIT-CLEAN

Full-surface audit of the #2168 fingerprint gate against the North Star bar (fail-closed, degrade-never-corrupt, no data loss, CRDT convergence). Verdict: **AUDIT-CLEAN — approve for merge.** Two non-blocking hardening follow-ups filed (#2176, #2177).

### 1. Both funnels gated identically (cross-backend parity)
`federation_receive::sync_push` (sqlite, ~:1520-1560) and `federation_signing_check::sync_push_via_store` (postgres, ~:372-412) carry byte-parallel gates: dim gate -> fingerprint equality (`receiver_fp == embedding_space_fingerprint(se.model)`) -> #1584 sanitize. The ONLY accepting path is exact fingerprint equality with the receiver's own derived fingerprint. On sqlite, `hnsw_updates.push` happens exclusively in the `Some(clean_shipped)` arm after a successful `db::set_embedding`, so a refused foreign vector structurally cannot reach the live HNSW index — including the quota-refusal short-circuit branch, which only indexes vectors already in `hnsw_updates`.

### 2. Fail-closed on ALL non-match cases
- Foreign fingerprint -> `return None` -> deferred local re-embed. Verified live (test 1).
- Empty/garbage `se.model` -> canonicalises to a fingerprint (`"#none"`, `"<junk>#none"`) that can never equal the receiver's non-empty fingerprint -> deferred re-embed. Structurally verified (single equality accept path); test-pin follow-up filed as #2176.
- Truly ABSENT `model` -> whole-body serde deserialization failure (`ShippedEmbedding.model` is a required field, no default — pre-existing #1566 wire contract; every embed-shipping binary sends it).
- No local embedder (`receiver_fp = None`) -> the dim gate already refused (`receiver_dim = None`).
No permissive/accept-with-WARN branch exists anywhere in the diff. M-STRONG-TYPES-GUARD honoured: false MISMATCH = safe local re-embed; false MATCH requires two distinct model ids canonicalising identically, which the head-token+lowercase parse cannot produce.

### 3. CRDT convergence / no data loss
The gate runs strictly AFTER the memory row applies (`merge_inbound` / `apply_remote_memory` -> `actual_id`/`applied_id`); the `None` arm pushes `(id, embedding_document(title, content))` onto `deferred_embed` — the row lands unconditionally, only the derived vector is regenerated from the durable TEXT under the receiver's own model. Re-embed rides `spawn_deferred_embedding_refresh(_via_store)` OFF the ack path (test 2 pins zero ack-path embeds on match; the pre-existing `slow_embedder_does_not_delay_sync_push_ack` still passes). Verified `applied == 1` on the mismatch path.

### 4. Fingerprint derivation — no false accepts
Receiver: `space_fingerprint()` = `embedding_space_fingerprint(model_description())`. Shipped: `embedding_space_fingerprint(se.model)` where a well-behaved peer ships its own `model_description()`. Same function, byte-identical canonicalisation (head token before " (", trim, ASCII-lowercase; scheme via the #1520 `model_requires_nomic_prefix` predicate, itself case-insensitive-contains so the lowercased id is scheme-stable). Distinct ids never collide. Residual (non-blocking, #2177): the scheme axis is model-implied on both sides, so it cannot discriminate same-id peers with divergent BINARY prefix behaviour — bounded today because #1520 predates #1566 embed-ship; explicit wire field is the v1.x evolution the issue itself anticipated. Note: `se.model` remains peer-CLAIMED — a lying enrolled peer can assert the local model id; that is the #2167/attested-provenance threat, out of #2168's automatic-corruption scope.

### 5. Reuse of the existing deferred path
The `None` arm is the UNCHANGED pre-existing dim-mismatch/no-vector fallback (`deferred_embed` -> `spawn_deferred_embedding_refresh`), both funnels — no divergent new path. The WARN is structured (M-LOG-STRUCTURED, `federation::attestation` target) naming peer + both fingerprints.

### 6. Test validity (falsification performed)
- `tests/federation_2168_embed_fingerprint.rs` 4/4 green; `space_fingerprint_2168_tests` 5/5; `federation_1566_embed_ship` 9/9 (its two verbatim-store cases now correctly ship a matching fingerprint — model is load-bearing).
- **Pre-fix falsification:** reverted ONLY the two funnel files to `origin/release/v1.0.0` and re-ran test 1 — it FAILS exactly as designed (`stored[0] must be the LOCAL re-embed fill, got 0.03608439` = the foreign unit-norm fill stored verbatim). The suite genuinely pins the vulnerability, not the implementation.
- Gates: `cargo fmt --check` clean; `qual_10_module_size_ceiling` + `qual_6_7_legacy_error_type_ceiling` green.

### 7. #2167 independence
The gate consumes only `se.model` + `app.embedder` — zero references to any provenance schema column; migration-free. Merge-order-independent of #2167. Coordination note for the #2167 implementer: stamp the local-space provenance on BOTH vector-landing paths this gate produces (verbatim store under a matching fingerprint, and the deferred local re-embed), and share the scheme vocabulary consts when promoting the wire field (#2177).

### 8. CI
All checks green at HEAD 446f7983: Check (ubuntu/macos/windows), Lint, MSRV, SAL-only gate, Postgres feature gate, Per-Module Coverage Thresholds, Dockerfile, mobile cross-compiles, bench, all 6 script gates.

Filed: #2176 (empty-model-string test pin, non-blocking), #2177 (explicit prefix-scheme wire field, non-blocking v1.x).

Scope: this audit covers the PR #2171 diff and its interaction surface (both receive funnels, HNSW insertion, deferred-re-embed reuse, wire back-compat, #2167 adjacency). It does not re-audit the pre-existing #1584 sanitizer, the outer federation auth stack (#29/#30/#43/#94), or the #2167 recall-side provenance work in flight.

Issue #1692, opened 2026-06-15
[HIGH] HNSW semantic-phase dimension mismatch uncounted — silent wrong cosine after embedder swap (violates #1584 de-silence contract) + seed-embedding finiteness


alphaonedev:
**Source:** v0.7.1 AI-NHI audit (Cluster E, pairs with the CRITICAL NaN panic). **Bucket:** v0.7.1 (#1683).

`src/storage/mod.rs:8466-8556` (HNSW branch) takes `cosine=1.0-hit.distance` (~8492), never checks dimensionality, never bumps `dim_mismatch_count` (threaded at 8457). The H7/#1584 de-silencing fires only on FTS (8965-8970) and linear-scan (8649-8655). `cosine_distance` (hnsw.rs:218-221) zips unequal-length slices → truncated-dot finite distance passes the >0.2 gate → a garbage hit ranks while `dim_mismatch_count` reports 0 after an embedder swap. Separately, **local/seed embeddings bypass non-finite sanitization** before entering the HNSW index (ingress to the CRITICAL panic).

**Proposed fix (~15-25 lines):** gate HNSW hits on `query_embedding.len()==stored_dim` and increment `dim_mismatch_count` identically to the FTS/linear-scan branches; finiteness-sanitize seed/local vectors at ingress.
Labels: bug, recall, honesty-gap, v0.7.1, auto-filed-by-agent

alphaonedev:
Fixed in v0.7.1 (released; tag v0.7.1, merged to main via PR #1702): HNSW semantic branch detects vector-dimension mismatch (cosine_similarity_checked) instead of silent wrong cosine (commit c85c1b58). Gates green at release (default 5613 / sal-postgres 6034 tests, 0 failed; clippy -D pedantic; fmt; cargo audit). Verified per prime directive pm-v3 (verify-before-claiming + recompile-retest).

PR #2175, opened 2026-07-17
[#2167] embedding vector-space provenance (schema v84) — complete S1-S9, both backends

### LINKED ISSUE #2167: [v1.0.0][data-integrity] Silent recall corruption: embedding-space provenance (model+prefix at same dim) is never persisted or checked — config-swap, interrupted reembed, restore, default drift (both backends)

alphaonedev:
Closes #2167 (embedding vector-space provenance, schema v84). **Complete — S1–S9 landed on BOTH backends.** Recall NEVER scores a stored vector from a different embedding space (a same-dim model swap); an incompatible vector never enters unflagged; degrade-never-corrupt; self-heals. Do NOT merge until the Fable re-audit of the new postgres-parity correctness surface clears and CI's per-module coverage sweep confirms the floors.

## Landed guarantee (S1–S9)

- **S1 schema v84** — per-row `embedding_space TEXT` (NULL = legacy/unverified) on `memories` + `archived_memories`, both backends; additive `ALTER ADD COLUMN` + ladder-owned partial index, no rebuild; `CURRENT_SCHEMA_VERSION` 83→84.
- **S2 fingerprint SSOT** — `embeddings::embedding_space_fingerprint` / `Embedder::space_fingerprint` (`<canonical_model_id>#<prefix_scheme>`), deliberately over-distinguishing (DEGRADED-not-WRONG); compiled-default pin.
- **S3 write funnels** — vector + space stamped in ONE statement at every embedding write; federation ACCEPT stamps the sender's CLAIMED space `mint(se.model)`.
- **S4 ★ recall core** — `cosine_similarity_space_checked` (space→dim→score); all 4 sqlite recall sites gate; pg `AND embedding_space = $fp` on every `<=>` site + static site-count pin; HNSW 3-layer boundary; telemetry counters + aggregated WARN.
- **S5 adoption + S6 census** — `adopt_legacy_embedding_space` ([G1] strict-off + [G2] no-differently-stamped guards) + `distinct_embedding_spaces` + boot maintenance in embedder→adopt→census order; heterogeneous corpus → loud WARN; `doctor` census section.
- **S7 reembed knobs** — `--stamp-only` / `--skip-current-space` / `--sleep-ms` / `--max-rows`.
- **S8 restore/migrate heal** — archive funnels carry `embedding_space`; restore classifies against active space (active → intact; foreign/NULL-with-vector → trio NULLed → backfill re-heals).
- **S9 dup-check space gate + pg `<=>` write-pin + fed HNSW insert-gate.**

## Postgres write-side + boot-heal parity — Fable audit blockers #2178 / #2179 / #2180

- **#2178** — `PostgresStore::store_with_embedding` (primary pg create anchor) now stamps `embedding_space` in the SAME statement as the vector on BOTH the INSERT and the `ON CONFLICT DO UPDATE` upsert-merge arm; a same-dim swap-back can never leave a stamp attesting a foreign vector. Space threads through a compile-enforced `store_with_embedding(…, space: Option<&str>)` signature. The dim-migration NULL path clears the stamp WITH the vector (no orphan stamp to falsely trip [G2]). Enumerated every pg writer of `embedding`: only `store_with_embedding` + the dim-migration were unstamped.
- **#2179** — the §5/§6 boot machinery ran against the LOCAL sqlite file even for postgres. Added pg twins `PostgresStore::{adopt_legacy_embedding_space,distinct_embedding_spaces,embedding_space_boot_maintenance}` run against the actual pg corpus at pg `serve` boot (adopt→census, [G1]/[G2] guards preserved EXACTLY; dim gate uses `vector_dims(embedding)` since a legacy row may carry NULL `embedding_dim`). The pg backfill predicate also targets `embedding IS NOT NULL AND embedding_space IS NULL`, so the serve-boot sweep re-embeds [G1]/[G2]-blocked NULL-space rows from durable text (the pg reembed/heal). Closes the silent, unhealable fleet-wide recall outage on postgres.
- **#2180** — the S1 `CORPUS_SCHEMA_VERSION` 83→84 bump regenerated `conformance/manifest.json` (schema_version-only; `corpus_digest` byte-identical, vectors/ untouched) so the per-module coverage sweep evaluates the new #2167 LOC.

## Tests / gates

- `tests/embedding_space_provenance_2167.rs` — **T-INV 13/13** green (`--features sal`).
- `tests/embedding_space_provenance_2167_pg.rs` — pg T-INV twins (`#[ignore]` + sal-postgres): the `<=>` exclusion, the INSERT+upsert stamp (#2178), the §5 adoption no-nuke (#2179), the [G2] multi-space refusal.
- Gates green: `cargo check --all-targets` (default + sal + sal,sal-postgres), `clippy --all-targets` (default + sal + sal-postgres) `-D warnings -D clippy::pedantic`, `fmt`, `check --examples`, `qual_10` (postgres.rs ceiling 27_250→27_600 lockstep) + `qual_6_7`, docs-vs-SSOT / hardcoded-literal / vendor-literal gates.

Data integrity (North Star): the memory TEXT is the durable source of truth; `embedding_space` is a disposable, regenerable-from-text provenance stamp. Every path fails toward DEGRADED recall, never WRONG — on both backends.


alphaonedev:
## Fable pre-merge audit: AUDIT-BLOCKED (3 blocking findings — #2178, #2179, #2180)

Line-by-line audit of the correctness core at HEAD 155c974b (S4 recall gate, S5 adoption, S8 restore heal, T-INV, both backends, CI-red disposition). Scope: `cosine_similarity_space_checked` + all four sqlite recall sites + caller enumeration of the whole `recall_hybrid*` family; the pg `<=>` predicate on every cosine site; the §5 G1/G2 adoption guards against the safety proof; all archive/restore funnels both backends; every remaining cosine consumer in the crate; the v84 migrations; the 13-test T-INV suite (run locally, sal); the failing CI jobs (logs pulled and reproduced locally).

### Verified correct (the core holds)
- **S4 sqlite** — comparator order space→dim→score, exact whole-token equality, `None`→`UnverifiedSpace` fails closed. FTS site: cosine 0.0 + count, keyword score kept (degraded-not-invisible). Linear scan: skip + count (`embedding_space` appended at index 19, `row.get(17)` pin intact). HNSW recompute: `_ = cosine;` — `hit.distance` is never trusted, and the #1692 None-arm stale-distance fallback is DELETED (missing/undecodable row-side vector → excluded + `unverified_space_count`). Aggregated WARN + both telemetry counters reconcile. Caller enumeration complete: HTTP (both branches), MCP, CLI, and the SAL sqlite adapter all pass `Some(fingerprint)` whenever a query embedding exists; the three `active_embedding_space: None` Filter sites are list-only (no embedding).
- **S5 adoption** — G2 (`EXISTS` differently-stamped) refuses BEFORE any stamping including the archived twin; G1 strict-off enforced in `embedding_space_boot_maintenance`; `WHERE embedding_space IS NULL` = monotone (never overwrites a stamp), idempotent, marker-free; boot order embedder→seed-active→adopt→census→index-seed at both sqlite boot sites. The proof properties (no-nuke, post-swap catch, monotone detectability, residual window) hold in code **on sqlite**.
- **S8** — 8 sqlite + 7 pg archive INSERT...SELECT funnels carry `embedding_space`; restore CASE NULLs the whole trio coherently on both backends (space==active keeps the vector intact; foreign OR NULL-with-vector → trio NULLed → `list_unembedded` backfill re-embeds from durable text; None-active keeps stamped, drops unverifiable). No excluded-forever, no scored-wrong, for RESTORED rows on both backends.
- **HNSW 3 layers** — boot seed SQL-filtered (`get_all_embeddings(conn, active)`, all callers thread it incl. the CLI-local index), federation insert-gates on claimed==active at BOTH `hnsw_updates` loops, row-side recompute post-filter. Federation ACCEPT stamps `mint(se.model)` claimed space; local re-embed stamps the receiver's active space.
- **T-INV** — 13/13 green locally (`--features sal`). T-INV-1 is non-vacuous: mixed active/foreign/NULL/cross-dim corpus, EXACT result-set assertion + exact 2/1/1 counter reconciliation — reverting the S4 gate makes the same-dim foreign rows (cosine 1.0) reappear and the assertion fail. pg `<=>` static pin (3 sites, forces gating of any future cosine query) + the `#[ignore]` live-pg T-INV twin both present.
- Migrations v84 clean both backends (additive, probe-guarded, no rebuild, no migration-time stamping); the pg ladder correctly re-pins the v83 arm to record literal 83. Fingerprint fold is case-insensitive (`eq_ignore_ascii_case`) and over-distinguishes in the safe direction. tool-count red: `conclusion=cancelled` push-event concurrency dup; the pull_request variant passed — ignorable as briefed.

### BLOCKING
1. **#2178 — pg `store_with_embedding` writes the vector UNSTAMPED.** The primary postgres create anchor (handlers/create.rs:1000) inserts `embedding` with no `embedding_space`, and its `DO UPDATE` arm replaces the vector via `COALESCE(EXCLUDED.embedding, …)` without touching the stamp — a stale stamp can end up attesting a foreign vector (lying stamp → scoreable as verified after a swap-back = a cross-space score path). Violates this PR's own §2 same-statement rule. The PR-body "SAFE gap, self-heals via S5/reembed" claim fails because neither S5 nor reembed exists on pg (finding 2).
2. **#2179 — no pg adoption/census/heal at all.** §5/§6/reembed are sqlite-only; the pg boot runs them against the LOCAL sqlite file. After the v84 upgrade every legacy pg embedded row is `NULL`-space → permanently excluded by `AND embedding_space = \$10` with no WARN, no census, no heal. Fail-closed (never wrong) but it breaks the §5 no-nuke proof obligation on an entire backend: a silent, unhealable, fleet-wide semantic-recall outage on postgres.
3. **#2180 — the Per-Module Coverage red is PR-INTRODUCED, not pre-existing.** Correcting the completion report: S1 (799eb4ef) bumped `CORPUS_SCHEMA_VERSION` 83→84 in `tests/conformance_corpus.rs` without regenerating `conformance/manifest.json` (still `\"schema_version\": 83`; `corpus_digest` byte-identical — zero encoder drift). Reproduced deterministically on the branch locally; cannot fail at base. Because the suite aborts there, the per-module floors were never evaluated for this PR's ~1400 new LOC. Fix is a one-line golden regen + commit, then confirm the sweep actually passes.

### Non-blocking (filed)
- **#2181** — residual ungated cosines: proactive-conflict scan pool, consolidate/curator stored-vs-stored clustering (advisory/merge decisions, Jaccard-bounded, reversible — but posture-inconsistent with the S9 dup-check rationale).
- **#2182** — spec §10 test gaps: HNSW/poisoned-index T-INV, T-hnsw-boot, T-INV-3, T-strict (code line-verified; tests should pin it).
- The PR title/body and the CHANGELOG lead line still say "PARTIAL — S1-S3 / IN PROGRESS"; refresh to match the landed S1-S9 before merge.

**Do NOT merge** until #2178, #2179, #2180 are resolved on this branch; re-audit of the pg-parity commits (funnel stamp + adoption twin) is required since they are new correctness-core surface.

PR #2187, opened 2026-07-18
fix(#2181,#2182,#2183): embedding-space residual cosine gate + pg FOREIGN heal

alphaonedev:
Addresses the THREE non-blocking follow-ups from the #2167 Fable pre-merge audit (base `release/v1.0.0` @ 924965c1). Data-integrity-adjacent: recall NEVER scores a cross-space vector, and no destructive MERGE / conflict advisory is taken on a meaningless cross-space cosine — degrade, never corrupt; both backends.

Closes #2181
Closes #2182
Closes #2183

## #2181 — residual ungated cosine sites gated (both backends)
- **proactive-conflict scan pool** (`src/storage/mod.rs::proactive_conflict_check`): the recency scan-fallback SELECT gains `AND (?N IS NULL OR embedding_space = ?N)` on `embeddings::active_embedding_space()` (the ANN-routed `_candidates` path is already effectively gated — `get_all_embeddings` only seeds active-space vectors). Prevents a FALSE 409 conflict advisory on a legitimate write against a foreign-space same-dim row.
- **consolidation clustering** — both the live autonomy `find_consolidation_clusters` (sqlite) and the curator `ConsolidationPass` / `ConsolidationClustering::cluster_memories` (SAL, both backends) now fetch each candidate's `embedding_space` alongside its vector (new `db::get_embedding_with_space` + SAL `MemoryStore::get_embedding_with_space`, sqlite + postgres overrides) and count the pairwise cosine ONLY when both stored vectors carry the SAME non-NULL space. Extends the #1774 missing-embedding-blocks-merge precedent to mismatched-space-blocks-merge.

## #2182 — spec §10 test-gap closures (all NON-VACUOUS)
In `tests/embedding_space_provenance_2167.rs`:
- **T-INV via a POISONED HNSW index** — a foreign-space vector + a vanished-vector row force-inserted into the graph are NEVER scored on `hit.distance` (the §3.3 layer-3 row-side post-filter + the deleted #1692 None-arm exclude + count them).
- **T-hnsw-boot** — `db::get_all_embeddings` seeds ONLY active-space rows.
- **T-INV-3** — a reembed killed at an arbitrary batch boundary leaves no vector-without-space and no scoreable cross-space row (the per-row single-statement re-stamp invariant).
- **T-strict** — `AI_MEMORY_REQUIRE_EMBED_MODEL_MATCH` [G1] disables boot adoption end-to-end through recall.

[...]

Issue #2183, opened 2026-07-17
[#2167 Fable audit] non-blocking: pg FOREIGN-stamped rows have no heal path (reembed CLI is sqlite-only) and the pg census WARN names it anyway


alphaonedev:
Residual from the PR #2175 re-audit of the #2179 fix (commit b08b62a5). The pg §5/§6/heal parity is correct for the legacy NULL-space corpus: adoption ([G1]/[G2] preserved byte-for-byte, `vector_dims(embedding)` dim gate), census WARN, and the widened `list_unembedded` predicate (`embedding IS NULL OR (embedding IS NOT NULL AND embedding_space IS NULL)`) which re-embeds [G1]/[G2]-blocked NULL-space rows from durable text at every serve boot.

Remaining gap (post-SWAP scenario only, not the upgrade scenario): rows stamped with a FOREIGN non-NULL space (after a same-dim model swap, exactly the state [G2] correctly refuses to adopt) are excluded from semantic recall on postgres with no heal path — `ai-memory reembed` (src/cli/commands/reembed.rs) is sqlite-only, and the pg backfill predicate deliberately targets only NULL-space rows. Worse, the pg census WARN (`PostgresStore::embedding_space_boot_maintenance`) tells the operator to "heal foreign-stamped rows with `ai-memory reembed`" — a command that does not operate on the pg corpus.

Fail-closed and loudly WARNed each boot (correct direction, no corruption); text is durable so nothing is lost. But at fleet scale a pg model swap currently has no sanctioned recovery. Fix options: (a) pg support in the reembed CLI (SAL-routed), or (b) an opt-in knob extending the pg backfill predicate to `embedding_space IS DISTINCT FROM $active` (paced), plus (either way) correcting the WARN text to name a command that actually works on postgres.

alphaonedev:
Fixed + merged via PR #2187. BOTH halves: (a) pg census/[G2]/strict WARNs no longer name the sqlite-only reembed — they state pg reality (NULL-space self-heals on backfill; FOREIGN via the knob); (b) new opt-in AI_MEMORY_PG_HEAL_FOREIGN_SPACE (default OFF, byte-identical) widens list_unembedded to re-derive FOREIGN-stamped rows from durable text + re-stamp active space. Fable-confirmed: re-derives from durable text (no data loss), single-statement re-stamp (no lying stamp), monotone + paced. Verified per pm-v3.

alphaonedev:
[...]

Issue #2182, opened 2026-07-17
[#2167 Fable audit] non-blocking: spec §10 test gaps — HNSW/poisoned-index T-INV, T-INV-3 kill-mid-reembed, T-hnsw-boot, T-strict


alphaonedev:
From the PR #2175 audit: `tests/embedding_space_provenance_2167.rs` (13/13 green, sal) proves T-INV-1 on the LINEAR-SCAN path (mixed 3-fingerprint corpus, exact result-set + counter reconciliation — non-vacuous), degraded-not-invisible, adoption a/d/e (G1/G2), T-funnel, T-restore x4, and the static pg `<=>` pin. Missing vs spec §10 (mem ac1d83b4):

- T-INV via the HNSW-recompute path incl. a deliberately POISONED index (foreign vector force-inserted) proving the §3.3 layer-3 row-side post-filter and the deleted #1692 None-arm (index hit whose row vector is gone → excluded, never scored on hit.distance). The code was line-verified correct in the audit (`_ = cosine;` — distance never trusted), but no test pins it against regression.
- T-hnsw-boot: `get_all_embeddings(conn, active_space)` seeds ONLY active-space rows.
- T-INV-3: reembed killed at an arbitrary batch boundary → every intermediate state has only consistently-stamped scorable rows (structurally safe today: per-row single-statement re-stamp; pin it).
- T-strict: AI_MEMORY_REQUIRE_EMBED_MODEL_MATCH posture (G1 covered indirectly via adoption_e; the recall-side strict/keyword-degrade behavior is untested — noting the §6 strict recall-withhold itself is still an open S6 residual).

alphaonedev:
Fixed + merged via PR #2187. Added the spec §10 regression tests (HNSW poisoned-index T-INV, T-hnsw-boot, T-INV-3 kill-mid-reembed, T-strict) + 2 clusterer pins — all non-vacuous (fail on pre-#2181 code), serialized via STRICT_FLAG_GUARD. Fable-confirmed non-vacuous. Verified per pm-v3.

alphaonedev:
Reopening — PR #2187 did NOT merge (conflict with release/v1.0.0 after #2184/#2185/#2186 landed). Resolving the conflict; will re-close on verified merge.

alphaonedev:
[...]

Issue #2181, opened 2026-07-17
[#2167 Fable audit] non-blocking: residual ungated cosine sites — proactive-conflict scan pool + consolidate/curator stored-vs-stored clustering


alphaonedev:
Non-blocking residuals from the PR #2175 line audit (recall + dup-check are fully gated; these are the remaining cosine consumers that can still compare cross-space pairs):

1. `proactive_conflict_check` bounded recency-scan pool (src/storage/mod.rs ~8194): SELECT has no `embedding_space` predicate, so `proactive_conflict_verdict` (~8393) can cosine a live query vector against a foreign-space same-dim stored vector. Bounded by the dim-skip + the #1579 Jaccard corroboration floor, and the worst case is a FALSE 409 conflict advisory on a write (degrade/availability, no corruption, no loss) — but posture-inconsistent with the S9 rationale that gated `check_duplicate` for exactly this reason. The ANN-routed path is effectively gated already (the index only holds active-space vectors post-#2167), so only the scan fallback needs `AND (?n IS NULL OR embedding_space = ?n)` on `embeddings::active_embedding_space()`.
2. Consolidate clustering (src/autonomy.rs ~380, `db::get_embedding` pairs) and curator clustering (src/curator/cluster.rs ~190): stored-vs-stored pairwise cosines with no space axis. A mixed-space corpus can produce meaningless similarities feeding a MERGE decision. Bounded: the Jaccard lexical prefilter must also pass, and consolidation archives its sources (reversible via restore — which now heals spaces). Fix shape: fetch `embedding_space` alongside the vectors and require equal non-NULL spaces for a cosine to count (the #1774 missing-embedding-blocks-merge precedent extends naturally to mismatched-space-blocks-merge).

Neither can produce a wrong RECALL score; filing for posture completeness (degrade-never-corrupt applies to merge/advisory decisions too).

alphaonedev:
Fixed + merged via PR #2187 (release/v1.0.0). Residual ungated cosine sites now embedding-space-gated on BOTH backends: proactive_conflict_check scan-fallback + both consolidation clusterers (autonomy sqlite via db::get_embedding_with_space, curator SAL via the new MemoryStore::get_embedding_with_space, fail-closed default space=None blocks the merge). Fable audit AUDIT-CLEAN (both-clusterer no-cross-space-merge proof-in-code, index alignment verified); Postgres gate + coverage green. Verified per pm-v3.

[...]

PR #2193, opened 2026-07-18
fix(#2188,#2189,#2190): embedding-space cosine-gating residuals from the #2187 Fable audit

### LINKED ISSUE #2188: [#2187 Fable audit] defense-in-depth: proactive_conflict_check_candidates verdict pool lacks the row-side embedding_space predicate
Residual from the PR #2187 pre-merge line audit of #2181 (degrade-class, does NOT block #2187 — the window pre-exists the PR and #2181 explicitly scoped the fix to the scan fallback).

**Gap.** `src/storage/mod.rs::proactive_conflict_check_candidates` (~8320) verifies ANN candidate ids with a point-lookup SELECT (`id IN (...) AND embedding IS NOT NULL AND live AND namespace = ?`) and `proactive_conflict_verdict` recomputes EXACT cosine from the freshly-read row-side blob — with NO `embedding_space` predicate. The #2181 rationale ("the index only holds active-space vectors post-#2167") gates index MEMBERSHIP (boot seed `get_all_embeddings` is active-only; federation §3.3 layer-2 insert-gate skips foreign claimed spaces; local live-embedder inserts are active by construction), but the verdict re-reads the ROW, and the row's space can drift from the index snapshot.

**Reachable window (narrow, operational).** A concurrent out-of-process `ai-memory reembed` (sqlite CLI) re-stamps rows to a NEW space while a live daemon still holds the OLD process-active fingerprint + OLD index vectors. A candidate id then resolves to a freshly re-embedded NEW-space row-side vector, cosined against the daemon's OLD-space query — a meaningless same-dim cross-space cosine that can surface a FALSE 409 conflict advisory on a legitimate write. Transient until daemon restart; advisory-only (no merge, no recall score, no corruption/loss). Note the scan fallback, `check_duplicate`, recall layer-3, and both clusterers all read the row-side space and are immune to this exact window — the candidates verdict pool is the one remaining row-side cosine consumer without the axis.

**Proposed fix (~4 lines + 1 test).** Add the same nullable gate the #2181 scan fallback got: `AND (?N IS NULL OR embedding_space = ?N)` on `embeddings::active_embedding_space()` to the `IN (...)` SELECT in `proactive_conflict_check_candidates` (src/storage/mod.rs:8333-8341), + a poison-pin test (foreign-space row id force-fed as a candidate → never scored). Posture completeness: S9/#2181 discipline applied uniformly to every cosine consumer.

Found by the Fable pre-merge line audit of PR #2187 (b45813d8).

### LINKED ISSUE #2189: [#2187 Fable audit] test-hygiene: STRICT_FLAG_GUARD serializes flip-vs-flip but not flip-vs-read in embedding_space_provenance_2167
Residual from the PR #2187 pre-merge line audit of #2182 (test-flake hazard only; no product-code defect; does NOT block #2187 — the hazard pre-exists the PR, which strictly improved it by adding the mutex for the flip-vs-flip case).

**Gap.** `tests/embedding_space_provenance_2167.rs` `STRICT_FLAG_GUARD` (line 26) is taken by BOTH tests that flip the process-global strict flag (`adoption_e_g1_strict_disables_adoption` line 372, `t_strict_disables_adoption_null_row_excluded_from_recall_2182` line 810) — flip-vs-flip is correctly serialized. But the NON-flipping tests that call `db::embedding_space_boot_maintenance` / recall (adoption a/d, T-restore x4, T-INV-1, T-funnel, ...) do not take the guard, so under the default parallel test harness one of them can observe another test's transient `Some(true)` strict window → adoption unexpectedly skipped → spurious assertion failure (an intermittent test is a real bug, per the #1724 lesson).

**Proposed fix (small).** Either (a) have every test that calls boot maintenance/recall take a shared read of the guard (make it an `RwLock<()>`: flippers take `write`, readers take `read`), or (b) wrap the flip in an RAII guard type whose Drop restores `None` and have all boot-maintenance tests acquire the same mutex. Option (a) keeps parallelism for the non-flipping majority. ~15 lines, test-file only.

Found by the Fable pre-merge line audit of PR #2187 (b45813d8).

### LINKED ISSUE #2190: [#2187 Fable audit] test-gap: AI_MEMORY_PG_HEAL_FOREIGN_SPACE widened list_unembedded arm has no live-pg test
Residual from the PR #2187 pre-merge line audit of #2183(b) (test-gap only; the widened SQL was line-verified correct in the audit — binds `$1`=cap / `$2`=active_fp, predicate `embedding IS NULL OR (embedding IS NOT NULL AND embedding_space IS NULL) OR (embedding IS NOT NULL AND embedding_space IS NOT NULL AND embedding_space <> $2)` — and default-OFF is byte-identical to the pre-PR SELECT, which existing pg tests cover; does NOT block #2187).

**Gap.** PR #2187 lands the opt-in `AI_MEMORY_PG_HEAL_FOREIGN_SPACE` knob widening `PostgresStore::list_unembedded` (src/store/postgres.rs ~14734), but `tests/embedding_space_provenance_2167_pg.rs` gained no twin: the knob-ON arm (foreign-stamped row returned by the scan; NULL-space + NULL-embedding rows still returned; active-space rows NOT returned; unseeded-fingerprint never widens) has zero coverage on the live-pg gate. An untested heal path can silently regress until an operator flips the knob at fleet scale. Also uncovered: the new `PostgresStore::get_embedding_with_space` override (real per-row space; NULL-embedding collapses to None).

**Proposed fix (~60 lines).** Add two `#[ignore]` sal-postgres tests to `tests/embedding_space_provenance_2167_pg.rs` (run via `--include-ignored` in the Postgres gate, the #1799 pattern):
1. `pg_heal_foreign_space_knob_widens_scan_2183` — seed active + foreign + NULL-space + NULL-embedding rows; knob OFF → foreign row absent from `list_unembedded`; knob ON + active fingerprint seeded → foreign row present, active row absent; end with a `set_embeddings_batch` re-stamp and assert the row leaves the scan (monotone).
2. `pg_get_embedding_with_space_returns_real_space_2181` — stamped row returns `Some((vec, Some(space)))`; NULL-space row returns `Some((vec, None))`; NULL-embedding row returns `None`.
Use the env-serialization guard pattern for the process-global env var.

Found by the Fable pre-merge line audit of PR #2187 (b45813d8).

alphaonedev:
Addresses the THREE non-blocking residuals the #2187 Fable pre-merge line audit filed against the just-merged #2167/#2181/#2182/#2183 embedding-space cosine-gating. North-star: recall/clustering NEVER scores a cross-space vector; degrade-never-corrupt; both backends where applicable.

## #2188 — defense-in-depth: candidate verdict pool gains the row-side `embedding_space` gate
`storage::proactive_conflict_check_candidates` (the ANN-routed candidate verdict pool) re-reads each row's blob and recomputes EXACT cosine in `proactive_conflict_verdict`, making it a row-side cosine consumer exactly like the #2181 scan fallback, `check_duplicate`, and both clusterers — and it was the LAST such consumer without the row-side `embedding_space` predicate. It now carries the same nullable `AND (?N IS NULL OR embedding_space = ?N)` gate on `embeddings::active_embedding_space()`, closing the false-409-advisory window a concurrent out-of-process `ai-memory reembed`'s stale-active fingerprint opens. Advisory-only, no corruption/loss (pre-existing window), but posture-complete now.

**Sqlite-only, noted honestly:** proactive-conflict has no postgres twin — it runs on the rusqlite `Connection` in `handlers/create.rs`; the postgres store never routes through it.

Test: `embedding_space_2188_candidates_path_excludes_foreign_space_row` in `tests/proactive_conflict.rs` (a FOREIGN-space candidate id is never scored; the same row re-stamped into the ACTIVE space DOES still 409 — proving the gate excludes on space, not the row).

## #2189 — test-hygiene: `STRICT_FLAG_GUARD` flip-vs-read hazard closed with an `RwLock`
The #2182 `Mutex` serialised flip-vs-flip but still let a non-flipping reader observe a torn `Some(true)` strict window (an intermittent test is a real bug, #1724). `STRICT_FLAG_GUARD` is now an `RwLock<()>`: flippers hold the EXCLUSIVE `write` lock for the whole strict window via the panic-safe RAII `StrictFlagSession` (restores the flag to `None` on drop even on a panicking unwind), and every non-flipping reader holds a SHARED `read` lock. Readers still parallelise with each other; only the flipper serialises against them. Test-only.

Regression pin: `strict_flag_session_restores_flag_on_drop_even_on_panic_2189` (injects a panic mid-flip and asserts the flag is restored to `None` on unwind, with poison-tolerant recovery).

## #2190 — test-gap: live-pg twins for the #2183(b) widened `list_unembedded` arm + the #2181 pg `get_embedding_with_space` override
Two twins in `tests/embedding_space_provenance_2167_pg.rs`:
- `pg_heal_foreign_space_knob_widens_scan_2183` — knob OFF → foreign row NOT swept; knob ON + UNSEEDED fingerprint → still not widened; knob ON + seeded active fingerprint → foreign row swept, active row NOT swept, NULL-space + NULL-embedding always swept; a `set_embeddings_batch` re-stamp makes the healed row LEAVE the scan (monotone → the sweep terminates).
- `pg_get_embedding_with_space_returns_real_space_2181` — stamped row → `Some((vec, Some(space)))`; NULL-space legacy row → `Some((vec, None))`; NULL-embedding row → `None`.

Both follow this file's coverage-correct **SKIP-IF-URL** convention (deliberately NOT `#[ignore]`, per the module header's rationale: the coverage job does not pass `--include-ignored`, so an `#[ignore]` twin would show 0% coverage). They are picked up by the CI Postgres feature gate AND the coverage job (both set `AI_MEMORY_TEST_POSTGRES_URL`), and self-skip cleanly with no live pg. Uses the #1799 env-serialisation guard for the two process-globals (`AI_MEMORY_PG_HEAL_FOREIGN_SPACE` + `active_embedding_space()`) the widened arm reads.

## Gates
- `cargo check --all-targets` (default + `--features sal` + `--features sal-postgres`): clean
- `cargo clippy --all-targets -- -D warnings -D clippy::pedantic` (default + sal + sal-postgres): clean
- `cargo fmt --all --check`: clean
- `cargo test --test proactive_conflict`: 15/15
- `cargo test --test embedding_space_provenance_2167`: 18/18 (incl. #2189 regression)
- `cargo test --test store_legacy_classifier_paths proactive`: 7/7 (no MCP-path regression)
- `cargo test --test embedding_space_provenance_2167_pg` (no live pg): 8/8 self-skip — the two new twins are recognised, not silently absent; they EXECUTE under the CI Postgres gate
- `qual_10` (storage/mod.rs 25765 < 25850 ceiling, no bump) + `qual_6_7` (no new legacy `Result<_,String>`): pass

Also corrects a stale doc-header line in the sqlite provenance test that mis-described the pg twin as `#[ignore]`-gated. CHANGELOG `[Unreleased]` entry added.

Closes #2188
Closes #2189
Closes #2190

https://claude.ai/code/session_0182xSSVeBxMzhqGaBgH4MVz


alphaonedev:
## Fable pre-merge LINE-AUDIT — VERDICT: AUDIT-CLEAN

Adversarial line-audit of `2e58dc2d` against #2188/#2189/#2190, both backends, North-Star lens (recall/clustering NEVER scores a cross-space vector; fail-closed; degrade-never-corrupt).

### 1. #2188 — candidates-pool `embedding_space` gate: FAIL-CLOSED PROOF ✅
- **Gate parity**: `proactive_conflict_check_candidates` (src/storage/mod.rs:8357) now carries the byte-same nullable predicate as the #2181 scan fallback (mod.rs:8214): `AND (?N IS NULL OR embedding_space = ?N)` on `embeddings::active_embedding_space()`. Every row-side cosine consumer (scan fallback, candidates pool, `check_duplicate`, both clusterers) is now uniformly gated.
- **Fail-closed semantics verified**: with an active space seeded, a FOREIGN-space row is excluded (`'active' = 'foreign'` → false) and a NULL-space row is ALSO excluded (`NULL = 'active'` → NULL → not true) — the gate is applied in SQL BEFORE `proactive_conflict_verdict` ever decodes a blob, so a cross-space vector can never be scored, and the worst case is a MISSED advisory (safe direction: a missed conflict allows a write; never a wrong 409). With NO active space seeded, `?N IS NULL` short-circuits to the documented legacy dim-only posture — identical to the scan fallback and to `t_inv_1_none_active_skips_gate_legacy_dim_only`.
- **NULL binds SQL NULL, not `""`**: the heterogeneous bind vec materialises `rusqlite::types::Value`; `active_space.map_or(Value::Null, Value::Text)` (mod.rs:8373). Param numbering audited: plain `?` IN-list occupies 1..len, explicit `?{len+1..len+3}`; `?{p_space}` used twice resolves to ONE parameter index; statement param count = binds.len(). Correct.
- **Positive control is real**: `embedding_space_2188_candidates_path_excludes_foreign_space_row` force-feeds the foreign row as an ANN candidate (verdict `None`), then re-stamps the SAME row into the active space and proves it DOES 409 — the gate excludes on space, not on any incidental row property. It also seeds the active space to the same "test-space" fingerprint every sibling row carries and restores `None` BEFORE asserting, so a failing assert can't leak the global into a parallel test.
- **Sqlite-only claim is honest**: codegraph callers = exactly one production caller, `proactive_conflict_check_with_index` (mod.rs:8297), invoked from `src/handlers/create.rs:1451-1452` on the rusqlite `Connection` (`&lock.0`). No `src/store/postgres.rs` twin exists; the pg store never routes through proactive-conflict.
- Local rerun: `cargo test --test proactive_conflict` → **15/15**.

### 2. #2189 — STRICT_FLAG_GUARD Mutex→RwLock: TORN-FLAG HAZARD CLOSED ✅
- All 18 test fns in `tests/embedding_space_provenance_2167.rs` enumerated: the 2 flippers (`adoption_e_g1_strict_disables_adoption`, `t_strict_disables_adoption_null_row_excluded_from_recall_2182`) hold `StrictFlagSession` (EXCLUSIVE write lock) for their entire strict window incl. the two-phase ON→OFF half; all 15 flag-influenced readers hold `strict_flag_read()` shared guards bound for the whole test body; the 18th (`pg_cosine_query_sites` static source pin) touches no flag. No reader can observe a half-flipped flag.
- **RAII restore is panic-safe and ordered correctly**: `Drop for StrictFlagSession` runs `set_strict_embed_model_match_for_test(None)` BEFORE the `_guard` field (the write guard) drops — Rust drop order (Drop::drop body, then fields) means the flag is restored while STILL holding the exclusive lock, so even the unwind window leaks nothing. Poison from a panicking flipper is recovered via `PoisonError::into_inner` on both acquire paths (the `()` payload carries no torn state; the flag itself was restored by the unwinding session's Drop).
- **Regression pin verified**: `strict_flag_session_restores_flag_on_drop_even_on_panic_2189` injects a mid-flip panic under `catch_unwind`, asserts the panic propagates, then asserts the flag reads OFF through a fresh (poison-tolerant) reader. Passes locally.
- **No deadlock**: lock order ACTIVE_SPACE → STRICT-read at all 4 dual-lock sites; flippers never take ACTIVE_SPACE_TEST_LOCK; no cycle.
- Local rerun: `cargo test --test embedding_space_provenance_2167` → **18/18**.

### 3. #2190 — pg twins ACTUALLY EXECUTE under the CI Postgres gate ✅ (the critical check)
- **Log-level proof, not convention-level**: the Postgres feature gate job for this PR head (run 29649984775, job 88094571723) shows `Running tests/embedding_space_provenance_2167_pg.rs` at 15:52:24 and `test pg_get_embedding_with_space_returns_real_space_2181 ... ok` + `test pg_heal_foreign_space_knob_widens_scan_2183 ... ok` at 15:52:25.
- The self-skip branch cannot have fired: the exact step that ran them sets `AI_MEMORY_TEST_POSTGRES_URL=postgres://ai_memory:ai_memory_ci@localhost:5432/ai_memory_ci` (ci.yml:697) inside a job whose readiness step logged `##[notice]postgres is up` at 15:28:57 in this same run — so `live_pg()` returned `Some` and the assertions executed against live pg. The coverage job sets the same var (coverage.yml:53), so the SKIP-IF-URL (not-`#[ignore]`) rationale in the module header is coverage-correct as claimed.
- **Heal semantics**: the knob-ON widened `list_unembedded` arm is scan-widening ONLY (no stamp of its own); the healed row leaves the scan after a `set_embeddings_batch` re-stamp (Phase D monotone assert → the sweep terminates), and the backfill re-derives vectors from the durable (id, title, content) text — no lying stamp. Knob-OFF and knob-ON-unseeded phases pin byte-identical legacy scope. The #1799 env-serialisation lock guards the two process-globals, and the pg gate runs `-- --test-threads=1`, so the `unsafe` env mutation is race-free in CI.

### 4. CI terminal state ✅
`gh pr checks 2193`: **25/25 pass** — including Postgres feature gate (35m7s), Per-Module Coverage Thresholds (36m53s), Check on all 3 OSes, Lint, MSRV, SAL-only gate, all 6 script gates. Terminal green, `mergeStateStatus: CLEAN`.

### Non-blocking nit (filed, does not gate merge)
#2194 — the OLD #2181 comment inside `proactive_conflict_check` (mod.rs:8204-8207) still claims the candidates path is "already effectively gated" via index membership — the exact rationale this PR's own #2188 comment rebuts. Comment-only drift; 3-line follow-up.

No non-fail-closed gate, no torn-flag reader hazard, no silently-skipped pg test. Merge-ready from this audit's standpoint.

PR #2065, opened 2026-07-15
feat(#1830): TRACT G16 durability-model disclosure anchor (no erasure cold tier)

alphaonedev:
Advances #1830 (TRACT-gap G16 — no (n,k) erasure-coded no-primary cold tier). **Does NOT close #1830** — see "Governance" below.

## Scope — the v1.0.0-correct slice (NOT the subsystem)
Codegraph: ZERO erasure/shard/Reed-Solomon code exists; durability today is local persistence + opt-in quorum replication. The full erasure subsystem is freeze-hostile (a T4 shard byte-format + no-primary coordination) and needs an **operator-gated** erasure dependency — it cannot ship autonomously under GA freeze. What ships:

- **`src/durability.rs`** — `DurabilityModel` (`#[non_exhaustive]`) pinning the three postures the substrate *actually* provides, with **deliberately no erasure variant**. `resolve_durability_model` computes the live posture from REAL config (`synchronous` level, `quorum_writes`, peers) — a genuine consumer, not a self-referential bool. The absent variant IS the machine-checked gap: wildcard-free exhaustive matches (`label`, `is_multi_node`) hard-break the build if an erasure variant is added, forcing conscious wiring + a §9 ledger update.
- **contract §9** — the HONEST durability taxonomy (default = local single-node SQLite / WAL / `synchronous=NORMAL`; `FULL`/asi-hard = power-loss-safe; opt-in **W-of-N quorum** federation, *not* full-copy, *not* wired into the store write path). Records the construction-independent v1.x acceptance invariant: **any k-of-n shards reconstruct**.

## What does NOT ship (by design)
**No erasure codec — not compiled, not test-only.** A hand-rolled or self-generated GF(2⁸) reference construction would freeze a shard byte-format the eventual audited crate won't match, ship error-prone finite-field math in the GA binary, and route around the operator dependency gate.

## Governance (why #1830 stays open)
Per the defer-rots vote finding: #1830 is on the **locked v1.0.0 set** and is **not** silently re-milestoned to v1.x (that would be the banned rot-defer). This PR ships the v1.0.0-shippable disclosure anchor; the erasure **subsystem** (codec + shard placement + no-primary coordination + reconstruction + repair) is genuinely BLOCKED on an operator **dependency-authorization** decision — filed as **#2064** (reed-solomon crate vs hand-roll; a Cargo dep add is operator-authority under sole-authority). That escalation is what makes this honest-blocked, not a dodge.

[...]

Issue #1873, opened 2026-07-03
[LOW][security][S5-governance-erasure-audit] #1850 residual — audit-trail anchors compare sequence only: same-length suffix rewrite on unsigned daemons undetected; sub-interval tail window


alphaonedev:
**Severity:** low  **Lane:** S5-governance-erasure-audit  **Class:** CWE-354  **Parent:** #1850 (pre-GA triage, epic #1853)

### Context
#1850's off-table watermark anchor (v0.8.1, default-on via `init_forensic_audit`) and the G5b dual-chain audit-head witness (#1822, opt-in key enrolment + K2 require-mode, now also force-emitted on daemon graceful shutdown) detect `signed_events` tail truncation that reaches at/below the last anchored head. Two residuals remain (assessed in the #1850 triage comment, 2026-07-02):

### Residual 1 — sequence-only anchor comparison misses same-length suffix rewrite
Both verifier verdicts compare **sequence only**:
- `TruncationCheck` (`src/signed_events.rs`): `Detected` iff `db_head < anchored_head`;
- `compute_witness_verdict` (`src/signed_events.rs`): `Detected` iff `witnessed head > surviving head` per chain.

Yet both anchors already RECORD the canonical head hash (`head_canonical_hash` in the forensic watermark payload; `signed_head_hash` / `revisions_head_hash` in the witness `resolution`). On an unsigned daemon (no enrolled daemon key → `verify_chain` verifier `None`), an attacker with DB write access can rewrite the whole suffix with recomputed `prev_hash` at the SAME length and both checks read clean.

**Proposed v1.0 fix (bounded, verifier-side only):** when the surviving row at the anchored sequence exists, recompute its `canonical_chain_bytes` hash and compare against the anchored/witnessed hash; mismatch → a new hash-mismatch `Detected` arm. No new high-water-mark scheme, no schema change. Requires the K3 postgres `verify_audit_trail` twin, CLI/report surface, and tests (`tests/audit_truncation_anchor_1850.rs` + `tests/audit_witness_truncation_1822.rs` extensions).

### Residual 2 — sub-interval tail window
[...]

Issue #1830, opened 2026-06-28
[TRACT-gap G16] No (n,k) erasure-coded no-primary cold tier


alphaonedev:
## Gap G16 — TRACT §26 canonical-gaps catalog (UNTRACKED → tracked per v0.8.1 W7)

**Priority tier:** P2  ·  **Milestone:** v1.0  ·  **TRACT ref:** L2 durable cold storage

### Gap statement
There is no (n,k) erasure-coded, no-primary cold tier — durability rests on full-copy replication, not erasure coding.

### Anchor(s) (CodeGraph-verified on `release/v0.8.0`)
- `erasure` / `no-primary` — 0 production hits

### Fix shape (from the canonical doc)
- an (n,k) erasure-coded cold tier with no single primary, so any k-of-n shards reconstruct

### Provenance
Filed per the **v0.8.1 (Patch 1)** work order §W7 and the §24 prime directive (every gap is a tracked defect — no UNTRACKED gap may rot in a queue; the v0.9.0 epic starts on a clean tracker). Sources: `docs/design/TRACT-v0.8.0-DEVELOPMENT-GAPS-CANONICAL-opus.md`, `docs/reviews/corpus-completeness-21-agent-OPUS.md`, ROADMAP §26. Orchestration hub: #1821.

🤖 Filed by Claude Code (Opus 4.8, 1M context).

alphaonedev:
**v1.0.0 disclosure-anchor shipped — PR #2065** (stacked on #2063→#2056→#2053). **This issue stays OPEN** (not re-milestoned) — the erasure *subsystem* is genuinely operator-gated; see below.

**What shipped (the v1.0.0-correct slice under GA freeze + sole-authority constraints):**
- `src/durability.rs` — a `DurabilityModel` enum (`#[non_exhaustive]`) pinning the three durability postures the substrate *actually* provides (local-single-node / power-loss-safe / opt-in quorum-replicated), with **deliberately no erasure variant**. `resolve_durability_model` computes the live posture from real config. The absent variant is the machine-checked gap: adding an erasure variant hard-breaks the wildcard-free exhaustive matches until the tier is wired + the ledger updated.
[...]

Issue #2211, opened 2026-07-18
[HIGH][#2006/PR#2205] v2 route silently bypasses the L1 restamp-by-default and ignores ImportArgs — a spine-less v2 bundle imports verbatim identity with zero verification


alphaonedev:
**Fable pre-merge audit finding on PR #2205 (`src/cli/io.rs::import_from_str` v2 routing), severity: HIGH (identity-laundering channel / silent security-default bypass).**

## Defect

The default L1 import deliberately **restamps `metadata.agent_id` with the caller's id** unless the operator opts in via `--trust-source` (an explicit provenance-hygiene default). The new v2 route is taken whenever the payload contains a `spec_version` key, and it **ignores `ImportArgs` entirely** (both the trust flag and the conflict disposition), always preserving identity verbatim.

Verbatim preservation is the correct contract for a *verified* integrity bundle — but the v2 gate verifies only the signed_events chain, and **an empty `signed_events` array passes trivially** (an empty chain is intact, gap-free, truncation-Unknown). So any JSON of the form `{"spec_version":"2", db_schema_version, memories:[…arbitrary agent_id claims…], links:[], portability_complete:false, conformance_level:"L1", conformance_by_class:{}, count:N}` imports **verbatim, with zero verification, silently bypassing the restamp default** — one attacker-controlled key flips the identity posture of the whole import. `metadata.agent_id` is load-bearing (ownership visibility, archive-ownership gates, entity binding, quota attribution).

Secondary: on the v2 path the imported memories are ALSO stamped with the destination node's vector-clock component (`storage::insert` → `stamp_version_vector`, the LOCAL-authorship chokepoint — src/storage/mod.rs comment explicitly says federation-received rows must NOT bump this node). So (a) the "preserves every field VERBATIM" claim is not quite true for `metadata`, and (b) the dest misattributes clock authorship for rows it did not author, which can skew #1757 vector-clock LWW/convergence decisions.

## Fix

- Gate verbatim-identity on an *earned* trust level: require either `--trust-source` OR a non-empty, re-verified spine covering the memories; otherwise restamp exactly like L1 (or refuse with a message naming the flag).
- Respect/replicate the ImportArgs conflict disposition, or document + error when flags are passed alongside a v2 envelope instead of silently ignoring them.
- Use remote-admission insert semantics (no local vector-clock bump), mirroring `insert_if_newer`'s posture.

Found by the Fable audit of PR #2205.

[...]

PR #2213, opened 2026-07-18
feat(#2064, #1830): opt-in (n,k) erasure-coded archive cold tier + DurabilityModel seam

### LINKED ISSUE #2064: [operator-decision] Erasure-coding dependency authorization for the G16 cold tier (#1830)
**Operator decision required** — surfaced by #1830 (TRACT-gap G16, 5-agent vote \`4d3ea1c5\`). This is NOT an agent-shirkable handoff: adding a Cargo dependency is explicitly operator-authority under the sole-authority rule, so an autonomous agent cannot make it.

## The decision
A real (n,k) erasure-coded no-primary cold tier (#1830) needs finite-field (Reed-Solomon / Cauchy) coding. Two paths, and the 5-agent vote unanimously rejected hand-rolling under GA freeze:

1. **Authorize a vetted crate** (recommended) — e.g. `reed-solomon-simd` or `reed-solomon-erasure` (SIMD-optimized, maintained, audited). Requires operator: (a) crates.io existence + maintainer audit, (b) `cargo audit` clean, (c) review of the introducing PR. This is the correctness-safe path.
2. **Hand-roll a dependency-free GF(2⁸) codec** — REJECTED by the #1830 vote: freezes a T4 shard byte-format the eventual audited crate won't match, ships error-prone finite-field math in the GA binary, and a codec bug corrupts the cold tier irrecoverably. Do not pursue without explicit operator override.

## Why this is blocking
The v1.0.0 slice for #1830 (contract §9 durability-disclosure anchor + the `DurabilityModel` enum-seam) has shipped. The erasure **subsystem** itself (codec + shard placement + no-primary coordination + reconstruction-on-read) cannot proceed until this dependency choice is authorized.

## Related
Subsystem tracker: #1830. Durability disclosure: `docs/spec/TRACT-L1-CLAIM-CONTRACT.md` §9. Sole-authority rule: `CLAUDE.md` §"Sole-authority operator".

alphaonedev:
Closes #2064

Implements the opt-in (n,k) Reed-Solomon erasure-coded archive cold tier (TRACT-gap **G16**, subsystem tracker **#1830**), rebased/merged onto `release/v1.0.0` HEAD (schema v85).

## What ships

- **`src/erasure/{mod,codec,store,archive_sync}.rs`** — the redundancy layer. When `AI_MEMORY_ERASURE_COLD_TIER` is truthy, a paced (`SWEEP_LIMIT_PER_TICK = 256`, oldest-first), idempotent, resumable sweep (embedding-backfill pattern, wired into the daemon gc loop + `ai-memory gc`) encodes each *committed* `archived_memories` row into `k` data + `m` parity shards (`AI_MEMORY_ERASURE_DATA_SHARDS`/`_PARITY_SHARDS`, defaults 4+2) under `AI_MEMORY_ERASURE_DIR` (default `<db>.erasure` sibling). Self-describing manifest (codec id, geometry, per-shard + whole-payload SHA-256, row schema version).
- **Reconstruct-on-read** inside `restore_archived`: an archived id whose DB row is GONE (partial DB loss) is re-materialized from any `k` of `k+m` *verified* shards, then the normal restore path runs unchanged.
- **`src/durability.rs`** — the #2065 `DurabilityModel` seam gains `ErasureCodedColdTier` through its wildcard-free exhaustive matches; `is_multi_node()` stays HONESTLY `false` (v1.0.0 placement is single-node — partial-disk-corruption protection, not whole-node loss; the no-primary multi-node placement is the tracked G16 residual).

## Data-integrity posture (North Star)

- The **archived DB row remains the durable source of truth**; shards are derived, regenerable redundancy.
- **Degrade, never corrupt.** Every shard is hash-verified; corruption is demoted to an *erasure* (never fed to the decoder as good); the reassembled payload must pass the whole-payload SHA-256 gate on every path. Loss/corruption beyond the `m`-shard budget FAILS LOUD (typed `ErasureError::InsufficientShards` / `PayloadHashMismatch`, tx rolled back, **zero partial state** — never partial, never wrong).
- Purge (destruction intent) removes the purged rows' bundles so purged content can't be resurrected from the redundancy layer.

## Opt-in posture (default-OFF, byte-identical when disabled)

Confirmed shippable under the campaign's TRUTH+PRESERVATION ruling:

- `erasure::erasure_cold_tier_enabled()` defaults **OFF** (`src/erasure/mod.rs` unit test `enabled_flag_default_off_and_truthy_grammar`).
- `store_for_conn()` returns `Ok(None)` when the flag is off → the gc sweep is a no-op.
- No new MCP tool / HTTP route / CLI subcommand / **DB migration** — no SSOT-count move.
- Integration test `disabled_by_default_no_bundles_and_no_resurrection` pins: no bundles written, byte-identical behavior. (Purge-side cleanup keys on directory *existence*, so a never-enabled deployment never creates the dir → truly inert; a *previously*-enabled-then-disabled deployment still honors destruction intent, a deliberate integrity feature.)

## Dependency vetting (operator-authorized 2026-07-18, sole-authority rule)

`reed-solomon-simd = "3.1"` — the #1830 2×5 adversarial vote (`4d3ea1c5`) UNANIMOUSLY rejected hand-rolling a GF(2^8) codec (a codec bug corrupts the cold tier irrecoverably). The vetted crate: AndersTrier/reed-solomon-simd, Leopard-RS GF(2^16), O(n log n), SIMD on x86-64 + AArch64, pure Rust (no build/install scripts), **zero RustSec advisories** (`cargo audit` clean), MSRV-compatible, `Cargo.lock` checksum-pinned. `src/erasure/` contains **zero** finite-field math of its own; the shard on-disk format is the crate's codec format (resolves the vote's T4 freeze concern). The authorizing comment is retained in `Cargo.toml`.

## Tests

`tests/erasure_cold_tier_2064.rs` (7 tests, all green): sweep+reconstruct-on-read after row loss; shard loss + in-place corruption up to parity budget; beyond-budget fails loud with zero partial state; purge removes bundles; owner-scoped restore refusal; pacing; default-OFF byte-identical. Plus `src/erasure/codec.rs` + `src/erasure/store.rs` unit suites and the `src/durability.rs` pins.

## Gates (all green locally)

`cargo fmt --check`; `clippy --all-targets --features sal -- -D warnings`; default-features `clippy --all-targets -- -D warnings`; `check --all-targets --features sal-postgres`; `check --examples`; `cargo audit` (clean, no new advisory); `qual_10_module_size_ceiling` (storage/mod.rs ceiling bumped 25_900→26_000 in lockstep); `qual_6_7_legacy_error_type_ceiling`; `migration_ladder_integrity`; vendor-literals / hardcoded-literals / docs-vs-ssot / migration-ladder shell gates.

Merged `origin/release/v1.0.0` (b8a95caa, #1873 audit-head hash-anchor) — clean 3-way merge (ort), both qual_10 ceilings preserved. One follow-up commit reuses `field_names::ARCHIVED_AT` at 3 sites to satisfy the hardcoded-literal ratchet.

## AI involvement (per docs/AI_DEVELOPER_WORKFLOW.md §8.2)

Authored by an AI NHI agent (Claude Opus 4.8) under operator authority in the v1.0.0 campaign autonomous loop: merged HEAD, ran the full CI-parity gate matrix, verified the opt-in/default-off posture, fixed the hardcoded-literal gate, and opened this PR. Human operator retains sole authority over merge/tag-cut. **Durability/data-integrity surface — a Fable audit is requested before merge.**


alphaonedev (commented):
## Campaign adversarial pre-merge audit — PR #2213 (erasure-coded archive cold tier, #2064/#1830)

Audit gate for the v1.0.0 clear-all campaign (durability/data-integrity surface; same gate class that caught #2207/#2208-#2211). All seven lenses covered: round-trip integrity, opt-in posture, crash windows, `reed-solomon-simd` usage, restore honesty, forget/tombstone interaction, test coverage. Every finding below was verified against the branch code (`campaign/2064-erasure-cold-tier` @ `70206c19`), the `reed-solomon-simd` 3.1.0 vendored source, and the surrounding HEAD funnels.

### Verdict: MERGE-AFTER-FIXES — F1 (HIGH) before merge; F2-F5 fixed or tracked as issues per the prime directive.

---

### What holds up (verified, not assumed)

- **Round-trip integrity is genuinely airtight at the codec layer.** Per-shard SHA-256 demotes corrupt-but-present shards to erasures before the decoder ever sees them (`src/erasure/codec.rs:404-428`); `valid < k` is a typed loud `InsufficientShards` (`codec.rs:422-428`); and the whole-payload SHA-256 gate runs on EVERY path including the all-shards-present fast path (`codec.rs:475-478`), so a reconstructed-but-wrong payload cannot be returned as truth. Exactly n-k+1 losses → loud error, pinned exhaustively by `any_k_of_n_reconstruct_exactly` (all 2^n masks) plus beyond-budget/corruption/tamper tests at codec, store, and integration levels.
- **`reed-solomon-simd` 3.1.0 usage is correct.** Verified against the crate source: shard size must be non-zero and even (`rate.rs:101`) — `shard_size_for` guarantees even ≥ 2, and `validate_manifest` re-enforces it on the read side; `MAX_SHARDS_PER_KIND = 256` keeps every (k, m) combo well inside the GF(2^16) support envelope (`recovery.next_power_of_two() + original <= 65536`, `rate_high.rs:19-25`); decoder errors surface as typed `ErasureError::Codec`, never swallowed; the decode index mapping (originals 0..k, recoveries 0..m) is correct, and returned-shard length is re-checked before reassembly (`codec.rs:463-471`).
- **Opt-in posture is real.** Default OFF, no schema change, no new tool/route/CLI surface; the disabled path is byte-identical (pinned by `disabled_by_default_no_bundles_and_no_resurrection`); the purge-side cleanup deliberately keys on the bundle DIRECTORY existing rather than the enable flag so disabling the feature does not strand destruction intent (`archive_sync.rs:109-122`, `storage/mod.rs:11976-11986`).
- **Write-side crash discipline (process crash) is sound.** Sweep-outside-the-archive-transaction means a rolled-back archive can never leave a phantom bundle; temp-dir assembly with manifest-written-last as the commit marker plus `rename` means a torn write never presents a manifest whose shards are absent (`store.rs:149-192`); reconstruct-on-read runs inside the caller's transaction so a failed reconstruct leaves zero partial state (pinned by `loss_beyond_budget_fails_loud_with_zero_partial_state`).
- **Restore honesty.** The bundle payload is a dynamic `SELECT *` snapshot of the archived row (all columns verbatim, BLOBs base64-wrapped, minting schema version recorded), so the v49/v85 archive-column-parity lessons are honored **by construction** — `lifecycle_state`, `valid_from`/`valid_until`, `cid`, `encrypted_envelope` all ride through, and the re-inserted row then flows down the UNCHANGED restore path (governance hook, collision check, cid re-mint, #2167 S8 embedding-space heal). The owner-scoped twin only materializes a bundle when the row is truly absent AND the bundle's own metadata passes the same ownership predicate as the SQL gate (verified field-by-field against `storage/mod.rs:11725-11738`) — no existence oracle, no cross-tenant materialization (pinned by `owner_scoped_restore_refuses_unowned_bundles`).
- **Forget interaction (the #2208 class).** `forget` archives (recoverable-delete contract, #1771/#1848: operator restore = authorized un-forget), so a bundle mirroring the archived row is consistent with the existing archive posture; the destruction point is purge, and all four purge funnels — including `auto_purge_archive` (verified: routes through `purge_archive`, `storage/mod.rs:11058-11063`) — remove bundles. Crypto-erase (#1956) overwrites the envelope on the LIVE row before archival, so an erased envelope never leaks a recoverable key into a bundle. Modulo F1 below, destruction intent flows through.
- **DurabilityModel seam** stays honestly `is_multi_node() == false` for the erasure tier with the G16 residual test-pinned; the resolver's precedence (quorum > erasure > power-loss-safe > local) is tested.

---

### Findings

#### F1 — HIGH — orphan bundles survive destruction and are resurrectable; the claimed self-healing ("caught by the next purge") is false

Concrete resurrection scenario (violates the PR's own stated invariant "purged content cannot be resurrected from the redundancy layer"):

1. Purge collects candidate ids, runs the autocommit `DELETE FROM archived_memories`, and only THEN removes bundles best-effort (`storage/mod.rs:12013-12019`). A crash between the DELETE commit and `remove_bundles_best_effort`, or a failed `store.remove()` (WARN-only), leaves the purged row's bundle on disk.
2. The claim that a leftover bundle "is caught by the next purge" (`src/erasure/archive_sync.rs:390-392` doc, `storage/mod.rs:11997` WARN) is **false**: `erasure_purge_candidate_ids` scans `archived_memories`, and the purged id is no longer there — no future purge ever revisits it. The orphan persists indefinitely.
3. `archive restore <id>` (or the owner via the caller-scoped twin) then triggers reconstruct-on-read (`archive_sync.rs:358-388`), which re-materializes the PURGED — possibly forgotten-then-purged, i.e. the exact #2208 resurrect-forgotten class — content back into `archived_memories` and on into the live table.

Additional members of the same orphan class, all closed by the same fix:

- **Cross-process race:** a CLI `ai-memory gc` sweep (separate connection/process) can complete `archived_row_payload` for row X, the daemon purges X + removes its bundle, then the CLI's `store.put(X)` lands afterward — re-minting a bundle for destroyed content. (The `"archived row vanished during bundling"` re-read narrows but does not close this window.)
- **Restore leaves a stale bundle:** `restore_archived` / `restore_archived_for_caller` delete the archived row (`storage/mod.rs:11686`, `:11876`) without touching the bundle. If the live row is later destroyed via a NON-archiving path (e.g. `gc` with `archive_on_gc=false`), the stale bundle resurrects an old version of the row via `archive restore`.
- **Crashed `.tmp-*` dirs are never cleaned:** the module doc claims "the next `put` for that id clears" the orphan temp dir (`store.rs:15-16`), but temp names embed nanos+pid (`store.rs:158-162`), so a later `put` uses a different name and the defensive `remove_dir_all` never matches — a slow, unbounded disk leak.

**Minimal fix (one lane closes all four):** add an orphan-reconciliation pass to the existing sweep — enumerate bundle-store entries; any bundle whose id has no `archived_memories` row AND no live `memories` row, older than a grace window (to avoid racing an in-flight archive), is removed with a WARN; also remove stale `.tmp-*` dirs past the grace window. Correct the two false doc/WARN claims, and (cheap, optional) have `restore_archived*` remove the bundle in the same flow that deletes the archived row. **Test that pins it:** mint a bundle, delete the archived row directly (simulating the crash orphan), run the sweep, assert the bundle is gone and `restore_archived` returns `Ok(false)`.

#### F2 — MEDIUM — sweep is O(N) per tick and the pacing bound only limits successes (MANAGEABLE-AT-SCALE gap)

`sweep_archive_bundles` scans ALL of `archived_memories` (`ORDER BY archived_at ASC`, no bare `archived_at` index — v56's is `(namespace, archived_at)`) every gc tick, and per already-bundled row performs a filesystem stat + manifest read + JSON parse (`archive_sync.rs:140-156`, `store.rs:125-130`). At North-Star scale (millions of archived rows) that is millions of file reads every 30 minutes, forever. Worse, `limit` bounds only `bundled` successes: under a systemic failure (disk full, non-UTF-8 row), EVERY row in the table is attempted each tick with per-row WARNs — the pacing bound evaporates exactly when the system is degraded. **Fix:** count attempts (`bundled + failed`) toward the limit, and persist/derive a resume frontier (last `(archived_at, id)` reached) so already-current rows are not re-probed every tick; add/reuse a supporting index.

#### F3 — MEDIUM — no fsync before the bundle rename: after a power loss the tier can silently believe rows are protected that are not

Shard files, the manifest, and the directory rename are written with plain `std::fs::write`/`rename` and never fsynced (`store.rs:149-192`). After a power loss, a sweep that already reported rows as `bundled` may leave torn/empty shard files behind a surviving manifest. `bundle_is_current` trusts the manifest stamp alone (`archive_sync.rs:174-184`), so the sweep never re-mints those bundles — and the corruption is discovered only at reconstruct time, i.e. exactly when the DB row is already lost and the redundancy is needed. Nothing ever returns wrong bytes (the hash gates hold), but the redundancy layer silently under-delivers its one job. **Fix:** `File::sync_all` each shard + manifest and fsync the parent directory around the rename (standard atomic-publish discipline), and/or add a slow scrub lane (a few bundles per tick, hash-verify + re-mint) — the scrub composes naturally with F1's reconciliation pass.

#### F4 — LOW — non-finite REAL column values silently encode as JSON `null` in the payload

`ValueRef::Real(x) => serde_json::Value::from(x)` (`archive_sync.rs:207`): `serde_json`'s `From<f64>` maps NaN/±Inf to `Value::Null`, so a non-finite REAL in an archived row would be silently altered to NULL in the bundle (and on any reconstruct). Near-unreachable given write-side validation, but it contradicts the "refuses (loud) … never guesses" posture the decode side correctly takes (`json_to_sql_value`). **Fix:** bail loudly on non-finite REAL during encode.

#### F5 — LOW — in-place mutation of archived rows silently stales bundles (`bundle_is_current` keys on `archived_at` only)

The live mutator today is the #2167 stamp `UPDATE archived_memories SET embedding_space = …` (`storage/mod.rs:13743`); a stamped row's bundle keeps the pre-stamp value and is still reported "current". Today this self-heals on restore (the S8 embedding-space CASE NULLs a foreign-space vector), but the invariant is implicit: any FUTURE in-place archived-row mutator (including a migration backfill over `archived_memories`) will silently diverge rows from their bundles. **Fix:** document the invariant at `bundle_is_current` ("archived rows are immutable post-mint except `embedding_space`; a new in-place mutator must bump `archived_at` or invalidate the bundle"), or key currency on a cheap row hash instead of the `archived_at` stamp.

#### F6 — NOTE (pre-existing, not this PR) — `resolve_durability_model` still has no production consumer

The PR correctly extends the #1830 seam (new variant, exhaustive matches, precedence tests), but the resolver remains test-pinned only — no capabilities/banner surface consumes it, so the §9 "disclosure" is a compile-time anchor rather than a live disclosure. Pre-existing #1830 shape; noting for the tracker, not against this PR.

---

### Test-coverage verdict (lens 7)

A test fails today for: loss up to and beyond the budget (codec + store + integration), corrupt-shard demotion, tampered manifest, foreign codec, wrong bundle id, hostile path ids, purge-removes-bundles, owner-scope refusal, zero-partial-state on loud failure, pacing/resume, default-off. **Missing** (all attached to the findings): the F1 orphan-crash-window test (bundle survives row destruction → sweep reconciles → restore refuses), a re-archive stale-stamp re-bundle test (`bundle_is_current`'s mismatch path is currently untested), and a torn-bundle-after-manifest-survives probe for F3.

---

**Verdict: MERGE-AFTER-FIXES.** Required before merge: **F1** (orphan reconciliation + the two false self-healing claims corrected + pinning test). **F2/F3** should land in this PR or as immediately-filed campaign issues (both bear on the PR's own MANAGEABLE/durability claims, not new scope); **F4/F5** fix-or-track. The codec/store core, opt-in posture, restore honesty, and forget/purge funnel coverage are otherwise merge-quality — no BLOCKED-class corruption path found: every failure mode identified degrades or resurrects, none returns wrong bytes as truth.

*Campaign audit gate, v1.0.0 clear-all. Audited: `gh pr diff 2213` + full read of `src/erasure/{codec,store,archive_sync,mod}.rs`, `src/durability.rs`, the restore/purge funnels in `src/storage/mod.rs`, `tests/erasure_cold_tier_2064.rs`, and the `reed-solomon-simd` 3.1.0 vendored source.*

alphaonedev:
## Durability-audit remediation — F1-F5 fixed at `1b4a37d7`

Thanks for the seven-lens pass. All six findings addressed in this PR (campaign clear-all: fix now, don't file-and-defer). Per-finding disposition:

### F1 — HIGH — orphan bundles resurrectable / false self-healing claim → **FIXED**
One lane closes all four members, as prescribed. New gc-tick **orphan-reconciliation** pass `archive_sync::reconcile_and_scrub`:
- **Purge-crash orphan / failed `store.remove`:** a committed bundle whose id has **no `archived_memories` row AND no live `memories` row**, older than a grace window (`RECONCILE_GRACE_SECS` = 1h, so an in-flight archive/reconstruct can't race the reaper), is removed with a WARN. This is the reaper the purged id never got (a purged id never re-appears in `archived_memories`).
- **Restore-leaves-stale-bundle:** `restore_archived` / `restore_archived_for_caller` now drop the now-stale bundle in-flow after the successful restore commit (the "cheap, optional" belt-and-suspenders), so a later non-archiving delete of the live row can't resurrect the old version; a survivor is still reaped by reconciliation.
- **Crashed `.tmp-*` dirs:** `ErasureStore::reap_stale_temp_dirs` sweeps them past the grace window (the "next put clears it" doc was false — temp names embed nanos+pid).
- **CLI-gc-vs-daemon-purge re-mint race:** the reconciliation is the backstop — a bundle re-minted for a destroyed row becomes an orphan (no rows) and is reaped next tick.

**False-claim corrections:** the `remove_bundles_best_effort` doc and the `erasure_purge_candidate_ids` WARN no longer claim "caught by the next purge" — they name the reconciliation pass.

**Pinning test** (`orphan_bundle_reaped_and_not_resurrectable`): mint a bundle, delete the archived row (crash orphan), reconcile → bundle gone AND `restore_archived` returns `Ok(false)`. FAILS on the old orphan-resurrection behavior.

### F2 — MEDIUM — O(N) per-tick + pacing evaporates under failure → **FIXED**
- `limit` now bounds **attempts** (`bundled + failed`), so a systemic failure (disk full / poison row) can no longer retry every row each tick — the bound holds precisely when degraded.
- A process-static **keyset `(archived_at, id)` resume frontier** (per store dir) skips the already-bundled prefix so it is not re-probed (stat + manifest read + JSON parse) every tick. Frontier advances only over the contiguous non-failed prefix; it resets on restart (a fresh oldest-first scan then re-covers everything — a clock-skew-skipped row self-heals across a restart). *Residual (noted honestly): I did not add a bare `archived_at` index this PR — the frontier's win is cutting per-row filesystem probes, which holds regardless of the planner; a supporting index is a clean follow-up.*

### F3 — MEDIUM — no fsync before rename → **FIXED**
`write_bundle_at` now `fsync`s each shard + the manifest + the temp dir + the parent dir around the rename (standard atomic-publish; dir-fsync is a `#[cfg(unix)]` no-op on Windows, per-file fsyncs still hold there). Plus a **scrub lane** (rides the reconciliation pass, `SCRUB_LIMIT_PER_TICK` = 16, rotating cursor for coverage) hash-verifies current bundles and re-mints a torn one **from the durable archived row** before a reconstruct needs it. Test `torn_bundle_detected_and_reminted_by_scrub`: delete m+1 shards → `get` fails loud → scrub re-mints → `get` clean, not degraded.

### F4 — LOW — non-finite REAL → JSON null → **FIXED**
`ValueRef::Real(x) if !x.is_finite()` now `bail!`s loudly on encode (matching the decode side's refuse-never-guess posture).

### F5 — LOW — `bundle_is_current` keys on `archived_at` only → **FIXED (option a, documented)**
Documented the archived-row immutability invariant at `bundle_is_current`: rows are immutable post-mint **except** the #2167 in-place `embedding_space` heal-stamp, which the restore S8 CASE neutralizes (NULLs a foreign/legacy-space vector) — so a bundle carrying the pre-stamp value re-materializes correctly. Any future in-place mutator MUST bump `archived_at` or invalidate the bundle. Chose (a) over a content-hash currency key deliberately: a content-hash would conflict with the F2 archived_at frontier (a below-frontier stamp change would be skipped), and the torn-bundle case a hash would otherwise catch is covered by the F3 scrub lane instead. Added the missing mismatch-path test `re_archived_row_rebundles_on_stamp_change`.

### F6 — NOTE — `resolve_durability_model` has no production consumer
Acknowledged, pre-existing (not this PR). The TRACT lane consumes the seam next.

**Gates:** `tests/erasure_cold_tier_2064.rs` 10/10, in-crate erasure suite 17/17, `cargo fmt`, `clippy --features sal` + default `-D warnings`, `check --features sal-postgres` + `--examples`, `qual_10` / `qual_6_7`, hardcoded/vendor-literal gates — all green. Ready for re-verification.

alphaonedev (commented):
## Campaign audit gate — re-verification of remediation commit `1b4a37d7` (F1-F5)

Re-verified against the actual code in the worktree at `1b4a37d7`, with a deliberate re-attack on each fix per the audit-gate protocol.

### Confirmed fixed

- **F3 (fsync + scrub) — FIXED, ordering verified crash-safe.** `store.rs:195-223`: each shard is fsync'd before the manifest (commit marker) is written, the manifest is fsync'd, the temp dir is fsync'd (making the entries durable) BEFORE the rename, and the parent dir is fsync'd after — the standard atomic-publish discipline, correctly ordered. The rotating scrub lane (`archive_sync.rs:548-557`, `612-629`) hash-verifies current bundles and re-mints a torn-beyond-budget bundle from the durable archived row; pinned by `torn_bundle_detected_and_reminted_by_scrub`. Bonus soundness property: the F2 frontier is process-static (in-memory), so ANY power loss implies a restart and a fresh full scan — a crash can never strand a lost-rename row behind the frontier.
- **F4 — FIXED.** Encode bails loudly on non-finite REAL (`archive_sync.rs:303-308`); the decode side (`json_to_sql_value`) is byte-unchanged, refuse-never-guess posture intact.
- **F5 — FIXED, and the documented invariant is honest.** The `bundle_is_current` doc (`archive_sync.rs:252-264`) does NOT claim absolute immutability — it names the #2167 in-place `embedding_space` heal-stamp as the single existing exception and correctly explains why it is harmless (the S8 restore CASE NULLs a foreign-space vector on the way back to live), then pins the obligation on any FUTURE mutator. No false claim. The previously-untested mismatch path is now pinned by `re_archived_row_rebundles_on_stamp_change`.
- **F1 partial credit:** restore paths drop the stale bundle in-flow post-commit (`storage/mod.rs:11692-11702`, `:11893-11900`), both false self-healing claims corrected (`archive_sync.rs:492-500`, `storage/mod.rs:12015-12017`), crashed `.tmp-*` dirs reaped past grace (`store.rs:346-371` — mtime updates on write, so in-flight writers are safe), and the orphan/re-mint pinning tests exist.
- **F2 partial credit:** the pacing limit now bounds attempts (`bundled + failed`, `archive_sync.rs:210`), and the keyset frontier eliminates the steady-state O(N) re-probe (pinned by the updated `already_current == 0` assertions).

### Remaining findings

#### R1 — BLOCKING — the orphan reaper hard-deletes the LAST SURVIVING COPY of a DB-lost archived row (wrong-direction data loss, worse than the defect it fixes)

The reap predicate (`archive_sync.rs:577-610`) — committed bundle AND no `archived_memories` row AND no `memories` row AND manifest older than the grace window — is **byte-for-byte indistinguishable from the disaster state the cold tier exists to survive**: partial DB loss of the archived row while the bundle is the only copy left. The grace window does not help: `manifest_age_secs` (`store.rs:335-339`) measures from MINT time, not from when the row vanished, so any bundle older than 1 hour whose row is lost is reaped on the very next gc tick (≤30 min), up to 512 per tick, permanently (`remove_dir_all`), WARN-labeled "destruction intent reconciled" when no destruction was ever intended.

Concrete scenario: an accidental `DELETE FROM archived_memories`, a migration bug, or corruption-repair row loss — exactly the #2064 protection target — and the daemon's own reconciliation lane destroys the redundancy layer faster than any operator can respond, then `archive restore <id>` returns `Ok(false)`. The test suite itself demonstrates the contradiction: `sweep_bundles_and_restore_reconstructs_lost_row` and `orphan_bundle_reaped_and_not_resurrectable` construct the IDENTICAL on-disk + DB state via the same `drop_archived_row`, and require OPPOSITE outcomes — disambiguated by nothing but elapsed wall-clock time. Under the North Star, "never cause unintentional data loss" outranks "purged content must not be resurrectable"; the pre-remediation defect required an explicit restore call to do harm, while this reap is automatic. This inverts the prime directive and blocks merge as-is.

**Fix (either lane closes it; both compose):**
1. **Durable destruction-intent journal (preferred discriminator).** Purge funnels append the candidate ids to a fsync'd write-ahead journal in the erasure dir BEFORE the `DELETE` (a crash after journal-write but before DELETE is harmless — the reconciler's row-exists check already skips live rows); `remove_bundles_best_effort` / the reconciler clear entries once the bundle is confirmed absent. The reaper hard-reaps ONLY journaled ids. The restore in-flow removal already covers the restore-staleness lane.
2. **Reap-to-quarantine for everything else.** A rowless, UN-journaled bundle is moved to a dot-led quarantine subdir (invisible to `get`/`list_committed_bundle_ids`, so it is not resurrectable — the resurrection lane stays closed) with a LOUD "no recorded destruction intent — possible DB loss, bundle QUARANTINED not destroyed" WARN, retained for a long operator-configurable window with an explicit restore/purge verb. This is the North-Star shape verbatim: destructive ops need explicit intent + reversibility.

Re-point `orphan_bundle_reaped_and_not_resurrectable` at a journaled purge orphan, and add the inverse pin: a rowless bundle with NO recorded intent must remain recoverable (quarantined, not destroyed) while `restore_archived` refuses to auto-materialize it.

#### R2 — MEDIUM — the reconcile scan is `take(scan_limit)` over stable readdir order: not eventually-covering past 512 bundles

`reconcile_and_scrub` examines `ids.iter().take(scan_limit)` (`archive_sync.rs:577`) over `list_committed_bundle_ids` (`store.rs:312-328` — raw `read_dir` order, which is roughly stable on ext4). There is no rotating offset for the SCAN (the scrub cursor rotates only WITHIN the first-512-derived candidates). A store with >512 committed bundles never examines listing positions beyond 512: orphans there are never reaped (F1 residual) and torn bundles there are never scrubbed (F3 residual). The "paced + eventually-covering" claim (`archive_sync.rs:53-55`) is false at exactly the North-Star scale the constant exists for. **Fix:** give the reconcile scan its own rotating cursor over the full listing (the `scrub_cursor` pattern, one line of modular arithmetic), or a persisted resume offset.

#### R3 — MEDIUM — a permanently-failing row pins the frontier and reopens unbounded per-tick probes (F2 partially reopened; F4 creates the poison class)

The frontier only advances over the contiguous clean prefix (`failed_seen`, `archive_sync.rs:231-238`) — correct for retry — but `already_current` rows do NOT count toward the attempt limit (`:210`). A row that fails bundling PERMANENTLY (non-UTF-8 TEXT, or precisely the non-finite REAL that F4 now loudly bails on) pins the frontier forever, and every tick then iterates the ENTIRE remainder of the table with a stat+manifest-read+parse per row (uncounted `already_current` probes) — the exact O(N)-per-tick-under-degradation behavior F2 was meant to close, now triggered by a single poison row. **Fix:** count `already_current` probes toward a (separate, larger) per-tick probe budget, or track persistently-failing ids in a bounded retry set the frontier may advance past.

#### R4 — LOW — an out-of-band FULLY-deleted bundle below the frontier is not re-minted until process restart

The frontier skips already-covered rows; the scrub only sees COMMITTED bundles (manifest present). A bundle wholly deleted out-of-band (not torn — gone) for a row below the frontier is invisible to both lanes, so its redundancy is not re-minted for the life of the process. Bounded: the durable row is intact, and the process-static frontier resets on restart (a fresh scan re-mints). Note it in the frontier doc; R2's rotating reconcile cursor could also probe row-existence in reverse to close it fully.

---

### Verdict: NOT MERGE-CLEAN — MERGE-AFTER-FIXES

**R1 is BLOCKING**: the remediation introduced an automatic destruction lane for last-copy bundles of DB-lost rows — unintentional data loss by the janitor, strictly worse on the North-Star axis than the resurrection defect it fixed. **R2/R3** should land with the R1 rework (all three touch `reconcile_and_scrub`); **R4** doc-note or fold into R2. F3, F4, F5, and the non-reap portions of F1/F2 are verified fixed with correct crash-ordering and honest documentation.

*Campaign audit gate re-verification, v1.0.0 clear-all. Verified in worktree wt-2064 at `1b4a37d7` (read-only).*

alphaonedev:
## R1-R4 remediation at `62ed739f` — journal + quarantine (the reaper no longer destroys DB-lost last copies)

You were right, and this was the important catch: the F1 reaper INVERTED the failure mode. Its predicate (rowless bundle + manifest ≥ grace) is byte-identical to the partial-DB-loss disaster the tier exists to survive, and grace measures from MINT time — so it auto-destroyed the last surviving copy of any DB-lost archived row. Under the North Star that is strictly worse than the resurrection defect it fixed. Redesigned per your prescription.

### R1 — BLOCKING — reaper destroys the last copy of a DB-lost row → **FIXED (journal + quarantine)**

**(a) Durable write-ahead purge-intent journal.** All four purge arms (`purge_archive` / `purge_archive_for_caller`, days + full) now `journal_purge_intent_best_effort` BEFORE the `DELETE` — fsync'd empty `.purge-intent/<id>` markers in the erasure dir. A crash after journal-write but before the DELETE is harmless: the reconciler's row-exists check skips a still-live row, and `compact_purge_journal` drops vestigial markers. `remove_bundles_best_effort` clears the marker on a successful remove (keeps it on failure so the reconciler reaps the journaled survivor).

**(b) Reap-to-quarantine for everything else.** The reconciler now discriminates the two byte-identical rowless states:
- **journaled** rowless bundle → HARD-reap (confirmed destruction; `remove_dir_all`);
- **un-journaled** rowless bundle → **QUARANTINE** — `rename` into a dot-led `.quarantine/<id>` holding pen, invisible to `get` / `list_committed_bundle_ids` (so `archive restore` can't materialize it — the resurrection lane stays closed), **retained on disk, never auto-destroyed**, with a LOUD WARN: *"no recorded destruction intent — possible DB LOSS; QUARANTINED (preserved, not destroyed); set AI_MEMORY_ERASURE_RECOVER_QUARANTINE=1 and restart, then archive restore."*

`try_restore_archived_row_from_bundle` also refuses a journaled id immediately (closes the purge→reconcile window — a purged id never resurrects even while its bundle lingers). New DR knob `AI_MEMORY_ERASURE_RECOVER_QUARANTINE` (env #144) moves quarantined bundles back to active and stops quarantining new ones, the operator's explicit "this was DB loss, keep it serveable" assertion. `ErasureStore` gains `journal_purge_intent`/`is_purge_intended`/`clear_purge_intent`/`compact_purge_journal`/`quarantine`/`list_quarantined_ids`/`recover_quarantined`.

**Inverted test pins (per your (c)):**
- `journaled_purge_orphan_is_reaped_and_not_resurrectable` — journal + drop row → hard-reaped, restore refuses.
- `unjournaled_dbloss_bundle_is_quarantined_not_destroyed_and_recoverable` — drop row with NO journal → bundle QUARANTINED (asserts `.quarantine/<id>/manifest.json` still on disk, `orphans_reaped == 0`), `restore` returns `Ok(false)` (invisible), then DR recovery → bundle back to active → `restore` reconstructs byte-exact. The two tests build the IDENTICAL on-disk+DB state via the same `drop_archived_row` and now require OPPOSITE outcomes — disambiguated by the journal, not wall-clock.

### R2 — MEDIUM — reconcile scan not eventually-covering past 512 → **FIXED**
Replaced `ids.take(scan_limit)` with a rotating window over the FULL listing (`reconcile_cursor().fetch_add(scan_limit) % n`, the existing cursor pattern), so every listing position is eventually examined. The scrub feeds off the same rotating window.

### R3 — MEDIUM — poison row pins frontier + reopens O(N) probes → **FIXED**
Added a per-tick `SWEEP_PROBE_BUDGET_PER_TICK` (4096) that counts EVERY probed row (incl. `already_current`); the sweep breaks on either the bundling `limit` or the probe budget. A permanently-failing row (non-UTF-8, or the F4 non-finite REAL) can no longer make the sweep re-scan the whole table each tick — the per-tick filesystem cost is bounded regardless of the frontier being pinned.

### R4 — LOW — out-of-band fully-deleted bundle below frontier → **doc + R2**
Documented in the frontier doc (durable row intact; process-static frontier resets on restart → re-mint), and R2's rotating reconcile scan probes such positions over successive ticks.

**Gates:** `tests/erasure_cold_tier_2064.rs` 11/11, in-crate erasure suite 17/17, `cargo fmt`, `clippy --features sal` + default `-D warnings`, `check --features sal-postgres` + `--examples`, `qual_10`/`qual_6_7`, hardcoded/vendor/docs-ssot literal gates — all green. Ready for the R1 re-verify.

alphaonedev (commented):
## Campaign audit gate — R1-focused re-verification of `62ed739f`

Re-attacked the journal/quarantine design per the gate protocol. Verified in the worktree at `62ed739f`.

### R1 core — FIXED, re-attack survived

1. **Crash after journal-write, before the DELETE (the inversion probe): SAFE.** The reconciler consults the journal ONLY for ROWLESS bundles — `row_exists(archived_memories)` is checked first (`archive_sync.rs` window walk) and a row-present bundle goes to the scrub lane untouched. A stale marker cannot cause a live-row bundle to be reaped.
2. **Journal durability ordering: CORRECT.** Markers are empty files whose EXISTENCE is the signal, so the single `fsync_dir(journal)` before the `DELETE` (`store.rs::journal_purge_intent`) is a valid durability scheme. Every failure direction points toward preservation: journal-write failure → WARN → the bundle is later QUARANTINED, not destroyed; marker-lost-after-acked-purge → quarantine-forever (accepted); marker fsync precedes the DELETE, so "delete durable but intent not" cannot occur.
3. **Quarantine escape: NONE FOUND.** `validate_bundle_id` rejects dot-led ids and path separators, so `get`/`put`/`remove` can never address `.quarantine/*` or `.purge-intent/*`; `list_committed_bundle_ids` skips dot-led entries, so neither the scrub, the rotating reconcile window, nor the sweep ever touches quarantined bundles; `try_restore` reaches only the active store. DR-mode ordering is right: recovery runs BEFORE the listing walk, and a just-recovered (still rowless, un-journaled) bundle hits the `recover_mode` early-return — no quarantine ping-pong within a pass. `recover_quarantined` refuses to clobber an active bundle. Both inverted test pins are present and assert the right discriminations (including restore refusing a journaled id pre-reconciliation, and byte-exact DR recovery).
4. **R2/R3: FIXED.** The reconcile cursor now rotates the scan window over the FULL listing (`(start + offset) % n`), covering stores > 512 bundles across ticks, and the scrub feeds from the rotating window. The sweep probe budget (`SWEEP_PROBE_BUDGET_PER_TICK = 4096`) bounds the pinned-frontier tick cost.
5. **Env knob discipline: CLEAN.** CLAUDE.md row 144 present and accurate (truthy grammar shared with the master switch via `env_truthy`, default OFF, source-pinned); CHANGELOG updated.

### Remaining findings

**R5 — MEDIUM (delayed-fuse variant of R1; small fix) — stale intent markers from an ABORTED purge are never cleared and later convert a genuine DB loss into hard destruction.** All four purge arms journal intent, THEN run the `DELETE`. If the `DELETE` errs (SQLITE_BUSY, I/O — a plain error return, not just a crash) or the process dies between journal-fsync and DELETE, the markers persist for rows that were never deleted. `compact_purge_journal` clears a marker only when the BUNDLE is gone (`!contains(id)`) — here bundle AND row both exist, so the marker lives forever. If that id later suffers real partial DB loss (the disaster lane), the reconciler sees rowless + journaled → HARD-reaps the last surviving copy, and `try_restore` refuses the id — precisely the forbidden direction, on a delay. Compound-rare (aborted purge AND later loss of the same id) but fail-open in the wrong direction and cheap to close.
**Fix (~5 lines):** in the reconcile window walk, when the ARCHIVED ROW EXISTS and `is_purge_intended(id)` and the marker is older than the grace window, clear the marker with a WARN ("aborted purge intent cleared") — a row-present marker means the purge never completed, and a real purge re-journals. The marker-age check removes the concurrent-purge race entirely; even without it the race direction is quarantine (preservation — safe). Pin with a test: journal → simulate DELETE failure (row stays) → reconcile clears the marker → later `drop_archived_row` → reconcile QUARANTINES (not reaps).

**R6 — LOW (track as issue) — quarantined purged content has no completion path for destruction intent.** When the journal write fails (the designed degrade) the purged bundle lands in `.quarantine/` and is NEVER auto-destroyed — the purged bytes persist on disk indefinitely, with `list_quarantined_ids` but no operator verb to inspect/purge the quarantine. Correct per "explicit intent + reversibility", but the destruction intent never completes for those bytes (privacy/lifecycle residual). Suggest a tracked issue for an explicit quarantine inventory + purge verb (or documented retention guidance).

**R3 residual — LOW (note/track)** — the probe budget bounds the tick but does not un-pin the frontier: with a poison row plus >4096 subsequent rows, rows beyond the budget (including NEWLY archived ones) get no bundle until the poison row — WARN-named every tick — is fixed. Bounded, loud, and the durable rows are intact; a future poison-id skip-set would let the frontier advance past permanent failures.

### Verdict: MERGE-AFTER-FIXES — R5 (small, required); R6 + R3-residual fix-or-file. The R1 blocking inversion is closed: the reaper can no longer destroy a last-copy bundle without a durably recorded destruction intent, the quarantine is escape-proof, and both direction-pins are tested. With R5 landed this surface is merge-quality.

*Campaign audit gate re-verification, v1.0.0 clear-all. Worktree wt-2064 @ `62ed739f`, read-only.*

alphaonedev:
## R5 fixed + R6/R3-residual filed — head `759f30b2`

Thanks for confirming R1 CLOSED. R5 landed in this PR; R6 and the R3-residual are filed as tracked follow-ups per the prime directive.

### R5 — MEDIUM (required) — aborted-purge stale marker → delayed-fuse hard-reap → **FIXED**
Exactly the delayed-fuse you flagged: a purge that journaled intent then had its `DELETE` abort (SQLITE_BUSY/IO) leaves a stale `.purge-intent/<id>` marker while row + bundle both survive; `compact_purge_journal` can't clear it (bundle exists), so a LATER genuine DB loss of that id would hard-reap the last copy.

Fix, per your prescription: in the reconcile walk, a **row-present (archived OR live) id whose marker is journaled AND older than the grace window** is a stale marker from an aborted purge → cleared with a WARN. A real purge re-journals; the age gate skips a concurrent in-flight purge (whose failure mode is quarantine — safe — anyway). New `ErasureStore::purge_intent_age_secs` + `ReconcileReport.stale_intent_cleared`.

**Pinning test** `aborted_purge_stale_marker_is_cleared_so_later_dbloss_quarantines`: journal intent (row survives) → reconcile clears the stale marker (`stale_intent_cleared == 1`) → a subsequent genuine DB loss of the same id **QUARANTINES** (`quarantined == 1`, `orphans_reaped == 0`), not hard-reap. Fails on the pre-R5 behavior.

### R6 — LOW — quarantined purged content has no operator inspect/purge verb → **FILED #2224**
Quarantine inventory + purge verb proposed (built on `list_quarantined_ids` + a new `purge_quarantined`); cites this re-verify.

### R3-residual — LOW — probe budget bounds cost but doesn't un-pin the frontier → **FILED #2225**
Bounded poison-id skip-set (the frontier may pass a persistently-failing row) proposed; cites this re-verify.

### Also folded: origin/release/v1.0.0 (`4a277bc8`)
The PR now merges the current HEAD (#1860 vectorlite, #1834 claim-bitemporal, #1390 SDK shims). CHANGELOG kept both sides; `qual_10` `storage/mod.rs` ceiling reconciled max-of-then-sum 26_100 -> 26_250 (merged file 26_143 = #2204 claim-bitemporal threading + #2064 erasure funnel wiring).

**Gates:** erasure suite 12/12, `cargo fmt`, `clippy --features sal` + default `-D warnings`, `check --features sal-postgres` + `--examples`, `qual_10` / `qual_6_7` / `migration_ladder_integrity`, hardcoded/vendor/docs-ssot/migration-ladder shell gates — all green. Ready for the final verdict check.

Issue #1005, opened 2026-05-21
v0.9 — Vector search minimal opt-in slice: VectorSearchIndex seam + §5.2 lazy-allowlist recall fix + G4 strict-dim + G2 capacity knob (full substrate → #1860)


alphaonedev:
# v0.9 — Vector search: minimal opt-in slice (seam + §5.2 recall fix + G4/G2 knobs)

> **REWRITTEN 2026-07-01** by the v0.9.0 design vote (15-agent codegraph-grounded review + 2×5 vote; 7 MODIFY / 3 ACCEPT / 0 DEFER, 10/10 `v090_fit=true` for this carve-down; operator-confirmed). The original full-substrate plan was drafted against a stale tree (phantom symbols, u64-vs-String-cid model, schema v34 vs live v74+) and is **deferred to v1.0 → #1860**, which preserves the original text plus the stale-anchor corrections. This issue now tracks ONLY the accepted v0.9 increment below.
>
> Every item mirrors the shipped G5b/G9/G10.1/#1828 inert-until-enabled idiom: **byte-identical to legacy behavior until a flag/allowlist is set.**

### 1. Seam: `VectorSearchIndex` trait
- Define `trait VectorSearchIndex` (NOT `VectorIndex` — that name is the concrete struct at `src/hnsw.rs:320` with 23+ callers).
- The existing `hnsw::VectorIndex` struct implements it verbatim as the inert default backend.
- Flip `Arc<tokio::sync::Mutex<Option<VectorIndex>>>` → `Arc<Mutex<Option<Box<dyn VectorSearchIndex>>>>` at the ~25 sites.
- Signatures: `fn insert(&self, id: String, emb: Vec<f32>)`; `fn search(&self, q: &[f32], k: usize, allowlist: Option<&[String]>) -> Vec<VectorHit>`.

### 2. §5.2 LIVE recall bug — namespace starvation (the one REAL gap from the original issue)
- Add `allowlist` to search; consume the `hnsw.search` iterator LAZILY until k in-namespace hits or exhaustion, replacing the global `k*2`/`ann_limit` cutoff.
- Anchors: `hnsw.rs:764`; `handlers/recall.rs:616`; post-filter `storage/mod.rs:11379` (re-verify — tree moves).
- `None` = today's behavior; zero migration.

### 3. G4: opt-in strict-dim guard
[...]

Issue #2208, opened 2026-07-18
[BLOCKING][#2006/PR#2205] v2 import resurrects destination-forgotten memories — dest forget_tombstones (and archived lifecycle) never consulted


alphaonedev:
**Fable pre-merge audit finding on PR #2205 (`src/portability/import.rs`), severity: BLOCKING (erasure-durability / data-integrity).**

## Defect

`import_full_envelope` → `apply_all_classes` step (2) claims *"tombstone-before-admit, so a forgotten row can never resurrect through the memory insert"* — but the `tombstoned` set is built **only from the BUNDLE's** `env.forget_tombstones`. The **DESTINATION's** `forget_tombstones` table is never consulted, and the memory insert goes through `storage::insert` (src/storage/mod.rs:1035), which — unlike the federation-receive funnel `insert_if_newer` (src/storage/mod.rs:12084, which calls `memory_is_tombstoned`) — performs **no tombstone check**.

## Reproduction (resurrection)

1. Export a v2 bundle from source containing memory X.
2. Import into dest. Later, dest operator runs `memory_forget(X)` → row deleted + dest tombstone recorded (the durable erasure receipt).
3. Re-import the same bundle: `tombstoned` (bundle set) misses X; `storage::get(conn, X)` returns `None` (row was erased); `storage::insert` re-admits X.

Result: the forgotten content is **resurrected**, AND the dest now holds a live row *alongside* its own FORGET tombstone for the same id — an internally contradictory state the substrate's own `memory_is_tombstoned` gate exists to structurally prevent (same posture as the evict tombstones: "cannot be resurrected"). This is a committed, silent violation of the forget contract on the highest-integrity import path.

Adjacent (same class, lesser severity): a memory the dest **archived** (row moved to `archived_memories`) also probes `None` on `storage::get` and re-admits as live, leaving the id present in both `memories` and `archived_memories`.

## Fix

In the memories loop, probe `storage::memory_is_tombstoned(conn, &mem.id)` (dest-side) in addition to the bundle set, and count into `tombstoned_skipped`. Consider an archived-row probe too. Regression tests: forget-at-dest → re-import → row stays gone, tombstone honored, report counts skip.

Found by the Fable audit of PR #2205; blocks merge of #2205 as-is.

alphaonedev:
[...]

PR #2204, opened 2026-07-18
feat(#1834): claim-bitemporal VALID-time read/write surface (residual of #2036)

alphaonedev:
Closes #1834
Closes #2207

## Summary

Re-lands the #2036 residual (killed by a session limit) on the current v85 tree: threads the schema-v79 `memories.valid_from` / `valid_until` columns (RFC3339 TEXT — the half-open `[valid_from, valid_until)` END-EXCLUSIVE VALID-time interval a claim is asserted to HOLD, distinct from the `created_at` transaction-time axis) through the full read/write surface on BOTH backends. The columns existed since v79 but were write-dead (no read/write surface). No new migration.

## Approach

- **`Memory` struct** gains `valid_from` / `valid_until` (`FIELD_COUNT` 28→30, `skip_serializing_if` so legacy wire shapes stay byte-identical when unset). SSOT reconciled: `Memory::FIELD_COUNT`, CLAUDE.md field-count narrative, `memory_field_count_invariant`, `check-docs-vs-ssot.sh` (mem_fields=30).
- **Persist** (both backends): `db::insert` / `insert_if_newer` / pg `store` / `store_batch` / `apply_remote_memory` — `valid_from` IMMUTABLE (stored/genesis value wins on upsert + federation merge, like `cid`); `valid_until` caller-closable (`COALESCE` on upsert; newer-wins under federation LWW so a peer that CLOSED a claim replicates the close).
- **`memory_update`** (HTTP `PUT`, MCP, SAL `UpdatePatch.valid_until`, sqlite `update_with_expected_version` + pg both funnels $12/$14) accepts an opt-in `valid_until` patch — `valid_from` is structurally absent from every UPDATE SET list, so the genesis assertion instant can never be rewritten.
- **Recall + list** gain an opt-in RFC3339 `valid_at` AS-OF instant (validated via the new `validate::validate_valid_at`): keyword `recall`, hybrid FTS + semantic linear-scan SQL predicates + the HNSW Rust re-filter, pg FTS/semantic pools, `build_list_query` / pg `list` — all apply the same NULL-guarded half-open window. `None` is byte-identical to pre-#1834. Threaded through `RecallRequest.valid_at` (all constructors), `ListQuery`/`RecallQuery`/`RecallBody`, MCP `memory_recall`/`memory_list`/`memory_update`, SAL `Filter.valid_at`.
- **Archive** round-trip already shipped (v85 #2035) — untouched.

## #2207 — same-`id` federation merge lane (Fable audit finding, folded in)

[...]

PR #2219, opened 2026-07-19
feat(#1860): opt-in vectorlite ANN backend (default-off, fail-closed derived index)

### LINKED ISSUE #1860: v1.0 — Persistent Vector Index Substrate (vectorlite) — deferred from #1005
# v1.0 — Persistent Vector Index Substrate (vectorlite) — DEFERRED from #1005

**Origin:** carved out of #1005 by the v0.9.0 design vote of 2026-07-01 (15-agent codegraph-grounded review + 2×5 vote; tally 7 MODIFY / 3 ACCEPT / 0 DEFER, 10/10 `v090_fit=true` for the carve-down). Operator-confirmed decision: v0.9.0 ships only a minimal inert opt-in slice (tracked in #1005, rewritten); the full substrate re-platforming below is v1.0 scope.

## Deferred to this issue (v1.0)

- vectorlite / any SQLite loadable-extension backend + `build.rs` native `.so`/`.dylib`/`.dll` acquisition. Blockers found: rusqlite 0.32 is `bundled` with `load_extension` OFF; extension availability breaks the 3-OS (Windows-required) CI matrix, 5 release channels, and reproducible builds.
- u64 re-keying of vector ids (live model is String cids).
- Signed `IndexInserted`/`IndexDeleted` events on the default write path (anti-idiom on-by-default; if pulled earlier it must be an opt-in additive `event_type` on `SignedEvent`, net-delta verify only, no caller `prev_hash`).
- Builtin persistent WAL backend — flagged RISK #1 in review: cross-domain 2PC hazard; as-written the fallback for high-assurance environments would get the WEAKER coherence guarantee.
- One-shot migration + rebuild primitive (original Tasks 3.1/3.2).
- rabitq-rs alternate backend.
- The 15-agent fan-out execution model (merge-order coupled; mislabelled parallel).

## Blocking gates carried forward

- Task 0.1 recall spike stays strictly blocking (standalone /tmp binary; R@5 within 1.0pt of the 97.8% baseline).
- NEW co-equal blocking pre-flight: the pinned extension must load + smoke-query on macOS AND Windows CI runners before any substrate work fans out — extension availability across the matrix/channels is the real plan-killer, not R@5.
- p95 ≤ 35ms is advisory only (no macOS bench runner exists).

## ⚠️ Stale-anchor warning on the original plan below

The plan was drafted against a stale tree and MUST be re-grounded before execution:
- Proposes a `VectorIndex` **trait** — the name collides with the concrete struct at `src/hnsw.rs` (23+ callers). v0.9 introduces the seam as `VectorSearchIndex` instead.
- References phantom symbols: `src/recall/mod.rs::filter_by_namespace_post_ann` (Task 2.3), a `MemoryStore` HNSW field (`MemoryStore` is a trait at `src/store/mod.rs`), `src/audit/signed_events.rs` (wrong path).
- Assumes u64 ids (live: String cids) and `&mut self` + `rusqlite::Transaction` (live: deliberate `&self` + interior-Mutex + `Arc<Mutex<Option<VectorIndex>>>` swap-on-read).
- Assumes schema v34; CURRENT_SCHEMA_VERSION was already 74+ at review time.
- 3 of its 4 gap justifications were stale at review time: G2 eviction already LOUD since v0.7.0 R3-S1; G3 async-masked, not a correctness gap; G4 already enforced at the write boundary + read recompute. The §5.2 post-ANN namespace-starvation hazard was REAL and is fixed by the v0.9 slice in #1005.

---

# Original #1005 plan (historical, un-re-grounded — see warning above)

# ai-memory v0.9 — Vector Index Substrate Development Plan

**Capability:** Replace the in-memory `instant-distance` HNSW with a persistent, transactionally-coherent, audit-chain-integrated vector index behind a swappable trait.

**Closes (from ROADMAP2.md §5.4):** G2 silent eviction at 100k, G3 cold-start O(N) rebuild, G4 mixed-dim silent tolerance, post-ANN namespace filter hazard (§5.2).

**Primary backend:** vectorlite (hnswlib + Google Highway SIMD) as SQLite extension.
**Fallback backend:** pure-Rust HNSW for locked-down environments.
**Pluggable via trait** so future quantization-optimized backends (rabitq-rs, etc.) drop in without architectural change.

**Execution model:** AI NHI multi-agent parallel. Wall-clock target ~7 hours; floor 4 hours, ceiling 10 hours.

**Audience for starter prompts:** Claude Code, Codex, or equivalent CLI-driven coding agents working against a checkout of ai-memory-mcp on a feature branch.

---

### Task 0.1 — Recall spike on real corpus (BLOCKING)

**Owner:** Single agent. Blocks all downstream work.
**Duration estimate:** ~45 minutes.
**Outcome:** Pass/fail decision on whether vectorlite holds recall on ai-memory's actual embedder dimensions before committing to the architecture.

**What "pass" means:**
- LongMemEval R@5 within 1.0 point of published v0.6.3 baseline (97.8%)
- R@10 within 0.5 points (99.0%)
- R@20 within 0.2 points (99.8%)
- p95 search latency ≤ 35ms at 100k vectors on M2/M3 reference hardware

**What "fail" means:** Stop. Re-run plan against rabitq-rs as primary backend. Same trait, same hour count.

**Starter prompt:**

> You are running a recall-validation spike for ai-memory's v0.9 vector index decision. Do not modify any production ai-memory code. Build a standalone Rust binary in `/tmp/vectorlite-spike/` that: (1) loads a snapshot of an ai-memory SQLite database — use the LongMemEval reference corpus if no production snapshot is available, available at the path documented in `benchmarks/longmemeval/README.md` in the ai-memory repo; (2) reads existing MiniLM-384d embedding BLOBs from the `memories` table; (3) inserts them into a vectorlite-backed SQLite table using the `vectorlite` extension loaded via `sqlite3_load_extension`; (4) runs the LongMemEval question set against both the existing instant-distance HNSW and the vectorlite index; (5) reports R@5/R@10/R@20 and p50/p95/p99 latency for both. Output a JSON report at `/tmp/vectorlite-spike/results.json` with the exact schema documented in `benchmarks/results/longmemeval_v0.6.3.json`. If vectorlite recall is more than 1.0 R@5 point below the baseline, fail loudly with a non-zero exit and a one-line summary on stderr. If it passes, exit 0 with a one-line summary on stdout. Do not commit anything. The output of this spike is a go/no-go signal for the rest of the v0.9 plan.

**Gate criteria:** JSON report exists, exit code 0, R@5 delta ≤ 1.0 point. If gate fails, escalate to operator with the JSON results and stop.

---

## 1. Foundation layer (parallel after gate clears)

The trait contract published by Task 1.1 is the dependency boundary for every subsequent task. Once 1.1 lands, all downstream agents can start in parallel.

### Task 1.1 — VectorIndex trait + capability envelope schema

**Owner:** Agent A (foundation).
**Duration estimate:** ~30 minutes to publish trait stub; ~3 hours to complete full impl with default backend wiring.
**Blocks:** Tasks 1.2, 1.3, 1.4, 2.1, 2.2, 2.3, 3.1.

**Deliverables:**
- `src/index/mod.rs` defining the `VectorIndex` trait
- `src/index/error.rs` with `IndexError` enum integrated into ai-memory's existing error hierarchy
- `src/index/event.rs` with `IndexEvent` variants for the `signed_events` chain
- Capabilities envelope v3 extension documented in `docs/capabilities-v3.md`
- The trait stub committed and pushed to `feat/v0.9-vector-index` within 30 minutes so downstream agents can start

**Starter prompt:**

> You are Agent A. Your job is to publish the VectorIndex trait contract for ai-memory v0.9 within 30 minutes so five other agents can start in parallel against it. Create `src/index/mod.rs` defining a `pub trait VectorIndex: Send + Sync` with these methods: `insert(&mut self, id: u64, vec: &[f32], txn: &mut rusqlite::Transaction) -> Result<IndexEvent, IndexError>`, `delete(&mut self, id: u64, txn: &mut rusqlite::Transaction) -> Result<IndexEvent, IndexError>`, `search(&self, query: &[f32], k: usize, allowlist: Option<&[u64]>) -> Result<Vec<(u64, f32)>, IndexError>`, `dim(&self) -> usize`, `embedder_version(&self) -> u32`, `verify(&self, id: u64) -> Result<IndexedState, IndexError>`, `rebuild(&mut self, new_embedder_version: u32) -> Result<RebuildHandle, IndexError>`. Define `IndexEvent` as an enum with `Inserted`, `Deleted`, `Rebuilt`, `MigrationCompleted` variants, each carrying enough metadata (id, dim, embedder_version, timestamp, prev_hash) to be hashed into the Ed25519 `signed_events` chain that already exists in `src/audit/signed_events.rs`. Define `IndexedState` as `{ id: u64, dim: u32, embedder_version: u32, indexed_at: i64, last_verified: i64 }`. Define `RebuildHandle` as a join-handle that exposes `status() -> RebuildStatus` and `cancel()`. Define `IndexError` integrating with the existing `AiMemoryError` hierarchy. Add the new module to `src/lib.rs`. Commit and push the stub at minute 30 so other agents can branch from it. Then continue implementing the default backend wiring (Task 1.2 hands off to you here) — wire `Box<dyn VectorIndex>` into the `MemoryStore` struct in `src/store/mod.rs`, replacing the current direct `instant_distance::Hnsw` field, with the existing `instant_distance` impl temporarily satisfying the trait as a transitional adapter so the codebase compiles. Do not delete `instant_distance` code yet — the migration agent needs it readable for one more release cycle. Run `cargo build --all-features` after every step; do not push if the build is red.

**Gate criteria:** Trait compiles, tests pass, `cargo doc` produces clean output, MemoryStore now holds `Box<dyn VectorIndex>`. Push to `feat/v0.9-vector-index` so other agents can branch.

### Task 1.2 — vectorlite backend implementation

**Owner:** Agent B. Starts at minute 30 once trait stub is pushed.
**Duration estimate:** ~3 hours.
**Depends on:** 1.1 trait stub.

**Deliverables:**
- `src/index/vectorlite.rs` implementing `VectorIndex` against the vectorlite SQLite extension
- Build script `build.rs` additions to compile and link vectorlite if not present
- Cargo feature flag `vectorlite-backend` (default-on)
- Loader-detection code that gracefully falls back if `sqlite3_load_extension` is disabled

**Starter prompt:**

> You are Agent B. Wait until Agent A has pushed the VectorIndex trait stub to `feat/v0.9-vector-index`. Branch from there as `feat/v0.9-vectorlite-backend`. Your job is to implement `VectorIndex` for vectorlite. Create `src/index/vectorlite.rs` with a `pub struct VectorliteIndex { conn: rusqlite::Connection, dim: u32, embedder_version: u32 }`. The vectorlite extension must be loaded via `sqlite3_load_extension` against the same connection ai-memory uses for the main database — do not open a second connection. Add a `build.rs` step that downloads vectorlite v0.2.0 from its GitHub release artifact, verifies the SHA256 (pin the hash explicitly), and places the platform-appropriate `.so`/`.dylib`/`.dll` at `target/<profile>/vectorlite_extension/`. Use feature flag `vectorlite-backend` defaulting to on. The insert implementation must take a transaction handle and insert into a vectorlite virtual table named `vec_memories` with schema `(id INTEGER PRIMARY KEY, embedding BLOB)` using the vectorlite-specific INSERT syntax; the search implementation must use `knn_search(embedding, query, k)` with an optional `rowid IN (...)` allowlist that materializes the allowlist into a temp table for sets larger than 100 ids and uses an inline `VALUES` clause for smaller sets. Implement `verify(id)` as a SELECT-and-hash against the embedding BLOB. Implement `rebuild(new_embedder_version)` as a stub that returns NotImplemented for now — Agent F owns the full rebuild primitive. The constructor must detect whether `sqlite3_enable_load_extension` returned an error and, if so, return `IndexError::ExtensionLoadDisabled` so the daemon can fall back to the builtin backend. Write unit tests covering: insert/search round-trip, delete-then-search-misses, dim-mismatch-rejected, allowlist-filtered search, verify-after-insert. Do not write integration tests yet — Agent G owns those. Run `cargo test --features vectorlite-backend` after every change.

**Gate criteria:** Unit tests pass, vectorlite loads cleanly on Linux x64 + macOS ARM64 in CI, `IndexError::ExtensionLoadDisabled` returned correctly when extension loading is blocked.

### Task 1.3 — builtin (pure-Rust HNSW) fallback backend

**Owner:** Agent C. Starts at minute 30 in parallel with Agent B.
**Duration estimate:** ~2 hours.
**Depends on:** 1.1 trait stub.

**Deliverables:**
- `src/index/builtin.rs` implementing `VectorIndex` against `hnsw_rs` (a pure-Rust HNSW)
- Persistence via SQLite-backed embedding BLOB column + on-disk graph serialization
- Cargo feature flag `builtin-backend` (default-on)

**Starter prompt:**

> You are Agent C. Wait until Agent A pushes the trait stub. Branch as `feat/v0.9-builtin-backend`. Your job is to implement the pure-Rust HNSW fallback. Use the `hnsw_rs` crate (verify it's still maintained; if not, use `instant-distance` itself wrapped in the new persistence story). Create `src/index/builtin.rs` with `pub struct BuiltinIndex` that persists the graph state to `<db_dir>/vector_graph.bin` and reads embeddings from a `memory_embeddings` SQLite table colocated in the main ai-memory database. The key correctness property: graph mutations must be flushed to disk before the SQLite transaction commits, so a crash mid-write leaves graph and SQL in a consistent state. Use a write-ahead-log pattern: append graph deltas to `vector_graph.wal`, fsync, commit SQL, then merge WAL into the main graph file on next startup or every N operations. For search, use the existing `hnsw_rs` k-NN with a Rust-side allowlist filter applied as a `Predicate` if the crate supports it, or as a post-filter over an over-fetched candidate set (over-fetch factor of 3x for selective allowlists, dynamically computed). Implement the same trait method surface as vectorlite. Write the same unit test set Agent B writes. The builtin backend is slower than vectorlite but must produce identical search results to within the recall noise floor of HNSW determinism. Add Cargo feature `builtin-backend` defaulting to on. Run `cargo test --features builtin-backend` after every change.

**Gate criteria:** Unit tests pass identical to vectorlite suite, persistence round-trips cleanly across daemon restarts, recall within 0.5 R@5 points of vectorlite on the same corpus.

### Task 1.4 — backend selection logic + capabilities envelope

**Owner:** Agent D. Starts at minute 30.
**Duration estimate:** ~1 hour.
**Depends on:** 1.1 trait stub.

**Deliverables:**
- `src/index/factory.rs` with backend auto-detection
- `--index=auto|vectorlite|builtin` CLI flag
- `[index]` config section parsing
- Capabilities envelope v3 reports active backend

**Starter prompt:**

> You are Agent D. Wait for Agent A's trait stub. Branch as `feat/v0.9-index-factory`. Your job is the small glue layer that selects between vectorlite (Agent B) and builtin (Agent C) at daemon startup. Create `src/index/factory.rs` with `pub fn build_index(config: &IndexConfig, conn: &mut rusqlite::Connection) -> Result<Box<dyn VectorIndex>, IndexError>`. The selection logic: if `config.backend == "vectorlite"`, attempt to load vectorlite; if it fails with `ExtensionLoadDisabled`, return error (explicit selection means do not silently fall back). If `config.backend == "builtin"`, use builtin unconditionally. If `config.backend == "auto"` (default), attempt vectorlite first; on `ExtensionLoadDisabled` or any vectorlite-specific load error, log a structured warning to tracing and fall back to builtin transparently. Add a `--index <backend>` CLI flag to the daemon entry point in `src/bin/ai-memory.rs`. Parse `[index]` section from `config.toml` using the existing config parser pattern. Extend the capabilities envelope at `src/capabilities/v3.rs` to add: `index_backend: "vectorlite" | "builtin"`, `index_persistence: "embedded" | "hybrid"`, `filtering_mode: "predicate_pushdown" | "post_filter"`, `embedding_dim_enforced: bool`. The capabilities envelope must be regenerated on every daemon boot, not cached. Add structured logging when fallback occurs so operators can audit which backend is actually running. Tests: factory selects correctly under all three flag values, fallback path triggers on simulated extension-disabled error, capabilities envelope reflects active backend.

**Gate criteria:** All three backend selection modes work, capabilities envelope correctly reports active backend, structured logging fires on fallback.

---

### Task 2.1 — signed_events extension for index mutations

**Owner:** Agent E. Starts at minute 30.
**Duration estimate:** ~2 hours.
**Depends on:** 1.1 trait stub, in particular the `IndexEvent` enum.

**Deliverables:**
- Schema migration extending `signed_events` to include index mutation rows
- Ed25519 signing wired into trait insert/delete/rebuild paths
- V08-PE-8 audit-trail verifier extended to walk index events as first-class

**Starter prompt:**

> You are Agent E. Wait for Agent A's trait stub. Branch as `feat/v0.9-index-signed-events`. Your job is to make every vector index mutation a first-class event in the Ed25519-signed audit chain that already exists in `src/audit/signed_events.rs`. Add a schema migration (next version after v34 sqlite / v33 postgres ladder — coordinate with `src/storage/migrations.rs::CURRENT_SCHEMA_VERSION`) extending the `signed_events` table to support new event kinds: `IndexInserted`, `IndexDeleted`, `IndexRebuilt`, `IndexMigrationCompleted`. Each row carries `event_kind`, `memory_id`, `dim`, `embedder_version`, `prev_hash` (linking to the previous row in the chain), `sequence` (monotonic), `signature` (Ed25519 over canonical-bytes), `timestamp`. Modify the `VectorIndex` impls (Agent B and Agent C work, which you'll coordinate with via daily-equivalent sync — share a Slack-equivalent channel) so that insert/delete/rebuild emit the event into the chain inside the same transaction as the index mutation. Extend `ai-memory verify-audit-trail` (the V08-PE-8 verifier) to recognize the new event kinds and validate that for every `IndexInserted` event, a corresponding `memories` row exists with the same id and matching embedder_version; for every `IndexDeleted` event, the index actually shows the id as absent. Add a test that intentionally corrupts the chain (modify one signed_events row by hand in a test DB) and verifies the verifier catches it. Coordinate with Agents B and C on the exact signing call site — the simplest design is a `SignedEventEmitter` injected into each `VectorIndex` impl constructor, so the impl calls `emitter.emit(IndexEvent::Inserted { ... }, txn)` inside its insert path. Run `cargo test --all-features` after every change.

**Gate criteria:** Index mutations appear in `signed_events`, signatures verify, audit-trail verifier catches injected corruption, chain hash continuity holds across all mutation types.

### Task 2.2 — embedding_dim + embedder_version schema and enforcement

**Owner:** Agent F. Starts at minute 30.
**Duration estimate:** ~1.5 hours.
**Depends on:** 1.1 trait stub.

**Deliverables:**
- Schema migration adding `embedding_dim INTEGER NOT NULL` and `embedder_version INTEGER NOT NULL` to `memories`
- Type-layer rejection of mismatched-dim writes via trait
- Capabilities envelope reports active embedder dim
- G4 closes

**Starter prompt:**

> You are Agent F. Wait for Agent A. Branch as `feat/v0.9-embedding-dim-enforcement`. Your job is to close G4 from the v0.6.3 audit — the silent tolerance of mixed embedding dimensions that produces cosine=0.0 on mismatch. Add a schema migration extending `memories` with `embedding_dim INTEGER NOT NULL` and `embedder_version INTEGER NOT NULL`. The migration must backfill these columns from the existing data: `embedding_dim` is computed as `length(embedding) / 4` for any existing rows where the embedding BLOB is non-null (since f32 = 4 bytes); `embedder_version` defaults to 1 for all pre-migration rows. After the migration, the `memory_store` path must read the active embedder's dim from a new `embedder_registry` table (also added in this migration) and refuse to write any memory whose embedding dim does not match the registered dim for the currently-active embedder. Mismatched writes return `MemoryStoreError::EmbeddingDimMismatch { expected, got }` — a typed error, not a silent zero score. Extend the capabilities envelope v3 to report `embedder_active: { name, dim, version }` and `embedder_dim_violations_count: u64` (a monotonic counter incremented per rejected write, persisted to a small bookkeeping table for the `ai-memory doctor` command). Test cases: write with correct dim succeeds, write with wrong dim returns typed error and increments counter, capabilities envelope reflects active embedder, migration on a database with mixed-dim history correctly identifies the latest dim as the active one and flags older rows for re-embedding via Agent G's rebuild primitive. Run `cargo test` after each change.

**Gate criteria:** Mismatched-dim writes rejected at API boundary, migration handles historical mixed-dim databases gracefully, capabilities envelope shows enforcement state, counter increments correctly.

### Task 2.3 — namespace pre-filter rewiring

**Owner:** Agent G. Starts at minute 30.
**Duration estimate:** ~2 hours.
**Depends on:** 1.1 trait stub and Agent B's vectorlite allowlist parameter.

**Deliverables:**
- Removal of the post-ANN Rust-side namespace filter from `src/recall/mod.rs`
- New filter path using `allowlist` parameter on `VectorIndex::search`
- Ship-gate test case proving small-namespace queries no longer return zero results
- §5.2 "production hazard" closes

**Starter prompt:**

> You are Agent G. Wait for Agent A's trait stub AND Agent B's vectorlite backend reaching the allowlist-parameter-functional state (Agent B will signal via a tag `vectorlite-allowlist-ready` on their branch). Branch as `feat/v0.9-namespace-prefilter`. Your job is to delete the post-ANN namespace filter that lives in `src/recall/mod.rs::filter_by_namespace_post_ann` and replace it with a pre-filter using the new `allowlist` parameter on `VectorIndex::search`. The current flow is: ANN returns top-50, Rust filters by namespace, returns up to k. The new flow must be: resolve the target namespace and its descendants via `build_namespace_chain` (already exists), query the SQLite-backed `memory_namespace_index` table to get the full set of memory_ids in scope, pass that set as the `allowlist` parameter to `VectorIndex::search`. For the vectorlite backend this materializes into `rowid IN (...)` predicate pushdown handled by the SQLite query planner. For the builtin backend Agent C handled this with a Predicate or over-fetch. Add a critical ship-gate test scenario: create a parent namespace with 50k memories and a child namespace with 50 memories; issue a recall scoped to the child namespace; verify that all 50 child memories are eligible candidates and the result is not dominated by parent-namespace top-50 noise. This test should currently fail on main and pass after this change. Also add a test ensuring that an allowlist larger than the candidate pool returns exactly `min(k, len(allowlist))` results, not k-padded-with-junk. Coordinate with Agent E on whether the allowlist construction itself should be a signed event (likely no — too granular — but document the decision). Update `docs/recall.md` to describe the new pre-filter semantics.

**Gate criteria:** The named ship-gate test passes, post-ANN filter code is deleted (not commented out), recall budgets in PERFORMANCE.md are not regressed on the unfiltered path.

---

### Task 3.1 — migration from instant-distance to active backend

**Owner:** Agent H. Starts at minute 30 in parallel with foundation work, but final integration waits for 1.2 and 1.3 to land.
**Duration estimate:** ~2 hours.
**Depends on:** 1.1 trait stub; converges with 1.2 + 1.3.

**Deliverables:**
- `src/migration/v0_8_to_v0_9_index.rs` migration script
- Idempotent, restartable, progress-reporting
- Degraded-read-only daemon mode during migration
- `ai-memory migrate-index` CLI surface

**Starter prompt:**

> You are Agent H. Wait for Agent A. Branch as `feat/v0.9-index-migration`. Your job is the one-shot migration from the existing `instant-distance` in-memory HNSW to whichever VectorIndex backend the factory selects (vectorlite or builtin). The migration runs on first daemon boot after v0.9 install: detect schema version, if pre-v0.9, enter a `migrating` capability mode where read traffic is served from the existing instant-distance index but writes are buffered to a write-ahead-log. Walk the `memories` table in batches of 1000, reading existing embedding BLOBs, inserting each into the new backend via the trait, emitting `IndexMigrationCompleted` signed events at batch boundaries. The migration must be idempotent: if interrupted, the next start resumes from the last completed batch (track via a `migration_state` table with `last_completed_memory_id` and `total_memories`). On completion, atomically swap: the new index becomes the active one, the buffered writes from the WAL replay, the old `instant-distance` state is dropped from memory but its serialized form (if any) is retained on disk in `<db_dir>/.archive/` for one release cycle so operators can roll back. Surface progress via the capabilities envelope (`migration: { in_progress: bool, completed: u64, total: u64, eta_seconds: u64 }`) and the `ai-memory doctor` command. Add an explicit `ai-memory migrate-index --dry-run` CLI subcommand that walks the corpus without writing, reporting estimated time and any flagged anomalies (e.g., rows with NULL embeddings that would be skipped, mixed-dim rows that need re-embedding via Agent I's rebuild primitive). Test cases: clean migration of a 10k-row corpus completes, interrupted-and-resumed migration produces identical end state, dry-run reports match actual run, signed events form an unbroken chain across the migration boundary. Run `cargo test --all-features` after every change.

**Gate criteria:** Migration is idempotent, restartable, emits proper signed events, capabilities envelope reflects progress, dry-run matches actual run behavior.

### Task 3.2 — re-embedding rebuild primitive

**Owner:** Agent I. Starts at minute 30 in parallel.
**Duration estimate:** ~3 hours. This is the longest single-agent branch.
**Depends on:** 1.1 trait stub; converges with 2.2 (embedder_version).

**Deliverables:**
- Background worker for live re-embedding
- Atomic swap semantics with read-during-rebuild
- `memory_reindex` MCP tool
- Status surface via capabilities envelope and `doctor`

**Starter prompt:**

> You are Agent I. Wait for Agent A's trait stub and Agent F's `embedder_version` column landing (Agent F will tag `embedder-version-ready` on their branch). Branch as `feat/v0.9-rebuild-primitive`. Your job is the long-running re-embedding primitive that ai-memory needs for embedder transitions — the operation that's missing today and is worth more than any compression ratio. Implement `VectorIndex::rebuild(new_embedder_version: u32) -> RebuildHandle` for both vectorlite and builtin backends (coordinate with Agents B and C — you contribute the rebuild impl into each backend, they don't block on you). The semantics: when called, spawn a background tokio task that walks the corpus in batches of 500, generates new embeddings using the embedder registered for `new_embedder_version`, writes the new vectors into the backend tagged with the new version. During rebuild, reads serve from the union of (old-version vectors not yet rebuilt) ∪ (new-version vectors already rebuilt) — implementations differ between backends but the trait contract is: search returns results that are eventually-correct, never empty during rebuild. On completion, atomic swap promotes new-version vectors to active, old-version vectors are tombstoned for one release cycle (so rollback is possible), the active `embedder_version` in `embedder_registry` advances. Failure during rebuild leaves the index in a recoverable state: the next rebuild call resumes, or `--rollback` reverts to the previous version. Add a `memory_reindex` MCP tool exposing this operation with parameters `{ new_embedder_version: u32, dry_run: bool, max_concurrency: u32 }`. Surface status via capabilities envelope (`reindex: { in_progress, current_version, target_version, completed, total, eta_seconds }`) and `ai-memory doctor`. Emit `IndexRebuilt` signed events at batch boundaries via Agent E's signing infrastructure. Test cases: full rebuild on a 10k corpus, rebuild interrupted mid-stream resumes correctly, reads during rebuild return non-empty results, dry-run does not modify state, rollback restores previous state. Run `cargo test --all-features` after every change.

**Gate criteria:** Rebuild completes correctly, reads during rebuild are non-empty, signed events form unbroken chain, rollback works, the MCP tool surface is correct and documented.

---

## 4. Verification + ship gate (parallel after integration)

These six agents launch only after Agents A–I have merged. They run concurrently.

### Task 4.1 — ship-gate phase 1–4 against new index

**Owner:** Agent J.
**Duration estimate:** ~1.5 hours.
**Depends on:** All foundation + audit-chain + migration work merged.

**Starter prompt:**

> You are Agent J. The v0.9 vector index work has merged to `feat/v0.9-vector-index`. Your job is to run the full ship-gate 4-phase certification against the new index and surface any regressions. Run `cargo run --bin ship-gate -- --phases all --backend vectorlite` and `--backend builtin`. Phase 1 (functional) must show CRUD operations and MCP handshake all green. Phase 2 (federation) must show W=2 of N=3 quorum holding with the new index. Phase 3 (migration) must demonstrate clean round-trip of the new index between SQLite and Postgres SAL adapters. Phase 4 (chaos) must show 50× kill_primary_mid_write convergence ≥0.995 — this is the critical one for the new transactional-coherence story since the whole point of moving the index inside SQLite is that crashes don't desync it. If any phase fails, file an issue tagged `v0.9-ship-gate-blocker` with the failing scenario, the diff in behavior vs. v0.6.3 baseline, and a hypothesis about the root cause. Do not attempt to fix — that's a downstream agent's job. The wall time budget is the v0.6.3 baseline (~14 minutes total); if any phase exceeds 2x its baseline wall time, that's also a regression to flag. Output a structured report at `benchmarks/results/ship_gate_v0.9.json` matching the v0.6.3 schema.

**Gate criteria:** All four phases green for both backends, wall times within 2x of baseline, structured report committed.

### Task 4.2 — A2A-gate certification

**Owner:** Agent K.
**Duration estimate:** ~1 hour.
**Depends on:** All merges complete.

**Starter prompt:**

> You are Agent K. Run the full A2A-gate matrix against v0.9: ironclaw × {off, tls, mtls}, hermes × {off, tls, mtls}, mixed-framework × {off, tls, mtls}. Use both vectorlite and builtin backends. The certification cell is ironclaw-mtls — 48/48 scenarios must pass on both backends. If any scenario fails, file `v0.9-a2a-gate-blocker` with the failing scenario id, the framework × transport cell, and the diff vs. v0.6.3 baseline. Mixed-framework cells were blocked on terraform topology at v0.6.3 baseline — that's still acceptable to skip, but confirm via the ship-gate harness that the skip is for the same terraform reason and not a regression introduced by v0.9. Output a structured report at `benchmarks/results/a2a_gate_v0.9.json`. Total wall budget ~28 minutes matching v0.6.3 baseline.

**Gate criteria:** ironclaw-mtls 48/48 on both backends, baseline cells match v0.6.3, no new mixed-framework regressions.

### Task 4.3 — LongMemEval full disclosure

**Owner:** Agent L.
**Duration estimate:** ~1 hour.
**Depends on:** All merges complete.

**Starter prompt:**

> You are Agent L. Run LongMemEval against v0.9 with the full reranker-variant disclosure that was a v0.6.3.1 commitment: reranker-on, reranker-off, curator-on, three variants × two backends (vectorlite, builtin) = six runs. Publish R@5, R@10, R@20, and throughput (keyword + LLM-expanded) for each. The v0.6.3 published baseline is R@5=97.8%, R@10=99.0%, R@20=99.8%. Any drop greater than 0.5 R@5 points on the default-config (vectorlite + reranker-on + curator-on) run is a release blocker — file `v0.9-recall-blocker`. Smaller drops are acceptable but must be documented honestly in the v0.9 release notes per the project's audit-honesty discipline. Output structured JSON at `benchmarks/results/longmemeval_v0.9_<variant>.json` for each variant. Regenerate the recall comparison chart at `docs/recall_v0.9.svg`. Update `benchmarks/longmemeval/README.md` with the new numbers and a one-paragraph honest summary of any regressions.

**Gate criteria:** Default-config R@5 within 0.5 points of baseline, all six variants reported, charts regenerated.

### Task 4.4 — PERFORMANCE.md baselines + bench CI guard

**Owner:** Agent M.
**Duration estimate:** ~1 hour.
**Depends on:** All merges complete.

**Starter prompt:**

> You are Agent M. Update `PERFORMANCE.md` with v0.9 baselines for both vectorlite and builtin backends on the M2 reference hardware. The relevant budgets that may shift: `memory_store` keyword/semantic/autonomous (insert-time impact of writing the new index event), `memory_recall` semantic/autonomous (the structural change from post-filter to pre-filter), `memory_search` keyword (unchanged but verify). Add new budgets for new operations: `memory_reindex` per-1000-memories (LLM-bound, document the expected range), `migration` per-1000-memories (one-shot, document expected wall time). Run `cargo bench --bench v0_9_index_budgets` to populate. Update `performance/baseline.json` with the new numbers. The CI guard at `.github/workflows/bench.yml` already fails PRs that exceed budget by >10% — verify it still triggers correctly on a synthetic 15%-regression test. Document in `PERFORMANCE.md` the expected difference between vectorlite (faster, persistent) and builtin (slightly slower, more portable) so operators can make informed selection.

**Gate criteria:** baselines updated, CI guard verified, documentation clearly compares backends.

### Task 4.5 — ai-memory doctor integration + V08-PE-8 verifier extension

**Owner:** Agent N.
**Duration estimate:** ~1 hour.
**Depends on:** All merges complete.

**Starter prompt:**

> You are Agent N. Extend `ai-memory doctor` and the V08-PE-8 audit-trail verifier with the new v0.9 surface. Add to `ai-memory doctor`: an "index drift" check that counts rows in `memories`, counts vectors in the active backend, counts `IndexInserted` − `IndexDeleted` events in `signed_events`, and flags any mismatch. An "embedder dim violations" check that reports the `embedder_dim_violations_count` from the capabilities envelope. An "index backend" status line showing the active backend, whether fallback occurred at boot, and persistence mode. A "rebuild status" section showing any in-progress reindex operation with progress and ETA. Extend `ai-memory verify-audit-trail` to walk the new `IndexInserted | IndexDeleted | IndexRebuilt | IndexMigrationCompleted` event types: verify Ed25519 signatures, verify prev_hash chain continuity across the boundary between memory events and index events, verify that for every `IndexInserted` there exists a corresponding `memories` row, verify that for every `IndexDeleted` the memory_id is absent from the active index (call the trait's verify(id) method). Add a test that injects a mismatch (delete a vector from the index without emitting the signed event) and confirms the verifier catches it. Update `docs/security/audit-trail-coverage.md` with the new event kinds.

**Gate criteria:** doctor reports all new fields correctly, verifier catches injected mismatches, audit-trail-coverage doc reflects new event surface.

### Task 4.6 — release notes, CHANGELOG, capabilities envelope v3 doc

**Owner:** Agent O.
**Duration estimate:** ~1 hour.
**Depends on:** All merges complete; ideally also after Agents J–N have reported so release notes can honestly state results.

**Starter prompt:**

> You are Agent O. Write the v0.9 release notes at `docs/v0.9.0/release-notes.md` and update `CHANGELOG.md`. The release notes structure follows v0.7.0's: executive summary in three sentences, the audit-gap-closures section (G2, G3, G4, §5.2 post-ANN filter), the new capabilities section (vectorlite + builtin backends, rebuild primitive, embedder-version tracking, signed index events), the migration section (how operators upgrade, what they'll see, rollback path), the cutline section (anything that slipped from this scope), the honest disclosures section (any recall regressions from Agent L's report, any performance shifts from Agent M's report, any new dependencies). Update `CHANGELOG.md` with the standard ai-memory format. Update `docs/capabilities-v3.md` with the full v3 envelope schema including all v0.9 additions. Update `ROADMAP2.md` to mark §5.4 G2, G3, G4 as SHIPPED with the v0.9.0 reference, and to mark the §5.2 post-ANN filter hazard as RESOLVED. The release notes must follow the project's honesty discipline — do not paper over any regressions Agents J–N reported. If LongMemEval R@5 dropped by 0.3 points on builtin backend, say so explicitly. Cross-link to the structured JSON benchmark results so readers can verify independently.

**Gate criteria:** Release notes are honest about regressions, CHANGELOG follows project format, ROADMAP2.md correctly updated, capabilities-v3.md complete.

---

### Task 5.1 — tag, sign, push to all 5 channels

**Owner:** Agent P (release agent).
**Duration estimate:** ~30 minutes if all channels are healthy.
**Depends on:** All 4.x verification tasks green.

**Starter prompt:**

> You are Agent P. v0.9.0 is ready to cut. Verify all of these are green before tagging: ship-gate 4/4 on both backends, A2A-gate ironclaw-mtls 48/48 on both backends, LongMemEval default-config R@5 within 0.5 of baseline, PERFORMANCE.md baselines updated and CI guard passing, doctor + verifier integration complete, release notes honest. If any of these is red, refuse to cut and escalate. If all green: GPG-sign the tag `v0.9.0` against the operator key, push to GitHub, and trigger the release pipeline that publishes to crates.io, Homebrew tap, Fedora COPR, Docker GHCR, APT PPA. After each channel publishes, run the smoke test (`memory_capabilities` returns valid v3 envelope showing `index_backend` populated correctly). If any channel smoke test fails, file `v0.9-distribution-blocker` with the channel name and the failure mode but do not roll back the tag — the standard procedure is to ship a v0.9.0.1 patch through the failing channel. Update the public landing pages (ship-gate, A2A-gate) by posting the new result JSON to the auto-update endpoint. Post a release announcement to the project's standard channels following the v0.7.0 template.

**Gate criteria:** Tag exists, signed, all 5 channels publish, smoke tests pass, landing pages updated.

---

## 6. Critical-path timing summary

```
Hour 0.0–0.75 ┃ Task 0.1 spike (BLOCKING, single agent)
              ┃   ├─ Gate: vectorlite recall holds
              ┃   └─ If fail: re-plan with rabitq-rs as primary
Hour 0.75–1.0 ┃ Task 1.1 trait stub (single agent)
              ┃   └─ Trait stub pushed at minute 30 of 1.1
Hour 1.0–4.0  ┃ PARALLEL FAN-OUT (9 agents on 9 tasks)
              ┃   ├─ Agent A: 1.1 (continues, full impl)
              ┃   ├─ Agent B: 1.2 vectorlite backend (3h)
              ┃   ├─ Agent C: 1.3 builtin backend (2h)
              ┃   ├─ Agent D: 1.4 factory + capabilities (1h)
              ┃   ├─ Agent E: 2.1 signed events (2h)
              ┃   ├─ Agent F: 2.2 dim enforcement (1.5h)
              ┃   ├─ Agent G: 2.3 namespace prefilter (2h, after B tags)
              ┃   ├─ Agent H: 3.1 migration (2h)
              ┃   └─ Agent I: 3.2 rebuild primitive (3h, after F tags)
Hour 4.0–5.5  ┃ INTEGRATION (single coordinator + fix agents)
              ┃   └─ Merge order: A → C → B → D → F → E → G → H → I
Hour 5.5–7.0  ┃ PARALLEL VERIFICATION (6 agents)
              ┃   ├─ Agent J: 4.1 ship-gate (1.5h)
              ┃   ├─ Agent K: 4.2 A2A-gate (1h)
              ┃   ├─ Agent L: 4.3 LongMemEval (1h)
              ┃   ├─ Agent M: 4.4 PERFORMANCE.md (1h)
              ┃   ├─ Agent N: 4.5 doctor + verifier (1h)
              ┃   └─ Agent O: 4.6 release notes (1h, blocks on J-N partial)
Hour 7.0–7.5  ┃ Task 5.1 release (single agent)
              ┃   └─ Five channels published, smoke-tested, landing pages updated
```

**Floor:** 4 hours if no surprises and the spike clears in 20 minutes.
**Expected:** 7 hours with normal integration friction (1–2 fix cycles).
**Ceiling:** 10 hours if a recall regression triggers a backend swap (re-plan against rabitq-rs).

---

## 7. Risk register

**Recall regression at the spike (Task 0.1).** Mitigation: pre-planned re-run against rabitq-rs as primary, same trait, same hour count. Decision belongs to the operator, not the spike agent.

**vectorlite extension loading blocked in target environment.** Mitigation: Agent D's factory falls back to builtin transparently. End user sees a structured warning, not a daemon failure.

**Windows ARM64 vectorlite untested.** Mitigation: Agent J's ship-gate runs on Windows ARM64 CI runner if available; if not, document the gap and ship builtin as default on Windows ARM64 for one release cycle while CI catches up.

**Migration interrupted on a 10M+ memory database.** Mitigation: Agent H's idempotent restartable design. The state table tracks last completed memory_id; resumption is automatic on next start.

**Audit chain hash mismatch under concurrent write load.** Mitigation: Agent E's signing emitter is sequenced through a single tokio task with bounded mpsc channel — no parallel signing, no chain race. This is correct-but-not-maximally-parallel; if throughput becomes the binding constraint later, a Merkle batching strategy can be added behind the same interface.

**Build-time download of vectorlite extension binary fails.** Mitigation: pin the GitHub release artifact hash, fail the build loudly if SHA mismatch, document the offline-build path using a vendored binary in `vendor/vectorlite/`.

---

## 8. Out of scope for v0.9 (explicitly deferred)

- **Quantization backends.** RaBitQ-IVF, TurboQuant, residual VQ — all pluggable via the trait but not shipped in v0.9. Belongs in v0.10 or later when corpus size demands it.
- **GPU acceleration.** Not a substrate concern. Commercial-tier deployment may add this behind the same trait.
- **Per-namespace HNSW shards.** Mentioned in ROADMAP2.md §5.4 G2 fallback; the namespace pre-filter solution (Task 2.3) addresses the same hazard more cleanly. Revisit if filter selectivity becomes a measured problem.
- **Asymmetric distance computation** (compressed corpus, full-precision query). Quantization-era concern. Defer.
- **Streaming consistency under data-dependent quantization** (December 2025 paper). Research direction; not yet a production primitive.

---

## 9. Definition of done

v0.9 ships when all of these hold simultaneously:

1. Tasks 0.1, 1.1–1.4, 2.1–2.3, 3.1–3.2, 4.1–4.6, 5.1 have closed against their gate criteria.
2. ROADMAP2.md §5.4 G2, G3, G4 are marked SHIPPED with v0.9.0 references.
3. ROADMAP2.md §5.2 "post-ANN namespace filter production hazard" is marked RESOLVED with v0.9.0 reference.
4. The VectorIndex trait is documented, the two backends ship, the factory selects correctly under all flag combinations.
5. The release notes honestly disclose any regressions found by Agents J–N.
6. Five distribution channels publish v0.9.0 with passing smoke tests.
7. Public landing pages (ship-gate, A2A-gate) reflect v0.9.0 results.

---

*Document classification: Public-facing development plan. Apache 2.0. v0.9.0 target.*

alphaonedev:
Closes #1860

## Summary

Ships the **vectorlite opt-in ANN backend** — the headline deliverable of #1860 (the vector-index substrate deferred from #1005) — as a coherent, **default-OFF, byte-identical-when-disabled** slice that honors the North Star (data integrity is the highest-order constraint): the ANN index is a **derived, disposable artifact**, and every failure path is **fail-closed / degrade-never-corrupt**.

[vectorlite](https://github.com/1yefuwang1/vectorlite) is an Apache-2.0 C++ SQLite **loadable extension** (hnswlib + Google Highway SIMD). There is **no Rust crate** for it, and upstream ships prebuilts only inside `vectorlite_py` wheels (4 platforms; no linux-arm64). A build-time native-lib acquisition step would break the 3-OS CI matrix, the 5 release channels, and reproducible builds — the exact blockers #1860 documents. This slice resolves that by doing **zero build-time acquisition**: the OFF-by-default `vectorlite` cargo feature only enables rusqlite's `load_extension` API (bundled SQLite unchanged) and loads an **operator-acquired** extension binary at runtime from `AI_MEMORY_VECTORLITE_EXTENSION` (pinned-SHA256 acquisition helper: `scripts/fetch-vectorlite.sh`). The feature is deliberately in **no CI job or release channel** — cross-matrix native-lib availability stays an operator/release-infra decision.

## Data-integrity contract (North Star)

- **Derived/disposable index.** In-memory (`Connection::open_in_memory`), rebuilt from the durable memory text/embeddings at every boot — no durable state, so the deferred on-disk-persistence + 2PC-coherence substrate re-platforming (RISK #1 in the #1860 review) is not pulled in.
- **Load/smoke failure at boot** → the backend-resolving funnels (`hnsw::boxed_configured_index` daemon / `hnsw::arc_configured_index` MCP stdio, both sharing `vectorlite::from_env`) log an ERROR and **fall back to the default pure-Rust HNSW backend**. Durable data untouched.
- **Hard failure mid-life** → **degrade in place** to a freshly-built default `VectorIndex`, re-seeded from a retained entry mirror (eviction sink re-wired; removed ids stay removed). Worst case = a brief not-fully-searchable window (fewer results), **never wrong results**.
- **Construction smoke-verifies** the full virtual-table contract (create / insert / `knn_search` / delete / re-insert) before the backend is eligible — an ABI-mismatched or non-vectorlite library is rejected at construction, never mid-life.
- **Strict-dim boundary is structural** (vectorlite's virtual table is dimension-strict — a #1005 G4 superset; never zip-truncates a distance). **Capacity** honors the #1005 G2 evict-oldest / `hard-fail-at-cap` semantics with the same `max_entries_reached` eviction-sink events.

## Files changed (9, +1111/-9 net vs base)

- `src/vectorlite.rs` (new, ~750 LOC) — `VectorliteIndex: VectorSearchIndex`, `try_with_path` / `from_env` / `boxed_from_env`, smoke-check, insert/search/remove/lifecycle, degrade-to-default self-heal, 5 unit tests.
- `src/hnsw.rs` — `boxed_configured_index` / `arc_configured_index` backend-resolving funnels; hoisted `pub const EVICTION_REASON_MAX_ENTRIES` SSOT (deletes the vectorlite local duplicate — no-hardcoded-literals rule).
- `src/lib.rs` — `#[cfg(feature = "vectorlite")] pub mod vectorlite;`
- `src/daemon_runtime.rs` / `src/mcp/mod.rs` — boot paths route through the funnels (identical to the default backend unless the feature + env knob select vectorlite).
- `Cargo.toml` — `vectorlite = ["rusqlite/load_extension"]` (no new dependency).
- `scripts/fetch-vectorlite.sh` — operator acquisition helper (pinned upstream SHA256s; fail-closed on digest mismatch).
- `tests/vectorlite_1860.rs` — fail-closed construction + funnel fallback (runs with no native lib), plus `#[ignore]`d live seam-parity + capacity/eviction suites.
- `CHANGELOG.md` — `[Unreleased]` entry.

No new MCP tool / HTTP route / CLI subcommand / DB migration — **no SSOT-count move**.

## Tests / gates (all green locally)

- `cargo fmt --all --check` ✅
- `cargo check --all-targets` (default) / `--features sal` / `--features sal-postgres` / `--features vectorlite` / `--examples` ✅
- `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` (default) + `--features vectorlite` + `--features sal -D warnings` ✅
- `cargo test --features vectorlite --test vectorlite_1860` → 1 passed, 2 ignored (live) ✅
- `cargo test --features vectorlite --lib vectorlite::` → 5 passed ✅
- `cargo test --test feature_flag_audit_arch_11` ✅ (feature registered)
- `cargo test --test qual_10_module_size_ceiling --test qual_6_7_legacy_error_type_ceiling` ✅
- lint gates: `check-hardcoded-literals` / `check-vendor-literals` / `check-docs-vs-ssot` / `check-migration-ladder` ✅
- `cargo audit` ✅ (exit 0; 2 pre-existing allowed advisories)

The live seam-parity + capacity/eviction suites (`#[ignore]`d) were verified locally against the pinned linux-x86_64 v0.2.0 extension binary per the wip checkpoint.

## Residual deferred scope (flag for operator — do NOT let it orphan)

This PR delivers the vectorlite **backend** slice only. #1860 also carried, deferred from #1005, a set of heavier substrate items that this slice deliberately does NOT ship (each a data-integrity-sensitive re-platforming, correctly left out rather than force-fit): on-disk persistent index + 2PC cross-domain coherence, u64 re-keying of vector ids, signed `IndexInserted`/`IndexDeleted` events, a builtin persistent WAL backend, the one-shot migration+rebuild primitive, and the rabitq-rs alternate backend. Recommend the operator spin a fresh **v1.x** tracking issue for these before this closes #1860, so nothing is orphaned (prime directive: every gap tracked).

## AI involvement (per docs/AI_DEVELOPER_WORKFLOW.md §8.2)

Authored by an AI NHI agent under operator authorization (v1.0.0 autonomous campaign). The vectorlite backend architecture (opt-in runtime-load, fail-closed derived index, no build-time acquisition) was designed + operator-authorized 2026-07-18 in the wip checkpoint; this PR completes the SSOT literal-dedup, merges the latest `release/v1.0.0`, and takes the slice through the full CI-parity gate set. No 5-agent crossroads vote was newly required — the architecture decision was pre-made under operator authorization, and the completion work (literal hoist, gate runs, merge) is mechanical / precedent-copying (T6-exempt).


alphaonedev:
## v1.0.0 campaign adversarial pre-merge audit — PR #2219 (#1860 vectorlite backend)

Audit gate per the campaign protocol (#2204/#2205/#2213/#2214/#2216 class). Audited against the actual worktree code, the live upstream v0.2.0 extension binary, and the seam consumers.

### Verdict: **MERGE-CLEAN** (0 blocking findings; 3 advisories, none merge-gating)

### Lens results

**1. Forget reaches the index — PASS.** There is **no persisted index in this PR**: `Connection::open_in_memory()` (`src/vectorlite.rs:197`), rebuilt from the durable, `#2167`-space-filtered seed query (`daemon_runtime.rs::load_boot_index_entries` → `db::get_all_embeddings(&conn, active_space)`) at every boot. The on-disk-persistence substrate stays deferred, so the #2208-adjacent persisted-residue class (crypto-erased embeddings surviving in a file) is structurally absent. `remove` reaches the backend through the same trait call sites as the default backend (MCP `src/mcp/tools/delete.rs:231`, HTTP `src/handlers/memories.rs:573`, consolidate) — the funnels change no call site. Ordering is right: `vl_remove` purges mirror+maps **before** the SQL DELETE (`src/vectorlite.rs:462-473`), so even a degrade fired mid-remove reseeds the fallback WITHOUT the removed id — a forgotten row can never resurrect across the degrade edge. Verified live: remove-then-search excludes the id (`tests/vectorlite_1860.rs:126-130`, run against the real binary — see evidence below).

**2. Derived-index honesty / fail-closed — PASS.** No durable artifact ⇒ no torn-file / stale-generation / restored-DB-phantom classes. Construction smoke-verifies the full SQL contract before eligibility; any load/smoke failure falls back to the default backend (verified: `try_with_path` on a non-extension file and a missing path both `Err`, funnel still returns a working default index). Every mid-life SQL error degrades in place to a default `VectorIndex` reseeded from the mirror; mirror-first mutation ordering in insert/evict/remove keeps the reseed consistent on every error edge I traced. Unknown rowids from a search are dropped rather than attributed (`src/vectorlite.rs:530-537`); rowids are monotonic and never reused, so a stale hit cannot alias a different memory.

**3. embedding_space (#2167) discipline — PASS.** The boot seed is space-filtered in SQL (layer 1), and `semantic_phase` re-verifies every candidate row-side via `cosine_similarity_space_checked` regardless of which backend produced the candidate ids — the index is candidate-selection only, so a cross-space vector cannot score. Strict-dim is structural (table pinned to first accepted dim; mismatched insert rejected, mismatched query returns empty — live-verified). G2 capacity semantics live-verified: cap-3 evicts m0,m1 oldest-first onto the eviction sink; hard-fail-at-cap refuses the overflow insert with nothing evicted.

**4. Opt-in posture / dependency — PASS.** `vectorlite = ["rusqlite/load_extension"]` only — **no new crate, Cargo.lock untouched** (verified), no build-time acquisition, feature in no CI job/release channel. Default build byte-identical: without the feature `boxed_configured_index`/`arc_configured_index` compile to exactly the prior constructor calls; with the feature but env unset, `from_env` → `None` → default (unit-pinned). Supply chain: **all 4 SHA256 pins in `scripts/fetch-vectorlite.sh` re-downloaded from upstream and verified byte-exact** (linux-x86_64 `85df3a88…`, macos-arm64 `668d5e95…`, macos-x86_64 `eb682db3…`, win-amd64 `a9a7c789…`); digest checked before unzip, mismatch aborts. The `unsafe load_extension` is operator-trust-bounded (env-pointed path, `LoadExtensionGuard` re-disables, private in-memory conn, all SQL crate-authored/parameterized — the only interpolations are a const table name, integer dim/cap, and i64 rowids).

**5. Visibility / §5.2 allowlist — PASS.** The index stores ids only; every hit is hydrated and passed through the same visibility + lifecycle ladder in `semantic_phase` as default-backend hits, so no cross-namespace/owner leak lane is added. Allowlist parity: live test passes, and I additionally probed the exact SQL shape (`knn_search(...) AND rowid IN (...)`) against the real extension with an allowlist whose members are NOT in the global top-k — vectorlite **pushes the rowid filter into the ANN walk** (returned the far allowlist members, not empty), so the #1005 §5.2 small-namespace no-starvation contract holds, it is not a post-filter.

**6. Tests — PASS.** Evidence run in this audit against the pinned upstream binary:
- `cargo test --features vectorlite --lib vectorlite::` → 5/5 (degrade paths, sink survival, env-unset, fail-closed construction)
- `AI_MEMORY_VECTORLITE_EXTENSION=… cargo test --features vectorlite --test vectorlite_1860 -- --include-ignored` → **3/3 including both live suites** (seam parity incl. forget-then-search + duplicate-replace + strict-dim; capacity/eviction/hard-fail)
- `cargo clippy --all-targets --features vectorlite -- -D warnings -D clippy::all -D clippy::pedantic` → clean (matters: the feature is in no CI job, so this gate is local-only)
- Lock-order check: `set_eviction_sink` drops the sink lock before taking state; `degrade`/`vl_evict_oldest` take sink while holding state but the sink lock scope never takes state — no inversion.

### Advisories (non-blocking; file as v1.x follow-ups)

- **A (LOW, seam-contract honesty).** `is_fully_searchable` returns `true` during an in-flight `seed_and_rebuild_async` window (`src/vectorlite.rs:588-595` + `607-625`): the seed thread inserts one-by-one, so partially-seeded coverage is reported as full. Consumers: `proactive_conflict_check_with_index` (storage/mod.rs:8288) may draw candidates from the partial index — explicitly documented safe-direction (a missed conflict only ALLOWS a write); semantic recall serves partial ANN + full keyword during the window — fewer results, never wrong ids. The boot loops themselves join the seed handle first, so the "index-backed" INFO line stays truthful. Minimal fix: an `AtomicUsize` pending-seed counter; `is_fully_searchable` → healthy && pending == 0, plus a test pinning it.
- **B (LOW, perf).** Under opt-in `AI_MEMORY_VECTOR_NAMESPACE_ALLOWLIST`, `vl_search` materializes `rowid IN (…)` as a joined string per query — O(allowlist) SQL build (~800 KB at 100k ids) per recall vs the default backend's lazy iterator. Correctness fine (pushdown verified); cost note only.
- **C (INFO).** `vl_evict_oldest` uses `Vec::remove(0)` — O(n) memmove per eviction at cap; a `VecDeque` would be O(1). Only bites in sustained past-cap churn.

Also endorsing the PR's own residual-scope flag: the deferred #1860 substrate items (on-disk persistence + 2PC coherence, u64 re-keying, signed index events, WAL backend, migration primitive, rabitq-rs) need a v1.x tracking issue before #1860 closes so nothing orphans.

Merge-gate disposition: **MERGE-CLEAN** — advisories A-C to a follow-up issue, not this PR.

PR #2205, opened 2026-07-18
feat(#2006): integrity-complete Portability-v2 exporter + FAIL-CLOSED importer

alphaonedev:
Closes #2006

## What

Rebuilt fresh on schema v85 (vote `34bbf781`). Closes the #1944 residue: the DEFAULT `ai-memory export` is a memories+links convenience view that DROPS the signed audit/governance spine (a durability data-loss on a pipe-to-file "backup"). This ships the integrity portability path — and its symmetric, fail-closed importer.

### Exporter — `ai-memory export --full`
Emits the complete Portability-v2 envelope (spec `docs/spec/PORTABILITY-V2.md`): `spec_version="2"`, `db_schema_version`, and every §V2-2 signed record class — `signed_events`, `memory_revisions`, `forget_tombstones`, `agent_lineage`, `model_attestations`, `governance_rules`, role `trust_anchors` — **byte-preserved** (a `HexBytes` newtype keeps every `Vec<u8>` signature/hash byte-exact instead of serde's default number-array). The `conformance_level` (L1/L2/L3) marker is **COMPUTED** from an in-export re-verify pass — a source whose audit chain is broken honestly downgrades to L1 — alongside a per-class `conformance_by_class`. Trust anchors carry enrolled-role PUBLIC keys only (a private key never crosses).

### ★ Importer — LOSSLESS + FAIL-CLOSED (the highest data-integrity path)
`import` detects a v2 envelope by `spec_version` and routes to `import_full_envelope`, a BULK WRITE into source-of-truth memory TEXT that is **ALL-OR-NOTHING**:
- Every class stages inside ONE transaction; the imported signed spine is re-verified (`verify_audit_trail`) BEFORE commit.
- A malformed / tampered / truncated bundle (a broken `prev_hash` link, a sequence gap, or detected tail-truncation) is REJECTED with the transaction rolled back — a rejected bundle applies **ZERO rows** (never a partial apply).
- Signed classes are RAW-inserted byte-preserved — the importer **NEVER re-signs**; `agent_lineage` `record_bytes` is recomputed byte-identically via the record's own canonical-CBOR encoder and its witnesses ride the `signed_events` array (never re-witnessed / double-anchored).
- Idempotent (id-keyed memories + `INSERT OR IGNORE` signed classes); tombstones-before-admit; governance rules verify-or-drop; trust anchors advisory-only.
[...]

Issue #2006, opened 2026-07-13
[v1.x][R72] Integrity-complete JSON exporter — emit+import round-trip of the signed record classes (deferred from #1944)


alphaonedev:
## Carrier for the integrity-complete JSON exporter deferred from #1944 (v1.0.0)

Per the v1.0.0 GA triage + the B_WARN 5-agent vote (`woaiwndla`, 10/10), landed via PR #2005: for v1.0.0 the JSON `export` command was DE-SILENCED (stderr WARN + in-band `export_scope`/`portability_complete:false`/`excludes` markers on CLI + HTTP) rather than made complete. The **integrity-complete exporter** (the A_EMIT scope) defers here.

### Deferred scope (v1.x)
- Extend `ai-memory export` to EMIT the omitted signed/tamper-evidence spine: governance rules, `memory_revisions`, `forget_tombstones`, lineage edges, agent/model attestations, `signed_events`.
- Symmetric IMPORT that round-trips those classes (L2/L3 conformance per docs/spec/PORTABILITY-V2.md).
- On landing: flip `portability_complete` → true, rescope the de-silencing WARN, update claims.

### Why safe to defer (v1.0.0)
§11.6 DISCHARGED (Portability Spec v2 FINAL + reader.py + reader.mjs + CC0 corpus); `ai-memory backup` (VACUUM INTO) = lossless path, `export-forensic-bundle` = signed egress. GA does NOT claim `export` round-trips the integrity spine.

Cross-ref: #1944 (closed), PR #2005, vote woaiwndla, docs/spec/PORTABILITY-V2.md §V2-7.

alphaonedev:
## v1.0.0-GA scope verification (autonomous perfect-endpoint sweep, 2026-07-14)

Code-verified that the **v1.0.0-GA-gating portion of #2006 is already satisfied** (via PR #2005, merged 2026-07-13). The integrity-complete exporter (A_EMIT) is correctly v1.x-deferred; no v1.0.0 code change is warranted here.

**Shipped (the GA gate — B_WARN de-silencing, vote `woaiwndla`):**
- `src/export_scope.rs`: `PORTABILITY_COMPLETE = false`, `SCOPE_MEMORIES_LINKS`, `OMITTED_SIGNED_CLASSES`, `export_scope_warn()`.
[...]

Issue #1859, opened 2026-07-01
[G13-mem / #1828a] Memory-derivation lineage-DAG — consolidation conserves + navigable provenance


alphaonedev:
Split from #1828 per the G13 design's §26.5 scope-honesty (2×5-vote accepted). **#1828's title is agent-identity lineage (key-loss = death → signed succession DAG); this child (#1828a) is the memory-DERIVATION provenance DAG** the v0.9.0 EPIC brief scoped G13 as. #1828 stays open for the identity-lineage facet (separate design).

**The real gap (precise):** most memory-derivation infra already ships (the `derived_from`/`reflects_on`/`derives_from` provenance relations exist in the CHECK allowlist; reflection/atomisation/promote/persona already write child→parent edges; `cid` is already the lineage node id; ancestry traversal exists on both backends). BUT `db::consolidate` **hard-DELETEs its source memories** (recording provenance in a non-navigable `metadata.derived_from` id array, precisely because a real edge would be `ON DELETE CASCADE`-killed) and does not emit a CONSOLIDATE revision leaf on SQLite. This makes a **multi-hop** lineage DAG impossible: `store A → reflect R (R reflects_on A) → consolidate [R]→C` hard-deletes R → CASCADE-deletes R's edge → C's ancestry to A is unreachable. Same class as the G7 (#1824) contradiction hard-delete.

**Fix (11-pass, both backends, additive/opt-in, lineage_dag default-on):** consolidation TOMBSTONES (not erases) sources under `consolidate_tombstone_sources`, writes navigable `derived_from` edges carrying `source_cid`/`target_cid` (additive columns), emits exactly-one CONSOLIDATE leaf per source (reconciling the pre-existing PG-emits/SQLite-doesn't divergence), a P-restricted acyclic ancestry/descendants query (SQLite CTE + Postgres AGE with Deferred→CTE read-your-own-write routing), and MCP/HTTP/CLI/SDK lineage surfaces. Proven acyclic (single-node) + multi-hop-queryable. Hub #1821. Ships with v0.9.0.

alphaonedev:
G13-mem shipped on release/v0.9.0 — all CI green on both commits.

**Commits (both SSH-signed, Good signatures):**
- e57d22b0 `feat(lineage)` — the full implementation (44 files, +3380/-102)
- 8c22d6d9 `test(lineage)` — per-module coverage lift (handlers/links.rs, store/mod.rs)

[...]

PR #2229, opened 2026-07-19
fix(#2215): Portability-v2 round-trip preserves the v75 lineage-DAG cid mirror

### LINKED ISSUE #2215: [#2006 residual] v2 import does not repopulate memory_links.source_cid/target_cid — imported edges lose the v75 tombstone-resilient lineage mirror
Observed while fixing #2208-#2211 on PR #2205 (`src/portability/import.rs`, links loop). Severity: LOW (degrade-not-corrupt fidelity gap).

## Gap

The v2 importer's raw `INSERT OR IGNORE INTO memory_links` carries 9 columns but NOT the schema-v75 `source_cid`/`target_cid` lineage-mirror columns (#1859 G13-mem). `MemoryLink` (the wire model) does not carry them either — they are stamped at link-write time from the endpoints' `memories.cid` — so every imported edge lands with NULL mirrors.

## Why this is degrade, not corrupt

The lineage traversal reads the stored mirror when present and falls back to a LEFT JOIN on the live endpoint rows (`src/models/link.rs` lineage view). So imported edges still resolve while both endpoints are live. What is LOST is the mirror's tombstone-resilience: after a source is later tombstoned at the destination, an imported edge cannot resolve stable node identity the way a natively-written edge can.

## Proposed fix (small)

In the import links loop, after the endpoint memories are staged, stamp the mirrors the same way the native link-write path does (probe each endpoint's `memories.cid` and bind `source_cid`/`target_cid` into the INSERT), gated on `lineage_dag_enabled()` parity with the native path. ~15 LOC in `src/portability/import.rs` + one regression test (import → tombstone source → lineage walk still resolves cid identity).

Found during the #2208-#2211 fix pass; deliberately not folded into that surgical security diff.

alphaonedev:
Closes #2215

## What

The Portability-v2 import (landed via PR #2205) did not repopulate `memory_links.source_cid`/`target_cid` — so every imported edge lost the schema-v75 (#1859 G13-mem) tombstone-resilient lineage mirror. A #2006 residual fidelity gap the #2205 re-audit confirmed was the ONLY remaining one.

Root cause was two-sided: the raw importer `INSERT INTO memory_links` carried nine columns but not the two cid columns, AND the wire model `MemoryLink` did not carry them either (they are stamped at link-write time from the endpoints' `memories.cid`), so even a fresh exporter dropped them. **Both sides are fixed for a lossless round-trip.**

### Exporter-side verdict — the twin fix WAS needed
`MemoryLink` had no `source_cid`/`target_cid` fields and `storage::export_links` did not project them, so the v2 envelope could not carry them at all. Fixed:
- `MemoryLink` gains `source_cid` / `target_cid` (`#[serde(default, skip_serializing_if = "Option::is_none")]`, so the wire is byte-identical for the common NULL-mirror row). These are advisory-resolution only (COND 2, #1859) — NOT part of the Ed25519 `SignableLink` preimage — so adding them is byte-compat with every shipped signature and federated peer.
- `storage::export_links` (used by the v2 exporter `emit::build_full_envelope` AND the SAL sqlite `export_links` delegate) now projects the two columns.

### Importer-side fix
The `import` links loop resolves the mirror, gated on `lineage_dag_enabled()` for byte-parity with the native `create_link_signed` write path (DAG OFF → NULL, the legacy shape):
1. **PREFER** the bundle's cid (a fresh exporter round-trips the source's real cid);
2. else **BACKFILL** from the just-staged endpoint's re-derived `memories.cid` (#1825 deterministic re-derivation — an older bundle that predates the exporter fix heals from the live endpoint);
3. else leave **NULL** — the pre-v75 legacy state. DEGRADE, do not INVENT.

The write lands on the sqlite `Connection` (v2 import is sqlite-only, like MCP); the other SAL/pg link read projections stay selective (`None`, consistent with the existing `signature`/`attest_level` selective-projection posture).

## Data-integrity posture (North Star)
The memory TEXT stays the durable source of truth; the cid mirror is a derived, regenerable content-address. The fix never fabricates a cid — it prefers the bundle, backfills only from a resolvable endpoint, and otherwise leaves NULL (which the query layer still resolves via the live endpoint LEFT JOIN while both endpoints are present). No `SignableLink` bytes change, so no attestation is invalidated.

## Tests
`tests/portability_lineage_cid_2215.rs`:
- `round_trip_carries_lineage_cids` — export a linked pair carrying cids → import into a FRESH dest → the imported edge carries the same `source_cid`/`target_cid`. FAILS pre-fix on BOTH the parsed-envelope assertion (exporter twin) AND the imported-row assertion (importer).
- `legacy_bundle_backfills_from_endpoint_cids` — an older bundle with the cids stripped backfills from the staged endpoints.
- `legacy_bundle_without_dag_leaves_null` — absent-cid bundle + DAG OFF binds NULL, not garbage.

Regression: `tests/portability_roundtrip_2006.rs` (2/2) + the `src/portability/import.rs` inline fail-closed suite (15/15) stay green.

## Scope / SSOT
No new MCP tool / HTTP route / CLI subcommand / DB migration (the schema-v75 columns already exist) — no SSOT-count move. `MemoryLinkRelation::COUNT` unchanged (no new relation). CHANGELOG `[Unreleased]` entry added.

## Gates (local, green)
`cargo check --all-targets` default + `--features sal` + `--features sal-postgres` + `--examples`; `cargo clippy --all-targets --features sal` AND default, `-D warnings -D clippy::pedantic`; `cargo fmt --all`; `qual_10_module_size_ceiling` + `qual_6_7_legacy_error_type_ceiling` (no bump needed).

## AI involvement

Implemented autonomously by Claude (Opus 4.8) as an ai-memory v1.0.0 campaign worker: issue + #2205 re-audit read, exporter-vs-importer root-cause, both-sides fix, three-branch regression tests, full local CI-parity gate run. Human operator authored the task scope and gates the merge. Per `AI_DEVELOPER_WORKFLOW.md` §8.2.


alphaonedev:
## v1.0.0 campaign pre-merge audit (Fable adversarial gate) — PR #2229 / #2215

**VERDICT: MERGE-CLEAN** (one pre-existing residual filed as #2233 — NOT a #2229 defect; two verified-acceptable observations below).

### Lens 1 — round-trip losslessness, both sides (and did the exporter need the twin fix?)
VERIFIED YES on both sides. The exporter twin fix was REQUIRED and correctly landed at the shared chokepoint: `emit::build_full_envelope` builds its `links` array via `crate::storage::export_links` (`src/portability/emit.rs:191`), and pre-fix that SELECT carried 8 columns with the wire model lacking the fields entirely — so every envelope link was cid-less regardless of the importer. Post-fix `export_links` (`src/storage/mod.rs:12316-12347`) selects `ml.source_cid, ml.target_cid` and maps them onto the two new `MemoryLink` fields; the importer (`src/portability/import.rs:467-522`) widens the raw 9→11-column `INSERT OR IGNORE` and binds the resolved pair. Wire compat is clean: the fields are `#[serde(default, skip_serializing_if = "Option::is_none")]`, so NULL-mirror rows (and every federation `sync` payload built from them, since `cli/sync.rs` shares `export_links`) stay byte-identical, older receivers ignore the unknown fields (no `deny_unknown_fields` on the model), and the fields are correctly OUTSIDE the Ed25519 `SignableLink` preimage (COND 2, #1859) so no shipped signature is invalidated.

### Lens 2 — legacy bundles: NULL, never fabricated
VERIFIED, with a precise reading of "not fabricated": a legacy bundle (no cids) triggers the step-2 BACKFILL from the just-staged endpoint's `memories.cid` — that value is the endpoint's own deterministic v74 genesis-derived cid (re-derived identically at the destination under `trust_source`), i.e. exactly what the native `create_link_signed` stamp would mint; it is derivation, not invention. When the endpoint carries no cid (pre-v74) or is absent, the mirror binds NULL (step 3) — and the edge itself is skipped by `OR IGNORE` when an endpoint is missing, so no orphan-mirror row can land. The DAG-off arm binds NULL unconditionally (native-off byte-parity). All three arms are test-pinned (see lens 5).

### Lens 3 — does mirror population bypass or respect the #2205-landed tombstone/verification gates?
RESPECTS THEM — the mirror cannot function as an admission side-channel. Ordering in `apply_all_classes` is gate-first: (1) `forget_tombstones` stage (with the #2208 live-probe refusal), (2) memories (tombstoned ids SKIPPED), then links. A link whose endpoint was tombstone-refused has no destination row, so the `INSERT OR IGNORE` skips the whole edge — the cid mirror rides ONLY on edges the gates already admitted; it never re-admits content, carries no content bytes (it is a `b3:` address), and the backfill probe (`read_memory_cid`) reads only rows the gated memories loop actually staged in the same tx. The spine re-verification lanes (`verify_audit_trail`, revision-chain, agent-lineage FORGED refusal) are untouched. Residual accepted-model note: step 1 PREFERS the bundle-carried cid verbatim, so an operator-doctored bundle can stamp a mirror that disagrees with the live endpoints' cids. This is consistent with the import module's established trust posture (it already imports `signature`/`attest_level`/content bytes verbatim — the L2 byte-preserving contract) and the mirror is advisory-resolution only, never attested (outside the `SignableLink` preimage) — preferring the bundle value is also load-bearing for FIDELITY (the mirror is a link-creation-TIME stamp; re-deriving from live endpoints would lose it for tombstoned-endpoint edges, the exact case the mirror exists for). No gate is weakened.

### Lens 4 — sqlite/pg parity
CLEAN. Portability emit/import is a rusqlite-`Connection` surface (no pg lane exists to diverge). The pg read projections (`get_links`/`list_links`/timeline, `src/store/postgres.rs`) return `None` for the new fields — matching the sqlite read projections (`storage::get_links` also `None`; the SAL migrate projection `src/store/sqlite.rs:599` `None`, the documented `attest_level` selective-projection precedent; a sqlite→pg `migrate` re-stamps mirrors natively at the destination write, equivalent under deterministic cids). Struct-literal updates across pg/sqlite/federation/hooks/migrate test fixtures are mechanical `None` adds. Both `--features sal` / `sal-postgres` builds compile per CI.

### Lens 5 — test quality
GOOD. `tests/portability_lineage_cid_2215.rs` pins all three arms through the REAL JSON wire form (serialize→deserialize of the envelope, so `skip_serializing_if` round-trip is exercised, not just in-memory structs): (a) `round_trip_carries_lineage_cids` fails pre-fix on BOTH sides (with only the model fields present, exporter-unfixed ⇒ `wire.source_cid == None` fails the first assert; importer-unfixed ⇒ dest edge NULL fails the second); (b) `legacy_bundle_backfills_from_endpoint_cids` pins the backfill arm by stripping the wire cids; (c) `legacy_bundle_without_dag_leaves_null` pins the required legacy-NULL case. The process-wide `AtomicBool` is correctly serialized via a file-local mutex with poison-recovery (the `lineage_dag_g13.rs` precedent; integration-test binary = own process, so no cross-file race).

### Filed residual (pre-existing, NOT introduced here) — #2233
`lineage_dag` is deliberately NOT boot-seeded in production (`src/daemon_runtime.rs` ~L1080-1095 discards the resolved values; `lineage_dag_enabled()` reads FALSE unseeded). PR #2229 correctly copies the native write-path gate, which means in production processes the import (like every native link write today) binds NULL mirrors until the documented seed follow-up lands — behavior-consistent and non-lossy (the mirror is derived/advisory), but the "tracked follow-up" named in the daemon_runtime comment had no discoverable tracker issue, so the audit filed #2233. No change requested on this PR.

Audit basis: worktree `.local-runs/wt-2215-lineage-mirror` @ `e2e9f1f2`. Campaign audit gate: v1.0.0 pre-merge adversarial batch (EPIC #1940 terminal-state directive, North-Star data-integrity + degrade-don't-invent lens). Known pre-existing windows failure (#2230) out of scope per dispatch.

Issue #2225, opened 2026-07-19
Erasure cold tier (#2064): probe budget bounds cost but a poison row still starves bundling of newer rows (R3-residual)


alphaonedev:
Tracked follow-up from the PR #2213 durability re-verify (Fable, R3-residual LOW). Cited: the #2213 R1-scoped re-verification (2026-07-19).

## Finding
The #2064 sweep's R3 fix (`SWEEP_PROBE_BUDGET_PER_TICK`, PR #2213) bounds the per-tick filesystem PROBE cost so a permanently-failing (poison) row — non-UTF-8 TEXT, or the F4 non-finite REAL that now bails loud — can no longer reopen the O(N)-per-tick re-probe. But the probe budget does NOT un-pin the keyset frontier: a poison row plus >4096 (`SWEEP_PROBE_BUDGET_PER_TICK`) successors past it means the frontier never advances past the poison row, so newer archived rows beyond the per-tick probe window are not bundled until the poison row is fixed. Bounded + loud (the poison row WARNs every tick, durable archived rows are intact — a redundancy gap, never corruption), but bundling of the tail is starved while the poison row persists.

## Proposed fix
A bounded process-static poison-id skip-set: after N consecutive failures a row is added to the skip-set and the frontier is allowed to advance PAST it (the auditor's F2/R3 "retry set the frontier may pass"), with the skipped ids re-attempted at a low cadence. Cap the set size (degrade to current behavior beyond the cap). Add a test: a poison row + tail rows → the tail eventually bundles while the poison row keeps WARNing.

Bounded, opt-in tier; redundancy-gap (degrade), not corruption.

alphaonedev:
Fixed via PR #2227, squash-merged to release/v1.0.0 at de795db4 (all CI green incl. macos after the /private/tmp symlink PathBuf-key test fix). The erasure sweep now maintains a bounded process-static poison-id skip-set (threshold 3 consecutive failures, cap 4096 FIFO-evict, WARN naming id+reason, restart-clears = self-healing) so a permanently-failing archived row can no longer pin the keyset frontier and starve tail bundling; skipped_poison surfaced in SweepReport. Starvation + restart-clear behaviors test-pinned. Verified per prime directive pm-v3 (memory cd8ede94).

Issue #2224, opened 2026-07-19
Erasure cold tier (#2064): quarantined-purge content has no operator inspect/purge verb (R6)


alphaonedev:
Tracked follow-up from the PR #2213 durability re-verify (Fable, R6 LOW). Cited: the #2213 R1-scoped re-verification (2026-07-19).

## Finding
The #2064 erasure cold tier's journal-write-failure degrade path (`journal_purge_intent_best_effort` WARNs and proceeds if the fsync'd `.purge-intent/` marker can't be written) means a genuinely-PURGED id whose marker never landed is treated as an UN-journaled rowless bundle → QUARANTINED (`.quarantine/`), not destroyed. That is the correct fail-safe (never destroy without recorded intent), BUT: quarantined purged content is then never auto-destroyed and there is no operator verb to inspect or purge the quarantine holding pen. Destruction intent never completes for those bytes.

Same gap applies to any bundle the reconciler quarantines that the operator later confirms WAS a purge (not DB loss) — they want to complete the destruction, and today the only path is a manual `rm -rf <dir>/.quarantine`.

## Proposed fix
A quarantine inventory + purge verb, e.g. `ai-memory erasure quarantine list|purge [<id>|--all]` (or an env-gated reconciler sweep), built on the existing `ErasureStore::list_quarantined_ids` + a new `purge_quarantined`. Emit an audit row per destroyed bundle. Weigh against the SSOT CLI-count surface (new subcommand) vs. an env-only knob mirroring `AI_MEMORY_ERASURE_RECOVER_QUARANTINE` (#2064 F1). Whatever ships, the quarantine WARN should name the inspect verb.

Bounded, opt-in tier; no data-integrity risk (quarantine PRESERVES data — the residual is that destruction-intent completion is manual).

alphaonedev:
**Recorded v1.x ruling** per the `b682c76a` TRUTH+PRESERVATION decision (2x5 adversarial vote, 2026-07-18).

This is v1.x-class because the current behavior — quarantined-purge content is never destroyed without a recorded intent — is the correct fail-safe under the North Star (worst case is retained-but-inspectable, never silently-gone). The fix here is a genuinely destructive operator verb (inspect/purge), which deserves the same `#2089`-class governance-gating treatment as skill hard-purge, not a bolt-on addition inside the freeze window.

Staying OPEN as the v1.x tracker for the quarantine inspect/purge verb.

PR #2262, opened 2026-07-19
fix: run erasure cold-tier sweep off the HTTP request lock (pre-ship 3x7)

alphaonedev:
## Summary

Pre-ship 3x7 code battery HIGH availability finding on the #2064 erasure cold tier: the daemon gc loop ran the erasure sweep (`archive_sync::gc_tick` — up to `SWEEP_LIMIT_PER_TICK`=256 Reed-Solomon encodes + per-shard fsyncs, plus the reconcile/scrub pass) while holding the ONE global `Db = Arc<Mutex<(Connection, …)>>` that every HTTP handler serializes on (`src/daemon_runtime.rs::spawn_gc_loop_with_shadow_retention`; lock type in `src/handlers/transport.rs`). With the advertised `AI_MEMORY_ERASURE_COLD_TIER` knob enabled and a backlog draining, the ENTIRE API stalled for seconds-to-tens-of-seconds every gc tick.

References #2064 (erasure cold tier), #2213 (audit invariants), #2225 (poison skip-set).

## Topology chosen: (a) dedicated connection, wrapped in `spawn_blocking`

- New `archive_sync::gc_tick_detached(db_path)` opens its **own** connection from the daemon's stored DB `PathBuf` and runs the unchanged `gc_tick` against it. The gc loop calls it **after** dropping the handler mutex, inside `tokio::task::spawn_blocking`, and **awaits** the join so two sweeps of one store can never overlap (the keyset frontier + rotating reconcile cursor are single-sweeper state).
- **Why (a) is self-contained:** the sweep is DB-READ-ONLY — its only writes are bundle files. It touches `archived_memories`/`memories` with SELECTs only; it needs none of the shared in-memory HNSW/handler state. Cross-connection concurrency with the purge/restore funnels is the **already-supported regime** (a CLI `archive purge` runs on its own process/connection against a live daemon — the reconciler's grace-window (`RECONCILE_GRACE_SECS`) and R5 stale-intent age gates explicitly reason about a concurrent in-flight purge). Chunked lock-drop (option b) was unnecessary: no step needs the shared connection.
- **Fail-closed fallback:** non-file-backed databases (`:memory:` / `mode=memory` URIs / empty path) stay on the legacy under-lock tick via `archive_sync::is_in_memory_db_path` — a fresh connection there opens a DIFFERENT, empty database, and the detached reconciler would mis-classify every live bundle as rowless (quarantining the lot). Classification is conservative: a false positive only re-serializes the sweep (correct, merely slower), never yields a wrong result. `gc_tick_detached` itself bails LOUD if handed such a path.

## #2213 invariants preserved (untouched, not re-proved)

- **Purge-intent journal ordering** — journal-BEFORE-`DELETE` lives in the purge funnels (`journal_purge_intent_best_effort` / `remove_bundles_best_effort`), which run on the mutator's own connection and are not modified.
- **Quarantine escape-proofing** — `reconcile_one_rowless` classification (journaled → hard-reap; un-journaled → quarantine, never destroy) keys on durable filesystem + DB state, never on connection identity.
- **Poison skip-set (#2225)** — process-static, keyed by store DIRECTORY; the dedicated connection resolves the identical `<db>.erasure` sibling, so frontier/poison/cursor semantics carry over byte-identically. Only WHERE the sweep's connection comes from changes.

## Tests

`tests/erasure_cold_tier_2064.rs` — suite now **17/17** (was 14):
- `detached_sweep_completes_while_handler_mutex_is_held` — the sweep encodes + fsyncs to completion **while the `handlers::Db` mutex is held** by the test (structurally impossible pre-fix: the sweep only ran inside the lock), and writes nothing to the DB (read-only proof via the shared connection afterward).
- `detached_sweep_refuses_non_file_backed_paths` — fail-closed refusal + `is_in_memory_db_path` classification + disabled-tier strict no-op.
- `gc_loop_runs_detached_erasure_arm_for_file_backed_db` — end-to-end daemon wiring through the production `spawn_gc_loop_with_shadow_retention`.

Also green: `recall_purity_p01` (15/15 — drives the restructured gc loop), `qual_10_module_size_ceiling` (daemon_runtime ceiling 10_900 → 10_950 in lockstep, dated comment), `qual_6_7_legacy_error_type_ceiling`, `cargo check --all-targets` on default + `--features sal` + `--features sal-postgres`, `cargo clippy --all-targets` pedantic on default + sal, `cargo fmt --check`, `cargo check --examples`.

No new MCP tool / HTTP route / CLI subcommand / DB migration — no SSOT-count move. CHANGELOG `[Unreleased]` Fixed entry included.

## AI involvement

Authored end-to-end by an AI agent (Fable 5, hard-coder tier) in the ai-memory v1.0.0 autonomous loop under operator-delegated authority; design choice (dedicated connection vs chunked lock-drop) was orchestrator-delegated between two prescribed options and is recorded in the commit message. Do NOT merge — flagged for Fable audit (concurrency + durability surface).


alphaonedev:
## Fable pre-merge audit — #2064 erasure sweep off the HTTP request lock

**VERDICT: MERGE-CLEAN** (5-lens adversarial pass, special weight on the durability-race lens; verified against the worktree code at `bafd6cd5`, all cited tests re-run locally).

### Lens 1 — lock correctness: PASS
- **No heavy IO under the shared lock on any file-backed path.** All three `gc_tick` call sites accounted for: the daemon's under-lock arm (`daemon_runtime.rs:3508`) is gated `!erasure_detached` and survives ONLY for non-file-backed DBs; the detached arm (`:3546`) runs strictly AFTER `drop(lock)` inside `spawn_blocking`; `src/cli/gc.rs:41` uses its own per-invocation connection and never touches the handler mutex.
- **Dedicated connection sees committed rows, no stale snapshot / BUSY thrash.** `storage::open` (`src/storage/connection.rs:121-129`) sets `journal_mode=WAL` + `busy_timeout=5000`; WAL readers never block on the handler connection's writes, and the connection is opened FRESH each tick (no long-lived read snapshot — staleness is bounded by one tick). `open()` runs pragmas only, no migration writes, so the per-tick open is safe against a live daemon.
- **State-key identity holds.** `resolve_dir_for_conn` keys off `conn.path()`; the dedicated connection is opened from the same stored `PathBuf` the handler connection was, so the process-static frontier/poison/cursor resolve the identical `<db>.erasure` dir. Single-sweeper is preserved: the join is awaited, so two daemon sweeps of one store can never overlap.
- **Fail-closed `is_in_memory_db_path` is correctly conservative** — a false positive only re-serializes (correct, slower); `gc_tick_detached` also bails loud on such a path independently of the caller gate.

### Lens 2 — #2213 durability invariants (highest-risk lens): PASS
Verified against `src/erasure/archive_sync.rs` line-by-line, not just the PR prose:
- **Journal-before-DELETE untouched** — the purge funnels (`journal_purge_intent_best_effort` / `remove_bundles_best_effort`) are unmodified and run on the mutator's own connection; `remove_bundles_best_effort` still clears the marker only on successful remove and keeps it on failure (reaper gets the journaled survivor).
- **Quarantine escape-proofing intact** — `reconcile_one_rowless` (journaled → hard-reap; un-journaled → quarantine-never-destroy; DR mode → leave active) keys on durable filesystem + DB state, never on connection identity. The live-row guard (a live-but-not-archived row's bundle is never reaped on a transient read race) still holds.
- **R5 stale-marker clear intact** — the row-present + marker-age >= grace clear at the reconcile walk is unchanged; the age gate still skips a concurrent in-flight purge.
- **The gc sweep call graph is DB-READ-ONLY**, verified: the only `INSERT` in the module (`insert_archived_row_from_payload`, line 628) is reachable solely from `try_restore_archived_row_from_bundle` (the restore path) — `sweep_archive_bundles` / `reconcile_and_scrub` / `scrub_one` / `bundle_one_row` do SELECTs only. The read-only proof in the new test (archived count observed through the SHARED connection after the sweep) pins this.
- **No new resurrection/destruction race.** Adversarial interleavings walked: (a) sweep reads a row mid-purge — SQLite tx atomicity means the sweep's SELECT sees pre- or post-commit state, never partial; (b) purge removes a bundle while the scrub `get`s it — the scrub's re-mint reads the (now deleted) archived row and errors into the absorbed WARN arm, no wrong re-mint; (c) restore-vs-sweep — the live-row guard covers it; (d) two reapers on one bundle — only the CLI-gc-vs-daemon cross-process case, pre-existing and unchanged (rename/remove races collapse to one winner + absorbed error).

### Lens 3 — correctness regression: PASS
`tests/erasure_cold_tier_2064.rs` **17/17** (was 14 — the three new tests are net-new coverage, none of the 14 prior pins were modified); `recall_purity_p01` **15/15** (drives the restructured gc loop end-to-end incl. the fold-before-gc pin); `qual_10_module_size_ceiling` 2/2 (daemon_runtime ceiling bumped in lockstep with a dated comment, house discipline). Frontier/cursor semantics carry over byte-identically — only WHERE the connection comes from changed.

### Lens 4 — shutdown: PASS
The gc `JoinHandle` is aborted at shutdown; aborting the async awaiter does not cancel a running `spawn_blocking` body — it runs to completion in the blocking pool and the dedicated `Connection` is a local dropped at the end of `gc_tick_detached` (clean close, and per-tick anyway: no long-lived handle to leak). A hard process kill mid-encode is exactly the crash case the F3 machinery already covers (temp-dir + per-shard/manifest/dir fsync + atomic rename + `reap_stale_temp_dirs`), so no half-written bundle is ever served.

### Lens 5 — the new tests genuinely demonstrate off-lock: PASS
`detached_sweep_completes_while_handler_mutex_is_held` holds the REAL `handlers::Db` mutex (`blocking_lock`) for the entire sweep and requires the sweep to complete bundling + fsync — under the pre-fix topology this call graph deadlocks by construction, so the test is structurally load-bearing, not decorative. `gc_loop_runs_detached_erasure_arm_for_file_backed_db` proves the PRODUCTION `spawn_gc_loop_with_shadow_retention` wiring drives the detached arm; `detached_sweep_refuses_non_file_backed_paths` pins the fail-closed classification + the disabled-tier strict no-op.

### Non-blocking observations (recorded, not fixes-required)
1. **OBS-1 — `row_exists` swallows query errors as `false`** (`archive_sync.rs:761-764`, pre-existing `.unwrap_or(false)`). Under the new topology a transient read error on the dedicated connection could misclassify a live row's bundle as rowless. Walked the consequences: un-journaled → QUARANTINE (preserved; the sweep re-mints from the durable row next tick — self-heals, zero loss); the journaled-aborted-purge corner → bundle reap but the durable row survives and re-mints. The durable truth is never destroyed, and WAL + 5s busy_timeout makes the trigger near-unreachable. A fail-closed refinement (propagate the error, skip the id this tick) is a clean hardening follow-up.
2. **OBS-2 — the HTTP-purge vs sweep re-mint interleaving is newly REACHABLE in daemon-only deployments** (pre-fix the mutex serialized them; the CLI-purge regime always had it, so it is not a new race CLASS). Worst case: purged content re-minted from the sweep's pre-purge row read, then QUARANTINED after the grace window — preserved-hidden, NOT resurrectable (`try_restore_archived_row_from_bundle` refuses journaled ids; quarantine is invisible to `get`/restore), i.e. destruction is delayed, not violated. The #2224 quarantine-inventory/purge verb is the tracked operator control. Consistent with the R1-adjudicated posture (preservation outranks prompt destruction).

North Star check: no path trades data integrity for the availability win — the fix is topology-only, the sweep stays read-only, and the fail-closed in-memory fallback degrades to the old serialized behavior rather than ever risking a wrong reconciliation. Merge-ready from this audit's perspective; merge remains operator-gated.

Issue #1834, opened 2026-06-28
[TRACT-gap G20] Claim-level bitemporal absent (links carry valid_from/valid_until, claims do not)


alphaonedev:
## Gap G20 — TRACT §26 canonical-gaps catalog (UNTRACKED → tracked per v0.8.1 W7)

**Priority tier:** P2  ·  **Milestone:** v1.0  ·  **TRACT ref:** L1 bitemporal Claim model

### Gap statement
Bitemporal validity (`valid_from`/`valid_until`/`observed_by`) exists on **links** but not on **claims/memories**, so a memory's assertion cannot itself be time-bounded or as-of queried.

### Anchor(s) (CodeGraph-verified on `release/v0.8.0`)
- bitemporal columns exist only on links — `src/models/link.rs:319-328`

### Fix shape (from the canonical doc)
- extend bitemporal validity to the Claim/memory level so assertions are as-of queryable, not just their edges

### Provenance
Filed per the **v0.8.1 (Patch 1)** work order §W7 and the §24 prime directive (every gap is a tracked defect — no UNTRACKED gap may rot in a queue; the v0.9.0 epic starts on a clean tracker). Sources: `docs/design/TRACT-v0.8.0-DEVELOPMENT-GAPS-CANONICAL-opus.md`, `docs/reviews/corpus-completeness-21-agent-OPUS.md`, ROADMAP §26. Orchestration hub: #1821.

🤖 Filed by Claude Code (Opus 4.8, 1M context).

alphaonedev:
Columns LANDED on main (`7575bb83`): v79 adds `memories.valid_from` + `memories.valid_until` (RFC3339, nullable, additive — mirrors the memory_links temporal pair) riding the coordinated crypto-core schema window per the §27 ordering gate. Write/read-surface plumbing (store/update params, recall surfacing) is follow-on within this issue.

alphaonedev:
v1.0.0 triage (2026-07-13): link-level valid_from/valid_until exist (field_names.rs:385-387 + link.rs:396 valid_from:Option<String>), but full CLAIM-level bitemporal (the six-verb Claim algebra's temporal dimension) is a v1.x completeness feature with no struct-wiring at the claim level today. Not a GA correctness/security gate (no 'defaults lie'). Moved v1.0→v1.x.

alphaonedev:
[...]

PR #2265, opened 2026-07-19
fix(#1834): canonicalize claim valid-time to UTC so string comparison is correct (pre-ship 3x7)

alphaonedev:
## Summary

The 3x7 pre-ship code battery found a HIGH correctness defect in the #1834 claim-bitemporal VALID-time feature (ON by default post-#2235): `valid_from` / `valid_until` / `valid_at` are compared as UNNORMALIZED RFC3339 strings BYTE-WISE on every surface — the sqlite list (`build_list_query` ?), keyword `recall` (?13), hybrid FTS (?13) + semantic linear-scan (?11) SQL predicates, the HNSW Rust re-filter, and the postgres `list` ($8) / `recall_hybrid` FTS ($10) / semantic ($11) `::text` binds. RFC3339 admits many renderings of the SAME instant (`Z` vs `+00:00`, non-UTC offsets, variable fractional digits) that order WRONGLY as bytes (`'...Z' > '...+00:00'`, `'+'` 0x2B < `'Z'` 0x5A), so:

- a `Z`-rendered `valid_until` at the boundary instant was wrongly INCLUDED against a `+00:00`-rendered `valid_at` — silently violating the documented END-EXCLUSIVE contract;
- a `Z`-rendered `valid_from` at the boundary was wrongly EXCLUDED — violating START-INCLUSIVE;
- a `+05:00`-offset bound mis-filtered by HOURS.

Default config, both backends. Under the North Star this is the forbidden class: WRONG results, not degraded results.

## Fix chosen: (a) canonicalize-at-boundary (over (b) instant-typed predicates)

T-trigger adjudication (recorded here per dispatch): **(a)** was chosen because it (1) heals the ON-DISK data so byte comparison becomes exactly instant comparison everywhere at zero per-row read cost, (2) keeps every SQL predicate shape unchanged (no per-row parse, no per-row Rust re-filter on the SQL paths — (b) would force parse-per-row on sqlite since SQLite cannot natively compare arbitrary-offset RFC3339 at micro precision), and (3) gives one SSOT (`validate::canonicalize_valid_time`) instead of N predicate-site parsers that can drift. The canonical rendering is `YYYY-MM-DDTHH:MM:SS.ffffffZ` (`chrono::SecondsFormat::Micros`, `Z`): fixed-width, so lexicographic order ≡ chronological order and byte equality ≡ instant equality; micros matches postgres `timestamptz` resolution for cross-backend parity.

## What changed

[...]

PR #2263, opened 2026-07-19
fix(security): v2 import — verify attestation + validate rows before persist (pre-ship 3x7)

alphaonedev:
Closes the TWO HIGH defects the pre-ship 3x7 security battery found at the #2006 Portability-v2 import site (`src/portability/import.rs`, the `apply_all_classes` memories/links loops feeding `storage::insert_imported`). Bundles are UNAUTHENTICATED input per the importer's own threat model — the same lane already ships the #2208/#2211 defenses — so both were GA-reachable. References #2006.

## HIGH-1 — wire attestation persisted verbatim → forged `agent_attested` rows

The v2 memories loop kept wire-supplied `metadata.attest_level` / `write_signature` / `agent_pubkey` UNTOUCHED. A crafted bundle minted `attest_level=agent_attested` rows that this node's trust surfaces (`row_is_agent_attested`, #1948 quarantine routing, ClaimView, forensics) then believed. The SECURE default `trust_source=false` made it worse: it restamped `agent_id` but kept the stale attestation.

**Fix — re-derive, never copy (the #1464 federation discipline, reused not reinvented):**
- Every staged row's attestation now goes through the SAME `identity::attest::stamp_attestation` gate `handlers::federation_receive::apply_inbound_write_attestation` uses, including its re-attribution rule (a restamped row never verifies the ORIGINAL claimant's signature against the new attribution — lands `claimed`).
- Presented `write_signature` is verified against the attributed author's **DESTINATION-enrolled** Ed25519 key, **snapshotted BEFORE the import transaction** (`snapshot_dest_enrolled_keys`) — so a crafted `_agents` registration row staged earlier in the SAME bundle can never self-enroll the key a later row "verifies" against.
- Valid → `agent_attested`. Absent signature / unenrolled author / re-attribution → `claimed` (counted `attestation_downgraded`, WARN). Presented-but-**FORGED** → per-row SKIP (counted `forged_signature_skipped`, WARN) — the #1464 invariant that a deliberately bad signature is always rejected, never downgraded into a stored row.
- Under the default restamp posture the wire `metadata.agent_pubkey` identity-key claim is STRIPPED (unauthenticated input must not seed the `db::agent_pubkey` enrolled-key lookup surface).
- Byte-preserving fast path: a row that neither asserts an `attest_level` nor presents a signature crosses untouched (trusted-backup round-trips stay byte-exact).

## HIGH-2 — zero input validation (L1 parity gap vs `src/cli/io.rs:316/330`)

[...]

Issue #2207, opened 2026-07-18
#1834 residual (PR #2204): same-id federation merge lane drops a peer's valid_until close — merge_memory local-wins + both full-row overwrite writers omit the v79 VALID-time columns


alphaonedev:
Found by the Fable pre-merge audit of PR #2204 (claim-bitemporal VALID-time surface, #1834). BLOCKING for #2204's stated contract.

## Contract vs implementation

PR #2204's CHANGELOG, PR body, and both SQL upsert comments assert:

> `valid_until` … newer-wins under federation LWW **so a peer that CLOSED a claim replicates the close**.

That guarantee is implemented ONLY on the `(title, namespace)` dedup arm (`db::insert_if_newer` sqlite / `apply_remote_memory` pg upserts — correct newer-wins `CASE` on `excluded.updated_at`). It is **absent on the same-`id` federation conflict path** — `merge_inbound`, the load-bearing #224 CRDT-lite reconciler (v0.8.0 Pillar-3) that handles every replication update to an already-replicated row, i.e. the PRIMARY lane a claim-close actually travels:

1. **`src/models/crdt_merge.rs:501-502`** — `merge_memory` resolves the new fields as
   ```rust
   valid_from: local.valid_from.clone(),
   valid_until: local.valid_until.clone(),
   ```
   `valid_until` is LOCAL-wins, so a remote row that closed the claim (newer `updated_at`, `valid_until = Some(T)`) merges to the local open value (`None`). (`valid_from: local` is fine — it matches the SQL arms' `valid_from = memories.valid_from` immutability.)
2. **`src/storage/mod.rs` `overwrite_full_row_by_id`** — the full-row UPDATE SET list (`tier = ?1 … encrypted_envelope = ?28`) was not extended: `valid_from`/`valid_until` are not written at all on the sqlite merge lane.
3. **`src/store/postgres.rs` `merge_inbound`** — the pg full-row UPDATE (`$2..$28`) has the same omission.

So even after fixing (1), the merge output would be dropped by (2)/(3). Both backends are at least consistent with each other (no cross-backend skew), but both contradict the documented LWW discipline.

## Impact (North Star)

[...]

PR #2270, opened 2026-07-19
fix(#2258): expose valid_from on the store write surfaces (MCP/HTTP/CLI)

### LINKED ISSUE #2258: bitemporal: valid_from unreachable from every advertised write surface (MCP/HTTP/CLI hardcode None) while docs claim caller-supplied backfill
**Severity: LOW** — from the v1.0.0 3x7 pre-ship battery. Verified still present at `release/v1.0.0` HEAD `8223e65e`.

**Location:** `src/mcp/tools/store/mod.rs:59` (`StoreRequest` — no `valid_from` field); every advertised write surface (MCP `memory_store`, HTTP store, CLI store) constructs the memory with `valid_from: None` (grep at HEAD shows only internal/CLI-kg paths ever set it).

**Failure scenario:** The bitemporal docs advertise caller-supplied `valid_from` backfill ("record when a fact BECAME true, not when it was stored"), but no public write surface accepts it — the field is unreachable except via import. An agent following the docs to backfill historical validity silently gets store-time semantics: durable, WRONG bitemporal data with no error (mixed/incompatible state, silently).

**Proposed fix (from the battery):** either (a) add an optional, validated (`validate_valid_at`) `valid_from` parameter to the store surfaces (MCP + HTTP + CLI), or (b) correct the docs to say backfill is import-only. NOTE: the in-flight docs-reconciliation lane may take path (b) for the DOC half — this issue tracks the missing WRITE-surface half so the decision is explicit, not defaulted.

alphaonedev:
## Summary

The #1834 claim-bitemporal VALID-time interval `[valid_from, valid_until)` had a durable persist + immutable-on-upsert layer (schema v79, both backends) and a `valid_at` recall filter, but was **unreachable from every advertised write surface** — MCP `memory_store`, HTTP `POST /api/v1/memories`, and CLI `ai-memory store` all hardcoded `valid_from: None` / `valid_until: None`. An agent following the bitemporal-backfill docs silently got store-time semantics: durable, WRONG bitemporal data with no error.

## Disposition — ADD the write surface (issue's preferred path)

Neither `valid_from` nor `valid_until` was reachable on CREATE (only `valid_until` was reachable on UPDATE via `memory_update`), so **both** are added for symmetry so the full claim interval can be set in one create call.

- **MCP** — `valid_from` + `valid_until` optional params on `StoreRequest`; parsed + RFC3339-validated (`validate::validate_valid_at`) in `parse_and_build_memory`. New `param_names::VALID_FROM` const (census 131 → 132, bumped in lockstep).
- **HTTP** — `valid_from` + `valid_until` on `CreateMemory`, validated in `validate_create`, threaded through **both** the sqlite and postgres create build sites.
- **CLI** — `--valid-from` / `--valid-until` flags on `StoreArgs`, validated, threaded into the built `Memory`.

The persist layer already stamps `valid_from` at create and preserves it **immutably on upsert** (both backends' `ON CONFLICT` keep the stored value; the MCP `on_conflict=merge` content-dedup detour never rewrites it), and `valid_until` stays updatable via `memory_update` — no persist-side change needed.

## Data-integrity (North Star)

A malformed bound is a loud typed error on every surface (RFC3339 validated up front), never a silently-stored value that would mis-filter `valid_at` recall. `valid_from` immutability is preserved by construction. No durable-truth destruction; the change is purely additive on the write surfaces.

## Token-budget / SSOT

- No new MCP tool / HTTP route / CLI subcommand — only optional params. `cli_subcommand_count_invariant` **unchanged**.
- Trimmed full-profile `tools/list` grew **6692 → 6712** cl100k tokens, within the existing **6750** ceiling — **no ceiling bump required**.
- Tool-list snapshots re-blessed (all 5 profiles); param-name census bumped in lockstep; CHANGELOG `[Unreleased]` entry added.

## Tests

`tests/valid_from_write_surface_2258.rs` (7 tests) — MCP + HTTP + CLI round-trip, RFC3339 rejection on each surface, `valid_from` immutable-on-merge, and `valid_at` recall participation (tolerant of the stored canonical form per the in-flight valid-time canonicalization lane).

Gates run green locally: `fmt`, clippy `--all-targets --features sal` + default, `cargo check --all-targets --features sal-postgres`, `cargo check --examples`, `budget_tokens` (incl. `--ignored` ceiling), `mcp_tools_list_snapshots`, `mcp_param_names_invariant`, `mcp_schema_handler_parity`, `cli_subcommand_count_invariant`, `qual_10_module_size_ceiling`, `qual_6_7_legacy_error_type_ceiling`, `claim_bitemporal_recall_1834`, `archived_valid_time_roundtrip_2035`.

Closes #2258

## AI involvement

Authored by Claude (Opus 4.8) as an autonomous v1.0.0 pre-ship fix worker.


alphaonedev:
## Pre-merge audit — #2270 (#2258 valid_from write surface)

**Verdict: MERGE-AFTER-FIXES** — one required fix (bulk funnel silent drop), one minor documentation gap. Everything else verified clean against actual code at `a9fc8405`.

### 1. Validation — PASS
`validate::validate_valid_at` (chrono `DateTime::parse_from_rfc3339`, `src/validate.rs:894`) is applied on all three surfaces BEFORE persist:
- MCP: `parse_and_build_memory` (`src/mcp/tools/store/validation.rs:311-331`) — trims, validates, rejects; empty string treated as absent.
- HTTP: `validate_create` (`src/validate.rs:1026-1034`) — both bounds.
- CLI: `src/cli/store.rs:259-277` — typed `invalid --valid-from/--valid-until` errors.

Malformed-rejection is fail-closed (the MCP test additionally asserts zero rows persisted after a rejection). **#2265 composition confirmed:** #2265 canonicalizes at the shared write funnels (`db::insert` / `insert_with_conflict` / `insert_if_newer`, `PostgresStore::{store,store_batch,merge_inbound}`, both update funnels) + a v86 healing migration. #2270 threads values INTO those funnels, so whichever lands second, surface-supplied bounds get canonicalized at admission — and any non-canonical rows written in the merge-order gap are healed by v86. The 2258 tests deliberately compare by parsed instant, never raw bytes, so they cannot false-fail under #2265.

### 2. Backend parity — PASS
- HTTP: both build sites threaded — `create_memory` sqlite (`src/handlers/create.rs:1294-1302`) and `create_memory_postgres` (`src/handlers/create.rs:820-828`). No silent drop on either.
- Persist layer: sqlite `db::insert` binds `mem.valid_from`/`mem.valid_until` (`src/storage/mod.rs:1332-1333`); `PostgresStore::store` binds them as $32/$33 in the 33-column INSERT. MCP stdio is structurally sqlite-only; CLI persists via `db::insert` (`src/cli/store.rs:419`).

### 3. Immutability (#1834) — PASS
- sqlite ON CONFLICT: `valid_from = memories.valid_from`, `valid_until = COALESCE(excluded.valid_until, memories.valid_until)` (`src/storage/mod.rs:1288-1295`).
- postgres ON CONFLICT (title,namespace): identical clause (`src/store/postgres.rs` store upsert, "#1834 — sqlite `db::insert` parity" comment). The `SQL_ARCHIVE_ON_CONFLICT_LAST_WINS` `valid_from = EXCLUDED.valid_from` at postgres.rs:114 is the #2195 archived_memories last-wins re-archive clause — archive-table semantics, not the live-row upsert; not a violation.
- MCP `on_conflict=merge` exact-dup detour goes through `db::update` with no valid-time args → both bounds left intact; pinned by `mcp_store_valid_from_immutable_on_merge_upsert` with an explicit comment that `valid_until` changes go through `memory_update`. Note the documented asymmetry: an HTTP/CLI re-store of the same (title,namespace) DOES set `valid_until` via the SQL COALESCE, while the MCP merge detour does not — deliberate and test-pinned, acceptable.

### 4. Interval sanity — MINOR (accidental, not deliberate)
No surface rejects `valid_from > valid_until`. An inverted interval is accepted and simply matches no `valid_at` (empty half-open interval — degraded results, never WRONG results, so no integrity hazard). But nothing documents this as intended: `validate_create` checks each bound independently, and neither the schemars descriptions nor the CLI help mention the inverted case. The existing `memory_update` `valid_until` patch has the same non-check (precedent), so this is consistent — but it should be made deliberate: either add a cheap `from <= until` check when BOTH are present at create (create is the one place both are in hand), or document acceptance. Suggest a small follow-up issue rather than blocking.

### 5. SSOT / budget — PASS
- Param census 131→132 with dated comment (`src/mcp/param_names.rs`); `VALID_FROM` is SSOT-aliased to `models::field_names::VALID_FROM`.
- Tool-list snapshots re-blessed on all 5 profiles; delta is exactly the two nullable-string params on `memory_store` — tool COUNT unchanged, no route/CLI/schema changes, so no other pins fire.
- Token budget: ceiling stays 6750 (`tests/budget_tokens.rs:110`); `budget_tokens` suite passes locally (6/6 + 1 opt-in ignored). Docs-vs-SSOT, C8, hardcoded-literal, L3, migration-ladder CI gates all green already.

### 6. Tests — PASS (with the funnel-coverage caveat below)
7/7 pass locally under `--features sal` (`cargo test --features sal --test valid_from_write_surface_2258`): MCP round-trip + `valid_at` end-exclusive filtering, MCP merge-immutability, CLI round-trip, HTTP round-trip, and malformed-rejection on each of the three surfaces (MCP one also asserts no row persisted). Note the suite is `#![cfg(feature = "sal")]` — a default-features run executes 0 tests; CI's sal matrix covers it.

---

### REQUIRED FIX — `POST /api/v1/memories/bulk` silently drops the new fields (the #2258 failure mode, funnel 4)

`bulk_create` (`src/handlers/memories_query.rs`) consumes the SAME `CreateMemory` DTO this PR extended, on BOTH backend branches — and both branches:
1. call `RequestValidator::validate_create(&body)` (lines 834 / 1102), which as of this PR REJECTS a malformed `valid_from`/`valid_until` per row, then
2. hardcode `valid_from: None, valid_until: None` in their `Memory` literals (postgres branch ~line 912, sqlite branch ~line 1141).

So bulk now errors on a malformed bound but silently discards a valid one — durable rows with the claim interval lost and no error, which is exactly the data-integrity gap #2258 exists to close (and the validated-but-dropped mismatch proves it is accidental, not a scope cut). Fix is mechanical: thread `body.valid_from` / `body.valid_until` at both bulk build sites (mirror the #1422 kind/citations resolve-before-literal pattern already in that function) + one bulk round-trip test. This mirrors the postgres-branch miss the single-create path already caught — same class, one funnel further out.

Minor sibling: after the bulk fix, the "MCP/HTTP/CLI" claim in the commit/CHANGELOG becomes fully true for the HTTP surface.

**Summary: land the bulk_create threading (+test), optionally file the interval-order follow-up, then this is merge-clean.** Local evidence: `cargo test --features sal --test valid_from_write_surface_2258` → 7 passed; `cargo test --test budget_tokens` → 6 passed. Audited read-only in `.local-runs/wt-fix-2258`; nothing pushed.

alphaonedev:
## Audit finding addressed — fourth funnel (bulk) + interval-sanity disposition (`f19d42ac`)

Thanks for the catch. The audit is correct: the single-create fix missed a **fourth** write funnel.

**`POST /api/v1/memories/bulk` (`handlers::bulk_create`)** consumes the same extended `CreateMemory` DTO and runs `validate_create` per row — so after this PR's DTO change it *did* start rejecting a malformed `valid_from`/`valid_until` — but then hardcoded `valid_from: None` / `valid_until: None` at **both** build sites (postgres branch ~`:912`, sqlite branch ~`:1145`). Validated-but-dropped is exactly the #2258 failure mode. Fixed by threading `body.valid_from`/`body.valid_until` into both `Memory` builds. No persist change needed — the ON CONFLICT clauses already preserve `valid_from` immutably + COALESCE `valid_until` on both backends.

**Interval-sanity MINOR (`valid_from > valid_until`):** disposition = **document as an intentional accept**, not reject. An inverted interval is a well-defined EMPTY half-open `[valid_from, valid_until)` that no `valid_at` instant satisfies, so the row is simply never returned on a `valid_at`-filtered recall — it DEGRADES to fewer results, never WRONG results (North Star), and matches the `memory_update` `valid_until`-patch precedent (which also does not enforce ordering). Rejecting would be inconsistent and unnecessary (the durable text is untouched — a caller assertion, not corruption). Recorded on `validate::validate_valid_at`'s doc.

**New tests (+3 → 10 total)** in `tests/valid_from_write_surface_2258.rs`:
- `http_bulk_store_valid_from_round_trips` — bulk POST with explicit `valid_from`/`valid_until` → read back → round-trips
- `http_bulk_store_rejects_malformed_valid_from` — malformed bulk row → `created: 0`, row not persisted
- `store_accepts_inverted_interval_as_empty_never_valid` — inverted interval accepted + stored, unfiltered recall returns it, every `valid_at` filters it out

Gates green: `fmt`, `clippy --all-targets --features sal` (exit 0), `qual_10_module_size_ceiling`, `cargo check --examples`, the full 10-test suite. Token budget unchanged this round (no MCP schema change; still 6712 < 6750). Not merging — over to the auditor for bulk re-verify.

alphaonedev:
## Audit follow-up — re-verify of f19d42ac

**Verdict: MERGE-CLEAN.** All four re-attack items verified against actual code.

1. **Both bulk build sites threaded.** `bulk_create` postgres branch (~:912) and sqlite branch (~:1145) now carry `body.valid_from`/`body.valid_until` into the `Memory` literal; `grep 'valid_from: None' src/handlers/memories_query.rs` returns nothing. Persist-side immutability/COALESCE unchanged and already verified on both backends.

2. **No fifth funnel.** `CreateMemory` is consumed by exactly two handlers: `create.rs` single-create (both backend branches, fixed in a9fc8405) and `memories_query.rs::bulk_create` (fixed here). Every other create-ish path carries a full `Memory` (federation `apply_remote_memory` — #2207 convergence lane; portability import — #2263 lane) so the bounds ride structurally, or synthesizes internal rows with no caller bounds by design (reflect/consolidate). Funnel coverage is complete.

3. **Empty-interval semantics correct.** The #1834 predicate everywhere is `(valid_from IS NULL OR valid_from <= t) AND (valid_until IS NULL OR valid_until > t)` (e.g. `src/storage/mod.rs:5262`, :6045, and the pg `::text` binds) — for non-NULL inverted bounds `from > until` there is no `t` with `t >= from AND t < until`, so no `valid_at` ever matches; unfiltered recall still returns the row. `store_accepts_inverted_interval_as_empty_never_valid` pins all three legs (verbatim persist, unfiltered-recall visible, every probed `valid_at` filtered). Durable text untouched — nothing anywhere rewrites content from the bounds. Doc note on `validate_valid_at` states the disposition explicitly. Deliberate-and-documented accept: satisfied.

4. **Bulk tests genuinely fail pre-fix.** Empirically falsified: with `src/handlers/memories_query.rs` reverted to a9fc8405 (test file kept at HEAD), `http_bulk_store_valid_from_round_trips` FAILS at `valid_from persisted` — the silent drop reproduced — while `http_bulk_store_rejects_malformed_valid_from` passes pre-fix (validate_create already rejected malformed at a9fc8405), which is precisely the validated-but-dropped signature the finding described. Handler restored; full suite at f19d42ac: **10/10 pass** under `--features sal`.

Token budget unchanged (no MCP schema delta this commit). Nothing further blocking from this audit; merge gated on CI as usual. Not merged by me.

PR #2275, opened 2026-07-19
fix(portability): export custody-file role trust anchors (#2245)

### LINKED ISSUE #2245: security(portability): trust-anchor export loads recorder/judge/stopper pubkeys env-only, dropping the custody-dir <label>.pub fallback
**Severity: LOW (security hardening)** — from the v1.0.0 3x7 pre-ship battery. Verified still present at `release/v1.0.0` HEAD `8223e65e`.

**Location:** `src/portability/emit.rs:133` (`role_pubkey_from_env` / `collect_trust_anchors`)

**Failure scenario:** The portability-v2 trust-anchor export re-implements role pubkey loading as env-var-only (`role_pubkey_from_env(RECORDER_PUBKEY_ENV)` etc. for recorder/judge/stopper), while the rest of the substrate resolves these roles via `load_enrolled_*_pubkey`, which ALSO falls back to the custody-dir `<label>.pub` files. A deployment enrolled via custody-dir files (no env vars set) silently exports an envelope MISSING those trust anchors — the export claims fewer enrolled roles than the node actually has, so the destination's advisory re-verify is weaker than it should be, with no warning.

**Proposed fix (from the battery):** Use the existing `load_enrolled_recorder_pubkey` / `load_enrolled_judge_pubkey` / `load_enrolled_stopper_pubkey` helpers (as already done for `witness`) instead of the env-only reimplementation, so both custody paths are honored and the anchor set matches the node's real enrollment.

alphaonedev:
Closes #2245.

Portability-v2 trust-anchor collection now uses the canonical enrolled recorder/judge/stopper public-key loaders, matching the existing witness path. Environment keys retain precedence, while custody-directory `<label>.pub` enrollments are no longer silently omitted from the export. Private keys are never read into the envelope.

Regression coverage creates public-only custody files for all three roles with their pubkey env vars unset, builds a full envelope, and asserts every exported role/key pair. The regression fails on the old env-only implementation.

Gates: formatting, dedicated integration test (5/5 including common harness tests), strict `clippy --all-targets --features sal` with all/pedantic warnings denied, examples/all-target check, and diff check. Commit is SSH-signed. No schema/wire/tool/route/CLI-count change.

**DO NOT MERGE: awaiting Fable re-audit.** Security/portability trust surface; keep draft until adversarial re-audit is clean.

Verified per prime directive pm-v3 (memory cd8ede94).


alphaonedev:
Status: awaiting-fable-reaudit. Security/portability trust surface; keep draft and do not merge until Fable posts a clean adversarial re-audit.

alphaonedev:
Exact-head adversarial merge gate completed for `83aa4b5d`.

The first three-pass audit found and fixed:
- MEDIUM: public trust-anchor export routed through a loader that materialized sibling private keys and could omit a valid public anchor when the unrelated private file was damaged.
- LOW: missing regression coverage for environment-pin precedence and malformed nonempty pins.

The amended implementation adds a public-only verifying-key loader, routes enrolled witness/recorder/judge/stopper verification through it, and proves custody export succeeds even with deliberately corrupt private siblings. Environment pins retain precedence; malformed nonempty pins do not silently fall back.

Final three fresh-context re-audits at the amended exact head: all DRY.

Local evidence:
- portability trust-anchor integration binary: 7/7
- identity keypair library tests: 48/48
- strict clippy with tests: green
- fmt and diff checks: green
- CodeGraph synchronized
- SSH signature verified

Awaiting the fresh exact-head GitHub matrix before ready/merge.

Issue #1779, opened 2026-06-23
reembed/backfill embed the EMPTY placeholder content of encrypted rows → overwrite good embeddings with title-only vectors (corpus-wide recall loss)


alphaonedev:
Auto-filed by destructive sweep #2 (Agent 6, conf 83).

## Defect
`get_memory_texts_batch` (`src/storage/mod.rs:8779`) does `SELECT id, title, content` with NO `open_content`/decrypt. For at-rest-encrypted rows (`AI_MEMORY_ENCRYPT_AT_REST=1`) `content` holds the empty placeholder (plaintext is in `encrypted_envelope`). `reembed` (`src/cli/commands/reembed.rs:154`) then embeds `(title, "")` = title-only, and `set_embeddings_batch_reembed` (`src/storage/mod.rs:8846`) has REPLACE semantics → a model-migration run **overwrites every encrypted row good store-time embedding with a degraded title-only vector**, corpus-wide, unrecoverable without decrypt-and-re-embed. The boot embedding-backfill (`src/mcp/mod.rs:2972`) has the same root cause (lower sev: only fires on never-embedded rows).

## Proposed fix
In the reembed/backfill SELECT path, when `encrypted_envelope IS NOT NULL`, decrypt via `open_content(envelope, agent_id)` (agent_id from `metadata.agent_id`) to recover plaintext before embedding; skip-with-WARN rather than embed an empty placeholder. Severity MEDIUM-HIGH. Evidence: audit mem 88dfef42.

alphaonedev:
Fixed via e2c644be on release/v0.8.0 — 8/8 CI green. The embed-fetch path (get_memory_texts_batch / get_unembedded_ids_batch[_after]) now decrypts at-rest-encrypted rows via a shared resolve_embeddable_content helper (open_content, agent_id from metadata) and SKIPS-with-WARN any row whose envelope won't decrypt — so reembed/backfill no longer embeds the empty placeholder + REPLACE-burns title-only vectors corpus-wide. Tests: get_memory_texts_batch_decrypts_encrypted_rows_1779. EPIC #1792 Batch C.

PR #2330, opened 2026-07-23
fix: postgres-parity chain — 8 cross-backend defects (FBL-07/10/11/21/23/25 + pg-lease + pg-quota)

### LINKED ISSUE #2310: postgres lease_acquire: two concurrent first-acquires both return Ok — unconditional ON CONFLICT DO UPDATE lets the loser silently steal a live lease
## Root cause
`SELECT ... FOR UPDATE` cannot lock an absent row. When no lease row exists, two concurrent `lease_acquire` transactions both read `None`, both pass the conflict check, and the second transaction's unconditional `ON CONFLICT (action_id) DO UPDATE SET holder = EXCLUDED.holder ...` overwrites the first holder's just-committed live lease. Both callers return `Ok(lease)` and both believe they exclusively hold the coordination action — a mutual-exclusion violation in the Pillar-1 lease primitive.

## Evidence
- `src/store/postgres.rs:20031-20036` — FOR UPDATE select returns None for an absent row (no gap lock)
- `src/store/postgres.rs:20037-20044` — conflict check only fires on an existing live row with a different holder
- `src/store/postgres.rs:20045-20058` — upsert overwrites holder/expires unconditionally on conflict
- Sequence: A SELECT→None; B SELECT→None; A INSERT + commit (holder=A, live); B's INSERT blocks on the PK until A commits, then the ON CONFLICT DO UPDATE arm fires and rewrites holder=B; both txs commit Ok. sqlx default READ COMMITTED.
- The sqlite twin is safe only via the single `Arc<Mutex<Connection>>` (`src/store/sqlite.rs:1309-1326`); the window recurs after every lease release/expiry-sweep DELETE, not just the first-ever acquire.

## Repro
sal-postgres build: create an action with no lease; two agents call `memory_lease_acquire` concurrently. Both receive Ok with themselves as holder; the leases row records only the second caller; the first executes the action on a lease it no longer holds — duplicate execution of a coordination action.

## Proposed fix
`src/store/postgres.rs::lease_acquire` (~15 LOC): make the upsert self-guarding — `ON CONFLICT (action_id) DO UPDATE SET ... WHERE leases.holder = EXCLUDED.holder OR leases.expires_at <= EXCLUDED.acquired_at` — and treat `rows_affected() == 0` as `StoreError::Conflict` (keep the FOR UPDATE pre-read as fast path). Regression test: two concurrent first-acquires assert exactly one Ok (live-PG gated on `AI_MEMORY_TEST_POSTGRES_URL`).

Fable v1.0.0 3x7 pre-ship review finding (concurrency-durability lens, verify 0.92). Part of the fix/postgres-parity-chain lane.

### LINKED ISSUE #2311: postgres tenant quota is check-then-record across separate transactions (TOCTOU) — concurrent writers overshoot the daily/storage caps sqlite's BEGIN IMMEDIATE closes
## Root cause
On postgres the #1795 enforcement seam reads the quota counters with a plain pooled SELECT (`check_memory_quota` — no transaction, no FOR UPDATE, no cap predicate) while the increment happens later inside the separate store transaction (`record_memory_quota_in_tx` / `record_memory_quota_batch_in_tx`, unconditional ON CONFLICT DO UPDATE). With a 16-connection pool, N concurrent tenant writes all pass the check below cap and all record — the exact each-passes-then-both-increment race the sqlite path closed with BEGIN IMMEDIATE (#628 H12; `src/quotas.rs:454-497` doc: "so concurrent writers cannot each pass the check and then both increment the counter past the cap").

## Evidence
- `src/store/postgres.rs:22106-22176` — `check_memory_quota`: bare `fetch_optional(&self.pool)`, compare in Rust
- `src/store/postgres.rs:13450-13549` — `record_memory_quota_in_tx` / `record_memory_quota_batch_in_tx`: unconditional increment, no ceiling re-check
- Call sites: `src/handlers/create.rs:986→1013`, `src/handlers/memories_query.rs:1054` (bulk), `src/handlers/power_consolidation.rs:379` — check on one pooled conn, record in a later tx

## Repro
Postgres-backed serve, agent at max_memories_per_day-1: fire K concurrent POST /api/v1/memories for the same (agent, namespace). All K pass the check, all K store and record -> counters land up to K-1 past the cap (larger in bytes on the bulk path). sqlite under identical load refuses at the cap.

## Proposed fix
Make the record statement enforce the ceiling atomically for TENANT callers: add WHERE ceiling guards to the ON CONFLICT DO UPDATE arms of both record fns, check `rows_affected() == 1`, and roll the store tx back with `StoreError::QuotaExceeded` when 0; keep `check_memory_quota` as the cheap fast-fail pre-check for the 429 shape. Enforcement is scoped by caller class (tenant `CallerContext` enforces; admin/curator `bypass_visibility` contexts stay record-only) so pg does not start enforcing on SAL-direct internal writes that sqlite neither charges nor enforces. Live-PG concurrency regression test (8 concurrent stores at ceiling-1 -> exactly 1 succeeds), gated on `AI_MEMORY_TEST_POSTGRES_URL`.

Fable v1.0.0 3x7 pre-ship review finding FBL-07 (concurrency-durability lens, verify 0.92-0.93). Part of the fix/postgres-parity-chain lane. Precedent copied: sqlite `quotas::check_and_record` atomic contract (#628 H12) — no new posture, so no T3 vote required.

### LINKED ISSUE #2312: forget pattern semantics diverge by backend: FTS5 token-AND (sqlite) vs ILIKE contiguous-substring (postgres) — same destructive wire call deletes different row sets
## Root cause
sqlite `forget`/`forget_count`/`forget_distinct_namespaces` resolve `pattern` through `forget_fts_query` -> `sanitize_fts_query(pat, false)` (phrase-quoted whitespace tokens, implicit AND, whole-token FTS5 match — #1601 deliberately made destructive matching conservative), while the postgres twin wraps the escaped pattern in `%...%` and matches `title ILIKE $3 OR content ILIKE $3` — a contiguous case-insensitive substring. postgres therefore over-deletes (pattern "cat" reaps rows containing "education") and under-deletes (multi-word patterns whose tokens are non-adjacent match on sqlite but not postgres) — unrecoverable under archive=false. The wire contract advertised by the memory_forget tool promises token-AND semantics ("every whitespace-separated token must match", src/mcp/tools/forget.rs:286-288).

## Evidence
- `src/storage/mod.rs:3874-3876` (forget_fts_query), `:7759-7812` (token phrase-quoting, implicit AND)
- `src/store/postgres.rs:18528` (`pattern_like = %escaped%`), predicate sites `:18573`, `:18608`, `:18615`, `:18639`, `:18668`, `:18692` (forget archive-copy / link snapshot / cid scrub / crypto-erase / DELETE) + `:18859-18860` (forget_distinct_namespaces)
- `src/store/postgres.rs:18515-18519` comment asserting the wire contract is preserved — it is not; #2032 LM1 only escaped LIKE metacharacters

## Repro
Store "education policy notes" on both backends. Forget pattern "cat": sqlite deletes 0; postgres deletes the row. Inverse: pattern "alpha beta" with non-adjacent tokens: sqlite deletes it, postgres does not. Mixed-backend fleets issuing one forget converge to different corpora.

## Proposed fix
Replace the ILIKE arm at all destructive predicate sites (+ forget_distinct_namespaces) with token-AND over the v57 stored `tsv` column: `tsv @@ plainto_tsquery('english', $3)` (plainto_tsquery is AND-joined and operator-injection-safe; mirrors the existing recall usage at postgres.rs:5584-5586). Empty/all-stripped patterns produce an empty tsquery matching NOTHING (fail-safe, mirroring sqlite's `"_empty_"` sentinel and preserving the #2032 LM1 wildcard-literal posture). Document the tokenizer residual (english stemming/stopwords vs unicode61). Cross-backend victim-set parity test (live-PG gated on `AI_MEMORY_TEST_POSTGRES_URL`).

Fable v1.0.0 3x7 pre-ship review finding FBL-10 (cross-backend-parity lens, verify 0.92). Part of the fix/postgres-parity-chain lane.

### LINKED ISSUE #2313: postgres hard forget (archive=false) still snapshots memory_links into archived_memory_links — permanent erasure leak sqlite deliberately avoids
## Root cause
`PostgresStore::forget` runs the #1771 `archived_memory_links` snapshot INSERT UNCONDITIONALLY (outside the `if archive` block that closes at src/store/postgres.rs:18585), while the sqlite twin gates the snapshot on `archive == true` with an explicit contract comment ("Runs only when `archive` so a hard (non-recoverable) forget keeps its documented edge-loss", src/storage/mod.rs:4626-4628). On postgres, an erasure-intent hard forget (archive=false) permanently retains the victims' full edge topology (source_id, target_id, relation, observed_by, signature, valid intervals) as orphan rows: no code path anywhere deletes from `archived_memory_links` (grep confirms zero `DELETE FROM archived_memory_links` sites; `archive_purge` deletes `archived_memories` only), and with archive=false no `archived_memories` row exists so restore can never consume them.

## Evidence
- `src/store/postgres.rs:18544-18585` (`if archive { ... }` archive-copy block) followed by `:18587-18624` (unconditional link-snapshot INSERT)
- `src/storage/mod.rs:4626-4676` (sqlite gates both snapshot statements on `if archive`)
- The postgres comment calls itself the "Postgres twin of the SQLite `archive_links_for_memory` snapshot" — parity intent, unconditional placement is the oversight
- Erasure-completeness posture context: the SAME pg forget path scrubs `cid_genesis` (#1825 G8 T7) and purges DLQ/dedup leaks (#1821 G30) precisely because forgotten-data lingering is a tracked defect class

## Repro
Postgres-backed daemon: store linked memories A and B; forget with archive=false matching A. `SELECT * FROM archived_memory_links WHERE source_id='A' OR target_id='A'` returns rows forever, unreachable by restore and untouched by purge. Same sequence on sqlite leaves zero rows.

## Proposed fix
Move the link-snapshot query inside the `if archive` block (~5-line move), matching the sqlite contract + the #1771 rationale (the snapshot exists so restore can re-insert). Live-PG regression test: archive=false forget leaves no `archived_memory_links` rows for the victims; archive=true forget preserves them (gated on `AI_MEMORY_TEST_POSTGRES_URL`).

Fable v1.0.0 3x7 pre-ship review finding FBL-11 (cross-backend-parity lens, verify 0.93/0.97). Part of the fix/postgres-parity-chain lane.

### LINKED ISSUE #2314: postgres federation receive routes memories through apply_remote_memory, bypassing the G30 tombstone resurrection guard, G29 redact, attestation sanitize, updated_at clamp, and same-id CRDT merge
## Root cause
The postgres-backed /sync/push funnel (`sync_push_via_store`) dispatches inbound memories through `MemoryStore::apply_remote_memory`, whose `PostgresStore` impl never received the receive-lane hardening wired into `merge_inbound` — so on a postgres receiver an inbound federated write bypasses the G30 forget-tombstone resurrection guard, the G29 secret-screen redact, the #1719 attestation sanitize, the #1755 updated_at clamp, and the same-id CRDT field-merge that the sqlite receive path enforces. The fully hardened `PostgresStore::merge_inbound` (all four guards + same-id merge) exists but has ZERO production callers.

## Evidence
- Routing: `src/handlers/federation_receive.rs:1044-1046` (StorageBackend::Postgres -> sync_push_via_store) -> `src/handlers/federation_signing_check.rs:354` (`app.store.apply_remote_memory`)
- Unguarded pg impl: `src/store/postgres.rs:16518-16859` — apply_remote_memory has no forget_tombstones check, no screen_storage_memory, no sanitize_inbound_attestation, no clamp_inbound_updated_at, and only an ON CONFLICT (title,namespace) arm (same-id/different-title pushes hit the id PK and error-skip forever)
- Guarded-but-dead pg impl: `src/store/postgres.rs:16861-16935` — merge_inbound carries tombstone check, screen, sanitize, clamp, merge_memory
- sqlite twin funnel: `src/handlers/federation_receive.rs:1511` (db::merge_inbound), with the guards inside insert_if_newer (`src/storage/mod.rs:12595-12620`) and the merge path (`:12900-12935`)
- pg forget writes forget_tombstones rows (`src/store/postgres.rs:~18742`) that nothing on the pg apply path consults
- The federation catchup puller (`src/federation/receive.rs::catchup_once_with_store`) also routes through apply_remote_memory, so the gap covers BOTH inbound lanes

## Repro
Receiver runs `serve --store-url postgres://...` in a 2-node federation. (a) Store memory M on the receiver, forget it (tombstone written); a stale enrolled peer re-pushes M -> forgotten/GDPR-erased row RESURRECTS (identical push against a sqlite receiver is dropped with the resurrection-guard INFO log). (b) Peer pushes secret-bearing content -> lands UNREDACTED on postgres. (c) Same-id/different-title push -> pg id-PK error, row skipped every push (permanent divergence) while sqlite field-merges. (d) Far-future updated_at wins LWW permanently on pg; sqlite clamps.

## Proposed fix
Two layers, closing every reachable path:
1. Switch `federation_signing_check.rs:354` to `app.store.merge_inbound(&ctx, &to_insert)` — the hardened impl with the same-id field merge; `merge_memory` preserves local lifecycle_state so the #1948 dequarantine-on-attest flow at :357-363 still holds.
2. Add the sqlite-`insert_if_newer`-parity guards (tombstone drop + secret-screen redact) to the top of `PostgresStore::apply_remote_memory` itself, so the catchup lane and the merge_inbound fresh-insert fall-through are guarded too (sanitize/clamp remain merge-path-only, exactly matching the sqlite placement).
Live-PG regression tests: tombstoned id is not resurrected via apply_remote_memory NOR merge_inbound; same-id merge path lands (gated on `AI_MEMORY_TEST_POSTGRES_URL`).

Fable v1.0.0 3x7 pre-ship review finding FBL-21 (cross-backend-parity lens, verify 0.93). Part of the fix/postgres-parity-chain lane.

### LINKED ISSUE #2315: postgres archive_by_ids deletes without AGE unprojection and archive_restore never re-projects — ghost nodes and permanently missing edges in the memory_graph projection
## Root cause
Two symmetric AGE-projection integrity gaps in the postgres archive funnels:
1. `PostgresStore::archive_by_ids` performs the same-tx cascade DELETE of archived rows but is the ONLY hard-delete path that never calls `unproject_memory_from_age` — on an AGE deployment the `memory_graph` projection retains the deleted node and its incident edges indefinitely (ghost nodes; AGE-routed kg_query returns live-looking edges to a non-live memory).
2. `PostgresStore::archive_restore` re-inserts the memory row and its preserved `archived_memory_links` edges via raw SQL only — it neither runs the sync `project_link_into_age` MERGE nor enqueues `kg_projection_outbox` rows — while forget/delete/gc/size_gc DO unproject, so restored edges are permanently missing from the AGE view with no reconciliation/heal verb anywhere in the codebase (the CTE fallback fires only on AGE runtime failure, never on a valid-but-empty Cypher result).

## Evidence
- `src/store/postgres.rs:21197-21315` (archive_by_ids: DELETE at :21303, zero unproject calls) vs the complete unprojection coverage on every sibling delete path: delete() :15901/:15918, apply_remote_deletion() :17211/:17225, forget() :18705, consolidate() :19370, run_gc() :20744, size_gc() :20903
- `src/store/postgres.rs:20926-21118` (archive_restore: memory INSERT + link re-insert :21089-21106; zero AGE calls); project_link_into_age callers are only link_internal / apply_remote_link / consolidate
- Helper doc `:13216-13236` (#1783) states the exact hazard: MERGE-only projection means a deleted memory leaves an orphan :Memory node + incident edges

## Repro
pg + AGE, sync projection mode: store A,B; link A->B (projected); (a) archive A via memory_archive/HTTP archive/CLI -> relational rows gone, AGE still returns the A node + A->B edge; (b) forget A with archive=true (unprojected correctly), then archive-restore A -> relational edge returns but AGE current-view kg_query(A) stays empty forever, while sqlite and the pg CTE-historical path both return it — permanent split-brain with no self-heal.

## Proposed fix
- archive_by_ids: `unproject_memory_from_age(&mut tx, id)` in the per-id loop next to the DELETE, mirroring the forget() call shape (~3 LOC).
- archive_restore: RETURNING on the link re-insert; when kg_backend == Age, project each restored edge — deferred mode enqueues `kg_projection_outbox` rows in the same tx (reusing the link_internal enqueue shape), sync mode projects via a SAVEPOINT best-effort (warn-and-continue on AGE runtime failure so the relational restore is never blocked by the derived graph — the #700/#1542 degrade posture).
AGE-gated regression test (`AI_MEMORY_TEST_AGE_URL`) asserting no ghost edges after archive_by_ids and edge visibility after forget->restore. Needs DO/AGE-node validation.

Fable v1.0.0 3x7 pre-ship review finding FBL-23 (cross-backend-parity lens, verify 0.93 both halves). Part of the fix/postgres-parity-chain lane.

### LINKED ISSUE #2317: postgres serve-boot embedding backfill embeds the encrypted-at-rest placeholder (no #1779 decrypt/skip parity) — writes title-only vectors and overwrites good embeddings on the heal arms
## Root cause
`PostgresStore::list_unembedded` returns raw `(id, title, content)` without selecting/decrypting `encrypted_envelope`, so on an at-rest-encrypted pg deployment the serve-boot SAL backfill sweep embeds `title + ""` (the seal placeholder that every pg write funnel stores in the content column per #2288/#2292), stamps it with the active `embedding_space` so the row permanently leaves the scan set, and on the embedding-not-null heal arms (NULL-space adoption gap + opt-in `AI_MEMORY_PG_HEAL_FOREIGN_SPACE`) REPLACES existing GOOD store-time vectors with title-only vectors — silently, corpus-wide, self-sealing. The sqlite #1779 fix (`resolve_embeddable_content`: decrypt-or-skip-with-WARN) was never mirrored onto the postgres SAL twin, and the `ai-memory reembed` CLI is sqlite-only so there is no recovery path on pg.

## Evidence
- `src/store/postgres.rs:15304-15380` (list_unembedded SELECTs only id,title,content; all WHERE arms admit encrypted rows)
- `src/store/postgres.rs:11789-11803` (seal_content_for_insert: content column stores the placeholder; plaintext lives only in encrypted_envelope)
- `src/store/mod.rs:3792-3827` (run_embedding_backfill_on_store embeds the triples verbatim; set_embeddings_batch stamps the active space at postgres.rs:15406)
- sqlite-only fix: `src/storage/mod.rs:13618-13669` (resolve_embeddable_rows / resolve_embeddable_content) with its own WARN text naming exactly this failure mode

## Repro
Postgres + at-rest encryption: (a) store rows while the embedder is down (embedding NULL, content sealed), boot serve with a working embedder -> title-only vector written and attested as healed, no WARN; (b) legacy good-vector NULL-space rows (or foreign-space rows under the heal knob) are re-derived from the placeholder and OVERWRITTEN.

## Proposed fix
Mirror #1779 on the SAL scan: make `crate::storage::resolve_embeddable_content` pub(crate) (shared home, no duplication); `PostgresStore::list_unembedded` additionally SELECTs `encrypted_envelope` + `metadata` and routes each row through the resolver, skipping undecryptable rows with a loud per-pass WARN naming the skipped count (the existing zero-progress break in run_embedding_backfill_on_store terminates an all-skipped pass — no spin). Live-PG regression test: sealed row + present key embeds the PLAINTEXT (not the placeholder); undecryptable sealed row is skipped and its existing vector is never overwritten (gated on `AI_MEMORY_TEST_POSTGRES_URL`).

Fable v1.0.0 3x7 pre-ship review finding FBL-25 (silent-failure lens, verify 0.93). Part of the fix/postgres-parity-chain lane.

### LINKED ISSUE #2318: postgres.rs v70 schema-history comment still calls the postgres archive-link snapshot/restore a tracked follow-up though it shipped (stale docs)
## Root cause
The v70 schema-history comments still describe the postgres `archived_memory_links` snapshot/restore wiring as "a tracked follow-up", but that wiring SHIPPED: `PostgresStore::forget` snapshots links (src/store/postgres.rs:18594-18624), `archive_by_ids` snapshots per-id (:21263-21278), and `archive_restore` re-inserts preserved edges (:21089-21106), pinned by `archive_restore_preserves_links_pg_1771`. Stale doc-comments on a data-integrity surface are a real defect per the prime directive (docs drift class).

## Evidence
- `src/store/postgres.rs:695-703` — "SQLite wires snapshot/restore this commit, postgres snapshot/restore is a tracked follow-up"
- `CLAUDE.md` §Database v70 paragraph — same stale sentence ("the postgres snapshot/restore wiring is a tracked follow-up")

## Proposed fix
Docs-only: correct both comments to state the postgres snapshot/restore wiring shipped (#1771 wiring in forget / archive_by_ids / archive_restore). No behavior change.

Grok v1.0.0 W1A3-01 finding, confirmed by the Fable 3x7 pre-ship review. Part of the fix/postgres-parity-chain lane.

alphaonedev:
POSTGRES-parity lane of the v1.0.0 100%% fix campaign. 8 fixes, one commit each, 1:1 issues #2310–#2318.

Closes #2310
Closes #2311
Closes #2312
Closes #2313
Closes #2314
Closes #2315
Closes #2317
Closes #2318

pg lease race, quota TOCTOU, forget token-AND parity, hard-forget erasure-completeness, G30/G29 federation guards, AGE archive_by_ids unproject + archive_restore reproject, backfill decrypt-or-skip (#1779 parity), v70 stale-comment. DRAFT — awaiting Fable review + CI (pg behavioral tests need live-PG/DO). Some tests #[ignore]-gated on AI_MEMORY_TEST_POSTGRES_URL.


alphaonedev:
Rebase status (Fable orchestrator): merged release/v1.0.0 (v87 schema anchor, PR#2354) into this branch at ad4f4ec7 — code files auto-merged cleanly; CHANGELOG keep-both; CLAUDE.md takes the release v87 paragraph with this branch's #2318 v70 correction grafted. Verified: sal-postgres compiles, fmt clean, anchor regression suite storage_chain_fbl 20/20 green, docs-vs-SSOT gate PASS (schema=87). Re-audit flag RESOLVED: pg lease_acquire IS #2310-guarded (FOR UPDATE pre-read + conditional ON CONFLICT arm + typed Conflict on rows_affected==0, pinned by lease_acquire_concurrent_first_acquire_single_winner_2310). REMAINING before merge: fold in the three pg-mirror gaps — FBL-03 (pg archive/restore kind_provenance carry), FBL-20 (pg apply_remote_memory/merge_inbound expiry lattice-join), FBL-33 (pg recall soft-loser 0.5 penalty) — as a follow-up commit on this branch.

alphaonedev:
Fable pre-merge review: PASS (final). 8 parity fixes (#2310-#2318) reviewed pre-rebase; this cycle added the v87 anchor merge (CHANGELOG keep-both + CLAUDE.md graft), the forget-tombstone literal hoist, the three pg-mirrors (FBL-20 GREATEST lattice-join on apply_remote_memory — merge_inbound already convergent via crdt_merge; FBL-33 ts_rank × SOFT_LOSER_SCORE_FACTOR; FBL-03 kind_provenance through all 7 archive snapshots + vocab-guarded restore), and the qual_10 lockstep bump. pg lease_acquire #2310 guard verified with its single-winner concurrency test. All checks green incl. the live-pg Postgres gate. Merging — this unblocks the postgres.rs chain in the campaign DAG.

Issue #18, opened 2026-04-02
[MEDIUM] Race condition in update_memory handler


alphaonedev:
Handler calls db::update then db::get separately. Client may receive stale data between operations.

**Status:** Accepted risk — single Mutex serializes all HTTP requests. Race requires multi-process access.

PR #2364, opened 2026-07-23
fix(#2359): offload TTL sweep locks per-chunk, never across the per-row sleep

### LINKED ISSUE #2359: Offload TTL sweep sleeps inside the global Db mutex — up to ~10s full HTTP stall per pass
## Root cause

`src/background/offload_ttl_sweep.rs::spawn` (~L54-76) acquires the **global daemon `Db` mutex** — the single `Arc<Mutex<(Connection, PathBuf, ResolvedTtl, bool)>>` that serves every HTTP handler — and holds it across the **entire** sweep pass. The `SweepAdapter::run_sweep` blanket impl calls `offload::sweep_expired(&self.0, now_unix, MAX_PER_RUN=1000, SLEEP_BETWEEN_DELETES=10ms)`, and `sweep_expired` (`src/offload/mod.rs` ~L631-651) does a **synchronous `std::thread::sleep(10ms)` between every one of up to 1000 deletes**. Worst case: ~10 seconds of wall-clock where the lock is held, so **every concurrent HTTP request stalls**, plus a tokio worker thread is parked on a blocking sleep inside an async task.

The module's own doc admits the hazard and defers it — never delivered:

> `src/background/offload_ttl_sweep.rs:48-52`: "the in-pass `std::thread::sleep` between deletes happens INSIDE that lock window ... v0.8.0 may move the per-row sleep outside the lock if the offload write volume grows."

## Evidence

- `src/background/offload_ttl_sweep.rs:54-76` — `spawn` holds `state.lock().await` across the whole `run_sweep` call.
- `src/background/offload_ttl_sweep.rs:95-97` — blanket impl passes `SLEEP_BETWEEN_DELETES` (10ms) as the in-pass sleep.
- `src/offload/mod.rs:631-651` — `sweep_expired` calls `std::thread::sleep(sleep_between_deletes)` per row, up to `max_per_run` rows.

## Repro

A daemon with ≥1 expired `offloaded_blobs` row: on the daily tick the sweep holds the shared connection mutex for `min(expired, 1000) × 10ms`. With a backlog of 1000 expired blobs that is ~10s during which `GET/POST /api/v1/*` all block on `state.lock()`.

## Proposed fix (~25 lines, `src/background/offload_ttl_sweep.rs` only)

Restructure `spawn` to lock **per chunk** instead of per pass:

```
loop {
    acquire lock;
    run a small chunk (cap 50 rows) with Duration::ZERO in-pass sleep;
    drop lock;                       // <-- lock released here
    tokio::time::sleep(10ms).await;  // <-- inter-chunk pacing, OUTSIDE the lock
} until cumulative MAX_PER_RUN reached or a chunk returns < requested
```

`offload::sweep_expired` already takes a per-run cap + sleep `Duration`, so **`src/offload/mod.rs` needs no change** — call it with `chunk_cap` and `Duration::ZERO`. Change `SweepAdapter` to a chunked method (`run_sweep_chunk(now, chunk_cap)`), update the blanket impl + doc-comment (remove the stale v0.8.0 promise, document the chunked-locking contract), and add a `#[tokio::test(start_paused = true)]` regression asserting the mutex is free during the inter-chunk sleep.

Files: `src/background/offload_ttl_sweep.rs` (impl + doc + test).

alphaonedev:
Closes #2359

## Problem

`src/background/offload_ttl_sweep.rs::spawn` acquired the **global daemon `Db` mutex** — the single `Arc<Mutex<(Connection, PathBuf, ResolvedTtl, bool)>>` that serves every HTTP handler — and held it across the **entire** sweep pass, while `offload::sweep_expired` runs a synchronous `std::thread::sleep(10ms)` between each of up to `MAX_PER_RUN=1000` deletes. Worst case: ~10s of wall-clock where every concurrent HTTP request stalls on `state.lock()`, plus a tokio worker thread parked on a blocking sleep inside an async task. The module doc admitted the hazard ("v0.8.0 may move the per-row sleep outside the lock") — never delivered.

Prime-directive relevance: this is a **manageability / DoS** defect (a self-inflicted thundering-stall on a shared mutex), not a data-integrity one. No data path changes; the durable memory TEXT and the reap semantics are untouched — only WHEN the lock is held changes.

## Fix (`src/background/offload_ttl_sweep.rs` only, ~25 LOC of behavior)

`spawn` now locks **per chunk**:

```
loop {
    chunk_cap = min(MAX_PER_RUN - swept, CHUNK_CAP=50)
    { lock = state.lock().await; lock.run_sweep_chunk(now, chunk_cap) }  // lock dropped at block end
    tokio::time::sleep(SLEEP_BETWEEN_CHUNKS=10ms).await                  // OUTSIDE the lock
} until MAX_PER_RUN reached or a chunk returns < chunk_cap (backlog drained)
```

- `offload::sweep_expired` is **unchanged** — it already accepts a per-run cap + sleep `Duration`; the chunked path calls it with `chunk_cap` and `Duration::ZERO` (no in-pass sleep).
- The `SweepAdapter` trait's per-pass `run_sweep(now)` became the chunked `run_sweep_chunk(now, chunk_cap)`; the production `(Connection, …)` blanket impl + all test adapters updated in lockstep. No external callers of the trait method exist (`spawn`'s call site in `daemon_runtime.rs` is unchanged — same signature).
- Consts: added `CHUNK_CAP=50`; renamed the now-inter-chunk `SLEEP_BETWEEN_DELETES` → `SLEEP_BETWEEN_CHUNKS=10ms` (only referenced in this module). Module doc rewritten: stale v0.8.0 promise removed, chunked-locking contract documented.

## Test

New `background::offload_ttl_sweep::tests::lock_released_between_chunks_2359` — a `#[tokio::test(flavor = "current_thread", start_paused = true)]` that drives the `spawn` loop into its first inter-chunk sleep (sweep mid-pass, backlog modelled by a full-chunk adapter) and asserts a contender can `try_lock()` the mutex during that sleep **without** advancing the clock past it. Under the pre-fix hold-across-the-whole-pass posture the lock would stay held for every remaining chunk + sleep, so the acquisition would fail.

## Gates

- `cargo clippy --lib -- -D warnings -D clippy::all -D clippy::pedantic` — clean
- `AI_MEMORY_NO_CONFIG=1 cargo test --lib -- background::offload_ttl_sweep offload::` — `test result: ok. 37 passed; 0 failed`

## AI involvement

Authored autonomously by Claude Opus 4.8 in the v1.0.0 GA fix campaign (roadmap-3x7 finding #18). Issue #2359 was filed by the same agent before the fix per the prime-directive discover→track→fix workflow. Base SHA: `994efd73`. This is a **draft** pending Fable review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Offload TTL sweep no longer holds the global `Db` mutex across its per-row sleeps — up to ~10s of full HTTP stall per pass eliminated** ([#2359](https://github.com/alphaonedev/ai-memory-mcp/issues/2359); v1.0.0 pre-ship, roadmap-3x7 #18). `src/background/offload_ttl_sweep.rs::spawn` acquired the single `Arc<Mutex<(Connection, …)>>` that serves EVERY HTTP handler and held it across the whole sweep pass, while `offload::sweep_expired` runs a synchronous `std::thread::sleep(10ms)` between each of up to `MAX_PER_RUN=1000` deletes — worst case ~10s where every concurrent HTTP request blocked on `state.lock()`, plus a tokio worker parked on a blocking sleep inside an async task (the module doc admitted the hazard and deferred it to a "v0.8.0" that never shipped). `spawn` now locks **per chunk**: it acquires the lock, deletes at most `CHUNK_CAP=50` rows with `Duration::ZERO` in-pass sleep, drops the lock, then performs the `SLEEP_BETWEEN_CHUNKS=10ms` pacing sleep OUTSIDE the lock via `tokio::time::sleep`, repeating until the cumulative `MAX_PER_RUN` cap is reached or a chunk drains the backlog — so concurrent HTTP handlers acquire the shared mutex between chunks. `offload::sweep_expired` is unchanged (it already accepts a per-run cap + sleep `Duration`); the `SweepAdapter` trait's per-pass `run_sweep` became the chunked `run_sweep_chunk(now, chunk_cap)`. Regression: `background::offload_ttl_sweep::tests::lock_released_between_chunks_2359` (a `#[tokio::test(start_paused = true)]` proving the mutex is free during the inter-chunk sleep while the sweep is mid-pass). No new MCP tool / HTTP route / CLI subcommand / DB migration — no SSOT-count move.
+
 ## [1.0.0] — 2026-07-21
 
 ### Security
--- a/src/background/offload_ttl_sweep.rs
+++ b/src/background/offload_ttl_sweep.rs
@@ -4,10 +4,26 @@
 //! v0.7.0 QW-3 — daily TTL sweep for `offloaded_blobs`.
 //!
 //! The sweep removes rows where `stored_at + ttl_seconds < now`,
-//! bounded at [`MAX_PER_RUN`] deletions per pass with a [`SLEEP_BETWEEN_DELETES`]
-//! gap between deletes so the connection lock window stays short
-//! under contended writes (matches the K2 pending-actions sweeper
-//! discipline).
+//! bounded at [`MAX_PER_RUN`] deletions per pass.
+//!
+//! ## Chunked-locking contract (#2359)
+//!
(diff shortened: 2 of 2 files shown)


alphaonedev:
Fable pre-merge review: PASS. Chunked-locking contract correct (lock scoped per chunk, pacing sleep outside via tokio::time::sleep, Err arm preserves per-pass warn+break semantics, final-chunk extra pacing sleep is lock-free and harmless). Trait rename run_sweep→run_sweep_chunk has zero external callers (grep-verified). Regression test deterministic under start_paused. Merging on CI green. Reviewed per Fable pre-commit gate (operator directive 2026-07-22).

Issue #2444, opened 2026-07-28
[dr][data-loss] `ai-memory backup` fails OPEN on a Postgres deployment — creates an empty SQLite file, VACUUMs it, writes a valid checksum manifest, exits 0


alphaonedev:
**Class:** unintentional total data loss / false assurance. **Verified by code inspection at `release/v1.0.0` @ `011dd8b9`.**

## The path

`BackupArgs` has **only** `--to` and `--keep` (`src/cli/backup.rs:24-33`) — there is no `--store-url` and no backend selector. Dispatch passes the sqlite path only (`src/daemon_runtime.rs:1627`).

`run_backup` then does (`src/cli/backup.rs:69-83`):

```rust
let conn = db::open(db_path).context("opening source DB for backup")?;
...
conn.execute("VACUUM INTO ?1", rusqlite::params![snapshot_path.to_string_lossy()])
    .context("VACUUM INTO failed")?;
```

and `db::open` is `Connection::open(path)` (`src/storage/connection.rs:145`), which **creates the file when absent**.

`grep -niE "postgres|store_url|backend" src/cli/backup.rs` returns **nothing**. There is no backend check and no refusal anywhere on the path.

## Failure scenario

An operator running a Postgres-backed daemon runs the product's own backup verb. They get:

1. a freshly-**created, empty** SQLite file at the default path,
2. `VACUUM INTO` a timestamped snapshot of that empty file,
3. a **valid sha256 manifest** over it,
4. **exit 0**, and
5. `--keep 48` rotation that will, over 48 runs, delete nothing real because nothing real was ever written.

The operator discovers this during a restore — which is the one moment it cannot be fixed.

## Fairness to the docs

The documentation is **not** the defect. `docs/production-deployment.md:84` correctly scopes `ai-memory backup` to *"SQLite deployments"* and prescribes `pg_dump` for Postgres at `:96-99`; the T3–T6 tier table specifies `pg_basebackup`/WAL archiving. `enterprise-deployment.md` §2.6 sits under Topology 1 (Singleton, SQLite).

**The CLI's silence is the defect.** A fail-open path on a destructive-adjacent operation is a structural problem regardless of whether the current doc set steers around it — docs are not a safety mechanism, and one binary serves both backends.

## Suggested fix

[...]

Issue #1927, opened 2026-07-07
[security][medium] src/daemon_runtime.rs:853 — Postgres DB password must be passed in --store-url process argument (visible in /proc


alphaonedev:
**Severity:** medium · **CWE-214: Invocation of Process Using Visible Sensitive Information** · **Location:** `src/daemon_runtime.rs:853` · dim: secrets

### Attack
The Postgres store URL is accepted ONLY as the `--store-url postgres://user:pass@host/db` CLI argument across `serve`, `curator`, `schema-init`, `migrate`, and `recover-previous-session`. The `AI_MEMORY_STORE_URL` environment fallback was intentionally dropped (documented at src/cli/curator.rs:311-314, commit 1e8ad69b), and there is no `--store-url-file`/env alternative (unlike the SQLCipher passphrase, which has `--db-passphrase-file`). On stock Linux (no `hidepid`), `/proc/<pid>/cmdline` is world-readable (mode 0444), so a co-tenant NHI agent or any other local UID runs `cat /proc/$(pgrep -f ai-memory)/cmdline | tr '\0' ' '` (or `ps auxww`) and reads the full DSN including the cleartext DB password while the daemon runs. Notably this is strictly worse than the removed env path: `/proc/<pid>/environ` is owner-only (mode 0400), whereas argv is world-readable — the removal of the env fallback forced the secret from a same-user-only sink into a world-readable one.

### Impact
Any local principal on a shared host (the co-tenant NHI agent in the threat model) exfiltrates the Postgres credential without any privilege, gaining full read/write to the memory database and its federated corpus. Also lands the password in shell history, systemd unit files, and process-listing telemetry.

### Fix
Reintroduce a non-argv credential channel: accept `AI_MEMORY_STORE_URL` from the environment (owner-only /proc/environ) and/or add a `--store-url-file` flag with a 0600 permission check mirroring `passphrase_from_file` (daemon_runtime.rs:2257). Support a passwordless URL plus a `PGPASSFILE`/separate-secret-file so the password never enters argv. Document that passing credentials via `--store-url` exposes them via /proc/cmdline.

---
_Found by the v0.9.0 multi-agent security review, adversarially verified. Part of #1918._

alphaonedev:
Fixed in 65772202 (security cluster S3, 8/8 CI green on release/v0.9.0). Verified: crypto sign/verify symmetry restored, HTTP enforcement gate #1924 completes #1885 for the HTTP write path, coverage floor restored.

PR #2476, opened 2026-07-29
fix(#2444): `ai-memory backup` fails CLOSED — no snapshot, no checksum, for a store it cannot capture

alphaonedev:
## The defect

`ai-memory backup` on a Postgres deployment **failed open**. `run_backup` took the `--db` path unconditionally and handed it to `db::open`, which **creates the file when absent** (`src/storage/connection.rs`) and then runs the bootstrap schema + the whole migration ladder on it. So the product's own backup verb:

1. created an **empty** SQLite database at the placeholder `--db` path,
2. `VACUUM INTO`-ed it to a timestamped snapshot,
3. wrote a **valid** sha256 manifest over those bytes,
4. rotated `--keep` against it, and
5. **exited 0**.

Every signal the operator had said the backup succeeded. The corpus lived in Postgres and was never captured. The discovery moment is the DR restore — the one moment it cannot be fixed.

### What the issue body did not say

- **The suggested "no `schema_version`" check would not have fired.** `db::open` runs `SCHEMA` + `migrate()` on the file it just created, so a freshly conjured database carries a full, current, valid schema. It is not distinguishable from a real database by any schema probe. The only honest discriminator is *it did not exist*, which must be checked **before** the open.
- **`restore` has the same defect, in the direction that hurts more.** Restoring a SQLite snapshot onto a Postgres-configured deployment copied the file to the placeholder path, printed `Restored`, and exited 0 while the real corpus was untouched — false assurance during a disaster rehearsal.
- **`restore` never validated the snapshot at all.** The sha256 only proves the bytes match a manifest *we* wrote over whatever was produced (and `--skip-verify` proves nothing), so a truncated or foreign file was copied straight over the live corpus after moving it aside.
- **`restore` orphaned the SQLite WAL sidecars.** It renamed `<db>` aside but left `<db>-wal` / `<db>-shm` in place next to the freshly copied snapshot, where SQLite can replay stale frames **into** the restored database. That is silent corruption of the restore, and it also left the moved-aside safety copy without its own WAL.

## The fix — fail CLOSED

- `resolve_sqlite_source` resolves the **configured** store through the same channels `serve` uses (`daemon_runtime::resolve_store_url`: `AI_MEMORY_STORE_URL_FILE` > `AI_MEMORY_STORE_URL` > `--store-url`, #1927), and refuses:
  - a `postgres://` store, naming `pg_dump` / `pg_basebackup` and `docs/production-deployment.md` (DSN password redacted via `logging::redact_url_password`);
  - an unrecognised scheme, rather than falling back to the local `--db` file;
  - an argv/env store-URL **disagreement**, rather than guessing which store to capture (`resolve_store_url` gives env precedence over argv, so a silent divergence was possible).
- `backup` refuses when the resolved SQLite file **does not exist** — before `db::open` can create it. No file, no snapshot, no manifest, non-zero exit.
- `restore` gained the same store guard, plus a **read-only structural probe** of the snapshot before the live corpus is touched, plus **`-wal` / `-shm` move-aside**.
- `BackupManifest` gained `backend` / `schema_version` / `memory_count` (`#[serde(default)]`, so pre-#2444 manifests still restore). Restore refuses a **cross-backend** snapshot and a **forward-schema** snapshot (which would open cleanly and then silently drop the newer columns on the next write).
- A zero-memory snapshot is **WARNed, not refused** — see the vote below.
- `SQLITE_URL_SCHEME` / `POSTGRES_URL_SCHEMES` / `is_postgres_url` hoisted from the `sal`-gated `crate::migrate` into ungated `crate::daemon_runtime` (re-exported verbatim) so the guard fails closed on the **default** build too. `storage::index_coverage::SQL_TOTAL_MEMORIES` promoted to `pub(crate)` so the corpus probe reuses the existing named statement instead of re-scattering the literal (pm-v3.1).

## 3x3 adversarial vote (this session)

Three concurrent agents, distinct lenses (DR-operator / CLI-surface + SSOT blast radius / red-team residual fail-open). Load-bearing outcome: the issue's suggested **`COUNT(*) == 0` refusal was rejected**. It cannot distinguish a legitimately fresh SQLite deployment from a wrong-store capture; on a Postgres host the local sqlite sidecar legitimately holds 0 memories while carrying the only on-disk copy of the `signed_events` governance spine (refusing it would be the data-loss it was meant to prevent); and its `--allow-empty` escape hatch would be baked into every cron template within a week, disarming the guard permanently. The store guard is the structural control; the row count is recorded in the manifest and warned on, as an honest signal.

The vote also carried `restore` into scope unanimously — shipping the backup guard alone would leave a DR procedure whose capture step fails loudly and whose restore step still succeeds silently.

## R-203 evidence

`tests/backup_fail_closed_2444.rs` drives the **real binary** and configures the store through the `AI_MEMORY_STORE_URL` env channel, referencing **no new struct field** — so the identical source compiles and runs against parent `47939793`.

Evidence is in a comment below.

## Residuals left open (filed, not hidden)

- **A Postgres deployment configured with `--store-url` on argv only, with `backup` run from a separate shell, has no channel to learn the backend.** The structural close is a durable backend marker written into the sqlite sidecar by a pg-backed `serve` — that is a new table + migration + cross-backend schema-version pins, deliberately out of scope for this fix.
- **`backup` still opens its source read-write**, so `db::open` runs the migration ladder as a side effect of taking a backup (a newer binary run as a backup cron can schema-upgrade a database a running older daemon holds open).
- **A pg deployment has no sanctioned way to back up its sqlite governance sidecar** (`signed_events`, `governance_rules`), which `pg_dump` does not cover.


alphaonedev:
## R-203 evidence

Same test file, same source, both commits. It references no new struct field and configures the store through the `AI_MEMORY_STORE_URL` channel that already existed at the parent, so it compiles and runs identically on both.

### Against parent `47939793` — 4 failed / 0 passed

```
test result: FAILED. 0 passed; 4 failed; 0 ignored; 0 measured; 0 filtered out
failures:
    backup_on_a_postgres_store_refuses_and_writes_no_artifact_2444
    backup_refuses_to_create_and_snapshot_a_missing_database_2444
    backup_still_succeeds_on_a_real_sqlite_deployment_2444
    restore_on_a_postgres_store_refuses_2444
```

The defect, verbatim from the failure output — `backup` against a Postgres-configured deployment:

```
Unexpected success
command=`AI_MEMORY_STORE_URL="postgres://ai_memory:hunter2@127.0.0.1:5432/ai_memory" ai-memory --db /tmp/.../placeholder.db backup --to /tmp/.../backups`
code=0
stdout=```
Snapshot: /tmp/.../backups/ai-memory-2026-07-29T181719Z.db
Manifest: /tmp/.../backups/ai-memory-2026-07-29T181719Z.manifest.json
SHA-256 : caccdbcbc8ce451db6d441c29a04d151d80bc4e89f0f2741c383b887891034d6
Bytes   : 798720
```
stderr=""
```

798720 bytes of a SQLite database the command created itself, a valid sha256 over it, exit 0, nothing on stderr. `placeholder.db` did not exist before the run.

And `restore` on the same store:

```
Unexpected success
code=0
stdout=```
Previous DB moved to /tmp/.../live.pre-restore-2026-07-29T181720Z.db
Restored /tmp/.../backups/ai-memory-2026-07-29T181720Z.db → /tmp/.../live.db
```
```

The positive control also failed at the parent (`manifest["backend"]`: `None` vs `Some("sqlite")`) — the manifest carried no provenance at all.

### Against this branch — 4 passed / 0 failed

```
running 4 tests
test backup_refuses_to_create_and_snapshot_a_missing_database_2444 ... ok
test backup_on_a_postgres_store_refuses_and_writes_no_artifact_2444 ... ok
test backup_still_succeeds_on_a_real_sqlite_deployment_2444 ... ok
test restore_on_a_postgres_store_refuses_2444 ... ok

test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Plus 11 unit tests in `src/cli/backup.rs` (`cargo test --lib cli::backup`: **18 passed; 0 failed** — the 7 pre-existing ones still pass unchanged).

## CI-parity run (local, exact CI flag set)

`cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic`, all four legs:

```
LEG=default      RC=0
LEG=sal          RC=0
LEG=sal-postgres RC=0
LEG=vectorlite   RC=0
```

- `cargo fmt --all --check` — OK
- `cargo check --examples` — OK
- `qual_10_module_size_ceiling` — 2 passed (`src/cli/backup.rs` carries no ceiling row; no bump needed)
- `qual_6_7_legacy_error_type_ceiling` — 2 passed (both handlers stay on `anyhow::Result`)
- `cli_integration` 20 passed · `cli_subcommand_count_invariant` 2 · `route_count_invariant` 8 · `migration_ladder_integrity` 1 — all green
- Shell gates, all PASS: `check-hardcoded-literals` · `check-vendor-literals` · `check-docs-vs-ssot` · `check-migration-ladder` · `check-l3-boundary` · `qc-codegraph-precheck` · `check-test-env-lock` · `check-test-stdin-reads`

SSOT: two new CLI **flags** on existing subcommands, no new subcommand / MCP tool / HTTP route / migration / schema bump. `EXPECTED_CLI_SUBCOMMANDS_DEFAULT` (89) / `_SAL` (91) / `EXPECTED_PRODUCTION_ROUTES_COUNT` (94) / the token-budget ceiling / the param-names census are all unmoved, and `check-docs-vs-ssot.sh` confirms it (`schema=87, full_tools=103, routes=94, cli_default=89, cli_sal=91`).

Issue #2493, opened 2026-07-30
[data-integrity][pg parity] 7 of 8 postgres delete/archive funnels leave a dangling namespace_meta.standard_id — #1642 was closed on one arm


alphaonedev:
## Summary

#1642 fixed the dangling-`namespace_meta.standard_id` defect on `PostgresStore::delete` only. **Seven other postgres funnels that hard-delete a live `memories` row do not clear it**, so governance keeps resolving a namespace standard whose backing memory no longer exists.

Found during a Fable 5 1×7 adversarial vote on #2488/#2491, when one arm (`apply_remote_deletion`) was proposed for inclusion in that PR and the sweep showed it was 1 of 7.

## The class

`PostgresStore::delete` clears it at `src/store/postgres.rs:16752` via `SQL_DELETE_NAMESPACE_META_BY_STANDARD_ID`, with the #1642 rationale at `:16745-16751`:

> pre-fix the postgres delete left a dangling `namespace_meta.standard_id` so governance kept resolving a standard whose backing memory no longer existed

Every other pg funnel that reaps a live row omits it:

| pg funnel | file:line | clears `namespace_meta`? |
|---|---|---|
| `delete` | `postgres.rs:16733` | **yes** (#1642) |
| `apply_remote_deletion` | `:18260` | **no** |
| `forget` | `:19602` | **no** |
| `consolidate` | `:19995` | **no** |
| `run_gc` | `:21961` | **no** |
| `size_gc` | `:22067` | **no** |
| `archive_by_ids` | `:22582` | **no** |
| `update_with_archive_on_supersede` | `:5246` | **no** |

sqlite clears it in four places — `db::delete` (`src/storage/mod.rs:3484`), `archive_memory_no_tx` (`:3838`), `archive_memory_for_caller` (`:4280`), `size_gc` (`:12235`) — so this is a genuine cross-backend divergence, not a shared gap.

## Impact

A namespace standard is not decorative. `read_namespace_policy` (`src/storage/mod.rs:16938-16941`) resolves the policy **through** `namespace_meta.standard_id` → `get(standard_id)` → `GovernancePolicy::from_metadata`. With the pointer dangling, `resolve_governance_policy` returns whatever that broken lookup yields — and `enforce_governance` (`:17360-17362`) is **allow-on-silence**:

```rust
let Some(policy) = resolve_governance_policy(conn, namespace) else {
    return Ok(GovernanceDecision::Allow);
};
```

[...]

Issue #2491, opened 2026-07-30
[federation][data-integrity] sqlite zero-config federated deletions are ALL silently refused — live delete-replication outage since #1934, replicas diverge permanently


alphaonedev:
> **⚠ CORRECTED 2026-07-30 — read this before implementing.** The original body of this
> issue asserted that a refused deletion left "the sender believing the tombstone
> replicated." **That mechanism is wrong and has been corrected below.** Since #2341,
> `success_report_non_ack_reason` (`src/federation/sync.rs:261-276`) treats `skipped > 0`
> inside a 2xx as a **non-ack**, so the origin records `Fail`. The defect is real but the
> mechanism differs — see "What the sender actually learns".
>
> This correction is in the **body** rather than only a comment because an implementer
> built from the stale body and shipped the stale claim into a CHANGELOG entry (caught on
> PR #2497 by an honesty-audit lens). A correction that lives only in a comment does not
> reach the person doing the work.

## Summary

On sqlite at HEAD, **every federated deletion in a zero-config mesh is silently refused** and counted as `skipped`. Deleted rows never propagate and replicas diverge permanently.

Introduced by `540465b5` (2026-07-07, the #1934 authz/isolation cluster) and never noticed.

## Mechanism

`src/handlers/federation_receive.rs:1767-1795` — the sqlite deletions loop calls `peer_attestation::namespace_allowed` **unconditionally**, with no enrolled-posture wrapper:

```rust
// src/federation/peer_attestation.rs:357-372
let Some(peer) = peer_header.map(str::trim).filter(|s| !s.is_empty()) else {
    return sync_trust_peer_bypass();          // header absent  -> false by default
};
match config.scope_for(peer) {
    Some(scope) => scope.allowed_namespaces.iter().any(|p| glob_match(p, namespace)),
    None => sync_trust_peer_bypass(),         // unenrolled/zero-config -> false by default
}
```

`sync_trust_peer_bypass()` is `matches!(env::var(...), Ok("1"))` — **default false**. So in zero-config (`has_allowlist() == false`), `scope_for` returns `None` and **every deletion is refused**. Same for a header-absent push.

## What the sender actually learns (CORRECTED)

- **The origin does NOT record success.** Since #2341, `skipped > 0` in a 2xx is a non-ack, so the quorum tracker records `Fail`.
[...]

Issue #2488, opened 2026-07-30
GA-BLOCKER (HIGH, CWE-284): postgres federated-delete namespace gate fails open for an enrolled peer that declares no allowed_namespaces


alphaonedev:
## Summary

The postgres federated-delete namespace gate landed with #2447 is **skipped entirely** for an enrolled peer whose `allowed_namespaces` is empty/absent — exactly the peer shape #2447's Layer 2 was voted in to refuse. `app.store.apply_remote_deletion` then runs unguarded, so such a peer can delete rows outside any namespace scope on a postgres-backed node by id.

Found by a Fable 5 audit of merge `458a8b43`; independently confirmed by reading `release/v1.0.0` @ `4f28f1d8`.

## Root cause — an optimization predicate used as an authorization predicate

`src/handlers/federation_signing_check.rs:633` wraps the whole gate:

```rust
if ns_scope_needs_existing {
    match resolve_stored_namespace(&app, body.sender_agent_id.clone(), del_id).await {
        ...namespace_allowed(...) -> skipped += 1; continue;
    }
}
match app.store.apply_remote_deletion(&ctx, del_id).await { ... }
```

`ns_scope_needs_existing` is `receive_auth::inbound_write_needs_existing_namespace` (`src/federation/receive_auth.rs:709-714`):

```rust
attest_cfg.has_allowlist() && peer_declares_namespace_scope(peer_id, attest_cfg)
```

and `peer_declares_namespace_scope` (`receive_auth.rs:690-699`) returns **false** when `scope.allowed_namespaces.is_empty()`.

That predicate is documented at `receive_auth.rs:701-707` as a **hot-path read-elision optimization**:

> \`false\` short-circuits a per-memory \`get\`-by-id on the federation hot path: when Layer 1 is not armed for this peer the stored namespace cannot change the verdict (Layer 2 refuses — or accepts — on the peer's scope alone). Zero-config deployments therefore pay ZERO extra reads.

[...]

Issue #2482, opened 2026-07-29
[CI-FLAKE] AI_MEMORY_AGENT_ID test lock guards only the writers — ambient-caller readers across the lib test binary can observe a half-applied ai:bob


alphaonedev:
**Class:** test-isolation / CI flake (cross-module, process-global env). **Found at `release/v1.0.0` @ `47939793`**, while landing #2448 (PR #2472). One instance is already fixed there; this issue tracks the residual class.

## Symptom

`Check (windows-latest)` went red on PR #2472 with a failure in a module that PR does not touch:

```
test mcp::delete::tests::double_delete_errors_second_time ... FAILED
thread 'mcp::delete::tests::double_delete_errors_second_time' panicked at src\mcp\tools\delete.rs:434:14:
first delete: "caller does not own this memory"
test result: FAILED. 6936 passed; 1 failed
```

The PR touched only `src/tls.rs`, `src/cli/sync.rs`, `src/security_profile.rs` and tests. It added one new lib unit test, which changed thread scheduling in the ~7000-test lib binary and surfaced a latent race.

## Root cause — half-locked, not unlocked

`AI_MEMORY_AGENT_ID` is process-global. Several MCP tool test modules transiently mutate it to exercise cross-owner gates, and they **do** correctly take `crate::identity::agent_id_env_test_lock()`. Example — `src/mcp/tools/delete.rs:705`:

```rust
let _envg = crate::identity::agent_id_env_test_lock();
…
unsafe { std::env::set_var("AI_MEMORY_AGENT_ID", "ai:bob") };   // cross-owner refusal
…
unsafe { std::env::set_var("AI_MEMORY_AGENT_ID", "ai:alice") }; // owner allowed
…
unsafe { std::env::remove_var("AI_MEMORY_AGENT_ID") };
```

**But the lock only serialises the writers against each other.** Every test that resolves its caller from the *ambient* env — and expects the trust-all `None` disposition — runs unlocked and concurrently, so it can observe the window where the var is `ai:bob`. The ownership gate then does exactly what it should (refuse), and the reader panics.

Reader-side exposure with a writer-side-only lock. A mutex is only a mutex if **every** participant takes it.

## Scope is the whole lib test binary, not one file

`cargo test --lib` is ONE binary, so a writer in any module can redden a reader in any other. Current writer/lock census:

| module | `agent_id_env_test_lock` sites | `set_var("AI_MEMORY_AGENT_ID")` sites |
[...]

PR #2497, opened 2026-07-30
fix(#2488,#2491): confine the federated DELETE lane on both backends via the shared by-id namespace verdict

alphaonedev:
> **Review status:** Fable 5 five-lens review complete; 4/5 lenses cleared, the bypass lens found a regression this PR introduced. All three blockers are addressed in `3c3878da` — see [§Review response](#review-response). The scope claim below is deliberately narrower than the first revision of this description.

## Summary

The federated `deletions[]` funnel was broken in **opposite directions** on the two backends. Neither was correct and they **disagreed**, so a mixed-backend mesh could not converge on what exists. Fixed together, because they are the same two lines on the two funnels and fixing either alone leaves them disagreeing.

| backend | defect | direction |
|---|---|---|
| postgres | **#2488** (CWE-284, GA blocker) — gate structurally UNREACHABLE for the enrolled-unscoped peer | fail-OPEN: a peer refused every write could still hard-delete anything by id |
| sqlite | **#2491** (data-integrity) — every ZERO-CONFIG deletion refused inside an HTTP 200 | silently LOSSY: permanent replica divergence, warn-only `Fail`, no retry |

**Root cause on pg — an optimization predicate used as an authorization predicate.** The #2447 gate wrapped itself in `receive_auth::inbound_write_needs_existing_namespace`, documented in that module as a hot-path READ-ELISION optimisation, which returns `false` for an enrolled peer whose `allowed_namespaces` is empty. That is a legal shape (`PeerScope::allowed_namespaces` is `#[serde(default)]`, so enrolling a peer purely for `allowed_sender_agent_ids` silently yields `[]`) and it is *exactly* the shape #2447 Layer 2 was voted in to refuse. In the WRITE lane a `false` verdict still falls through to Layer 2; in the DELETE lane it skipped **all** checking.

**Root cause on sqlite.** The #1934 gate called `peer_attestation::namespace_allowed` unconditionally, with no enrolled-posture wrapper. Its `scope_for(peer) == None` arm falls through to `sync_trust_peer_bypass()` (default **false**), so under zero-config every deletion was refused and counted `skipped` in a 200. Since #2341 a `skipped > 0` report **is** a non-ack, so the origin does not mis-count it as replicated — but `broadcast_delete_quorum` handles that `Fail` with a warn-only log and **no retry, no DLQ enqueue** (#2498), `/sync/since` carries no tombstones, and `QuorumPolicy::majority` can still satisfy `W` from the local commit plus one healthy peer. So the operator's DELETE returns 200 while a replica keeps the row forever: permanent, silent divergence. `receive_auth.rs` had documented this exact trap for the write lane since #2447; the delete lane never got the `has_allowlist()` guard.

## The fix

Both funnels are guarded on the ENROLLED posture (`PeerAttestationConfig::has_allowlist()`) and route through the shared `receive_auth::inbound_by_id_namespace_authorized`, so every lane that reaches the verdict returns the same answer for the same peer shape, and the empty-scope disposition honours `AI_MEMORY_FED_REQUIRE_PUSH_NAMESPACE_SCOPE` (env row 147) instead of hard-coding it. The stored-namespace probe stays **elided** on the Layer-2-only shapes, so the fix costs **zero extra reads** versus the broken code.

Precisely which lanes: `memories[]` + `deletions[]` on **both** backends, and `archives[]` + `restores[]` on sqlite. On postgres those two subcollections never reach the verdict at all — the funnel reports them `unsupported_on_postgres` (fail-closed, so no hole, but "all four lanes on both backends" would be false as written).

`inbound_by_id_namespace_authorized` now takes `stored_namespace: Option<&str>` (not `unwrap_or("")`), and `None` arriving while Layer 1 *is* armed fails **closed**.

### It also removes a permanent erasure-denial primitive

Both lanes probed the target row through a FULL-ROW read (`db::get` / `MemoryStore::get` → `SELECT * FROM memories WHERE id = $1`) to learn one TEXT column. Both full-row mappers are pinned to `DecryptFailurePolicy::FailClosed`, so a row whose at-rest envelope will not open on this node returned `Err`, landed in the gate's fail-closed arm, and became **permanently un-deletable by federation with no operator escape hatch** (`AI_MEMORY_STRICT_DECRYPT_READS` only hardens). The row whose content this node cannot read is the row an operator most urgently wants gone, and it was the one row the gate made immortal — behind an HTTP 200. The pg shape also shipped the `embedding vector(N)` (~12 KB at 3072-d), the stored `tsv`, full `content` and `encrypted_envelope` per probed id.

Both lanes now use a **scalar** namespace projection, which cannot fail for the decrypt reason: sqlite reuses `db::namespace_by_id` (whose doc has warned about routing a namespace gate through `get` since #2447 — the delete lanes were the only lanes ignoring it); postgres gets a new `MemoryStore::namespace_by_id` trait method, T1 public-contract, default composing `get` so third-party impls keep compiling, both shipped adapters overriding it (pg reuses `SQL_SELECT_NS_VERSION_BY_ID`, the shape `apply_remote_deletion` already used internally).

## Scope — the destructive-capability class is NOT retired

What shipped is **the `deletions[]` funnel confined on both backends**. `apply_remote_deletion` is one of several receive-side funnels that can reach a hard `delete`, and the review probed two others end-to-end to a destroyed row under **default config** with a correctly-enrolled, correctly-scoped, TOFU-passing peer:

- `pending_decisions[]` reaches `storage::delete` on an attacker-chosen `memory_id` with **no namespace check** — `pending_author_authorized` inspects only `requested_by` and the payload's `metadata.agent_id`, never `pa.namespace`, and never resolves `pa.memory_id`.
- `namespace_meta[]` has **no scope gate at all**, letting a peer rebind a foreign namespace's governance standard to an approver it controls and then self-approve (`set_namespace_standard` permits cross-namespace binding; `ApproverType::Agent` compares ids and never calls `is_registered_agent`).

Tracked as **#2478** / **#2479**, both sqlite-reachable, and **neither increments the `deleted` counter** — so such a delete is invisible in the 200 response, the refusal WARN, and the DLQ cause set this PR extends. Namespace scope is the control on this funnel; it is not yet the control on the capability. This is stated in the CHANGELOG entry and in the `apply_remote_deletion` doc comment so the class cannot be read as closed.

## Review response

**BLOCKER 1 — a regression this PR introduced, now fixed.** For {allowlist configured, `X-Peer-Id` absent, knob 147 falsy} the parent REFUSED and the first revision ALLOWED: `has_allowlist()` entered the gate, `peer_declares_namespace_scope(None)` was false, the `None` arm let the knob decide. The lens probed an **anonymous** push hard-deleting a `secure/ops` row (`deleted: 1`) on a node whose only enrolled peer is scoped `public/*`, with `AI_MEMORY_FED_SYNC_TRUST_PEER` removed. It also falsified this module's own claim that the knob's blast radius is ONE config shape.

Three distinct peer shapes reach Layer 2 under an enrolled posture and the knob was written for exactly one. New `peer_enrolled_in_allowlist` predicate splits them; header-absent and not-in-a-non-empty-allowlist are now refused **unconditionally**, before the knob is consulted, with a shape-naming WARN. The enrolled-unscoped rows — the #2488/#2491 closures — are unaffected, and the knob's documented blast radius is now structurally enforced rather than asserted. The malformed-header case is covered by the same arm: a whitespace-only header collapses to `None` in `extract_peer_id`, and the #1056 guard is `if let Some(peer_id)`, so "malformed" and "absent" are indistinguishable there. The tightening lands on the shared verdict, so the WRITE lane gets it too — deliberate, fail-closed, and it makes the lanes agree.

**BLOCKER 2 — the untested shapes.** Probing corrected my assumption rather than confirming it: a **present-but-unlisted** peer never reaches the namespace gate, because #1056 refuses it with a typed `401 x_peer_id_not_in_allowlist` before any subcollection loop, and that guard has no knob (it fires on `has_allowlist()` regardless of `AI_MEMORY_FED_REQUIRE_PEER_ENROLLMENT`). So the new cells assert the **honest** disposition — 401 + row survives, under both knob states, both allowlist shapes, both backends — and the gate's own refusal of that shape is pinned purely in the unit test as defence-in-depth. That split matters: the ordering is the *only* thing making the gate arm unreachable, so asserting both layers means neither can be removed silently. The header-absent shape *does* reach the loops and is the reachable regression; it is pinned on both backends.

**BLOCKER 3 — narrowed the claim.** See §Scope above; applied to the CHANGELOG heading, a new CHANGELOG scope bullet naming #2478/#2479, and the `src/store/mod.rs` architectural comment.

**Non-blocking, same pass.** Rot fix — `src/store/{sqlite,postgres}.rs` join the foundational `__ALL__` list (an edit to either impact-selected 58/527 and 21/527 *without* the pin covering it, because the tokeniser matches `postgres`, not `_pg`; `--self-test` 12/12). F1 — the "origin was told the erasure had propagated" sentence was false since #2341 and is corrected in the CHANGELOG, the test module doc, and the assertion message. F2 — the four-lanes claim is qualified. The closed DLQ cause set in CLAUDE.md row 84 + `docs/federation.md` had drifted behind the code **twice** (`unenrolled_author_strict` from #1464/#1801 was never mirrored; `namespace_probe_unresolvable` is new) and now names `classify_quarantine_cause` as the SSOT. #2498 cited where the delete-lane DLQ enqueue was "tracked separately".

## Tests — R-203 evidence

**The obvious approach was worthless.** `tests/l07_3_chunk_d_http_surface.rs:2811` already asserted `deleted >= 1` on a zero-config sqlite push and was **green over the broken lane**, because `build_router_fixture` pre-sets `AI_MEMORY_FED_SYNC_TRUST_PEER=1` and masks the refusal (`install_federation_legacy_bypass_pg` does the same on pg). Both new files build their own router with that var explicitly removed.

Verified locally against parent `4f28f1d8` and against the first revision `6d73ff0b`, on live PostgreSQL 16 + `vector` + `age`:

| cell | at `4f28f1d8` | at `6d73ff0b` | now |
|---|---|---|---|
| `zero_config_federated_deletion_applies_2491` (sqlite) | **RED** | green | green |
| `header_absent_federated_deletion_disposition_2491` (sqlite) | **RED** | green | green |
| `enrolled_unscoped_…_applies_under_knob_off_2488` (**sqlite**) | **RED** | green | green |
| `undecryptable_envelope_row_…_2488` (sqlite) | **RED** | green | green |
| `enrolled_unscoped_federated_deletion_refused_2488_pg` **(the #2488 exploit)** | **RED** | green | green |
| `undecryptable_envelope_row_…_2488_pg` | **RED** | green | green |
| `header_absent_…_refused_under_enrolled_posture_2497` (sqlite) | green | **RED** | green |
| `malformed_whitespace_peer_id_…_2497` (sqlite) | green | **RED** | green |
| `header_absent_…_refused_under_enrolled_posture_2497_pg` | green *(also allowed at parent — #2488 fail-open for this shape too)* | **RED** | green |
| `unknown_peer_…_refused_under_enrolled_posture_2497` (both backends) | green | green | green — pins the 401 ordering |
| `enrolled_unscoped_…_refused_by_default_2488` (sqlite) | green | green | green — control |
| `enrolled_scoped_…_confinement_controls_2488` (both backends) | green | green | green — #1934 control |
| `zero_config_federated_deletion_still_applies_2488_pg` | green | green | green — availability guard |
| `enrolled_unscoped_…_applies_under_knob_off_2488_pg` | green **for the wrong reason** — at parent the gate sat inside the elision predicate, so the deletion applied *regardless of the knob*; the sibling exploit cell with the knob ON is RED there, proving the knob had zero effect | green | green — knob-reachability pin |

Six cells red at `4f28f1d8` → green; three red at `6d73ff0b` → green. The unit-test arms (`inbound_by_id_verdict_option_contract_2488`) cannot be proven red by stashing `src/`, because they live in the same file as the fix — so they were proven by **mutation** instead: disabling only the new unconditional-refusal branch (`if false && !peer_enrolled_in_allowlist(...)`) while keeping every test arm fails at `peer=None require=false stored=None`, then reverts to green.

`tests/security_authz_isolation_cluster.rs` (the #1934 pin) passes **UNEDITED** — 5/5. Also re-run green: `federation_write_ns_scope_2447` 6/6, `…_2447_pg` 4/4, `l07_3_chunk_d_http_surface` 161/161, `handler_postgres_branches_fake_pg` 90/90, `cov_ga2_federation` 25/25, `cov_ga2_pg_federation` 11/11, `erasure_fanout_g30` 7/7, both qual ceilings.

Every assertion is on **DB row state**, counters secondary — `skipped` increments for several unrelated reasons. On pg the row probe is raw SQL, because `MemoryStore::get` both folds a #910 visibility denial into `NotFound` and fails closed on the very undecryptable row under test. Posture teardown is an **RAII drop guard** (the #2482 analogue). pg tests use `#![cfg(feature = "sal-postgres")]` + runtime soft-skip, deliberately **not** `#[ignore]` (the PR pg job does not pass `--include-ignored`).

Honest note: `store_parity_gaps` has 4 pre-existing `#[ignore]` pg failures (`pg_parity_gap_1_version`, `pg_parity_gap_3_recall_observations`, `pg_parity_private_leak_and_bypass_a7_1720`, `trait_update_bumps_version_1024`) that reproduce identically at parent and are untouched here.

## Out of scope (separately tracked)

- Restoring per-row ownership on the federated delete path — **refuted** by the vote; the doc comment landed here explains why so it is not re-filed.
- The `namespace_meta.standard_id` dangling-clear — a 7-arm pg class, **#2493**.
- Delete-lane DLQ enqueue in `broadcast_delete_quorum` — **#2498**; tombstones / anti-entropy for the delete verb.
- `pending_decisions[]` / `namespace_meta[]` namespace gating — **#2478** / **#2479** (see §Scope).

## Gates

`cargo fmt --all --check`; `cargo check --all-targets` on default + `sal` + `sal-postgres`; `cargo check --examples`; **clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` on all four legs** (default, `sal`, `sal-postgres`, `vectorlite`); both qual ceilings (no bump needed); `check-hardcoded-literals.sh`, `check-vendor-literals.sh`, `qc-codegraph-precheck.sh`, `check-docs-vs-ssot.sh`, `check-l3-boundary.sh`, `check-migration-ladder.sh`, `ci-test-impact.sh --self-test` (12/12). No SSOT count drift: no new MCP tool / HTTP route / CLI verb / schema version / response field / env row, no new `for_admin` site.

## AI involvement

Implemented autonomously by Claude Opus 5. The design was **not** chosen by the implementer: it was decided by a completed Fable 5 1×7 adversarial vote (`4d3ea1c5`), 7/7 unanimous for variant B′, recorded on #2488 and in ai-memory `e5f15dda`. The vote's `TOP_RISK` — an unconditional probe still routed through `store.get`, which would have turned the decrypt trap into a fleet-wide undeletable-row class on both backends while every existing test stayed green — is the shape this PR deliberately does not take, and the undecryptable-envelope cells are the pins that prove it. A subsequent five-lens review of the *implementation* found the header-absent regression and the scope overclaim; both are fixed in `3c3878da` rather than argued with.


alphaonedev:
## ⛔ BLOCKING — Fable 5 adversarial-bypass lens: **BYPASS FOUND** (0.93, probed end-to-end). Do not merge as-is.

Four of five review lenses cleared this PR (SAL trait SOUND-WITH-RESIDUALS 0.9; blast radius CLEAN 0.87 with every pre-existing test run green and unedited; honesty audit 3 non-load-bearing overclaims; CI pg job independently confirmed 5/5 executed on live postgres, `0 ignored`, `0 filtered out`).

The fifth found a **regression this PR introduces** and a **working exploit chain against the new code**. The `deletions[]` work is correct and worth landing. The framing is not.

---

## BLOCKER — this PR widens the header-absent shape from REFUSE to ALLOW

Verified independently at the source, not taken from the lens:

**Parent** (`4f28f1d8:src/handlers/federation_receive.rs`) called `peer_attestation::namespace_allowed(peer_header, &row.namespace, &attest_cfg)` **unconditionally**. For an absent header, `namespace_allowed` (`src/federation/peer_attestation.rs:362-364`) does:

```rust
let Some(peer) = peer_header.map(str::trim).filter(|s| !s.is_empty()) else {
    return sync_trust_peer_bypass();     // false by default
};
```

→ **REFUSED.**

**This PR:** `has_allowlist()` is true → gate entered → `peer_declares_namespace_scope(None, cfg)` is false → falls to the `None =>` arm → `layer2_unscoped_peer_authorized(lane, id, None, peer_id, require_push_namespace_scope)`. With knob 147 falsy → **ALLOW.**

The lens probed it: an **anonymous** push (no `X-Peer-Id`) hard-deleted a `secure/ops` row — `deleted: 1` — on a node whose only enrolled peer is scoped `public/*`. `AI_MEMORY_FED_SYNC_TRUST_PEER` was explicitly removed from the environment, so this is not the legacy bypass.

Reachability needs two documented operator rollout knobs: `AI_MEMORY_FED_TRUST_BODY_AGENT_ID=1` (without it, `403 peer_id_header_missing` — control probe confirmed) **and** `AI_MEMORY_FED_REQUIRE_PUSH_NAMESPACE_SCOPE=0`. Both are shipped rollout hatches, and the combination now yields unbounded federated delete.

**And it falsifies the knob's own documentation.** `src/federation/receive_auth.rs:645-651` asserts: *"Its entire blast radius is ONE config shape — the operator … left `allowed_namespaces` empty/absent."* After this PR the knob also governs **header-absent** and **peer-not-in-a-non-empty-allowlist**. That is the exact defect class this campaign exists to close, introduced by the PR that closes two instances of it.

Also worth pinning: a **malformed** peer id (e.g. `"   "`) is dropped by `extract_peer_id` (`federation_receive.rs:49-63`) → `None`, and the #1056 TOFU guard at `:950` is `if let Some(peer_id)`, so it is **skipped entirely** and the request is treated as anonymous. "Malformed header" and "no header" are indistinguishable to the enrollment gate.

### Required in-PR fix

Make Layer 2 refuse the **header-absent** and **not-in-a-non-empty-allowlist** shapes **unconditionally**, so the knob means only what its doc says. That restores parent behaviour for those shapes (no regression), keeps the documented blast radius true, and leaves the #2488/#2491 closures intact — the peer-shape matrix shows the enrolled-unscoped rows are unaffected by this change.

Alternative — re-scope the row-147 doc claim to match the widened reach — is **not** acceptable here: it would document an anonymous-delete capability as intended behaviour.

---

## The `deletions[]` fix is correct — the SCOPE CLAIM is what must change

The lens's census of all 13 `SyncPushBody` subcollections found **`pending_decisions[]` reaches `storage::delete` on an attacker-chosen `memory_id` with zero namespace gate**, and probed the full chain to a destroyed row under **default config** with a correctly-enrolled, correctly-scoped, TOFU-passing peer:

- `pending_author_authorized` (`federation_receive.rs:627-647`) checks only `pa.requested_by` and `payload.metadata.agent_id` — **`pa.namespace` is never checked and `pa.memory_id` is never resolved**
- `execute_pending_action` (`storage/mod.rs:18058-18064`) does `"delete" => { delete(conn, &mid)?; }` where `mid = pa.memory_id`, **decoupled even from `pa.namespace`**
- the approver is selected by `resolve_governance_policy(conn, &pa.namespace)` — an **attacker-supplied** namespace
- and `namespace_meta[]` (`:2495-2520`) has **no scope gate at all**, while `set_namespace_standard` (`storage/mod.rs:16736`) explicitly permits cross-namespace binding — so the peer rebinds `secure/ops`'s governance standard to its own, declaring `approver: {agent: "ai:evil"}`, then self-approves. `ApproverType::Agent` (`storage/mod.rs:17925-17930`) compares ids and never calls `is_registered_agent`.

**Detection gap:** in both variants the response envelope reports `"deleted": 0, "skipped": 0, "noop": 0` — the `deleted` counter only tracks `deletions[]`. A hard delete of a foreign-namespace row is invisible in the 200 response, the WARN log, and the DLQ cause set this PR extends.

These lanes are tracked as #2478 and #2479, but **neither issue describes a working end-to-end delete chain** — I am escalating both with the probed evidence. sqlite-only: on postgres these land in `unsupported_on_postgres` (`federation_signing_check.rs:792-798`), which inverts the cross-backend-parity claim for this lane.

### Required before merge

1. **Fix the header-absent regression** (above). In-PR, blocking.
2. **Narrow the claim.** CHANGELOG, PR body, and the `src/store/mod.rs:1744-1748` architectural comment must not assert *"CWE-284 federated-delete namespace confinement"* or that namespace scope is *the* control on this lane while a second receive funnel into the same `delete` enforces nothing. Say what shipped: **the `deletions[]` funnel is confined on both backends.**
3. **Do not let this retire the class.** #2478/#2479 must be referenced as open holes in the same destructive capability, in the CHANGELOG entry itself.

Plus the already-identified prose fixes: F1 (the `"origin was told the erasure had propagated"` sentence — false since #2341, corrected at source in #2491's body), F2 (`"all four lanes on both backends"` — on postgres `archives[]`/`restores[]` never reach the verdict), and the closed DLQ cause set in CLAUDE.md env row 84 + `docs/federation.md:450-452`.

---

## What cleared

Recording this so the blocking finding is not read as a verdict on the whole PR:

- **Probe elision — sound.** Verified read + probed: when `peer_declares_namespace_scope` is false, `layer2_unscoped_peer_authorized` reads `namespace` only for the WARN field, never for the verdict. No shape exists where the elided read changes the outcome.
- **`None`-while-armed fails closed, and is not knob-overridable** — probed both `req=1` and `req=0` → `false` both times.
- **Scalar projection** — no exploitable disagreement; the archived-row direction is under-reach (safe). One functional note: an archived row is now permanently un-hard-deletable by federation.
- **Glob edges** — all 14 behave; no bypass.
- **`AI_MEMORY_FED_SYNC_TRUST_PEER` correctly no longer rescues this lane.**
- **TOCTOU** — batch-internal not exploitable (`memories[]` precedes `deletions[]` and #2447 blocks the relocate; `namespace_meta[]` runs *after* `pending_decisions[]`, which is why the exploit needs two pushes). Cross-process is available but requires an out-of-band relocation the attacker cannot perform. Low; a same-tx check-and-delete would close it.

alphaonedev:
## Review complete — 5/5 lenses. Consolidated required-changes list.

Final lens (R-203) verdict: **SATISFIED, 0.88.** It executed every claimed red-at-parent cell in both directions against live postgres, then ran **mutation testing** to prove the suite discriminates rather than merely passes:

| mutation | observed |
|---|---|
| M1 — sqlite probe reverted `db::namespace_by_id` → `db::get` | **exactly 1 failure**: the undecryptable cell |
| M4 — funnel guard + verdict short-circuit both neutered (#2491 reintroduced) | **exactly 2 failures**: `zero_config_…` + `header_absent_…` |
| M5 — `layer2_unscoped_peer_authorized` always-allow | **exactly 1 failure**: `enrolled_unscoped_…_refused_by_default` |

Clean, minimal, non-overlapping. Also confirmed by execution rather than by reading the comment that claims it: the `Drop for PostureGuard` genuinely fires on unwind (real `Drop` impl, bound as a **named** `let _posture` so it lives to scope end, declared **after** the lock so reverse-drop clears posture before releasing it, and `panic = "unwind"` is retained project-wide at `Cargo.toml:539-547`); fixtures are clean — both files build their own routers and `set_posture` `remove_var`s **both** `SYNC_TRUST_PEER` and `TRUST_BODY_AGENT_ID`; and the enrolled-unscoped cells hand-build the allowlist with `allowed_namespaces` **omitted** rather than reusing `set_scoped_posture`, which is precisely how the original defect shipped green.

Both self-disclosures verified. The pg knob-off proof is mechanical: at parent the gate sat inside the elision predicate (false for an unscoped peer), so the deletion applied **regardless of the knob** — and the sibling exploit test with the knob **on** is RED at parent, so the knob demonstrably had zero effect there. One table row refuted as **under-specified**, not fabricated: `enrolled_unscoped_…_applies_under_knob_off_2488_pg` is GREEN at parent (the row names no backend; the sqlite twin *is* red).

---

## REQUIRED before merge

### 1. BLOCKING — fix the header-absent regression

Make Layer 2 refuse the **header-absent** and **not-in-a-non-empty-allowlist** shapes **unconditionally**, so `AI_MEMORY_FED_REQUIRE_PUSH_NAMESPACE_SCOPE` means only what `receive_auth.rs:645-651` says it means. Restores parent behaviour for those shapes; the peer-shape matrix shows the enrolled-unscoped rows — the #2488/#2491 closures — are unaffected.

### 2. BLOCKING — the untested shape the two lenses converged on

The bypass lens found header-absent now **allows** where the parent refused. The R-203 lens found the not-in-allowlist / header-absent **refusal** path has **zero assertions on either backend, on any lane** — and `2447`'s tests never use a non-allowlisted peer header either. Both point at the same hole from opposite sides.

Its failure scenario is the one to prevent: an operator reports deletions silently stopped replicating, someone reads the `None` arm as over-refusal and returns `true` for the not-in-allowlist shape, and **any unknown peer regains unbounded delete-by-id on an enrolled node while all 11 tests plus the unit pin stay green.** #2488 reopened for a different peer shape, with the R-203 proof intact and blind to it.

Add: one cell per backend pushing `Some("ai:unknown")` under both `UNSCOPED_ALLOWLIST` and `SCOPED_ALLOWLIST`, asserting the row **survives**; plus a `None`-peer and unknown-peer arm in `inbound_by_id_verdict_option_contract_2488`.

### 3. BLOCKING — narrow the scope claim

The bypass lens probed `pending_decisions[]` → `storage::delete` on an attacker-chosen `memory_id` with **zero** namespace gate, under default config, to a destroyed row — using the ungated `namespace_meta[]` lane to inject its own approver policy first. Escalated on #2478 / #2479.

So the CHANGELOG, PR body, and the `src/store/mod.rs:1744-1748` architectural comment must not assert *"CWE-284 federated-delete namespace confinement"* or that namespace scope is **the** control on this lane. State what shipped: **the `deletions[]` funnel is confined on both backends**, and reference #2478/#2479 as open holes in the same destructive capability **in the CHANGELOG entry itself**. This must not retire the class.

### 4. Rot fix — impact selection does not protect the probe-shape pin

Verified: `src/store/postgres.rs` → impact-selected 58/527, and `federation_delete_ns_scope_2488_pg` is **not** in the list (the heuristic matches the `postgres` token, not `_pg`). Same for `src/store/sqlite.rs` → 21/527 without `…_2488`.

So a future PR deleting the `namespace_by_id` override at `postgres.rs:18265` — letting the `get`-composing trait default take over — **reinstates the pg erasure-denial trap without re-running the test that pins it.** Add both adapter files to the foundational `__ALL__` list, or add the two test-binary names to an always-run set.

### 5. Prose and doc corrections

- **F1** — CHANGELOG bullet + the `zero_config_…_2491` assertion message: *"the origin was told the erasure had propagated"* is false since #2341 (`sync.rs:261-276` makes `skipped > 0` a non-ack). Replace with *"the origin logs a warn-only `Fail` and never retries."* My error, corrected at source in #2491's body.
- **F2** — *"all four lanes on both backends"*: on postgres `archives[]`/`restores[]` never reach the verdict — `unsupported_on_postgres` (`federation_signing_check.rs:792-798`). Fail-closed, so no hole, but the claim is false as written.
- **Closed DLQ cause set** — CLAUDE.md env row 84 and `docs/federation.md:450-452` enumerate `quota|unenrolled_peer|id_drift|permanent|peer_removed|other` and weren't updated. That enumeration already omitted `unenrolled_author_strict` at parent, so this extends a pre-existing drift; fix both while there.
- **The red-at-parent table** — name the backend on the knob-off row, and mark it green-at-parent-for-the-wrong-reason rather than RED.

---

## Recorded so the blockers aren't read as a verdict on the whole PR

Cleared across the other four lenses: SAL trait **SOUND-WITH-RESIDUALS 0.9** (decrypt trap verified real and removed, probed on live pg, cross-adapter agreement on every edge, all SSOT pins green, C8 clean at 20 sites); blast radius **CLEAN 0.87** (every pre-existing test **run** green and **none edited** — `security_authz_isolation_cluster` 5/5, `2447` 6/6, `l07_3_chunk_d` 161/161, fake-pg 90/90, `store::` 339, `ci-test-impact --self-test` 12/12); honesty audit 3 non-load-bearing overclaims; and I independently confirmed the CI pg job **executed** the new binary — `running 5 tests … 5 passed; 0 failed; 0 ignored; 0 filtered out`, no soft-skip.

Three independent live-postgres executions now agree on the exploit closure. The `deletions[]` work is correct. Items 1–3 are what stand between it and merge.

One property worth carrying forward rather than fixing: M2/M3 showed the zero-config guard is **doubly** implemented (funnel-level and verdict-level), so either single site can be deleted with the suite still green. That is defence-in-depth, not a test defect — but no test tells you which layer is holding, so a future single-site removal gets no signal.

alphaonedev:
## Review response — all three blockers fixed in `3c3878da`

The bypass lens was right, and blocker 1 is a regression I introduced. Not arguing with any of it. Per-cell evidence below; PR description rewritten with the corrected table.

### 1. BLOCKER — header-absent regression: FIXED

Three distinct peer shapes reach Layer 2 under an enrolled posture, and knob 147 was written for exactly one:

| shape | `scope_for(peer)` | governed by knob 147? |
|---|---|---|
| enrolled, `allowed_namespaces: []` | `Some(empty)` | **YES** — the #2447 Layer-2 shape |
| peer id ABSENT / whitespace-only | n/a | **NO** — refuse unconditionally |
| peer NOT in a non-empty allowlist | `None` | **NO** — refuse unconditionally |

New `peer_enrolled_in_allowlist` predicate splits them. The bottom two are now refused **before** the knob is consulted, with a shape-naming WARN. The enrolled-unscoped rows — the #2488/#2491 closures — are unaffected.

I did **not** re-scope the row-147 doc claim. I made it true, and said so at the claim site: the blast radius is now structurally enforced, with a pointer to the predicate and to the regression that skipping the split caused. The tightening lands on the shared verdict, so the WRITE lane gets it too — deliberate: fail-closed, and it makes the lanes agree. The malformed-header case falls in the same arm (whitespace collapses to `None` in `extract_peer_id`, and #1056 is `if let Some(peer_id)`, so absent and malformed are indistinguishable there).

### 2. BLOCKER — untested shapes: ADDED, and probing corrected my assumption

I wrote the unknown-peer cells expecting "row survives via the namespace gate". **Probe said otherwise:** a *present-but-unlisted* peer never reaches the namespace gate — #1056 refuses it with a typed `401 x_peer_id_not_in_allowlist` before any subcollection loop, and that guard has **no knob** (fires on `has_allowlist()` regardless of `AI_MEMORY_FED_REQUIRE_PEER_ENROLLMENT`, which my posture sets to `0`). Confirmed at `federation_receive.rs:951-968` after the probe, not before.

So the cells assert the honest disposition — **401 + row survives**, under both knob states × both allowlist shapes × both backends — and the gate's own refusal of that shape is pinned **purely** in `inbound_by_id_verdict_option_contract_2488` as defence-in-depth. That split is the point: the ordering is the *only* thing making the gate arm unreachable, so relaxing #1056 or adding a surface that skips it makes the gate the last line. Both layers asserted ⇒ neither can be removed silently, which is exactly your failure scenario.

The header-absent shape **does** reach the loops and is the reachable regression — pinned on both backends.

### 3. BLOCKER — claim narrowed

CHANGELOG heading is now "federated `deletions[]` funnel namespace confinement"; a new CHANGELOG bullet states that the class is **not** retired and names #2478/#2479 with the mechanism (including that neither increments `deleted`, so such a delete is invisible in the 200, the WARN, and the DLQ cause set this PR extends). Same narrowing on the `apply_remote_deletion` doc comment: "the control on the `deletions[]` funnel", with an explicit "do NOT read it as 'this method is confined'" section.

### Per-cell evidence

Three commits matter: `4f28f1d8` (original parent), `6d73ff0b` (first revision, regression present), `3c3878da` (now). Live PG16 + `vector` + `age`.

| cell | `4f28f1d8` | `6d73ff0b` | `3c3878da` |
|---|---|---|---|
| `header_absent_…_refused_under_enrolled_posture_2497` (sqlite) | green | **RED** | green |
| `malformed_whitespace_peer_id_…_2497` (sqlite) | green | **RED** | green |
| `header_absent_…_refused_under_enrolled_posture_2497_pg` | green *(also allowed at parent — the #2488 fail-open covered this shape too)* | **RED** | green |
| `unknown_peer_…_2497` (sqlite, 4 sub-cells) | green | green | green — pins the 401 ordering |
| `unknown_peer_…_2497_pg` (4 sub-cells) | green | green | green — pins the 401 ordering |

The unit-test arms can't be stash-proven (they live in the same file as the fix), so I mutation-tested them instead — `if false && !peer_enrolled_in_allowlist(...)`, every test arm intact:

```
test federation::receive_auth::tests::inbound_by_id_verdict_option_contract_2488 ... FAILED
#2497: under an enrolled posture a header-absent / unenrolled / whitespace-only peer
must be REFUSED unconditionally — peer=None require=false stored=None
```

Reverted → green. The arms sweep `{unscoped, scoped} × {None, unenrolled, "   "} × {knob on, off} × {stored None, in-scope, out-of-scope}` on **both** the by-id and write entry points, plus a zero-config guard so the #2497 tightening cannot re-break #2491.

`tests/security_authz_isolation_cluster.rs` still passes **UNEDITED** (5/5). Re-run green: `2447` 6/6, `2447_pg` 4/4, `l07_3_chunk_d` 161/161, fake-pg 90/90, `cov_ga2_federation` 25/25, `cov_ga2_pg_federation` 11/11, `erasure_fanout_g30` 7/7, both qual ceilings. New suites: sqlite 9/9, pg 7/7.

### Non-blocking items

- **Rot fix:** `src/store/{sqlite,postgres}.rs` added to foundational `__ALL__`, with the reason in the comment (58/527 and 21/527 without the covering pin, because the tokeniser matches `postgres`, not `_pg`). `--self-test` 12/12.
- **F1:** corrected in three places — CHANGELOG, test module doc, and the `zero_config_…_2491` assertion message. Thanks for flagging it as yours; I'd built from the issue body and would have shipped it.
- **F2:** qualified — `memories[]` + `deletions[]` both backends, `archives[]`/`restores[]` sqlite only.
- **DLQ cause set:** fixed in CLAUDE.md row 84 **and** `docs/federation.md:452`, noting it had drifted twice (`unenrolled_author_strict` was never mirrored at all) and naming `classify_quarantine_cause` as the SSOT so the mirror relationship is explicit.
- **Table row:** knob-off row now names the backend and is marked green-at-parent-for-the-wrong-reason, with the mechanism.
- **#2498** cited where I'd written "tracked separately".

All four clippy legs green, three `cargo check` legs + examples green, fmt clean, six gate scripts pass. Not merging — yours to re-review.

alphaonedev:
## Re-review of `3c3878da` — verified by direct execution (subagent lenses are 529-blocked; I ran these myself)

Two of the three open questions from the review are now answered with evidence. The third needs adversarial probing and remains open.

### 1. Regression CLOSED — both backends, run directly

sqlite, `--features sal`:
```
header_absent_federated_deletion_refused_under_enrolled_posture_2497 ... ok
malformed_whitespace_peer_id_is_treated_as_absent_and_refused_2497 ... ok
unknown_peer_federated_deletion_refused_under_enrolled_posture_2497 ... ok
zero_config_federated_deletion_applies_2491 ... ok
test result: ok. 9 passed; 0 failed; 0 ignored; 0 filtered out
```

postgres, live PG on `--features sal-postgres`, `--test-threads=1`:
```
enrolled_unscoped_federated_deletion_refused_2488_pg ... ok        ← the original exploit
header_absent_federated_deletion_refused_under_enrolled_posture_2497_pg ... ok
unknown_peer_federated_deletion_refused_under_enrolled_posture_2497_pg ... ok
zero_config_federated_deletion_still_applies_2488_pg ... ok        ← availability control
test result: ok. 7 passed; 0 failed; 0 ignored; 0 filtered out
```

`0 ignored`, `0 filtered out`, no `SKIP` lines — these executed. Both availability controls (`zero_config_*`) pass on both backends, so the fix did not regress in the other direction.

### 2. The 401-ordering question — ANSWERED: the protection does NOT rest on statement order

I asked whether `peer_enrolled_in_allowlist` would still catch an unlisted peer if #1056's ordering ever changed, or whether everything rested on statement order across the ~780 lines between the guard (`federation_receive.rs:948`) and the first subcollection loop (`:1742`).

It does not. The predicate (`receive_auth.rs:744`) performs its **own** lookup:

```rust
peer_id.map(str::trim).filter(|p| !p.is_empty())
       .and_then(|p| attest_cfg.scope_for(p))
       .is_some()
```

No dependency on #1056 having run. It is consumed at the **top** of `layer2_unscoped_peer_authorized` (`:957`), before `require_push_namespace_scope` is read, so the refusal genuinely precedes the knob. The trim-and-filter also matches `extract_peer_id`, so a whitespace-only header collapses to `None` and lands in the same arm — the gap the doc correctly notes #1056 cannot catch, since it is `if let Some(peer_id)`.

### 3. Mutation test — the predicate is SINGLY implemented, and the disclosure was accurate

The first review found the zero-config guard was **doubly** implemented (funnel-level and verdict-level), so either site could be deleted with the suite still green. I checked whether the new predicate has the same property. Mutated its only gate site to `if false && !peer_enrolled_in_allowlist(...)` and ran the sqlite suite:

```
header_absent_federated_deletion_refused_under_enrolled_posture_2497 ... FAILED
malformed_whitespace_peer_id_is_treated_as_absent_and_refused_2497 ... FAILED
unknown_peer_federated_deletion_refused_under_enrolled_posture_2497 ... ok
(7 others ok)  test result: FAILED. 7 passed; 2 failed
```

Reverted clean, zero remnants.

**Two things this establishes.** The predicate is the *sole* guarantee for header-absent and malformed — deleting it breaks exactly the two cells that pin it, so a future removal cannot pass silently. And `unknown_peer_*` **surviving** the mutation independently confirms the implementer's own disclosure: that shape is caught by #1056 upstream, not by this predicate, so its cell pins the **ordering** rather than the predicate. That was reported accurately rather than glossed, and the mutation proves it.

### Still open — needs adversarial probing, not test execution

A fresh bypass hunt on the `6d73ff0b..3c3878da` delta. Tests passing is not the same as no bypass existing; a seven-lens review previously cleared a configuration with three defects. That work is queued and #2497 stays held until it lands.

Also unchanged and out of scope by design: the `pendings[]` / `pending_decisions[]` / `namespace_meta[]` chain probed to a destroyed row in the first review (#2478 / #2479). Confirmed disclosed in the CHANGELOG entry rather than silently left, including that neither lane increments `deleted` — so those deletions are invisible in the 200 response, the WARN, and the DLQ cause set this PR extends.

### Note on the earlier instruction I gave

The consolidated list told the implementer to make Layer 2 refuse header-absent **and** not-in-a-non-empty-allowlist as if they were one shape. They are mechanically different: an unlisted peer is refused upstream by #1056 with no bypass knob, while header-absent skips that guard entirely. The implementer probed rather than following my framing, found the asymmetry, and split them with the predicate above — which is why the fix is correct and my instruction was not. Recording it because the mutation result is only interpretable with that distinction in hand.

alphaonedev:
## Follow-up: the rot fix I asked for has a cost I did not weigh, and a better mechanism already exists in the script

Two more unrelated flakes surfaced on the reruns, both positively identified as environmental and both with prior flake-fix commits:

| test | mechanism | prior fix |
|---|---|---|
| `e2_post_ship_dry_run` | shells out to `cargo build` a crate with **no dependency on `ai-memory`**; failed on `crates-io … curl failed … [16] Error in the HTTP2 framing layer`. Rerun got past it — transience proven. | `693d38b1` (#1742 / #1492) — scoped to llvm-cov only |
| `k7_hmac` | `50 × 100ms` = a hard **5s** budget for a cross-thread wiremock roundtrip, on a runner that just built 529 binaries and is running them at `--test-threads=1` beside a postgres container | `cec84179` "bypass wiremock pool" (#1201) |

Neither is touched by this PR. Now filed as **#2500** (the e2 class) with **#1742** superseded honestly.

### The part that is my fault

The implementer correctly observed that the pg gate's full-suite fail-fast fragility **is a direct consequence of the rot fix this review demanded** — and pointedly declined to quietly revert that fix to turn a red green, which is the right call.

I asked for the three federation files to be added to the foundational `__ALL__` list so a future PR deleting the `namespace_by_id` override could not reinstate the erasure-denial trap without re-running the test that pins it. That reasoning still holds. What I did not weigh is the cost: **`__ALL__` runs all 529 binaries**, and the suite carries enough load-sensitive flakes that a full run is close to a coin flip. So the mechanism I required makes the gate *least* reliable on exactly the changes that most need a trustworthy signal — the foundational ones. That is the same selection effect #2500 documents, and I created a fresh instance of it while fixing a different one.

### The better mechanism was already in the script, and in the original options

`scripts/ci-test-impact.sh:146` defines `ALWAYS_RUN_PARITY_TESTS`, applied at `:230` as *"ALWAYS include the parity invariants (existence-gated)."*

Adding `federation_delete_ns_scope_2488` and `federation_delete_ns_scope_2488_pg` to that list is **strictly better** than what I asked for:

| | `__ALL__` on 5 files | `ALWAYS_RUN_PARITY_TESTS` |
|---|---|---|
| pin fires when those files change | yes | yes |
| pin fires when **any** file changes | no | **yes** — stronger guarantee |
| binaries run | **529** | 2 |
| exposes the suite's flake surface | yes, every time | no |

A stronger pin at roughly 1/265th the blast radius. The R-203 lens actually offered both options — *"add both adapter files to the foundational `__ALL__` list, **or** add the two test-binary names to an always-run set"* — and I relayed both without indicating a preference, so the broader one was a reasonable read of an ambiguous instruction. The preference should have been mine to state.

### Requested change, non-blocking

Move the pin from `__ALL__` to `ALWAYS_RUN_PARITY_TESTS`: drop the five `return 0` foundational arms added for #2488 (the three federation files and the two SAL adapters) and add the two test-binary names to the always-run list instead. Keep the explanatory comments — they record *why* the pin exists, which is the part that rots first.

Not blocking the merge: the current form is **safe**, just costly, and it fails in the loud direction. But it should not ship as the permanent mechanism, because a gate that is unreliable on foundational changes trains people to re-run reds without reading them — which is precisely how the #2497 near-miss (a security fix whose heavy gates never ran) happened in the first place.

### Standing

Still held on the outstanding item, unchanged: the adversarial bypass hunt on the `6d73ff0b..3c3878da` delta. Everything else I have verified directly — 16/16 cells on both backends, the predicate independently load-bearing, and singly-implemented by mutation. The reds are not the reason this is held.

alphaonedev:
## Re-review of `3c3878da` — regression CLOSED (0.97). One in-PR change required; two new defects filed separately.

The lens that found the original bypass re-ran its own probe across all three trees:

```
parent 4f28f1d8:  deleted=0 skipped=1 survived=true    REFUSED
rev1   6d73ff0b:  deleted=1 skipped=0 survived=false   ALLOWED   <- the regression
HEAD   3c3878da:  deleted=0 skipped=1 survived=true    REFUSED
```

Exactly the shape claimed. It also validated its own harness first — the same run at parent reproduces the #2491 silent-lossy bug and its fix at HEAD, proving the probe measures the real lane rather than a mock.

**Full peer-shape matrix at HEAD: no shape allows where it should refuse.** Zero-config keeps applying in both header states, `["**"]` still works as the documented recovery for an identified peer, out-of-scope stays refused, and the knob governs exactly its one documented shape.

Two corrections to earlier claims, both in the honest direction:

- **The mutation claim was too pessimistic.** The implementer said the new arms couldn't be stash-proven because they share a file with the fix. They don't — the two load-bearing cells live in `tests/federation_delete_ns_scope_2488.rs`, a different file from `src/federation/receive_auth.rs`, and they **do** discriminate: mutating the gate site fails `header_absent_…_2497` and `malformed_whitespace_…_2497` while `unknown_peer_…_2497` passes. Better evidence than was claimed for it.
- **The 401-ordering answer is lane-dependent, sharper than my "no".** For the *gated* lanes the predicate stands alone — if #1056 moved after the loops, `deletions[]`/`memories[]`/`archives[]`/`restores[]` would still refuse. But for `pendings[]`, `pending_decisions[]`, `namespace_meta[]`, `links[]`, `signals[]` and the checkpoint lanes, the 401 at `:952` is the **entire** protection across a ~790-line gap with no second check. The load-bearing half is the half this PR does not gate — which is why the lens's probes had to use an *enrolled* peer.

---

## REQUIRED in-PR: F3 — the write lane's availability changed and the CHANGELOG does not say so

The new refusal arm sits in the **shared** `layer2_unscoped_peer_authorized`, so `memories[]` moved with it:

```
X12a  memories[] write — header-absent + allowlist + knob=0 + trust_body=1
  parent 4f28f1d8:  applied=1 skipped=0 row_present=true
  rev1   6d73ff0b:  applied=1 skipped=0 row_present=true
  HEAD   3c3878da:  applied=0 skipped=1 row_present=false
```

Intentional, asserted by the delta's own unit arm, and correct in the security direction. The problem is disclosure: **`grep -c 2497 CHANGELOG.md` returns 0.** The remediation exists only in source doc comments.

So an operator reading the CHANGELOG still believes `AI_MEMORY_FED_REQUIRE_PUSH_NAMESPACE_SCOPE=0` is a general rollout hatch, and on upgrade **loses writes as well as deletes** for the (allowlist configured + header-less legacy peers + knob=0) posture — silently, inside HTTP 200, with no DLQ enqueue (#2498). That posture is precisely the staged rollout the knob exists to serve.

Add to the CHANGELOG entry: the #2497 contract (the knob governs one shape; header-absent and not-in-allowlist are unconditional), and that `memories[]`/`archives[]`/`restores[]` availability changed for that posture. One paragraph. Nothing in the code needs to move.

Also worth a line in the refusal WARN: for a header-absent push under `["**"]`, its first remedy ("enroll the peer") is **inactionable** — you cannot apply a per-peer scope to a peer you cannot identify. Only removing the allowlist works. Correct behaviour, misleading guidance.

---

## Filed separately — both pre-existing, both probed at default config

**#2503 — an in-scope, correctly-ALLOWED delete strips a denied namespace's governance policy.** `storage::delete` (`src/storage/mod.rs:3484`) runs `DELETE FROM namespace_meta WHERE standard_id = ?1` with no namespace predicate; `set_namespace_standard` permits cross-namespace binding; `resolve_governance_policy` fails **open**. A `public/*` peer deleting a `public/ok` row cleared `secure/ops`'s policy — and via the global `*` standard, the whole namespace tree — reported as `deleted: 1, namespace_meta_cleared: 0`. Needs neither #2478's lane nor #2479's.

**This is the killer objection, and it lands on the CHANGELOG's own framing.** The entry says *"Namespace scope is the control on the `deletions[]` funnel."* It isn't. **Namespace scope controls which row the funnel erases; it does not control what that erasure destroys.** The disclosure paragraph closes the two holes it found and inadvertently certifies a third running straight through the funnel it just confined. Worth reflecting in the entry alongside F3.

**#2504 — one malformed character in `AI_MEMORY_FED_PEER_ATTESTATION` disables the entire gate.** `from_env` falls back to empty on parse error; `has_allowlist()` is the master switch for both this gate *and* the #1056 401. Probed: malformed JSON → `deleted=1`, `secure/ops` destroyed. And the two lanes take opposite postures from that one value — `/sync/since` read DENIED, federated DELETE ALLOWED. The WARN calls the fallback "default-deny," true for the read lane and false for this one.

---

## Confirmed clean

The `pendings[]`/`namespace_meta[]` chain is **disclosed, not silently left** — `CHANGELOG.md:21` names #2478 and #2479, states both are sqlite-reachable at default config, and states neither increments `deleted`. `git diff --stat` confirms the delta touches no handler file, so those lanes were not partially altered.

And the probes that found nothing, which is the other half of the value: trim asymmetry between the sibling predicates (byte-identical); whitespace/CRLF/oversize peer-id smuggling past `validate_agent_id`; 14 glob edges including `public/a/../../secure/ops` traversal; archived-row divergence; `attest_cfg` freshness between the 401 and the loops; and monotonicity of the delta at Layer 2 — HEAD's allowance set is a strict subset of rev1's, confirmed cell-by-cell rather than argued from source shape.

**Merge gate:** F3's CHANGELOG paragraph. That is the last item; everything else is verified or filed.

Issue #2050, opened 2026-07-15
[v1.0][CI-BLOCKER] paste git-dependency revision 6a30252 no longer fetchable → all fresh-resolve CI jobs fail repo-wide


alphaonedev:
## Severity: HIGH — blocks CI green on every PR (GA blocker)

`Cargo.toml` pins `paste = { git = "https://github.com/alphaonedev/paste", rev = "6a30252" }` (the RUSTSEC-2024-0436 workaround). That revision **no longer exists** in the fork:

```
Updating git repository https://github.com/alphaonedev/paste
error: failed to load source for dependency `paste`
  unable to update https://github.com/alphaonedev/paste?rev=6a30252#6a302522990cbfd9de4e0c61d91854622f7b2999
  failed to clone into: /home/runner/.cargo/git/db/paste-...
  revision 6a302522990cbfd9de4e0c61d91854622f7b2999 not found
```

## Impact
Every CI job that does a **fresh** cargo dependency resolve fails: `Lint (fmt+clippy)` (fresh 1.96 toolchain), `MSRV`, `Check (ubuntu/macos/windows)`, `Postgres feature gate`, `Cross-compile`, `Code Coverage`, `Dockerfile build`. Jobs only pass when they hit a **warm** `~/.cargo/git` cache that still has the old clone — so it presents as intermittent but is a hard breakage (the rev is gone). Observed red on PRs #2037 and #2043 (and any PR); **not** caused by the PR contents (no manifest changes).

## Fix options (pick one)
1. **Re-pin** `paste` to a rev that exists in `alphaonedev/paste` (re-push `6a30252`, or pin a current SHA/tag).
2. **Migrate to `pastey`** — the community-maintained fork already flagged in the `Cargo.toml` comment as the RUSTSEC-2024-0436 target (vendor into `vendor/pastey/` + `path=`, or fork to `alphaonedev/pastey` + tag-pin).
3. If `paste` is only reached transitively, evaluate dropping the pin for the upstream crate if the advisory is acceptable.

Whichever: bump `Cargo.lock` + re-run the full matrix. This is a one-PR manifest fix but a hard GA prerequisite — GA cannot cut on a red spine.

## Note
Local builds pass because dev caches are warm. Verify the fix on a cold runner (or `cargo clean` + fresh `~/.cargo/git`).

alphaonedev:
[...]

PR #2516, opened 2026-07-30
ci(#2512): supply-chain gate banning git-source cargo dependencies — nightly red is stale `main`, not a release-branch defect

### LINKED ISSUE #2512: certified-AGE nightly hard-red since 2026-07-28: vendored alphaonedev/paste fork rev 6a302522 unreachable — plus AGE pin drift (CI 1.6.0 vs SSOT 1.7.0)
**Found during L1-a ledger verification (2026-07-30), verified first-hand by Fable 5.**

## Defect 1 — nightly red, not flake
`postgres-parity-nightly` has failed 3 consecutive runs (`30331785458` 07-28, `30425388574` 07-29, `30516559819` 07-30). Failed job `Postgres + Apache AGE + pgvector (certified stack, live)`, step `Run AGE-gated Cypher/KG + pgvector suites`:

```
error: failed to load source for dependency `paste`
  unable to update https://github.com/alphaonedev/paste?rev=6a30252#6a302522...
```

The vendored fork revision is unreachable (deleted/GC'd/never-pushed commit in `alphaonedev/paste`). Consequence: **the ONLY certified-AGE CI leg has produced no evidence for ≥3 days** — combined with the AGE `kg_query` silent-CTE-degrade issue (filed alongside this one), no AGE-specific behavior currently has trustworthy CI evidence at all.

The `Postgres parity (#[ignore]-gated, live PG)` job in the same runs is also failing — triage whether it shares the root cause.

## Defect 2 — version drift
SSOT provisioning pins AGE **1.7.0** (`deploy/do-1461/provision/lib.sh:212` at `3d4f9baf`); coverage CI runs **1.6.0** (`.github/workflows/coverage.yml:153`). CI certifies a different AGE than production provisions.

## Fix direction
Restore/repin the `paste` fork rev (or repoint to a reachable rev with lockfile update), get the nightly green, then reconcile the AGE pin in one direction with the SSOT gate extended to cover it (no-hardcoded-literals enforcement pattern).

alphaonedev:
Addresses **#2512 defect 1 only** (the nightly red). The AGE 1.6.0/1.7.0 pin drift (defect 2) is deliberately untouched and remains open — that is a separate decision.

`Closes` keywords are inert on a non-default base, so this does not auto-close #2512.

---

## Root cause: the fix already landed here; the nightly does not run from here

The `paste` fetch error is **not a live defect on `release/v1.0.0`**. #2051 fixed it on this branch on 2026-07-15 by vendoring `paste` 1.0.15 into `vendor/paste/` (commit `6ccf22db`), and `Cargo.toml` @ `3d4f9baf` reads `paste = { path = "vendor/paste" }`. `cargo metadata --offline` resolves `paste 1.0.15` from `path+file://…/vendor/paste`; `Cargo.lock` records it with **no `source` key at all**. There is nothing to re-pin.

`postgres-parity-nightly.yml` triggers on `schedule:`, and **GitHub runs scheduled workflows only from the DEFAULT branch** — `main`. `main` is at `ba24ab2c` and still carries:

```toml
paste = { git = "https://github.com/alphaonedev/paste", rev = "6a30252" }
```

`git merge-base --is-ancestor 6ccf22db origin/main` → **NO**. `origin/release/v1.0.0` → **YES**. `git ls-tree origin/main vendor/` is empty. So the nightly resolves the deleted fork on every run, forever.

**Correcting the issue body:** the nightly has been red since **2026-07-15**, not 2026-07-28. Sixteen consecutive scheduled failures; the three runs cited in #2512 are the most recent three. Last green: run `29308129651`, `main` @ `e440b30752`, 2026-07-14 — both jobs green. Run `29675020389` (07-19) onward all fail. #2050 was filed 2026-07-15 14:50 UTC and closed 16:53 the same day by #2051 — so the nightly went red the same day the bug was found and fixed, and stayed red for 15 days because the fix went to the wrong branch for that workflow's trigger.

### Yes, the `#[ignore]`-gated parity job shares the root cause

Both jobs die on the identical error, before any postgres/AGE work. Verbatim from run `30516559819`:

| job | id | failing step | error |
|---|---|---|---|
| `Postgres + Apache AGE + pgvector (certified stack, live)` | 90787626633 | 9 — `Run AGE-gated Cypher/KG + pgvector suites` | `revision 6a302522990cbfd9de4e0c61d91854622f7b2999 not found` |
| `Postgres parity (#[ignore]-gated, live PG)` | 90787626700 | 8 — `Run` | identical |

```
Updating git repository `https://github.com/alphaonedev/paste`
error: failed to load source for dependency `paste`
  unable to update https://github.com/alphaonedev/paste?rev=6a30252#6a302522
  failed to clone into: /home/runner/.cargo/git/db/paste-eb7868dc19b70f56
  revision 6a302522990cbfd9de4e0c61d91854622f7b2999 not found
```

Cargo resolves `[patch.crates-io]` before it does anything else, so **any** cargo invocation on `main` fails. The parity job's postgres service came up healthy first — the failure is unrelated to postgres. One fix clears both jobs. It is not an AGE-specific or `#[ignore]`-specific problem.

### Why the rev is unreachable: the repository is gone, not the ref

Not a force-push, not a deleted branch, not an unpushed commit — **the whole repository was deleted**:

```
$ git ls-remote https://github.com/alphaonedev/paste
remote: Repository not found.

$ git ls-remote git@github.com:alphaonedev/paste     # authenticated
ERROR: Repository not found.

$ gh api repos/alphaonedev/paste
{"message":"Not Found","status":"404"}

$ gh repo list alphaonedev --limit 200 | grep -i past
(no match — 14 repos, none named paste or pastey)
```

Authenticated and anonymous agree, so it is deletion, not a visibility flip. Cargo's `revision … not found` wording is misleading here: the clone itself failed. GitHub does not preserve a deleted repository's objects, so the pin is unrecoverable **as a pin** — permanently, by anyone.

**Why the last-green nightly could still fetch it:** it could not. The fork was already gone on 2026-07-15; run `29308129651` (07-14) predates the deletion. There is no cache story on the CI side — those are cold runners.

**Why dev boxes mask it:** `~/.cargo/git/db/paste-eb7868dc19b70f56` persists after the remote dies. On this host that clone still exists and still contains the commit:

```
$ git -C ~/.cargo/git/db/paste-eb7868dc19b70f56 cat-file -t 6a302522990cbfd9de4e0c61d91854622f7b2999
commit
$ git -C ~/.cargo/git/db/paste-eb7868dc19b70f56 for-each-ref
6a302522990cbfd9de4e0c61d91854622f7b2999 commit  refs/commit/6a302522990cbfd9de4e0c61d91854622f7b2999
```

Cargo reuses that cache and never contacts the network, so every warm machine stays green while every cold runner fails. That is exactly why #2050 first read as flake against two unrelated PRs.

---

## Supply-chain guardrail: provenance and the inter-rev diff

**No new code is introduced by this PR** — it changes no `Cargo.toml`, no `Cargo.lock`, and no `.rs` file. But the guardrail question about the already-vendored code is answerable in full, and the answer is clean.

**(a) The old rev IS fetchable — from upstream.** `6a302522990cbfd9de4e0c61d91854622f7b2999` is `dtolnay/paste`'s current `master` **and** `HEAD`:

```
$ git ls-remote https://github.com/dtolnay/paste
6a302522990cbfd9de4e0c61d91854622f7b2999   HEAD
6a302522990cbfd9de4e0c61d91854622f7b2999   refs/heads/master

$ git fetch --depth=1 https://github.com/dtolnay/paste 6a302522990cbfd9de4e0c61d91854622f7b2999
 * branch  6a302522990cbfd9de4e0c61d91854622f7b2999 -> FETCH_HEAD    # succeeds
```

So the "fork" was an **unmodified mirror of upstream master**, and the pinned rev survives upstream (`paste` is archived/unmaintained, so `master` has not moved since 2024). The full tree diff of upstream `6a302522` against the in-tree `vendor/paste/`:

```
$ diff -r --brief <upstream-6a302522> vendor/paste -x .git
Only in <upstream-6a302522>: .github
```

**Zero source difference.** Every `.rs`, `Cargo.toml`, `build.rs`, `LICENSE-*`, `README.md`, and `tests/` file is byte-identical; the only omission is the upstream `.github/` CI directory, which is not compiled. Since a git SHA commits to its tree, this is a cryptographic identity argument, not a spot check: `vendor/paste/` **is** the code that rev pinned.

**(b) Provenance of the code now in-tree:**

| field | value |
|---|---|
| commit | `6a302522990cbfd9de4e0c61d91854622f7b2999` |
| contained in | `dtolnay/paste` `refs/heads/master` (also `HEAD`) — no tag points at it |
| subject | `Add unmaintained note` |
| author | David Tolnay `<dtolnay@gmail.com>`, 2024-10-07 01:16:43 +0200 |
| committer | David Tolnay `<dtolnay@gmail.com>`, 2024-10-07 01:16:43 +0200 |
| crate version | `paste` 1.0.15 (matches the `Cargo.lock` entry) |

The stale local fork clone in `~/.cargo/git/db/` yields byte-identical author/committer/subject metadata for the same SHA, independently corroborating that the fork carried no rewrite.

**Cross-check on the RUSTSEC decision (not undone here).** The `Cargo.toml` comment block explains the fork existed as the RUSTSEC-2024-0436 workaround: `paste` 1.x is unmaintained, `[patch.crates-io]` rejects a same-source registry-to-registry rename so `pastey` cannot be patched in directly, and the two viable paths were vendoring or a fork under our org. #2051 took the vendoring path — the one that comment already called "the audit-clean path" — so the RUSTSEC posture is preserved and strengthened, not reverted. Migration to `pastey` remains open and out of scope.

---

## What actually needs to happen to get the nightly green — Fable-executed

I cannot push `main`, so I am reporting rather than doing. **Nothing needs pushing to `alphaonedev/paste`** — recreating that repository and re-pushing the lost commit would work mechanically, but it would re-adopt the exact dependency shape #2051 deliberately abandoned and that this PR's gate now bans. Recommend against.

The needed action is a **commit on `main`** forward-porting #2051 — `Cargo.toml` line 519, `vendor/paste/**` (23 files), and the `Cargo.lock` entry. Minimal and mechanical; a full `release/v1.0.0` → `main` merge is the GA cut and is not appropriate for this.

Two things worth flagging beyond defect 1:

- **This is a class, not an instance.** *Every* `schedule:`-triggered workflow in this repo runs from `main`, which is a pre-v1.0.0 branch. Any nightly gate whose subject changed on `release/v1.0.0` is either testing stale code or red for stale-code reasons. Same family as the known "`Closes #N` is inert on a non-default branch" problem. A `ref:`-pinned checkout in the scheduled workflows would fix it, but that change is itself only effective once it is on `main`.
- **A green proof run needs the `main` commit first.** I have not dispatched `postgres-parity-nightly` (out of scope per my constraints), and dispatching it now would just reproduce the same red from `main`.

I also cannot certify the nightly goes green on the paste fix alone — the AGE pin drift (defect 2) touches the same job's version assertion, and #2512 notes the AGE `kg_query` silent-CTE-degrade issue against the same suite. Those are separate.

---

## This PR: the guard for the class

`scripts/check-git-dependency-sources.sh` — HARD-BLOCK, wired into `c8-precheck.yml` as `Git-dependency-source supply-chain gate (#2050/#2512)`.

No cargo dependency may resolve from a `git` source. Registry (crates.io) or in-tree `path = "vendor/<crate>"` only. Scans every `Cargo.toml` for `git = "…"` (inline-table **and** `[dependencies.x]` section form) and every `Cargo.lock` for `source = "git+…"`, `vendor/` included — a vendored crate that itself reaches for a git source re-imports the whole problem. Ratchet allowlist at `scripts/qc-allowlists/git-dependency-sources-allow.txt`, **currently empty by design**; adding an entry requires naming who guarantees that remote stays reachable for the lifetime of every artifact built from it.

**Why a static ban and not a reachability probe.** Probing remotes at PR time would make this repo's CI depend on third-party availability, go red for network reasons, and teach reviewers to ignore a red gate. The ban is offline and deterministic, and it states the actual rule: third-party source we depend on lives in-tree under `vendor/` where it is reviewable, diffable, and rebuildable with no network — the posture #2051 already adopted, now structural instead of a convention in a manifest comment.

**It removes both properties that hid this.** (1) It fires on every PR, warm cache or cold — the previous signal only existed on cold runners, and a gate that only fires on a cold runner is a gate that does not fire. (2) A future `git = ` pin cannot reach `main` without tripping this on the way in.

`--self-test` (run as its own CI step, per repo convention) plants three violation shapes and one control:

1. the exact #2050 inline `git` + `rev` pin under `[patch.crates-io]`
2. the `[dependencies.x]` section form with a bare `git =` line
3. a `Cargo.lock` `source = "git+…"` — this one matters independently: a `Cargo.toml` fix with a stale lock still breaks `--locked` cold resolves
4. a clean control carrying `repository`/`homepage` URLs plus `path = "vendor/paste"`, which must **not** trip

**North Star.** The failure prevented is not a wrong answer — it is a build that cannot be reproduced, which erases the ability to rebuild, audit, or bisect any artifact this repo has ever shipped. Refusing the un-vendorable edge up front is the fail-closed choice; discovering it on a cold runner months later is not. Worst case for a contributor is a loud, actionable refusal naming both remedies.

---

## Verification

Local, on this branch:

```
scripts/check-git-dependency-sources.sh              → OK (registry + vendor/ only)
scripts/check-git-dependency-sources.sh --self-test  → OK (3 violation shapes caught, clean control passed)
scripts/check-git-dependency-sources.sh --bogus      → usage, exit 2
bash -n scripts/check-git-dependency-sources.sh      → clean
python3 -c "yaml.safe_load(c8-precheck.yml)"         → 13 jobs parse
scripts/check-required-contexts.sh                   → OK (new job needs no mirror entry; rule (a) is one-directional)
scripts/check-hardcoded-literals.sh                  → PASS
scripts/check-docs-vs-ssot.sh                        → PASS
scripts/check-vendor-literals.sh                     → PASS
cargo fmt --all --check                              → clean
cargo metadata --offline                             → paste 1.0.15 from path+file://…/vendor/paste
```

The new job is not yet a required context; that mirror addition is an intentional Fable/branch-protection call, not something to launder in from a PR (see the `required-contexts-release.txt` header on why the mirror is hand-authored from intent).

**Compile gates.** This diff touches no `.rs` file and neither `Cargo.toml` nor `Cargo.lock` (`git diff --name-only` + untracked: zero matches for `\.rs$|Cargo\.(toml|lock)$`), so `cargo check`/`clippy` across the four feature legs and the `qual_10` / `qual_6_7` ceilings are structurally unaffected. The default clippy leg was run anyway as a sanity check.

**SSOT.** No MCP tool, HTTP route, CLI subcommand, or config knob is touched, so no tool-count / `EXPECTED_PRODUCTION_ROUTES_COUNT` / `param_names` / token-budget pin moves. `CHANGELOG.md` `[Unreleased]` carries the entry.


alphaonedev:
**Fable 5 review finding addressed** — `4d993be7` (follow-up commit, SSH-signed).

`--self-test` staged fixtures via a bare `mktemp -d` (system /tmp), violating the zero-strike hard rule and breaking step with `check-migration-ladder.sh:645` / `check-required-contexts.sh:578`, which both stage under the repo. Fixtures now stage at `${ROOT}/.local-runs/gitdep-selftest-$$` (`mkdir -p` + trap-scoped `rm -rf`); `.local-runs/` is gitignored at `.gitignore:53` and `$$` avoids concurrent-run collisions. No `mktemp` call remains.

**One consequential follow-on, flagged because it is not a pure path swap.** `scan()`'s own `find` excluded `'*/.local-runs/*'`, so relocating the fixtures there made every one of them invisible to the function under test — the three violation assertions would have passed **vacuously** while the gate detected nothing. That is precisely the tautology hazard `required-contexts-release.txt` warns about in its header. The `.local-runs` exclusion is now **anchored to the scanned root** (`"${dir}/.local-runs/*"`), which is also just more correct — the intent was always "skip the scratch dir at the scan root", and the live scan (`$dir` == `$ROOT`) still skips it. `*/target/*` stays a free glob because nested build dirs (`fuzz/target`, `loadtest/target`) are real. A do-not-simplify note records why.

**Proven by mutation, not assertion.** Two mutants of the committed script, each required to fail:

| mutant | result |
|---|---|
| manifest detector neutered | `SELF-TEST FAIL` rc=2 |
| `.local-runs` exclusion reverted to a free glob | `SELF-TEST FAIL` rc=2 |

The second is the direct proof that the anchoring is load-bearing and the note is accurate rather than decorative.

Green on the committed script: `--self-test` OK (3 shapes caught, clean control passed), live scan OK, `--bogus` → usage rc=2, trap leaves no residue, no /tmp write on any path. Re-ran `cargo fmt --all --check`, the c8-precheck YAML parse (13 jobs), `check-required-contexts.sh`, `check-hardcoded-literals.sh` — all green. Still zero Rust and zero Cargo manifest change.

**CI: 30/31 pass**, 1 skipping (`Regenerate bench baseline`, expected on a non-bench diff). The one red `tool-count grep gate` (run `30552146645`, 2s) is a **concurrency-coalescing artifact, not a failure**: the `push` and `pull_request` events fired 2s apart and share a concurrency group key, so `cancel-in-progress: true` cancelled the push run — `gh pr checks` renders `cancelled` as `fail`. Its `pull_request` twin (run `30552149627`) **passed in 11s**. `c8-precheck.yml`'s header documents this coalescing as intended.

**One deviation from the review instruction, deliberate:** landed as a follow-up commit rather than `--amend`. The branch head was already pushed, so amending would require a force-push, which this lane's hard constraints forbid unconditionally and which CLAUDE.md §"Pushing" permits only on explicit operator authorization. Same end state, auditable history.

Head SHA: `4d993be7ba9070088b3c0fb408af372411bf011a`

Issue #2602, opened 2026-07-31
[determinism] list / memory_load_family have no tiebreak past (priority DESC, updated_at DESC), so the row chosen at rank k among ties is plan-dependent


alphaonedev:
## Summary

Both list-shaped orderings — `crate::storage::build_list_query`'s `SQL_LIST_ORDER_LIMIT` and `PostgresStore::list`'s `ORDER BY priority DESC, updated_at DESC` — stop at two keys, and `crate::mcp::handle_load_family` uses the same pair. Neither adds a deterministic final tiebreak (e.g. `, id ASC`).

Rows sharing `(priority, updated_at)` are therefore ordered by whatever the planner produces. With `LIMIT k` applied to that order, **which** tied row lands at rank `k` — i.e. which rows are in the page and which fall off — is not specified and can differ run to run, backend to backend, and after an index or plan change.

Ties are not exotic here: `priority` is a small integer (1-10) and bulk/imported/federated corpora routinely land many rows in the same `updated_at` millisecond.

## Why now

Pre-#2580 the postgres `memory_load_family` truncated at rank `k` only *after* a 1000-row fetch, so the tie boundary sat at rank 1000 and was rarely observed. #2580 pushes the predicate into SQL with `LIMIT k`, which moves the boundary to rank `k` — the same class of nondeterminism, but now on the hot boundary. (This is not a regression: the pre-fix code was equally unspecified at rank `k` after its in-Rust `sort_by`, which is *stable* over a plan-dependent input and therefore preserves arbitrariness rather than removing it.)

## Proposed fix

Append `, id ASC` as the final ORDER BY key on both adapters and in `handle_load_family`, in ONE change so the backends cannot drift.

Caveats to handle deliberately, not silently:
- This **changes bytes** for tied rows — it replaces "unspecified" with "fixed", which is the correct direction but is a wire-visible change on an always-on tool and belongs in the CHANGELOG.
- On sqlite it may cost the `idx_memories_ns_list_order` early-stop unless the index is extended; measure with `EXPLAIN QUERY PLAN` before/after (the #1579 A2 pins live in `tests/issue_1579_storage_perf.rs`).
- Needs a ties-seeded regression asserting the same page across both backends and across repeated runs.

Discovered by the #2580 perf work (5-agent adversarial vote `4d3ea1c5`; raised by the semantics-preservation lens).

alphaonedev:
[...]

Issue #2601, opened 2026-07-31
[correctness] sqlite memory_load_family applies scope=private visibility AFTER the SQL LIMIT, so it silently under-returns; postgres applies it before


alphaonedev:
## Summary

`crate::mcp::handle_load_family` (`src/mcp/tools/load_family.rs`) runs `LIMIT ?4` (= the resolved `k`) **in SQL**, then applies `crate::visibility::is_visible_to_caller` as a **Rust post-filter** over the returned page (#1555). Any row in the top-`k` that the caller may not see is dropped *after* the window closed, so the response carries **fewer than `k` rows even when more visible matches exist**.

`PostgresStore::list` puts the scope=private predicate **in SQL, before `LIMIT`** (the `$6` clause), so the postgres branch of the same handler returns a full `k` window for the same data.

## Impact

Same inputs, same corpus, different `count` per backend on an **always-on core-profile tool** (`memory_load_family`, also reached via `memory_smart_load`). It is a DEGRADE (fewer results), never wrong results — but it is silent, and it is a cross-backend split.

Worst case is a multi-tenant namespace where another tenant's private rows outrank the caller's: the caller can be served an empty page while their own family-tagged rows sit at rank `k+1`.

## Note on the postgres side

Postgres is not fully immune either, for a different reason: the SQL `$6` clause covers only the private arm, while the Rust `is_visible_to_caller` re-check is **strictly narrower** — #1921 added subtree gating for `team`/`unit`/`org` scope that has no SQL twin. So a `team`-scoped row outside the caller's subtree passes SQL, consumes a `LIMIT` slot, and is then dropped in Rust.

#2580 handles this for `memory_load_family` specifically by re-asking once at the historical `MAX_BULK_SIZE` window whenever the narrowed set is short of `k`, so that path never under-returns relative to pre-#2580. **Every other `MemoryStore::list` caller still has the gap**, and the sqlite `handle_load_family` path is not covered at all.

## Proposed fix

Two candidates, worth a vote:

1. Push the FULL visibility predicate (including the #1921 team/unit/org subtree gate) into SQL on both backends, so `LIMIT` and the Rust re-check agree and the re-check becomes a true no-op belt-and-suspenders. Largest correctness win; the subtree gate needs a namespace-prefix predicate.
2. Generalise the #2580 escalation (fetch `k`, re-ask at a bounded larger window when the narrowed set is short) into the SAL so every `list` caller inherits it.

[...]

Issue #2600, opened 2026-07-31
[data-integrity] memory_load_family / memory_smart_load bypass the #1948 fail-closed lifecycle allow-list on sqlite — quarantined + tombstoned rows are readable through an always-on core-profile tool


alphaonedev:
## Summary

`crate::mcp::handle_load_family` (`src/mcp/tools/load_family.rs:321-333`) — the SQL behind **both** the MCP `memory_load_family` tool **and** the sqlite branch of `POST /api/v1/memory_load_family` (and, via routing, `memory_smart_load`) — carries **no `lifecycle_visible_clause`**:

```sql
SELECT id, tier, namespace, ... FROM memories
 WHERE (?1 IS NULL OR namespace = ?1)
   AND json_extract(metadata, '$.family') = ?2
   AND (expires_at IS NULL OR expires_at > ?3)
 ORDER BY priority DESC, updated_at DESC
 LIMIT ?4
```

`grep -n lifecycle src/mcp/tools/load_family.rs` returns only three hits, all of them the `core/lifecycle/graph/...` family-name list in doc comments. There is no gate.

## Why this matters

v1.0.0 R19/A3 (#1948) introduced `crate::models::lifecycle_visible_clause` as a **fail-CLOSED allow-list** (`lifecycle_state IN ('open','active','blocked','done','abandoned')`, NULL = visible-legacy) and applied it across every read/egress lane precisely so that:

- `lifecycle_state = 'quarantined'` — the route-IN disposition for provenance-less inbound federated memories under `AI_MEMORY_FED_QUARANTINE_UNATTRIBUTED` (env #123) — is **structurally hidden from this node's local view**; and
- `lifecycle_state = 'tombstoned'` — the #1859/#1875 consolidate/append-only disposition — does not resurface as live content.

`memory_load_family` is an **always-on core-profile tool** (one of the 7 in `--profile core`), so this is not an exotic path. An operator who enabled quarantine expecting unattributed federated content to be invisible can still be served it by name through the loader, on the default backend.

## Cross-backend divergence

The postgres branch of the same handler routes through `MemoryStore::list`, whose SQL **does** carry `lifecycle_visible_clause("")` (`src/store/postgres.rs`, the `list` builder). So today:

| backend | `memory_load_family` honours the #1948 allow-list? |
|---|---|
| postgres | yes (inherited from `list`) |
| sqlite | **no** |

[...]

Issue #2578, opened 2026-07-31
perf(postgres): v56 composite list indexes never landed on pg — list/session_start scan the whole namespace (8.5 ms -> 0.167 ms, 51x)


alphaonedev:
## Summary

The v56 composite list-ordering indexes (#1579 A2 / B6d) **were never created on postgres**. CLAUDE.md states this explicitly: *"`migrate_v56()` is a postgres version-stamp no-op"*, with the indexes shipped only in `migrations/sqlite/0047_v56_list_composite_indexes.sql`.

Consequence: on postgres, a `LIMIT 10` namespace-scoped list **reads every row in the namespace and sorts them**. It is O(namespace size) where sqlite is O(limit).

## Evidence — the plan (live pg tier, 8,004 rows, atlas-corpus = 7,855)

`EXPLAIN (ANALYZE, BUFFERS)` of the exact statement the daemon issues (captured from `log_min_duration_statement=0`):

```
Limit (actual time=11.019..11.023 rows=10 loops=1)
  Buffers: shared hit=2136
  ->  Incremental Sort (actual time=11.017..11.018 rows=10 loops=1)
        Sort Key: priority DESC, updated_at DESC
        ->  Index Scan using memories_priority_idx on memories
              (actual time=0.111..8.795 rows=7855 loops=1)
              Filter: (namespace = 'atlas-corpus' ...)
              Rows Removed by Filter: 149
Execution Time: 11.097 ms
```

`namespace` is a **Filter, not an Index Cond** — 7,855 rows read to return 10.

Existing pg indexes on `memories` include `memories_priority_idx` and `memories_namespace_idx` but **no** `(namespace, priority DESC, updated_at DESC)` composite.

## Evidence — counterfactual (isolated copy; the live table was never modified)

```sql
CREATE TABLE perf_scratch AS SELECT * FROM memories;   -- 8,004 rows
-- production index set (priority only):
Index Scan using pla_priority ... rows=7855 ... Execution Time: 8.474 ms
-- with the v56 composite index sqlite already has:
CREATE INDEX pla_ns_list ON perf_scratch(namespace, priority DESC, updated_at DESC);
Index Scan using pla_ns_list ... rows=10  ... Execution Time: 0.167 ms
```

**8.474 ms → 0.167 ms = 51x**, rows read 7,855 → 10, and the sort node disappears entirely.

## Evidence — end-to-end, both backends (HTTP, keep-alive, n=300)

[...]

PR #2603, opened 2026-07-31
perf(#2584,#2580): stop re-reading the whole forensic log at every process start; push memory_load_family's family predicate into SQL

### LINKED ISSUE #2580: perf(postgres): memory_load_family fetches 1000 full rows (~982 kB) to return 0 — metadata.family filter not pushed to SQL despite a GIN index
## Summary

`POST /api/v1/memory_load_family` on postgres fetches **1,000 full memory rows** from the database — regardless of the requested `k` (default 20) — then filters them **in Rust** on `metadata.family`. On the 8k-row corpus it costs **54.3 ms p50** and returns **0 rows**.

## Root cause

`src/handlers/power_consolidation.rs:938-968`:
```rust
let filter = crate::store::Filter {
    namespace: body.namespace.clone(),
    ...
    limit: MAX_BULK_SIZE,          // = 1000, regardless of `k`
    ...
};
return match app.store.list(&ctx, &filter).await {
    Ok(all) => {
        let mut filtered: Vec<Memory> = all
            .into_iter()
            .filter(|m| {
                m.metadata.get("family").and_then(serde_json::Value::as_str)
                    == Some(family_name)
            })
            .collect();
```
The in-code comment acknowledges it: *"Pull a generous superset via the SAL trait then filter on `metadata.family` in memory; the trait filter axes don't yet include metadata fields."*

## Evidence

Captured from postgres `log_min_duration_statement=0` — exactly **one** statement per request:
```
duration: 43.345 ms  execute sqlx_s_24: SELECT * FROM memories
    WHERE namespace = $1 ... ORDER BY priority DESC, updated_at DESC LIMIT $5
  Parameters: $1 = 'atlas-corpus', $5 = '1000'
```

End-to-end (HTTP, keep-alive, n=300):

| surface | postgres p50 | sqlite p50 |
|---|---|---|
| `memory_load_family` | **54.27 ms** | 12.71 ms |

CPU attribution over 400 requests: **16.45 ms of 53.99 ms (30.5%) is daemon CPU** — deserializing and discarding 1,000 rows — and 37.5 ms is waiting on the DB.

Volume: the 1,000 rows carry **~982 kB** of content+metadata per call (avg content 389 B). In this corpus **0 rows carry `metadata.family` at all**, so 100% of the transfer is discarded.

A `memories_metadata_gin` GIN index on `metadata` already exists and is unused by this path.

## Compounding
The statement is the same one from the postgres missing-v56-composite-index issue, so it also scans the whole namespace and sorts — at `LIMIT 1000` instead of 10, which is why it costs 43 ms rather than 11 ms.

## Reproduction
```bash
curl -s -XPOST http://127.0.0.1:PORT/api/v1/memory_load_family \
  -H 'Content-Type: application/json' -H 'X-Agent-Id: a' \
  -d '{"family":"core","namespace":"atlas-corpus"}'
```

## Proposed fix
1. Push the family predicate into SQL — `metadata @> jsonb_build_object('family', $x)` (GIN-indexable) or `metadata->>'family' = $x` — by adding a metadata-equality axis to the SAL `Filter`.
2. Bound the fetch by the **resolved `k`** (clamped 1..100), not `MAX_BULK_SIZE`.
3. Push `ORDER BY priority DESC, updated_at DESC LIMIT k` down as well, removing the in-Rust sort.

Classification: **algorithmic** — unbounded materialization + filter/sort in Rust that the DB can do.
Backends affected: **postgres** primarily (sqlite path is 12.7 ms and should be checked for the same shape).

Filed by the 3x3 adversarial performance audit, LENS A (read path).

### LINKED ISSUE #2584: perf(cli): ~130 ms fixed overhead per CLI invocation vs a 4.45 ms process floor (79% of syscall time in futex); root cause needs a profiler
## Summary

Every `ai-memory` CLI invocation that opens the database pays a **fixed ~130 ms of process overhead** before doing any work, against a **4.45 ms** bare-process floor. The actual query is sub-millisecond. For agent fleets that shell out to the CLI, ~97% of each invocation is overhead.

## Evidence (n=40 per row, 3 warm-ups, release build)

| invocation | p50 | p95 |
|---|---|---|
| `--version` (no DB open) | **4.45 ms** | 4.82 |
| `list --namespace atlas-corpus --limit 10` | **132.45 ms** | 135.07 |
| `search 'body bravo' --namespace … --limit 10` | 133.24 ms | 135.69 |
| `stats` | 141.57 ms | 144.52 |
| `recall 'body bravo' --namespace … --limit 10` | 154.75 ms | 176.73 |

The same `list` over HTTP against the same corpus is **0.62 ms**. So ~131 ms is invocation overhead, not query cost.

## What it is NOT (ruled out by measurement)

| hypothesis | test | result |
|---|---|---|
| corpus-dependent | empty DB | `list` = **132.98 ms** (unchanged) |
| boot inventory / index warm-up | `AI_MEMORY_BOOT_ENABLED=0` | 133.33 ms (unchanged) |
| both | empty DB + `BOOT_ENABLED=0` | 131.86 ms (unchanged) |
| fsync-bound | `strace -e fsync,fdatasync` | **0 fsyncs** |

## What the syscall profile shows

`strace -c -f ai-memory list --limit 10` on an empty DB — 2,299 syscalls, 61.9 ms total syscall time:
```
% time     seconds  usecs/call     calls    syscall
 78.89    0.066857        1172        57    futex
  8.33    0.007063        7063         1    epoll_wait
  1.76    0.001491          29        51    mprotect
  1.53    0.001297          19        65    mmap
  1.10    0.000935          62        15    clone3
  0.97    0.000820           0      1508    read
```
**79% of syscall time is `futex`** (57 calls, ~1.17 ms each) with **15 threads spawned** — i.e. thread-synchronisation / runtime-startup bound.

## Honest limitation

I could **not** attribute the 130 ms to a specific subsystem. `perf` is not installed on this host and a profiler build was out of scope for an audit-only pass. Candidates consistent with the evidence — bootstrap DDL replay on every open (CLAUDE.md: *"Both adapters replay their bootstrap on EVERY open, ALWAYS before `migrate`"*), tokio/rayon runtime construction, Ed25519 key load — are **hypotheses, not findings**. This issue asks for that attribution to be done with a profiler.

## Reproduction
```bash
AI_MEMORY_NO_CONFIG=1 AI_MEMORY_DB=/path/corpus.db \
  hyperfine --warmup 3 'ai-memory list --namespace atlas-corpus --limit 10' 'ai-memory --version'
```

## Proposed next step
Profile a `cargo build --release` with debug symbols under `perf record` / `cargo flamegraph` for a single `ai-memory list` on an empty DB, and attribute the futex time. Then decide whether the fix is lazy runtime construction, skipping bootstrap DDL replay when `schema_version` already matches, or deferring thread-pool creation until a consumer exists.

Classification: **constant** (fixed per-invocation), but dominant for CLI-driven workloads.
Backends affected: **sqlite / CLI surface** (CLI cannot target postgres — `--store-url` is wired on `serve`/`curator` only).

Filed by the 3x3 adversarial performance audit, LENS A (read path).

alphaonedev:
Closes #2584
Closes #2580

Two independent per-invocation costs on always-on paths. Both were **reproduced by execution** against `release/v1.0.0` @ `93a30bd5` before any code was written, and in both cases the issue's own diagnosis turned out to be partly wrong — the corrections are the substance of this PR.

---

### Reproduce-or-refute: **REPRODUCED, root cause DIFFERENT from the hypothesis**

`src/main.rs::init_forensic_audit` → `governance::audit::init` → `read_chain_tail` streamed the **entire** newest `forensic-<date>.jsonl` through `serde_json::from_str::<ForensicDecision>` to keep only the last row's hash. That runs on **every process start**: every CLI invocation, every `ai-memory mcp` stdio boot, every `serve`/`curator` boot, and **every firing of the `governance check-action` PreToolUse hook**.

**Attribution** (`strace -f -tt -T`, installed v1.0.0 release binary, 16,556,309-byte same-day forensic log):

- **2,021 of the process's 2,052 `read` syscalls** were 8 KiB reads of that one file (2021 x 8192 = 16,556,032 bytes = the whole file).
- A **254 ms contiguous window of the 305 ms straced run**, bracketed *exactly* by `openat(".../forensic-2026-07-31.jsonl", O_RDONLY)` at t=20.42 ms and the `statx(".../recorder-keys")` that follows it inside `init` at t=274.58 ms.
- No gap >3 ms anywhere on the main thread — the cost is diffuse parse work, not one blocking call.

**Controlled A/B**, `AI_MEMORY_AUDIT_DIR` pointed at an empty directory (load-independent syscall counts + wall clock, n=25, `ai-memory list --limit 10` on an **empty** database):

| | p50 | p95 | p99 | `read` syscalls |
|---|---|---|---|---|
| real audit dir (16.5 MB log) | **173.97 ms** | 176.44 | 177.66 | **2,052** |
| empty audit dir | **10.84 ms** | 11.83 | 12.34 | **28** |
| `--version` floor | 4.41 ms | 4.92 | 4.97 | — |

**~94% of the reported per-invocation overhead was this one scan.**

### Two corrections to the issue

1. **It was never a constant.** The cost is O(TODAY'S LOG SIZE) — it grows all day and resets at UTC midnight. #2584 measured 132 ms; hours later on the same host it measured 174 ms. Any fixed-ms framing is wrong.
2. **The `futex` evidence was an artefact.** The "79% of syscall time in futex, 15 threads" profile is `strace -c` billing the **parked** time of the tokio multi-threaded runtime's workers (14 `clone3` = `available_parallelism()` on this 14-core host). The timeline shows the runtime is not even constructed until **after** the scan finishes (first `clone3` at t=274.88 ms). Thread synchronisation was not the cost. Lens A was right to refuse to guess.

### The fix cannot change an answer

`read_chain_tail` now reads a bounded **256 KiB suffix** backwards to the last parseable row. Because every line after the last parseable one is *by definition* unparseable, a backward walk from EOF meets those same lines first and returns the very same row. When the window holds no parseable row (a tail damaged beyond 256 KiB, e.g. a SIGKILL mid-append) it **falls back to the original full forward scan, retained verbatim** — so the value is identical for **every** input and the worst case is the pre-fix behaviour, never a guess. Blank-line skipping, `\r\n` trimming and invalid-UTF-8 tolerance mirror `BufReader::lines()` line for line.

**Regression coverage** (`src/governance/audit.rs`, R-203 fail-at-parent):
- an **equivalence matrix over 10 tail shapes** (empty file, no trailing newline, blank lines, CRLF, truncated last row, trailing garbage, all-garbage, invalid-UTF-8 tail, ...) asserting bounded == full scan on each;
- a guard asserting the **bounded path alone** resolves the tail of a file three windows long — reverting to a full scan reds it;
- a fail-safe asserting an all-garbage window **declines** (`None`) while the exact fallback still recovers the true tail.

---

### Reproduce-or-refute: **REPRODUCED, plus an unreported correctness defect underneath it**

`memory_load_family` is an always-on **core-profile** tool (also reached via `memory_smart_load`), so every default Postgres deployment paid this on every call: `MAX_BULK_SIZE` (=1000) FULL rows regardless of the requested `k`, ~982 kB, to return **zero** rows.

`EXPLAIN (ANALYZE, BUFFERS)` on the live 8k-row Atlas corpus:

| shape | exec | buffers | rows |
|---|---|---|---|
| pre-fix `LIMIT 1000`, filter in Rust | 3.807 ms | 321 | **1000** (~982 kB, all discarded) |
| `metadata @> jsonb_build_object(...)` (**chosen**) | **0.286 ms** | **11** | **0** |
| `metadata->>'family' = $v` (**rejected**) | 17.759 ms | 2,755 | 0, `Rows Removed by Filter: 10724` |

The containment shape is served by the pre-existing `memories_metadata_gin` index via a Bitmap Index Scan; the `->>` form is not indexable and degenerates to a full namespace scan. *(Note the pre-fix figure is 3.8 ms rather than the issue's 43 ms because a sibling lane's #2578 composite index has since landed on this shared instance — the row/byte volume, which is the defect, is unchanged.)*

### Design — 5-agent adversarial vote `4d3ea1c5`

**Form tally: 3-1 for a new `MemoryStore::load_family` trait method; SYNTHESIS overrode it** in favour of a `Filter::metadata_eq` axis. Recorded honestly because it is a tally override:

- The three C-voters all argued from the same premise — a `Filter` field an adapter silently ignores returns a **superset** (wrong results) — and all three *also* independently required the in-Rust re-check that removes exactly that failure mode.
- The fourth lens measured what the other three did not consider: a bespoke `load_family` query must **re-derive five independent hardening predicates** (#1948 fail-closed lifecycle allow-list, the expiry guard, the scope=private `$6` clause, the #1834 AS-OF window, the #2383 undecryptable-row skip). One omission returns rows the caller must never see. Riding the shared `list` builder inherits all of them, and every future one, for free.

Under the North Star (never wrong results), a concrete drift risk outweighs a hypothetical fail-open that the unanimous cross-cutting mitigation already closes.

### What landed

- **postgres**: `metadata @> jsonb_build_object($9::text, $10::text)`, emitted only when the axis is set so every pre-#2580 `list` shape keeps its byte-identical SQL and cached plan.
- **sqlite**: the twin `EXISTS (SELECT 1 FROM json_each(memories.metadata) WHERE key = ? AND value = ? AND type = 'text')`, mirroring the sibling `tags_any` fragment.
- Both bind the **key as a query parameter**, never interpolated into a JSON path — pinned by a test that matches a key containing `"`, `$` and `[0]` literally.
- `db::list` keeps its historical 11-arg shape (a new `db::list_filtered` carries the axis), so the **~30 existing call sites are untouched**.

### Proof that semantics are unchanged

- **12-shape type-exactness matrix**, asserted **identical on sqlite and on live Postgres**: `"core"` matches; `"graph"` / `"corex"` / `"Core"` / `""` / `123` / `null` / `true` / `["core"]` / `{"x":"core"}` / nested `{"a":{"family":"core"}}` / absent all do not. `json_each.type='text'` and jsonb containment both restrict to JSON **string** values, so the SQL is byte-exact with the Rust predicate it replaces (`metadata.get(k).and_then(Value::as_str) == Some(v)`). jsonb `@>`'s documented array/nesting special cases are pinned **against the live planner**, not against the docs.
- **Fail-closed**: the handler RETAINS the in-process re-check (`MetadataEq::matches`), mirroring `PostgresStore::list`'s `is_visible_to_caller` re-apply. The pushdown is strictly **narrowing** — an adapter that ignored the axis could only degrade, never widen.
- **Unset axis is inert** on both backends (asserted).

### A correctness defect fixed, not just a slow path

Applying `LIMIT 1000` **before** the family filter silently dropped family rows below global rank 1000 — Postgres returned **fewer** rows than SQLite for identical data (SQLite's `handle_load_family` always pushed the predicate down). A live-Postgres regression seeds 1050 untagged high-priority rows + 3 low-priority family rows and asserts the legacy pipeline finds **zero** while the pushdown finds all three. The change is a strict **superset**: restricting the ordering to family rows can only move them **up** in rank, so every row the old code returned is still returned, in the same order.

### `k` is bounded without risking an under-return

The SAL `list` applies the strictly-narrower Rust `is_visible_to_caller` (#1921's team/unit/org subtree gate has **no** SQL twin in the `$6` clause) **after** the SQL `LIMIT`, so a bare `LIMIT k` could return fewer rows than the old 1000-row window. The handler therefore asks for `k`, and re-asks **once** at the historical `MAX_BULK_SIZE` window when the narrowed set is short — at most two round trips, the second carrying only family-tagged rows.

---

## Cross-backend

`Filter::metadata_eq` is implemented in **both** adapters in this PR, in each one's existing hardened `list` query, so there is no backend on which the axis is silently ignored.

## Gates

`cargo fmt --all`; `cargo check --all-targets` and `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` on **all four legs** (default / `sal` / `sal-postgres` / `vectorlite`) — clean; `qual_10` (ceiling `src/storage/mod.rs` 27_700 -> 27_820, dated rationale in the table) + `qual_6_7`; `docs-vs-ssot`, `hardcoded-literal`, `vendor-literal`, `l3-boundary`, `migration-ladder` gates all PASS. No MCP tool / HTTP route / CLI verb / schema version added, so no SSOT pin moves.

Live-Postgres tests run green against PG 18.4 + AGE 1.7.0 + pgvector 0.8.6 (3 passed).

## Residuals filed

- **#2600** [data-integrity] `memory_load_family` / `memory_smart_load` bypass the #1948 fail-closed lifecycle allow-list **on sqlite** — quarantined + tombstoned rows are readable through an always-on core-profile tool. Pre-existing and unchanged by this PR (the postgres branch already inherited the clause from `list`); surfaced while proving the pushdown preserves `list`'s hardening.
- **#2601** [correctness] sqlite `handle_load_family` applies scope=private visibility **after** the SQL `LIMIT` and silently under-returns; postgres applies it before. Includes the #1921-subtree residual that affects every other `list` caller.
- **#2602** [determinism] no tiebreak past `(priority DESC, updated_at DESC)`, so the row at rank `k` among ties is plan-dependent on both backends.

## AI involvement

Authored by Claude Opus 5 under operator direction. Design resolved by a 5-agent adversarial vote (`4d3ea1c5`); the tally override is documented above. All measurements are first-hand against `release/v1.0.0` @ `93a30bd5` and the live PG 18.4 instance.


alphaonedev:
## #2584 — measured before/after on the real corpus

Both binaries are **`--release`** builds of the same profile. `BEFORE` = the installed `/usr/local/bin/ai-memory` v1.0.0 (pre-fix); `AFTER` = this branch. Same host, interleaved, same empty database (`AI_MEMORY_DB` -> an empty file, so the corpus is not a variable), same real audit directory holding a **16,856,956-byte** same-day `forensic-2026-07-31.jsonl`.

### Load-independent evidence (syscall counts — build-profile and load independent)

`strace -c -f … list --limit 10`, against the real forensic log:

| | `read` calls | `futex` calls |
|---|---|---|
| BEFORE | **2,087** | 57 |
| AFTER | **30** | 57 |

The `futex` count is **identical** across the fix — which is the direct refutation of the issue's thread-synchronisation hypothesis. What changed is the ~2,057 8 KiB reads of the forensic log; `futex` "seconds" in `strace -c` is the tokio workers' *parked* time, not work.

### Wall clock (n=40, 3 warm-ups, `load average: 11.52` on 14 cores — the box is running six lanes, so p95/p99 carry scheduler noise; `min` is the cleanest signal and the effect size dwarfs the noise either way)

| invocation | | min | p50 | p95 | p99 |
|---|---|---|---|---|---|
| `list --namespace … --limit 10` | BEFORE | 165.06 | **184.87** | 334.52 | 365.08 |
| | AFTER | 12.12 | **23.14** | 26.78 | 28.45 |
| `stats` | BEFORE | 177.86 | **343.04** | 529.45 | 544.04 |
| | AFTER | 16.29 | **23.59** | 32.96 | 40.31 |
| `--version` (floor, no DB open) | AFTER | 4.39 | 7.72 | 10.82 | 72.48 |

**`list` p50 184.87 -> 23.14 ms (8.0x). `stats` p50 343.04 -> 23.59 ms (14.5x).**

### The number was never fixed

The same `list` measured **132 ms** in the original report, **173.97 ms** earlier in this session, and **184.87 ms** here — because the forensic log grew from 16,556,309 to 16,856,956 bytes between those two measurements. The cost is O(today's log size), it climbs all day, and it resets at UTC midnight. After the fix it is O(1) in log size: a bounded 256 KiB suffix read.

### Where the residual ~19 ms goes

`AFTER list` p50 23.14 ms against a `--version` floor of 7.72 ms leaves ~15 ms of genuine boot: the tokio multi-threaded runtime construction (14 `clone3` = `available_parallelism()`), the sqlite bootstrap DDL replay + migration check, and the query. That is the real per-invocation floor, and it is now the honest one — #2584's "4.45 ms floor vs 130 ms" framing compared `--version` (which exits inside `Cli::parse()`, before the runtime, the DB and the audit sink exist) against a full boot, so ~94% of the gap it attributed to boot was this single scan.

alphaonedev:
## #2584 — chain-tail equivalence proved against the REAL production log

The unit tests prove equivalence on synthetic tails. This is the same claim tested end-to-end on the actual 16.9 MB forensic log this host has been writing all day, through the real `governance check-action` CLI path (the one the PreToolUse hook invokes).

Method — the live file is never touched; each binary gets its own byte-identical **copy** via `AI_MEMORY_AUDIT_DIR`:

```
seeded copies: 39,660 lines / 16,936,801 bytes each  (tailA, tailB)

  BEFORE  /usr/local/bin/ai-memory (v1.0.0, pre-fix, full forward scan)
            governance check-action --kind bash --command "echo tailcheck"  -> rc=0, appends into tailA
  AFTER   this branch (bounded 256 KiB suffix read)
            governance check-action --kind bash --command "echo tailcheck"  -> rc=0, appends into tailB
```

Result — the `prev_hash` each binary computed from its own reading of that 39,660-line tail:

| | lines after append | `prev_hash` written |
|---|---|---|
| BEFORE (full scan) | 39,661 | `2b6f1c0956ae3e305e921bd8a2088e0fed66e4256f4696cc78f83008a778b3b1` |
| AFTER (bounded) | 39,661 | `2b6f1c0956ae3e305e921bd8a2088e0fed66e4256f4696cc78f83008a778b3b1` |

**Identical.** The bounded read resolves exactly the row the full scan resolved, on real production data, so the hash chain links continuously across the change — an upgraded daemon's first appended row chains onto the pre-upgrade tail with no break, and `verify_since` walks straight through.

alphaonedev:
## #2580 — measured on the EXACT production SQL, and a correction to my own earlier caveat

My first `EXPLAIN` used an ad-hoc query with literal values, which postgres plans as a **custom** plan. `sqlx` issues **prepared statements**, and postgres settles those onto a **generic** plan after a few executions — a different plan, and the one that actually runs in the daemon. Re-measured properly: both SQL shapes reconstructed byte-for-byte from `PostgresStore::list` (namespace sargable arm, tier, expiry, since/until, the `$6` scope=private clause, `$7` agent_id, the #1834 `$8` AS-OF window, the #1948 lifecycle allow-list, `ORDER BY priority DESC, updated_at DESC LIMIT $5`), run via `PREPARE`/`EXECUTE` under `plan_cache_mode = force_generic_plan`, warm cache, live 7,855-row `atlas-corpus`.

### BEFORE — pre-#2580 shape (`LIMIT 1000`, no metadata predicate)

```
Limit (actual time=33.453..33.602 rows=1000.00 loops=1)
  -> Sort (actual time=33.451..33.513 rows=1000.00 loops=1)
       Sort Key: priority DESC, updated_at DESC
       -> Index Scan using memories_namespace_idx on memories
            (actual time=0.032..9.131 rows=7855.00 loops=1)
Execution Time: 33.695 ms
```

### AFTER — #2580 shape (GIN containment pushdown, `LIMIT k=20`)

```
Limit (actual time=0.680..0.682 rows=0.00 loops=1)
  -> Sort (actual time=0.678..0.680 rows=0.00 loops=1)
       -> Bitmap Heap Scan on memories (actual time=0.671..0.672 rows=0.00 loops=1)
            Recheck Cond: ((namespace = $1) AND (metadata @> jsonb_build_object($9, $10)))
            -> BitmapAnd (actual time=0.651..0.652 rows=0.00 loops=1)
                 -> Bitmap Index Scan on idx_memories_ns_list_order   (rows=7855.00)
                 -> Bitmap Index Scan on memories_metadata_gin        (rows=0.00, 0.043 ms)
Execution Time: 0.737 ms
```

| | exec | rows to the daemon |
|---|---|---|
| BEFORE | **33.695 ms** | **1000** (~982 kB, 100% discarded) |
| AFTER | **0.737 ms** | **0** |

**45.7x on the database side, and the ~982 kB transfer + deserialise disappears entirely.** The plan confirms `memories_metadata_gin` is genuinely engaged under the generic plan, `BitmapAnd`-ed with the namespace index.

### Correction to my earlier comment

I previously wrote that the pre-fix shape now measures 3.807 ms "because a sibling lane's #2578 composite index has landed, so the issue's 43.345 ms is stale." **That was wrong, and the direction matters.** Under the generic prepared-statement plan the pre-fix shape does **not** use `idx_memories_ns_list_order` at all — it falls back to `memories_namespace_idx` plus an explicit `Sort` of all 7,855 rows, costing **33.7 ms warm**. That is squarely in line with the 43.345 ms this issue captured from `log_min_duration_statement`, which was also a real prepared statement. #2580's original measurement was accurate; my simplified reproduction was the misleading one.

Both bounds, for completeness:

| plan | BEFORE | AFTER |
|---|---|---|
| generic (what `sqlx` runs) | 33.695 ms | 0.737 ms |
| custom (ad-hoc literals) | 3.807 ms | 0.286 ms |

The pushdown wins decisively under either.

alphaonedev:
## Disclosure: my benchmark daemon wrote embeddings into the shared test corpus. Reverted; corpus restored.

Recording this in full because the affected database is a **shared measurement fixture** other perf lanes are using, and one of them (#2585) measures the embedding column payload specifically.

### What happened

My first #2580 end-to-end harness started `ai-memory serve --store-url postgres://…` with `AI_MEMORY_NO_CONFIG=1`, `AI_MEMORY_BOOT_ENABLED=0` and `AI_MEMORY_INFERENCE_EGRESS=deny`, believing that was inert. It was not:

```
INFO ai_memory::daemon_runtime: embedder loaded (all-MiniLM-L6-v2 (384-dim, local)) — tier=semantic
INFO ai_memory::store: embedding backfill: wrote 100/100 embeddings this pass (400 total)
```

Two wrong assumptions:

1. **`AI_MEMORY_NO_CONFIG=1` does not mean "no features"** — it means "ignore `config.toml`", and `serve` then falls back to the compiled-in default tier, which is **`semantic`**. `serve` deliberately takes its tier from config only (no `--tier` flag, per #703).
2. **`AI_MEMORY_INFERENCE_EGRESS=deny` only blocks REMOTE inference.** The semantic tier fell back to the bundled **local** all-MiniLM-L6-v2 model, so an embedder existed and the background backfill sweep ran. **`AI_MEMORY_BOOT_ENABLED=0` does not gate that sweep.**

### Scope

400 rows gained a 384-dim vector — 251 of them in `atlas-corpus`, the rest across small legacy test namespaces. `embedding_space` was correctly stamped (`sentence-transformers/all-minilm-l6-v2#none`), so the #2167 space gate was intact and no wrong results were possible at any point.

Pre-run state was verifiably **zero** rows with embeddings anywhere: the daemon logged exactly 400 writes and the post-run total was exactly 400.

### Remediation

Snapshotted the 400 affected ids to an audit CSV, then in one transaction cleared `embedding` / `embedding_dim` / `embedding_space` on exactly those rows:

```
rows_with_embedding_before_revert=400
rows_with_embedding_after_revert=0
atlas_corpus_rows=7855   (unchanged)
```

Per the project's North Star, embedding vectors are **derived, disposable and regenerable** from the memory TEXT — which was never touched — so clearing them is not data loss; it restores the "loaded with embeddings off" state the corpus is documented to be in and that sibling lanes measure against.

### Prevention

The harness now runs under an isolated `HOME` containing a config pinned to `tier = "keyword"`. Verified by dry run: `embedder disabled — tier=keyword`, zero backfill log lines, `rows_with_embedding=0` afterwards.

### Consequence for this PR's numbers

The BEFORE run I had already taken (p50 56.29 ms — coincidentally very close to #2580's reported 54.27 ms) was contaminated: a background embedding backfill was consuming CPU and hitting the same postgres instance throughout. **I am discarding it and re-running both arms under the keyword-tier harness.** The #2584 measurements are unaffected (sqlite CLI, no daemon), as are the SQL-level `EXPLAIN` figures (measured directly in `psql`).

### Note for other lanes

If you are benchmarking `serve` against the shared postgres instance, `AI_MEMORY_NO_CONFIG=1` + `BOOT_ENABLED=0` + `INFERENCE_EGRESS=deny` is **not** a read-only posture. Pin `tier = "keyword"` in a config under an isolated `HOME`, or your daemon will write vectors into the corpus you are measuring.

alphaonedev:
## #2580 — end-to-end A/B, and a wire-level semantics diff that found the real defect

Two `--release --features sal-postgres` daemons built from the same tree state: **BEFORE** = `release/v1.0.0` @ `93a30bd5`, **AFTER** = this branch. Same host, same live PG 18.4 instance, keyword-tier harness (verified zero writes), interleaved to bracket drift. `POST /api/v1/memory_load_family {"family":"core","namespace":"atlas-corpus"}`, HTTP keep-alive, n=300, 20 warm-ups.

| run | min | p50 | p95 | p99 | mean | resp bytes | count |
|---|---|---|---|---|---|---|---|
| BEFORE #1 | 47.26 | **53.84** | 73.73 | 78.26 | 56.08 | 75 | 0 |
| AFTER #1 | 1.88 | **2.59** | 3.82 | 4.17 | 2.73 | 75 | 0 |
| BEFORE #2 | 46.26 | **51.81** | 59.38 | 62.94 | 51.88 | 75 | 0 |
| AFTER #2 | 1.67 | **2.53** | 3.82 | 4.27 | 2.68 | 75 | 0 |

**p50 53.84 -> 2.59 ms (20.8x). p99 78.26 -> 4.17 ms (18.8x).** Both arms measured twice, interleaved; BEFORE drifted 53.84/51.81 and AFTER 2.59/2.53, so the separation is ~20x against ~4% run-to-run noise.

The BEFORE figure independently reproduces this issue's reported **54.27 ms p50** almost exactly.

## Wire-level semantics diff — 74 request shapes, both daemons

8 families x 3 namespaces x 3 `k` values, plus the `namespace`-omitted variants, against a purpose-seeded 44-row namespace containing duplicate `(priority, updated_at)` pairs, wrong-typed `family` values (`123`), a nested `{"nested":{"family":"core"}}` decoy, an EXPIRED row, a TOMBSTONED row, a `scope=collective` row owned by another agent, and a foreign-owned `scope=private` row.

```
identical shapes: 72/74   differing: 2/74

shape                     BEFORE  AFTER  relationship
fam=core  ns=<none>            5      9  AFTER ⊇ BEFORE, BEFORE is an exact ORDERED PREFIX
fam=graph ns=<none>            4      8  AFTER ⊇ BEFORE, BEFORE is an exact ORDERED PREFIX
```

**72 of 74 shapes are byte-for-byte identical** — including every namespace-scoped shape, the private/collective visibility carve-outs, the expired and tombstoned decoys, and the type-exactness decoys.

The **2 that differ are exactly the `namespace`-omitted ones**, and they differ in the direction the fix intends: the pre-fix path applied `LIMIT 1000` across the whole 8,724-row table *before* filtering on family, so the seeded low-priority rows fell below rank 1000 and were **silently dropped**. In both cases the BEFORE list is an **exact ordered prefix** of the AFTER list — every row the old code returned is still returned, in the same position, with the dropped rows appended in correct rank order. **No row is ever lost.**

This is the same defect the live-PG regression `pg_pushdown_recovers_family_rows_below_the_legacy_1000_row_window_2580` pins with a synthetic 1050-row fixture — but here it surfaced *naturally*, over HTTP, on the real corpus. That is the important part: **`memory_load_family` / `memory_smart_load` called WITHOUT a namespace has been returning incomplete results on any deployment holding more than 1000 memories** — which is essentially every real one. It is an always-on core-profile tool. That impact is not recorded in #2580 and is arguably more serious than the latency it reports.

## Corpus integrity

The instance is exactly as found: `atlas_corpus_rows=7855`, `rows_with_embedding=0`, all scratch namespaces (`l4-2580-wire`, `pushdown_ns_2580*`) removed, no daemons or listeners left behind. The one unintended write this lane made (the embedding backfill disclosed above) was reverted before these measurements were taken.

Issue #2550, opened 2026-07-31
bulk_create silently drops top-level `scope` (collective→private) and `kind_provenance` on BOTH backends — single-create honours both


alphaonedev:
## Summary

`POST /api/v1/memories/bulk` (`handlers::bulk_create`, `src/handlers/memories_query.rs:723`) re-implements the create path instead of reusing the single-create field-resolution stage, and in doing so **silently drops two caller-supplied fields that `POST /api/v1/memories` honours**:

1. **`scope`** — accepted by the DTO, never read, never validated. The row lands `scope_idx = 'private'` (the `COALESCE(metadata->>'scope','private')` generated-column default) instead of the caller's `collective`.
2. **`kind_provenance`** — the single-create path stamps `metadata.kind_provenance = "declared"`; the bulk path leaves the column `NULL`.

Symptom 1 is **security-relevant, not cosmetic**: `scope=private` is the input to the read-path ownership filter (`crate::visibility::is_visible_to_caller`). A caller who asks for a fleet-visible `collective` write gets an owner-only row, and **every other agent silently sees zero results** — a wrong-visibility outcome reported as `200 {"created": N, "errors": []}`.

Found while loading a 7,912-row corpus into the certified postgres tier. Verified **by execution on both surfaces against the identical source record**, not by reading.

## Reproduction (executed)

Daemon: `ai-memory serve` @ `release/v1.0.0` `b2a9f5bd`, built `--features sal-postgres`, store = `postgres://…:5433/ai_memory_test` (PG 18.4 + AGE 1.7.0 + pgvector 0.8.6).

Same record, two surfaces:

```
# A) POST /api/v1/memories        (single-create)
# B) POST /api/v1/memories/bulk   (bulk)
{"title":"US Stocks Rise Most of the Time", …, "scope":"collective", "kind":"concept", …}
```

Resulting rows:

| surface | `scope_idx` | `kind_provenance` | `metadata.scope` |
|---|---|---|---|
| `POST /memories` (single) | `collective` | `declared` | `"collective"` |
| `POST /memories/bulk`     | **`private`** | **`NULL`** | **absent** |

```
$ psql -c "select namespace, scope_idx, coalesce(kind_provenance,'<NULL>')
           from memories where namespace in ('atlas-pilot-single','atlas-pilot-bulk')"
 atlas-pilot-single | collective | declared
 atlas-pilot-bulk   | private    | <NULL>
```

[...]

Issue #2541, opened 2026-07-31
[SECURITY] the MCP namespace-standard bind is ungated when the caller simply omits agent_id, and the unowned-claim branch rewrites a foreign row's owner + scope


alphaonedev:
Found by the #2537 5-agent adversarial vote (`4d3ea1c5`, attacker + precedent lenses). **Not fixed there** — #2540 fixes the READ side; this is the WRITE side of the same object. Verified at `release/v1.0.0`.

## Hole 1 — the #929 ownership gate is caller-opt-in

`src/mcp/tools/namespace.rs::handle_namespace_set_standard`:

```rust
let identity_claimed = params
    .get(param_names::AGENT_ID)
    .and_then(|v| v.as_str())
    .is_some_and(|s| !s.is_empty());
if identity_claimed && let Ok(Some(existing_mem)) = db::get(conn, id) {
    // ... the #929 owner check lives HERE
}
```

The gate fires **only when the caller volunteers `agent_id`**. Omit the field and there is no gate at all: any MCP caller can bind any memory id as any namespace's standard. The in-code comment justifies this as "the caller is identifying as the daemon process itself" and leans on the HTTP entry being the load-bearing gate — but MCP stdio is a first-class tenant-facing surface, and this inverts the campaign's own rule: **key the exemption on the authenticated principal, never on a self-asserted request field the caller chooses to include.** A caller who wants to be gated includes `agent_id`; a caller who does not, does not.

## Hole 2 — the unowned-claim branch mutates a row the caller does not own

Same function, the `is_unowned` arm (and its HTTP twin in `handlers/hook_subscribers.rs`): when the bound memory's `metadata.agent_id` is empty or `"system"`, the substrate rewrites that row's `agent_id` **to the caller** and `or_insert`s `scope: "shared"`.

Two consequences, neither consented to by anyone:

1. **Ownership transfer.** A row nobody owned is now owned by whoever bound it first.
2. **Silent confidentiality downgrade.** `src/visibility.rs` maps an unknown scope string (`Some(None)`) to `true`, so a row that was default-private becomes **broadly visible on every read path** — and `scope=shared` is the documented federation shareable shape, so `/sync/since` will project it to every allowlisted peer. That is irreversible across the federation boundary: un-stamping the local row does not recall the peer copies.

The target population is exactly the legacy/system-authored rows — including curator reflections synthesized over other tenants' memories.

[...]

Issue #2445, opened 2026-07-28
[dr][docs-lie] there is NO downgrade guard — an older binary silently opens and writes a NEWER database, while production-deployment.md promises a refusal exists


alphaonedev:
**Class:** silent corruption on rollback + a false documented guarantee. **Verified at `release/v1.0.0` @ `011dd8b9`, both backends.**

## The code

Both migrate entrypoints treat "database newer than binary" as "nothing to do":

```rust
// src/storage/migrations.rs:1531-1533
if version >= CURRENT_SCHEMA_VERSION {
    return Ok(());
}
```

```rust
// src/store/postgres.rs:1644-1646
if current_version >= CURRENT_SCHEMA_VERSION {
    return Ok(());
}
```

An **older** binary opening a **newer** database silently proceeds to read and write it. Grepping `src/` for any refusal (`database is newer` / `too new` / `downgrade refus`) finds only a comment in `src/cli/boot.rs:52` and a test comment at `src/handlers/tests.rs:12126`. **No such refusal exists.**

## The documentation promises the opposite — twice

`docs/production-deployment.md:108`:

> Migrations are forward-only and run automatically on the first daemon start after an upgrade. **The substrate refuses to start against a database newer than the binary expects (downgrade refusal)** and progresses through every intermediate version on upgrade — never skips.

`docs/production-deployment.md:272`, inside the **rollback runbook itself**:

> Rollback: stop the daemon, restore the pre-upgrade snapshot, downgrade the binary. **The substrate refuses to start against a database newer than the binary expects, so a partial rollback fails loudly rather than silently corrupting data.**

The exact failure mode the doc promises protection from is the one that is unguarded — and the promise is made at the precise moment the operator is performing the rollback. An operator reading `:272` will reasonably skip the manual precaution the guard is supposed to make unnecessary.

## Why this matters at fleet scale

Rollback is the second half of **every canary deployment**. This is among the most-exercised paths in fleet operations, it has zero structural protection, and it carries a documented guarantee that actively discourages compensating care. A newer schema written by an older binary is exactly the silent-corruption class the north-star directive prohibits.

## Suggested fix

Refuse in both `migrate` entrypoints when `version > CURRENT_SCHEMA_VERSION` (strictly greater; `==` remains the no-op fast path), with the observed and expected versions in the error. Roughly 10 LOC per backend. This belongs in v1.0.0 — it is smaller than the doc change that would otherwise be required to retract the guarantee.

[...]

Issue #1921, opened 2026-07-07
[security][high] src/visibility.rs:61 — team/unit/org scope treated as world-readable — is_visible_to_caller returns true for


alphaonedev:
**Severity:** high · **CWE-863 Incorrect Authorization (scope isolation not enforced)** · **Location:** `src/visibility.rs:61` · dim: tenant-isolation

### Attack
is_visible_to_caller() short-circuits `if scope != "private" { return true }`, so scope in {team,unit,org,collective,shared,legacy} passes for ANY caller with no namespace-subtree check. This predicate is the ONLY scope gate on the paths that never invoke the SQL visibility_clause: HTTP GET /api/v1/memories (list_memories, src/handlers/memories_query.rs:148 postgres and :195 sqlite — the code comment itself notes 'db::list does NOT apply the visibility-prefix filter'), the SAL adapters get()/list() (src/store/sqlite.rs:173 and :274; postgres twins), the entity-alias fallback walk (src/handlers/kg.rs:357/383), and the kg_query per-node filter (src/handlers/kg.rs:1384). A co-tenant NHI agent (any authenticated X-Agent-Id, or anonymous when no api_key) sends `GET /api/v1/memories?namespace=<victim-org>/<unit>/<team>&limit=1000` and receives every team/unit/org-scoped Memory in the victim's subtree — full title + content + metadata — even though src/models/namespace.rs:42-49 documents Team/Unit/Org as subtree-restricted isolation levels.

### Impact
Full-content cross-tenant disclosure of every non-private (team/unit/org/collective) memory across all tenants via a standard list/get endpoint. The private-scope owner fix (#1720) is intact, but the equally-documented team/unit/org boundary provides zero confidentiality on these read paths.

### Fix
Route list()/get() (and any is_visible_to_caller call used as the sole scope gate) through the same subtree enforcement the recall/search SQL visibility_clause uses, keyed on the AUTHENTICATED principal's namespace prefixes (never a request-supplied as_agent). Concretely, change is_visible_to_caller to take the caller's (private/team/unit/org) prefixes and apply matches_subtree for Team/Unit/Org instead of returning true, mirroring storage::is_visible().

---
_Found by the v0.9.0 multi-agent security review, adversarially verified. Part of #1918._

alphaonedev:
fixed in 540465b506ad168c5e81be0b8d34ca9bbafc26f6

PR #2643, opened 2026-08-01
fix(#2538,#2633): close the named-approver self-approval hole and stop an unknown scope token publishing a row

### LINKED ISSUE #2538: [SECURITY][CWE-862] ApproverType::Agent approves on a bare string compare — no is_registered_agent, no self-approval reject, both backends
Named in #2479's body as a compounding fact and confirmed on BOTH backends during the #2479 5-agent adversarial vote (`4d3ea1c5`), which ruled 5/5-consistent that it is a SEPARATE node rather than fixable inside a federation-confinement PR. Verified at `release/v1.0.0`.

## The hole

The three `ApproverType` arms are hardened inconsistently. `Agent` — the arm an operator picks precisely when they want ONE named approver — is the weakest.

**sqlite**, `storage::approve_with_approver_type` (`src/storage/mod.rs:17933-17947`):

```rust
ApproverType::Agent(required) => {
    if approver_agent_id != required {
        return Ok(ApproveOutcome::Rejected(format!(
            "designated approver is '{required}'; got '{approver_agent_id}'")));
    }
    let ok = decide_pending_action(conn, pending_id, true, approver_agent_id)?;
    ...
}
```

**postgres**, `PostgresStore` (`src/store/postgres.rs:23488-23502`): byte-for-byte the same shape.

A bare string compare. Specifically, compared with its siblings:

| arm | registered-agent check | self-approval reject |
|---|---|---|
| `Human` | yes (`storage/mod.rs:17918`, pg `:23472`) | yes (`:17913`, pg `:23468`) |
| `Consensus` | yes (`:17960`, pg `:23506`) | n/a (distinct-voter check) |
| **`Agent`** | **NO** | **NO** |

So:

1. **No `is_registered_agent`.** The bar is "claim the string", not "the operator pre-registered this id" — exactly the bar #216 raised for `Consensus` and #1787/#1793/#1796 raised for `Human`. Any caller who can reach an approve surface as `X-Agent-Id: <required>` approves.
2. **No self-approval reject.** If `requested_by == required`, the requester approves its own pending. `SELF_APPROVAL_REFUSED` is never consulted on this arm. That is the "then self-approve" half of the #2479 attack chain, and it survives #2479 on every surface that is not the federated `namespace_meta[]` lane.

## Why it was not folded into #2479

#2479 closes the BINDING half: a scoped federated peer can no longer point a foreign namespace's standard at an approver it controls. It does not touch the approve path, and the vote was consistent that it must not:

- The arm is reached from EVERY approve surface — MCP `memory_pending_approve`, the HTTP approve/governance handlers, the CLI, and the federated `pending_decisions[]` lane (#2478) — on BOTH backends. It is not confinable to a receive lane.
- Adding `is_registered_agent` is a governance-semantics TIGHTENING: any deployment whose designated approver was never registered starts failing approvals. That needs the same treatment `Human` got — a surface split (`ApproveSurface::Http` unconditional vs `LocalOperator` opt-in, `storage/mod.rs:17906-17911`), or an env knob and a WARN cycle — plus postgres parity and its own test matrix.

## Proposed fix

Mirror the `Human` arm exactly, both backends, in one PR:

1. `src/storage/mod.rs` `ApproverType::Agent` arm — add the `approver_agent_id == pa.requested_by` reject and the `is_registered_agent` check, gated on the same `enforce_human_gate` surface predicate (`ApproveSurface::Http` unconditional; `LocalOperator` keyed on `resolve_read_visibility_caller().is_some()`, so the single-operator default cannot self-lock).
2. `src/store/postgres.rs` — the same two checks, UNCONDITIONAL, matching how the pg `Human` arm already differs (`:23466-23476` documents why: the pg SAL is reachable only through the multi-tenant HTTP daemon).
3. Tests: an `Agent`-arm self-approval cell and an unregistered-approver cell per backend, plus a control that a REGISTERED, DISTINCT designated approver still approves. Both backends, in the shape of the existing `human_arm_self_approval_gated_under_opt_in_1787` / `http_surface_rejects_self_approval_without_env_opt_in_1796` cells (`src/storage/mod.rs:25831`, `:25921`).
4. CHANGELOG under the governance lane; note the tightening and the remedy (`ai-memory agents register <id>`).

Estimated ~60 LOC production + ~200 LOC tests across two backends.

Related: #2479 (the binding half), #216 (`Consensus` hardening), #1787 / #1793 / #1796 (`Human` hardening — the precedent to copy), #1957 (signed approver quorum), #2478.

### LINKED ISSUE #2633: visibility.rs: an UNKNOWN metadata.scope token widens a row to world-readable (FBL-14 fail-open) — a typo makes a private row public
**Not covered by any existing issue.** Confirmed by direct read at HEAD `5449b6da`.

`src/visibility.rs:99`:

```rust
Some(None) => true,   // present but unknown / legacy (e.g. "shared") → broadly visible
```

## The defect

Any writer who can place an **arbitrary string** in `metadata.scope` — **including a typo** (`"privat"`, `"sharedd"`, `"Private"`) — makes the row **world-readable on every non-SQL read path**.

The mechanism is "unknown token ⇒ widest posture", which is precisely what the house FBL-14 rule forbids: *an unrecognised value must never silently widen a posture; it fails closed.* That rule is honoured consistently elsewhere — `receive_auth.rs::env_flag_default_on` and `resolve_fed_sig_flag` both keep the default on an unrecognised token, and `AI_MEMORY_INFERENCE_EGRESS` explicitly WARNs and fails closed to `deny` on a typo'd restriction intent.

## Why the existing justification does not cover it

The comment justifies the arm as the #948/#978 federation `shared` shape. That specific value is legitimate — but it is being carried by a **catch-all** rather than a closed-set variant, so the arm admits every unknown string, not just `shared`.

## Blast radius

- It is the lever **#2541's ownership-transfer hole** pulls: that path `or_insert`s `scope: "shared"` on a row it re-stamps, and `shared` is the documented `/sync/since` shareable shape — so the widening is **irreversible across the federation boundary**.
- **#2550** is the same field failing the *other* way (`collective → private` drop). One field, two opposite silent failures, no closed-set validation on either side.
- Private-by-default is the substrate's documented posture (a row with **no** scope key defaults to private). A row with a *misspelled* scope key is world-readable. Those two behaviours are one character apart.

## Proposed fix

Add a closed-set `Shared` variant, keep `None`/absent → private, and make the catch-all `_ => deny`. Do **not** leave the widening arm as the fallback. Validate `scope` on the write path against the closed set so a typo is refused at ingress rather than silently honoured at egress.

Found by the 3x7 issue audit, security/confinement lens.

alphaonedev:
Two authorization defects with the widest reach in the backlog. Both are small; both are severe; both are closed on **both backends**.

Closes #2538
Closes #2633

---

## #2538 — named-approver authorization was a bare string compare (CWE-863)

`ApproverType::Agent(required)` — the arm an operator selects *precisely to name a single approver* — read:

```rust
if approver_agent_id != required { reject } else { decide_pending_action(.., true, ..) }
```

The `Human` arm three lines above carries BOTH a `SELF_APPROVAL_REFUSED` check and `is_registered_agent` (#1787/#1796/#1793). The `Consensus` arm below carries the registered check (#216). The named-approver arm had **neither**. Since `X-Agent-Id` is a *claimed, not attested* principal, a namespace whose designated approver was also the requester had its governance gate satisfied **by the requester**, and a policy naming an agent the operator never registered was satisfiable by anyone who could type that string.

### The trap, and how it is pinned

`approve_with_approver_type` is overridden by **`SqliteStore` only**. PostgreSQL falls through to the trait default in `src/store/mod.rs`, which delegates to `MemoryStore::governance_approve_with_consensus` — a method `PostgresStore` implements with its own byte-for-byte duplicate of the same Human/Agent/Consensus state machine. **A fix landed in the sqlite override alone would never execute on postgres.**

`tests/authz_named_approver_2538_pg.rs` therefore drives *both* postgres entry points and asserts they agree. At the parent commit its failure message is literally:

```
[trait-default approve_with_approver_type] expected self-approval rejection, got Approved
```

— the trap demonstrated by execution rather than described in a comment.

### Disposition: mirror the `Human` arm, do not invent a third shape

The surface-keyed predicate is hoisted out of the Human arm into the shared `enforce_approver_identity_gate`:

| Backend / surface | Posture |
|---|---|
| sqlite, `ApproveSurface::Http` | enforced unconditionally |
| sqlite, `ApproveSurface::LocalOperator` | enforced under the `AI_MEMORY_AGENT_ID` opt-in |
| postgres | enforced unconditionally (SAL reachable only via the multi-tenant HTTP daemon) |

In the single-operator MCP/CLI trust-all default the requester and the approver both resolve to the **same** durable `host:<hostname>` id (#1720 B1), and this arm admits exactly **one** approver — so an unconditional reject-self would leave the lone operator's own queued actions unapprovable with no adversary present.

**Residual, stated rather than hidden:** under the opt-in, a policy whose designated approver *is* the requester makes that action unapprovable by anyone. That is the intended reading of a separation-of-duties control — the action stays `pending` and rejectable (`reject_pending` is not approver-gated), never wrongly approved — and policy change is not itself a `GovernedAction`, so there is no deadlock-on-deadlock.

**Liveness is pinned, not assumed:** both suites assert a registered non-requester designated approver still approves, that the pre-existing wrong-approver refusal fires **first** and is byte-unchanged (so the new checks never leak whether some other id is registered), and that the sqlite `LocalOperator` opt-OFF path is byte-identical to before.

---

## #2633 — an unknown scope token made a row world-readable (CWE-863 / FBL-14)

`src/visibility.rs` read `Some(None) => true` — "present but unknown ⇒ **widest** posture". A row with **no** scope key defaulted to owner-keyed private; a row with a **misspelled** one (`"privat"`, `"sharedd"`, `"Private"` — the parse is case-sensitive) was public. One character apart, opposite postures.

**On postgres this predicate is the SOLE scope gate** — `PostgresStore` ships no SQL `visibility_clause` and filters `list`/`search`/`get_many`/kg in Rust through this function — so the widening was unmitigated there.

The house rule is the opposite, and is honoured elsewhere: `receive_auth::env_flag_default_on` keeps the secure default on an unrecognised token; `AI_MEMORY_INFERENCE_EGRESS` WARNs and fails closed to `deny` on a typo.

### The closed set

The legitimate #948/#978 federation shareable token stays broadly visible via `visibility::LEGACY_BROAD_SCOPES` (currently exactly `"shared"`); every other unrecognised token WARNs (`tracing` target `visibility.unknown_scope`) and falls through to the **absent-key default**.

The disposition is `private_visible`, **not a bare deny** — deliberately. The narrowest posture that closes the cross-tenant hole still lets the row's own owner (and an inbox target) read it, so a typo costs the writer reach, never access to their own data. Failing closed to `false` would make a misspelled row unreachable by the one principal who can repair it. A dedicated test asserts a typo'd row is byte-indistinguishable from an absent scope key across five caller shapes — that equivalence **is** the fix.

### Blast-radius check (the requested one), and why `shared` is grandfathered rather than reified

Both alternatives were on the ballot; both lost.

- **Migration** mutates durable rows and is irreversible across the federation boundary — peers keep emitting `shared`.
- **A sixth `MemoryScope::Shared` variant** is worse than it looks: `MemoryScope::from_str` is **not** visibility-private. `governance::refusal::required_scope_refusal` funnels through it and coerces an unparseable token to `Private`, so a *parsing* `Shared` variant flips every `scope:"shared"` write from **satisfying** a `required_scope = private` policy to being **refused** by it. The rows carrying `scope:"shared"` are the `_standard:<ns>` governance-policy **placeholders** (`hook_subscribers` first-write + the #2541 ownership-restamp; `mcp::tools::namespace` set-standard restamp) — the policy **carrier** itself — so that route bricks `set_standard` in exactly the namespaces pinning a required scope. It would also make `"shared"` serde-reachable as a `CorePolicy::required_scope` value, widening a v1.0.0 public contract.
- **Denying `shared` outright** would have been a data-integrity defect strictly worse than the disclosure: on postgres the `set_standard` handler *locates* the existing placeholder through `store.list(..)` — i.e. through this very predicate. Make it invisible and the lookup misses, the handler takes its first-write branch, and the namespace acquires a **second** governance-standard row under a different owner. Governance policy split-brain.

Consequently the fix is confined to the one predicate that has the defect. **Untouched:** `MemoryScope`, `VALID_SCOPES`, `MemoryScope::COUNT`, `all()`, `all_strs()`, the "5 visibility scopes" docs narrative, `storage::is_visible`, the SQL `visibility_clause`, and every SSOT pin (`check-docs-vs-ssot.sh` still reports `scope=5`). `src/storage/mod.rs` is not touched by this half at all — no contention with the 18 open issues on that file.

**#2550 coordination:** lane A owns `handlers/memories_query.rs`; this PR does not touch it.

### The second funnel — and the reusable lesson: **a comment claiming parity is not parity**

Auditing every reachable read path (rather than fixing the one the issue named) turned up a SECOND copy of this predicate. `<PostgresStore as MemoryStore>::find_paths` filters each returned path by resolving every graph node's metadata, and carried its own inline check:

```rust
// v0.7.0 F-E3 fix (#1436): route through the canonical predicate ...
// The string compare against `MemoryScope::Private.as_str()` preserves the
// pre-refactor permissive semantics (any non-private scope is visible,
// matching `is_visible_to_caller`).
if scope != MemoryScope::Private.as_str() { true } else { owner == caller }
```

**That comment asserted parity with `is_visible_to_caller`. It was not parity, and it had not been for a long time.** The predicate had drifted **twice**:

1. **#1921 (two releases earlier)** — `team` / `unit` / `org` are all `!= "private"`, so they returned `true` here. They had been world-readable on the postgres kg path ever since #1921 restricted them to the caller's namespace subtree *everywhere else*. The comment was never updated, so the drift was invisible to every subsequent reader.
2. **#2633 (this issue)** — an unrecognised token is also `!= "private"`, so a typo'd scope published the node here exactly as it did in the canonical predicate. **A fix confined to `src/visibility.rs` would have closed the site the issue named and left this one open.**

Both are cross-tenant disclosure on a postgres-backed daemon: `find_paths` returns memory ids, and an id is the key to `GET /memories/{id}`.

**The lesson, stated plainly because it generalises well beyond this file: a comment claiming parity is not parity.** A hand-copied predicate under a comment saying "matches X" is strictly worse than an obvious duplicate — the duplicate invites scrutiny, the annotated copy deflects it, and it keeps deflecting it through every later change to X. The #951 note this module opens with already says drift between copies of this predicate is a real defect; what this case adds is that **a copy which merely LOOKS canonical is the most dangerous kind**, because it survives exactly the review that would have caught a raw duplicate.

**Independently corroborated from a different direction.** Lane D's `/sync/push` subcollection census found `action_transitions` unconfined: that lane verifies an Ed25519 signature against the actor's *enrolled* key, is replay-guarded, and even **reads** the local action's namespace to build the signable — and then never gates on it. Different subsystem, different reviewers, same deflection mechanism: there, the crypto arm makes the lane *look* protected; here, a comment claiming parity made a drifted copy *look* canonical. Two independent instances of **the thing that makes it look checked is what stopped anyone checking it** is a pattern, not a coincidence — and it argues that "this site is already handled" should be treated as a claim to verify rather than a fact to inherit, especially when the reassurance is a comment or an adjacent control rather than the gate itself.

The copy's stated justification — *"we only have the metadata blob, not a full `Memory` struct"* — was answerable in one line: select the two extra scalar columns the predicate needs (`SELECT id, namespace, metadata`) and call a new field-level `visibility::is_visible_by_fields`, which `is_visible_to_caller` now also delegates to. **There is again exactly one implementation of this rule, so the next change to it cannot drift.** The fix is strictly narrowing (fewer paths returned, never more) and is pinned against the live tier by `tests/scope_find_paths_visibility_2633_pg.rs`, including a liveness leg that passes in *both* states so a predicate that simply returned `false` could not masquerade as a fix.

### Ingress validation, scoped so federation still converges

**This is a primary hole, not defence-in-depth.** `validate_scope` has seven call sites, but every one of them guards the **top-level `scope` parameter** only; a caller who wrote the key straight into `metadata` was ungated on **HTTP create, HTTP update and MCP update**. The handler says so outright at `handlers/create.rs`: *"inline metadata.scope still works; top-level is a shortcut"*. This is established by execution rather than by argument — `validate_create_gates_inline_scope` and `validate_update_gates_inline_scope` both FAIL at parent `5449b6da`, which is what an ungated funnel accepting a typo looks like.

`validate::validate_metadata_scope` closes them on the caller-origin funnels: `validate_create` (HTTP `POST /memories` + `/bulk`), `validate_update` (HTTP `PUT`), and the MCP update path (which validates inline rather than through the DTO validator). The MCP **store** path already resolved inline `metadata.scope` through `validate_scope` and needed no change; the CLI exposes no inline-metadata surface.

Accepted set = `VALID_SCOPES` ∪ `LEGACY_BROAD_SCOPES`, because the substrate itself writes `scope:"shared"` onto the governance placeholders.

It is **deliberately not** wired into `validate_metadata` / `validate_memory`, which sit on the federation-receive, import, reflect and consolidate funnels: a refusal there would **diverge replicas** (the #1821 secret-screen lesson — caller-origin refuses, federation/recovery degrades). An inbound peer row with a typo'd scope is stored verbatim and simply reads as private locally; a regression test pins that.

### Fixtures corrected, not masked — and why `public` is NOT enumerated as a legacy alias

**Five** in-repo fixtures stamped `scope: "public"` — a token that is not a `MemoryScope` and only ever passed the visibility gate *because* unrecognised meant world-readable. They now use `collective`, the real world-readable scope they always meant.

One of the five (`tests/session_start_scope_private_1420.rs`) asserts the row is visible to a **non-owner**, so the naive reading of its failure is "add `public` to the broad set alongside `shared`". That reading was considered and **rejected on evidence**:

| | `shared` | `public` |
|---|---|---|
| Written by production code? | **Yes — 5 sites** (`hook_subscribers` first-write + the #2541 restamp; `mcp::tools::namespace` restamp) onto the `_standard:<ns>` governance placeholders | **No.** Zero sites in `src/` |
| Rows in the live 8,724-row certified corpus | the governance placeholders | **Zero** (`collective` 7,861, absent 863) |
| Documented as a legacy alias? | **Yes** — #948/#978, named in the `visibility.rs` module doc | Never named anywhere |
| `validate_scope` disposition | refused, but written by the substrate itself | **refused, and written by nobody** |

The last row decides it. **`src/validate.rs:2115` is a PRE-EXISTING test asserting `validate_scope("public").is_err()`.** Enumerating `public` would leave the substrate *refusing to let an operator write the token* while *publishing to everyone any row that carries it* — the write/read incoherence this issue exists to remove, re-created one token to the left. The fixture reached the DB through `db::insert` **directly, bypassing the validator** (its own parameter is named `scope_public`), so it is evidence of a fixture authored against the bug, not of a legacy alias with real data behind it.

`personal` was checked for the same reason and is likewise asserted invalid by `test_invalid_scope`, so there is no third alias and unknown -> owner-keyed-private is right for it by design.

This coupling is now **pinned rather than remembered**: `broad_set_does_not_contradict_the_write_validator` asserts the read-path broad set never contradicts the write validator, so any future widening has to be an explicit decision instead of a side effect.

---

## Design decision

3x3 adversarial vote, 9 lenses, `4d3ea1c5` protocol (attacker/CWE-863; data-integrity & irreversibility; fleet operability; backward-compat regression; house-rule precedent; minimum-diff/blast-radius; governance semantics; testability & R-203; steelman/red-team).

- **#2633:** Option B (recognise the closed legacy set, narrow the catch-all, freeze the other predicates) — **8-1**. The dissent (lens 9, self-flagged, confidence 45) argued for converging all three predicates; its own killer objection concedes that widening a currently-closed read gate to pre-empt a hypothetical refactor trades present least-privilege for speculative safety.
- **#2538:** Option S1 (surface-keyed, mirroring the Human arm) over S2 (unconditional) — **7-2**. Adding `is_registered_agent` to the arm: **6-1 in favour**.

Two lenses independently surfaced the `governance::refusal` landmine that killed the enum route, and one surfaced the postgres `set_standard` duplicate-placeholder hazard that killed the deny route. Both changed the implementation shape from the one originally proposed.

---

## R-203 evidence (parent `5449b6da`)

Each regression test was run against the parent-commit shape of the code it guards, then against the fix.

| Test | At parent | After |
|---|---|---|
| `storage::tests::agent_arm_refuses_self_approval_and_unregistered_2538` | **FAIL** — `expected self-approval rejection, got Approved` | pass |
| `storage::tests::agent_arm_local_operator_keeps_opt_in_2538` | pass (no-regression leg) | pass |
| `named_approver_cannot_self_approve_pg` (live PG) | **FAIL** — `[trait-default approve_with_approver_type] expected self-approval rejection, got Approved` | pass |
| `unregistered_named_approver_refused_pg` (live PG) | **FAIL** — `expected unregistered rejection, got Approved` | pass |
| `registered_non_requester_named_approver_still_approves_pg` (live PG) | pass (liveness leg) | pass |
| `scope_unknown_token_2633` — 5 of 10 assertions | **FAIL** (`typo_scope_is_not_world_readable`, `typo_scope_degrades_to_owner_keyed_private_not_to_nobody`, `typo_scope_matches_absent_key_posture_exactly`, `validate_create_gates_inline_scope`, `validate_update_gates_inline_scope`) | 10/10 pass |

**Postgres tier hygiene:** run against live PG 18.4 + AGE 1.7.0 + pgvector 0.8.6 with `--include-ignored --test-threads=1`. **No schema change** — the #2445 guard is untouched and the tier is left exactly as found. All namespaces/ids are uuid-randomised, and cleanup now also reaps the agent-registry rows that live in the shared `_agents` namespace; verified `residual=0/0/0` on `memories` / `namespace_meta` / `pending_actions` after a full run.

---

## Gates

- `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` — **PASS on all four legs**: default, `--features sal`, `--features sal-postgres`, `--features vectorlite`
- `cargo check --all-targets` (default + `sal` + `sal-postgres`), `cargo check --examples`
- `qual_10_module_size_ceiling` — bumped `src/storage/mod.rs` 27_900 → 28_120 in lockstep with a dated comment stating the measured figure and what the added lines are (mostly the WHY on an authorization posture change)
- `qual_6_7_legacy_error_type_ceiling` — pass
- `check-hardcoded-literals.sh`, `check-vendor-literals.sh`, `check-docs-vs-ssot.sh` — PASS (`scope=5` unchanged, confirming zero SSOT drift)
- Regression sweeps: `mcp::` 1294, `handlers::` 532, `storage::tests` 307, `store::` 182, `validate::` 166, `l07_3_chunk_d_http_surface` 159, `sync_since_visibility_gate_948`, `federation_legacy_row_visibility_978`, `memory_scope_count_invariant`, `store_legacy_classifier_paths`, `proptest_validate`, `governance::refusal` — all green

**No new MCP tools, HTTP routes, CLI verbs, or schema versions**, so no SSOT-pin reconciliation was required beyond the module-size ceiling. CHANGELOG `[Unreleased]` updated.


alphaonedev:
## Follow-up commit `60924eeb` — a SECOND reachable funnel for #2633

Auditing every reachable read path (per the standing rule that the review gate will find a missed one, so find it first) turned up a second copy of the visibility predicate that a fix to `src/visibility.rs` alone would **not** have closed.

`<PostgresStore as MemoryStore>::find_paths` filters each returned path by resolving every graph node's metadata, and carried its own inline predicate:

```rust
if scope != MemoryScope::Private.as_str() { true } else { owner == caller }
```

…under a comment asserting parity with `visibility::is_visible_to_caller`. It had drifted **twice**, and the comment concealed both:

1. **#2633** — an unrecognised `metadata.scope` token is `!= "private"`, so a typo'd node was treated as world-readable here exactly as it was in the canonical predicate.
2. **#1921 residual** — `team` / `unit` / `org` are also `!= "private"`, so they stayed **world-readable on the postgres kg path** long after #1921 restricted them to the caller's namespace subtree on every other read path.

Both are cross-tenant disclosure on a postgres-backed daemon: `find_paths` returns memory ids, and an id is the key to `GET /memories/{id}`.

The copy's stated justification — *"we only have the metadata blob, not a full `Memory` struct"* — is answered by selecting the two extra scalar columns the predicate actually needs (`SELECT id, namespace, metadata`) and calling a new field-level `visibility::is_visible_by_fields`, which `is_visible_to_caller` now also delegates to. There is again **exactly one** implementation of this predicate — the #951 rule the module doc opens with. Strictly narrowing: fewer paths returned, never more.

### R-203 (parent `5449b6da`, live PG 18.4)

| Test | At parent | After |
|---|---|---|
| `find_paths_drops_typo_scope_node_pg` | **FAIL** — `path through a typo'd-scope node must be dropped (scope="privat"), got [["fp-a-…", "fp-m-…", "fp-b-…"]]` | pass |
| `find_paths_drops_team_scope_node_outside_subtree_pg` | **FAIL** — `path through an out-of-subtree team node must be dropped, got [[…]]` | pass |
| `find_paths_keeps_visible_nodes_pg` | pass | pass |

The third is the **liveness leg** and passes in *both* states deliberately — without it, a predicate that simply returned `false` would look like a successful fix. It covers a `collective` middle node, an owner reading their own `private` middle node, and a stranger being denied that same node.

Gates re-run after the change: fmt clean; clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` **PASS on all four legs**; `store::` 182, `visibility::` 12, `scope_unknown_token_2633` 10/10, `qual_10` green. Schema untouched (v88); scratch pg rows reaped, residual 0.

alphaonedev:
## Base merge + the `Check` regression — fixed at `1d62a02c`

Base-merged `origin/release/v1.0.0` (`e31dea74`, merge not rebase; CHANGELOG conflict resolved by keeping **both** `[Unreleased]` sections). The `Check (ubuntu/macos)` failure was a real regression in my #2633 change, and it should have been caught locally — `session_start_scope_private_1420` was not in my filter set. It is now, along with **52 other scope/visibility-touching test binaries** enumerated mechanically rather than picked by hand.

### I diverged from one instruction — flagging it rather than quietly doing either thing

The direction was to enumerate `public` alongside `shared` as a broadly-visible legacy alias, on the precedent that `shared` is already enumerated, and that this needs no vote. **I did not do that**, because verifying the premise turned it over. Please overrule me if you still want it — it is a one-line change to `LEGACY_BROAD_SCOPES` plus deleting one test — but here is what I found first:

| | `shared` | `public` |
|---|---|---|
| Written by production code? | **Yes — 5 sites** (`hook_subscribers` first-write + #2541 restamp; `mcp::tools::namespace` restamp) onto the `_standard:<ns>` governance placeholders | **No.** Zero sites in `src/` |
| Rows in the live 8,724-row corpus | (the governance placeholders) | **Zero** (your fact 3: only `collective` 7,861 + absent 863) |
| Documented as a legacy alias? | **Yes** — #948/#978, named in the `visibility.rs` doc | No — never named anywhere |
| `validate_scope` disposition | refused, but written by the substrate itself | **refused, and written by nobody** |

The decisive one is the last row. **`src/validate.rs:2115` is a PRE-EXISTING test that asserts `validate_scope("public").is_err()`.** Enumerating `public` as broadly-visible would leave the substrate *refusing to let an operator write the token* while *publishing to everyone any row that carries it* — the write/read incoherence #2633 exists to remove, re-created one token to the left.

And the row in the failing test is not evidence of a legacy alias: `seed_memory` reaches the DB through **`db::insert` directly, bypassing `validate_scope`**, and its own parameter is named `scope_public`. It is the same fixture-authored-against-the-bug shape as the **four** `scope: "public"` fixtures already corrected in the first commit; this is the fifth. That is why I read "do not fix this by making the test pass" as pointing *away* from enumerating the token.

**`personal`, as you asked:** checked, and it is not a third alias — `test_invalid_scope` asserts it invalid too, so unknown -> owner-keyed-private is the right answer for it by design, not by accident.

New pin `broad_set_does_not_contradict_the_write_validator` asserts the read-path broad set never contradicts the write validator, so if you do want `public` in, that test is where the widening becomes an explicit decision instead of a side effect.

### Correcting the scope of #2633, as requested

The premise *"the write path is already gated, so the read-path arm is defence-in-depth"* is **incomplete**, and this is now in the CHANGELOG. `validate_scope`'s call sites only ever guarded the **top-level `scope` parameter**. A caller writing `metadata.scope` **inline** was ungated on **HTTP create, HTTP update, and MCP update** — the handler comment at `handlers/create.rs` says so outright: *"inline metadata.scope still works; top-level is a shortcut"*. `validate_create_gates_inline_scope` and `validate_update_gates_inline_scope` **FAIL at parent `5449b6da`**, which is the proof those funnels accepted a typo from an ordinary caller. So #2633 was a primary hole on three caller-facing surfaces, not only defence-in-depth.

MCP **store** was already gated (it resolves `params["scope"]` OR `metadata.scope` before validating) and the CLI exposes no inline-metadata surface. Federation receive stays ungated **deliberately**: refusing there diverges replicas, the #1821 lesson.

### What I ran, and what I observed

All under `( umask 022; ... )` with `AI_MEMORY_NO_CONFIG=1`.

- `session_start_scope_private_1420` — **5/5** (was 4/5, with the exact assertion quoted)
- `scope_unknown_token_2633` — **11/11**
- **52 scope/visibility-touching binaries**, enumerated by `grep -rlE '"scope"|META_KEY_SCOPE|is_visible_to_caller|scope_public|visibility' tests/` rather than chosen by hand: `integration` 216, `recall_purity_p01` 17, `store_legacy_classifier_paths` 16, `visibility_private_leak_1720` 15, `issue_2059_2060_covenant_write_gates` 13, `serve_integration` 10, `store_parity_gaps` 10, `ns_standard_read_leak_2537` 9, `federation_pending_ns_scope_2478` 9, `expiry_canonical_ingress_2418` 9, `import_memories_admin_gate_956` 8, `federation_legacy_row_visibility_978` 8, `a2a_campaign_round1` 8, ... — **zero failures**
- **Live PG 18.4:** `cov_ga2_pg_handlers_1` 20, `scope_private_sal_level_visibility` 12, `recall_purity_p01_postgres` 11, `required_scope_postgres` 10, `postgres_owner_gate_1412` 8, `federation_write_ns_scope_2447_pg` 4, `scope_find_paths_visibility_2633_pg` 3, `authz_named_approver_2538_pg` 3 — zero failures, schema untouched (v88), scratch rows reaped
- clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` **PASS on all four legs**; `cargo fmt --all --check` clean; `cargo check --examples`
- `qual_10` / `qual_6_7` / `memory_scope_count_invariant` green; hardcoded-literal, vendor-literal, docs-vs-SSOT, migration-ladder gates **PASS**

### R-203 recap with parent SHA and observed text (parent `5449b6da`)

| Test | Observed at parent |
|---|---|
| `storage::tests::agent_arm_refuses_self_approval_and_unregistered_2538` | `expected self-approval rejection, got Approved` |
| `named_approver_cannot_self_approve_pg` | `[trait-default approve_with_approver_type] expected self-approval rejection, got Approved` |
| `unregistered_named_approver_refused_pg` | `expected unregistered rejection, got Approved` |
| `find_paths_drops_typo_scope_node_pg` | `path through a typo'd-scope node must be dropped (scope="privat"), got [[...]]` |
| `find_paths_drops_team_scope_node_outside_subtree_pg` | `path through an out-of-subtree team node must be dropped, got [[...]]` |
| `scope_unknown_token_2633` | 5 of 10 fail, incl. `validate_create_gates_inline_scope` / `validate_update_gates_inline_scope` |

Head SHA: `1d62a02c`. CI re-running now.

*(Replaces the previous comment, which was mangled by shell backtick interpolation.)*

alphaonedev:

**Control:** Capacity authz — #2538 named-approver self-approval hole + #2633 unknown `metadata.scope` world-readable fail-open.

### Code
- `ApproverType::Agent(required)` now shares the Human-arm gate: `SELF_APPROVAL_REFUSED` + registered-agent check (`storage::approve_with_approver_type`).
- Unknown/typo scope tokens degrade to owner-keyed **private** (not world-readable) via `visibility::is_visible_*` + SQL/Postgres predicate convergence (`find_paths` etc.).
- Caller-origin writes: `validate_metadata_scope` refuses invalid inline `metadata.scope` at create/update ingress.
- Federation still accepts unknown scope (degrade-not-diverge) — intentional.
- Rebased onto `f95d889e` (post Gate1+Gate2); CHANGELOG keep-both.
- Local: `agent_arm_*_2538` ok; `scope_unknown_token_2633` 11 ok; visibility 12 ok; clippy pedantic clean.

### Security (CWE-863)
- Closes self-approval on designated-approver policies (separation of duties).
- Closes misspelled scope → public/world-readable publication.
- Residual: bulk funnel #2644 still separate; capacity train continues after this merges.

### Merge bar
- Await required checks green on tip `0d523072`; merge commit only; close #2538/#2633 with evidence after merge.

alphaonedev:
# Code + security review — PR #2643 (#2538 named-approver + #2633 unknown scope)

**Control under review:**
1. CWE-863 named-approver self-approval / unregistered agent on both backends
2. Unknown `metadata.scope` token must not widen to world-readable; find_paths single predicate

**SHA:** dc8fa5dd (re-synced onto tip after cert-note)

## Code
- Agent arm gains SELF_APPROVAL_REFUSED + is_registered_agent on sqlite override AND postgres trait path
- Unknown scope → WARN + owner-keyed private (closed set); LEGACY_BROAD_SCOPES preserves "shared"
- find_paths uses shared is_visible_by_fields (no inline drifted copy)
- validate_create/update gate inline metadata.scope
- R-203 tests both backends (authz_named_approver_2538_pg, scope_unknown_token_2633, scope_find_paths_visibility_2633_pg)

## Security
| Threat | Disposition |
| Requester self-approves when named as ApproverType::Agent | refused |
| Unregistered agent string as named approver | refused |
| Typo scope publishes row | fails closed to private |
| Postgres find_paths bypass of scope narrowing | single predicate |

Issue #2595, opened 2026-07-31
perf(postgres): governance policy re-resolved from scratch on every write — 6 statements + a throwaway transaction to learn 'no policy', 22% of a single store


alphaonedev:
## Summary

Every postgres write resolves the namespace governance policy from scratch: **1 `parent_namespace` lookup + 3 `standard_id` lookups + a dedicated `BEGIN`/`ROLLBACK` transaction = 6 statements**, even when the namespace has **no standard configured at all** and the verdict is the allow-on-silence default. Measured contribution to a single store: **0.97 ms of a 4.44 ms write (21.8%)**.

## Evidence

Paired-baseline differential (same daemon generation, baseline measured immediately before and after, 250 writes each, tap-free):

```
AI_MEMORY_PERMISSIONS_MODE=off   baseline=4.409/4.463 (mean 4.436)  knob=3.467  delta=-0.969 ms (-21.8%)
```

Wire-tap statement shape on a single `POST /api/v1/memories` into `perfb/e1-kw`, a namespace with **no** `namespace_meta` row:

```
BEGIN                                                              -- enforce_governance_action opens its own tx
SELECT parent_namespace FROM namespace_meta WHERE namespace = $1   -- build_namespace_chain_in_tx
SELECT standard_id      FROM namespace_meta WHERE namespace = $1   -- chain walk: '*'
SELECT standard_id      FROM namespace_meta WHERE namespace = $1   -- chain walk: 'perfb'
SELECT standard_id      FROM namespace_meta WHERE namespace = $1   -- chain walk: 'perfb/e1-kw'
ROLLBACK                                                           -- resolved to Allow; nothing was written
```

Six round trips, one whole transaction, to learn that no policy exists. `enforce_governance_action` opens the transaction (`src/store/postgres.rs`, the H10 same-snapshot fix) *before* it knows whether it will need to write a `pending_actions` row — and in the overwhelmingly common `Allow` case it never does.

This is the same walk that #2589 shows running **per row** in bulk (5000 statements for a 1000-row batch) and #2590 shows running **per entry** in federated receive.

## Not an indexing problem

The lookups are sargable and index-backed at scale — I verified this rather than assuming it. At 100,000 `namespace_meta` rows:

```
EXPLAIN (ANALYZE, BUFFERS) SELECT standard_id FROM namespace_meta WHERE namespace = 'perfb/e1-x';
[...]

Issue #2589, opened 2026-07-31
perf(postgres): bulk create pays 7 SQL round trips PER ROW for governance+quota — 97% of bulk DB time, ceiling 943 rows/s


alphaonedev:
## Summary

`POST /api/v1/memories/bulk` batches the **INSERT** but not the **gate**. Every row in a batch pays **7 postgres round trips** for governance-policy resolution + quota check before the single multi-row INSERT runs. Measured on the wire: the per-row gate is ~**100% of the marginal cost** of a bulk ingest; the batched INSERT is ~6%.

## Evidence (PostgreSQL wire-protocol tap, steady state, tap-free timings)

Statement counts are protocol-level and load-independent. Exactly **7.0 SQL statements per row** at every batch size:

| batch | rows/s | ms/row | SQL statements per row |
|---|---|---|---|
| 10 | 763.6 | 1.29 | 7.40 |
| 50 | 871.0 | 1.14 | 7.08 |
| 100 | 837.8 | 1.19 | 7.04 |
| 250 | 916.9 | 1.08 | 7.01 |
| 500 | 891.3 | 1.11 | 7.01 |
| 1000 | **925.0** | 1.07 | **7.005** |

Linear fit over batch 1→1000: `batch_ms = 2.60 + 1.0603 * n` ⇒ steady-state ceiling **943 rows/s**.

The 7 per-row statements (verbatim from the tap):

```
BEGIN                                                        -- governance tx
SELECT parent_namespace FROM namespace_meta WHERE namespace = $1
SELECT standard_id     FROM namespace_meta WHERE namespace = $1   -- x3 (chain: '*', 'perfb', 'perfb/ns')
ROLLBACK                                                     -- governance tx resolved to Allow, nothing written
SELECT CASE WHEN date_trunc('day', day_started_at) = ... FROM agent_quotas WHERE agent_id=$1 AND namespace=$2
```

Per-statement wall time (wire tap, batch 1000, 3000 rows):

```
0.747 ms  x3.00  SELECT standard_id FROM namespace_meta
0.350 ms  x1.00  SELECT ... FROM agent_quotas            (quota check)
0.330 ms  x1.00  BEGIN
0.309 ms  x1.00  ROLLBACK
0.265 ms  x1.00  SELECT parent_namespace FROM namespace_meta
--------  -----
2.001 ms         == per-row gate ==
[...]

Issue #2449, opened 2026-07-28
[SECURITY][supply-chain] install.sh fails OPEN on a missing checksum — and release.yml never publishes .sha256 for the tarballs install.sh downloads, so the fail-open branch is the NORMAL path


alphaonedev:
**Class:** supply-chain integrity — the primary install path never verifies. **Verified at `release/v1.0.0` @ `011dd8b9`.**

## The fail-open

`install.sh:188`:

```sh
if download "$CHECKSUM_URL" "$TMPDIR/${ASSET}.sha256" "checksum" 2>/dev/null; then
    ... verify ...
fi
```

When the checksum download fails, the entire verification block is **skipped** and installation proceeds. `install.sh:186-215` prints a warning and installs the binary anyway.

## Why this is the normal path, not an edge case

`.github/workflows/release.yml` generates `.sha256` assets for exactly three things:

```
:265   sha256sum ai-memory.cdx.json > ai-memory.cdx.json.sha256          # SBOM
:397   shasum -a 256 ai-memory-ios.xcframework.tar.gz > ...sha256        # iOS
:548   shasum -a 256 ai-memory-android.tar.gz > ...sha256                # Android
```

The desktop/server tarballs that `install.sh` actually downloads get **no published `.sha256` asset**. Line `:631` does compute `SHA=$(sha256sum "ai-memory-${TARGET}.tar.gz" ...)` — but that value is substituted into the **Homebrew formula** (`:656-669`), not uploaded as a release asset.

So `$CHECKSUM_URL` 404s on every ordinary install, the `if` is false, and **no install via `install.sh` has ever verified the binary it installed.**

## Why this matters

This directly contradicts the project's own fail-closed directive, on the one path that puts a binary on an operator's machine. It is also a supply-chain gate that an enterprise review checks first — and `grep -rn 'cosign|sigstore|attest-build-provenance|slsa' .github/` returns **zero** hits, with ROADMAP §17 conceding that SLSA-style signed build-provenance is "tracked for v1.x" and reproducible builds are "NOT claimed".

## Suggested fix

1. **Emit `.sha256` for every published tarball** in `release.yml`, alongside the SBOM/iOS/Android ones that already exist.
2. **Make `install.sh` fail CLOSED**: a missing or unverifiable checksum must abort the install, not warn. Allow an explicit `--skip-checksum` for operators who genuinely need it, so bypassing is a named decision.
[...]

PR #2644, opened 2026-08-01
fix(#2550,#2551,#2552,#2588,#2594): make bulk_create reuse the create funnel and report the truth

### LINKED ISSUE #2550: bulk_create silently drops top-level `scope` (collective→private) and `kind_provenance` on BOTH backends — single-create honours both

### LINKED ISSUE #2551: bulk_create reports created=rows-SENT not rows-PERSISTED — 7912 sent / 7912 reported / 7855 landed, 0 errors (#2490 false-success class, write path)
## Summary

`POST /api/v1/memories/bulk` reports `created: <number of rows SENT>`, not the number of rows **persisted**. When a batch contains two rows sharing `(title, namespace)`, the upsert collapses them — correctly — but the response still counts both. An operator reconciling `sent == created` gets a **false green**.

Measured at corpus scale: **sent 7,912 → reported `created: 7,912` → 7,855 rows in the table.** 57 rows silently unaccounted for, across 8 batches that every one returned `200 {"errors": []}`.

This is the write-path sibling of #2490 (a zero-count export exiting 0). Same class: *a wrong number reported as success*, where the wrongness is indistinguishable from correctness without an out-of-band count.

## Reproduction (executed) — minimal

```bash
curl -s -X POST http://127.0.0.1:9177/api/v1/memories/bulk \
  -H 'Content-Type: application/json' -H 'X-Agent-Id: probe' -d '[
  {"title":"dup-title","content":"first version",       "namespace":"probe-ns", …},
  {"title":"unique-a", "content":"a",                   "namespace":"probe-ns", …},
  {"title":"dup-title","content":"second version wins", "namespace":"probe-ns", …}]'
```

```
{"created":3,"errors":[],"pending":[]}
```

```
$ psql -c "select title, content from memories where namespace='probe-ns' order by title"
 dup-title | second version wins
 unique-a  | a
(2 rows)
```

**3 reported, 2 persisted, 0 errors.**

## Reproduction — at scale (the load that surfaced it)

Corpus: 7,912 records, single namespace, 49 `(title,namespace)` groups carrying 57 duplicate rows.

```
batch 0..7  → status 200, errors 0, created 1000,1000,1000,1000,1000,1000,1000,912
SUMMARY     → sent_total 7912, created_total_reported 7912
psql        → select count(*) from memories where namespace='atlas-corpus'  →  7855
```

Breakdown of the 57 (verified): 10 collapsed **intra-batch** (deduped before the INSERT — `version` stays 1) and 47 collapsed **cross-batch** (a later batch upserted an earlier row — `version` 2 or 3, confirmed by `select version, count(*) … group by 1` → `1:7808, 2:44, 3:3`).

## Root cause

`src/store/postgres.rs::store_batch` (`:15271`) deliberately keeps the LAST occurrence of each `(title, namespace)` so the single multi-row INSERT never hits *"ON CONFLICT DO UPDATE command cannot affect row a second time"*. That part is correct and well-commented. But it then explicitly re-expands the result to input arity:

```rust
// Map (title, namespace) -> returned id so every input position
// (including the ones that collapsed into a later duplicate) can
// resolve to its persisted id in 1:1 input order.
```

so `ids.len() == memories.len()`, and the handler does:

```rust
// src/handlers/memories_query.rs:1122
match app.store.store_batch(&ctx, &allowed).await {
    Ok(ids) => ids.len(),   // <-- becomes the reported `created`
```

The 1:1 alignment is a legitimate contract for *callers who need per-input-row ids*. The bug is that the **handler reuses that length as a persisted-row count**, which it is not. The cross-batch case is not even visible to `store_batch` — an upsert of a pre-existing row is indistinguishable from an insert in the returned vector.

## Proposed fix

Make the envelope tell the truth. Minimal, additive, no wire break:

1. `store_batch` returns (or the handler derives from the existing `RETURNING`) the count of **distinct persisted ids** — `ids.iter().collect::<HashSet<_>>().len()` at the handler is a one-line lower bound that fixes the intra-batch half immediately.
2. For the cross-batch half, the `ON CONFLICT … RETURNING` arm can surface insert-vs-update per row via the standard `xmax = 0` discriminator, letting the envelope report `created` and `updated` separately.
3. Envelope becomes `{"created": N, "updated": M, "collapsed": K, "errors": [], "pending": []}`. `created + updated + collapsed + errors.len() == sent` is then a reconcilable invariant an operator (or a fleet orchestrator) can actually assert on.

Until then the only reliable reconciliation is an out-of-band `SELECT count(*)`, which defeats the point of the count in the response.

## Why it matters at ASI scale

Bulk ingest is the fleet-orchestration surface. A loader that trusts `created` — the obvious thing to trust — cannot detect partial application, and the discrepancy grows with corpus duplication rate. Here it was 0.72% of rows; nothing in the response bounded it.

Cross-reference: #2490 (false-success class, export/import side), #1481 (the batching change that introduced the single multi-row INSERT).

---
🤖 Filed autonomously per the track-and-fix prime directive. Verified by execution per pm-v3.3 step 7 (freshly-built `release/v1.0.0` binary at `b2a9f5bd`, `--features sal-postgres`, live PG 18.4).

### LINKED ISSUE #2552: bulk_create collapses every row error to `"validation failed"` — no field, no row index; single-create returns the actionable message
## Summary

When a row fails validation inside `POST /api/v1/memories/bulk`, the error returned to the caller is the bare string **`"validation failed"`** — no field name, no reason, no row index. The identical body posted to `POST /api/v1/memories` returns the actionable message. A 1,000-row batch that fails therefore yields 1,000 identical, information-free strings and the operator has no way to know *which field* to fix.

## Reproduction (executed)

Same record, two surfaces. Daemon `release/v1.0.0` `b2a9f5bd`, `--features sal-postgres`, live PG 18.4.

**Bulk** (5 verbatim corpus records):

```json
{"created":0,"errors":["validation failed","validation failed","validation failed",
                       "validation failed","validation failed"],"pending":[]}
```

**Single-create**, same first record — two successive, genuinely different failures:

```json
{"error":"invalid source 'atlas-knowledge-card' — must be one of: user, nhi, claude, hook, api, cli, import, consolidation, system, chaos, notify"}
```
```json
{"error":"source URI must start with one of: uri:, doc:, file:"}
```

Both were **distinct, independently-fixable** defects in the payload. The bulk surface rendered both as the same five words. Diagnosing a 7,912-row corpus load required bisecting field-by-field through the single-create endpoint — the bulk endpoint contributed nothing.

## Root cause

`src/handlers/memories_query.rs:875` (postgres branch; the sqlite branch mirrors it):

```rust
if let Err(e) = validate::RequestValidator::validate_create(&body) {
    tracing::warn!("bulk_create(postgres): validate_create failed: {e}");
    errors.push(super::sanitize_bulk_row_error(&e.to_string()).to_string());
    continue;
}
```

`sanitize_bulk_row_error` exists for a good reason — issue #851 established that echoing a caller-influenced title paired with a raw error leaks presence/shape of server-side fields. But it collapses the *entire* error to a constant, discarding the parts that carry no caller-controlled data.

The typed error already has the structure needed to do better. `validate::ValidationError { field, reason }` (`src/validate.rs`) carries **explicit field attribution**, and `RequestValidator::validate_create` returns exactly that type. The `field` component is a fixed, server-authored identifier from a closed set — it is not caller-influenced and leaks nothing. It is being thrown away.

## Proposed fix

Return the field attribution, keep the reason sanitized:

```rust
errors.push(json!({
    "index": row_index,                 // which row in the submitted array
    "field": e.field,                   // server-authored, closed set — safe
    "error": sanitize_bulk_row_error(&e.to_string()),
}));
```

`errors[]` is already a JSON array in the envelope, so richer elements are additive; a caller reading `errors.len()` is unaffected. The `index` matters as much as the `field` — with 1,000 rows in flight, "row 417 failed on `source_uri`" is actionable and "validation failed" is not.

If echoing even the field name is judged too much, the `index` alone is a strict improvement: it lets the caller re-submit that single row to `/memories` and read the full message there.

## Why it matters at ASI scale

Bulk ingest is the fleet-orchestration write surface, and it is the one surface where a human is *least* likely to be in the loop. An error channel that cannot distinguish "your `source` value is not on the allowlist" from "your `source_uri` needs a scheme prefix" is not fleet-orchestratable — it forces every failure into a manual bisect against a different endpoint. That is the MANAGEABLE-AT-SCALE leg of the North Star, not a cosmetic concern.

Cross-reference: #851 (the sanitization this preserves), #2551 (the sibling envelope-truthfulness defect on the same handler).

---
🤖 Filed autonomously per the track-and-fix prime directive. Verified by execution per pm-v3.3 step 7.

### LINKED ISSUE #2588: data-integrity: bulk write returns 200 OK with "internal error" when the quota rejects every row — 31k rows silently dropped, zero logs
## Summary

`POST /api/v1/memories/bulk` returns **HTTP 200 OK** when 100% of its rows were rejected by the per-agent write quota, reports each rejection as the opaque string `"internal error"`, and **logs nothing at all**. A corpus-ingest job silently loses every row past the quota while being told it succeeded.

The single-row `POST /api/v1/memories` on the *same* daemon, *same* agent, *same* quota returns a fully typed **429 `QUOTA_EXCEEDED`**.

Found by the LENS C leg of the 3x3 adversarial performance audit on `release/v1.0.0` @ `93a30bd5` — encountered while loading a test corpus, where **31,000 of 32,000 rows were silently dropped behind an HTTP 200**.

## Measured evidence

Daemon booted with `AI_MEMORY_MAX_MEMORIES_PER_DAY=5`, agent `qtest`:

```
single-row POST /api/v1/memories  #1..#5 -> HTTP 201
single-row POST /api/v1/memories  #6     -> HTTP 429
  {"agent_id":"qtest","code":"QUOTA_EXCEEDED","current":5,"limit":"memories_per_day",
   "error":"QUOTA_EXCEEDED: agent qtest namespace q hit memories_per_day (current=5, max=5)"}

bulk POST /api/v1/memories/bulk (2 rows, same agent, quota already exhausted) -> HTTP 200
  {"created":0,"errors":["internal error","internal error"]}

daemon log lines matching /quota/i during the whole run: 0
```

At corpus scale the same shape:

```
{"off": 31500, "sent": 500, "status": 200, "created": 0, "n_err": 500, "err0": ["internal error"]}
{"SUMMARY": {"requested": 32000, "created": 1000, ...}}
```

HTTP 200. 31,000 rows gone. No server-side trace of why.

## Root cause

`src/handlers/errors.rs:52` `sanitize_bulk_row_error` classifies per-row bulk errors against an allowlist of substrings and falls through to `"internal error"`:

```rust
pub fn sanitize_bulk_row_error(raw: &str) -> &'static str {
    let lower = raw.to_ascii_lowercase();
    if lower.contains("cannot be empty") || lower.contains("exceeds max") || … { return "validation failed"; }
    if lower.contains("already exists in namespace") || … { return "conflict: already exists"; }
    if lower.contains("not found")                        { return "not found"; }
    if lower.contains("denied by governance") || …        { return "forbidden"; }
    if lower.contains("quorum") || …                      { return "replication unavailable"; }
    "internal error"
}
```

There is **no quota arm**. The quota error text — `"QUOTA_EXCEEDED: agent qtest namespace q hit memories_per_day (current=5, max=5)"` — matches none of the branches, so it becomes `"internal error"`.

Note this is not a data-leak tradeoff: the *identical* text is already returned verbatim to an unauthenticated caller by the single-row 429 path, so classifying it costs no confidentiality.

## Why this matters (North Star)

This is a **silent, unintentional loss of durable writes**. The caller is told 200 and has no signal to retry, back off, or raise the quota; the operator has no server-side log line to correlate. Three independent failures compound:

1. **Wrong status.** 200 for a request where every row failed.
2. **Erased actionable detail.** A typed, non-sensitive, retryable `QUOTA_EXCEEDED` is flattened to `"internal error"`.
3. **Zero observability.** Nothing at any log level; the documented `ai_memory_federation_*`/quota telemetry does not cover this funnel.

CLAUDE.md env #49/#84 describe quota behaviour as a "clean 429" (the in-repo `loadtest` harness even carries a `quota-throttle` reason tag for 429). The bulk surface neither 429s nor logs.

## Reproduction

```bash
AI_MEMORY_MAX_MEMORIES_PER_DAY=5 ai-memory --db /tmp/x.db serve --port 9103 &
for i in $(seq 1 6); do
  curl -s -o /dev/null -w "%{http_code}\n" -X POST :9103/api/v1/memories \
    -H 'Content-Type: application/json' -H 'X-Agent-Id: qtest' \
    -d "{\"namespace\":\"q\",\"title\":\"t$i\",\"content\":\"c\",\"tier\":\"mid\"}"
done   # 201 x5 then 429
curl -s -w '\n%{http_code}\n' -X POST :9103/api/v1/memories/bulk \
  -H 'Content-Type: application/json' -H 'X-Agent-Id: qtest' \
  -d '[{"namespace":"q","title":"b1","content":"x","tier":"mid"}]'
# -> {"created":0,"errors":["internal error"]}  /  200
```

## Proposed fix

1. **`src/handlers/errors.rs:52`** — add a quota arm before the fallback, e.g.
   `if lower.contains("quota_exceeded") || lower.contains("quota") { return "quota exceeded"; }`. ~3 lines.
2. **Status honesty in the bulk handler** — when `created == 0 && !errors.is_empty()`, do not return 200. Return the typed status that matches the dominant cause (429 for quota) or at minimum a 207/4xx; a request in which nothing was created is not a success. ~15 lines.
3. **Log it.** Emit one WARN per bulk call summarising `{created, failed, dominant_cause}` — one line per call, not per row, so a 500-row rejection cannot flood. ~5 lines.
4. Consider surfacing the structured per-row cause code (not just a label) in the bulk response so a client can programmatically retry.

Suggested regression pin: a test that exhausts the quota, POSTs a bulk batch, and asserts the response is **not** 200 and the per-row error is **not** `"internal error"`.

## Related

- env #49 `AI_MEMORY_MAX_MEMORIES_PER_DAY`, env #52 `AI_MEMORY_MAX_PAGE_SIZE`, env #84 (CLAUDE.md).
- `tests/handler_error_sanitization.rs` — the existing pin on this classifier's allowlist.

🤖 Filed by the LENS C leg of the 3x3 adversarial performance audit.

### LINKED ISSUE #2594: bulk create performs NO embedding on either backend while single create embeds inline — bulk-ingested rows are silently invisible to semantic recall
## Summary

`POST /api/v1/memories/bulk` performs **no embedding at all** on either backend, while `POST /api/v1/memories` embeds inline. Rows ingested in bulk are therefore invisible to semantic recall until the paced backfill sweep reaches them; rows ingested singly are searchable immediately. Same endpoint family, same tier, two different visibility contracts, and nothing on the wire says so.

## Evidence

`src/handlers/memories_query.rs::bulk_create` contains **zero** references to `app.embedder` (the whole function; both the postgres branch and the sqlite branch). Confirmed:

```
$ grep -n "embedder" src/handlers/memories_query.rs
1467://   - embed_create_before_lock   (1) no embedder ⇒ (None, Indexed)     <- a comment, not a call
```

versus the single-create path:

```
src/handlers/create.rs:303    match app.embedder.as_ref().as_ref() {          # sqlite
src/handlers/create.rs:1029   match app.embedder.as_ref().as_ref() {          # postgres
```

The postgres bulk branch persists via `store_batch`, which has no embedding parameter at all (`MemoryStore::store_batch(&self, ctx, memories)`), whereas single-create uses `store_with_embedding(ctx, mem, embedding, space)`.

Wire-tap confirmation: a 1000-row bulk request issues exactly one `INSERT INTO memories` with no `embedding` / `embedding_space` binding, and 7 gate statements per row (see #2589) — no embed call, no vector column written.

## Why this matters

- **Silent semantic invisibility.** A caller who ingests 100k rows via bulk gets a `{"created": 100000}` success envelope and a corpus that returns nothing from semantic recall until the sweep catches up. The sweep is paced at `AI_MEMORY_EMBED_BACKFILL_BATCH` (default **100 rows/pass**, env #38), so a 100k-row bulk ingest needs ~1000 passes.
- **It is only reachable while a daemon runs.** The backfill sweep is a `serve`-boot/loop concern; an MCP-stdio or CLI-only topology that bulk-ingests never runs it.
- **The response is not truthful about it.** The bulk envelope is `{"created", "errors", "pending"}` — there is no `embed_status`, while the single-create path carries an `EmbedStatus` (`Indexed` / degraded) precisely so callers can tell.

## Interaction with the other write-path findings

This is the *mirror image* of #2593 (single-create embeds synchronously and pays 203 ms/write for it). Between them the two paths bracket the correct answer:

- single-create: embeds inline, correct visibility, terrible latency
- bulk-create: no embed, great throughput, silently wrong visibility

The consistent resolution is the one #2593 proposes — **take embedding off both write paths** and let the (already-existing) backfill sweep own vectorisation for *both* — provided the sweep is made reachable from every topology and the response tells the caller the row is not yet semantically visible.

## Proposed fix

1. Make the contract **explicit and identical on both endpoints**: either both embed inline or (preferred) neither does, with the backfill sweep as the single vectorisation owner.
2. Add an `embed_status` / `pending_embeddings` field to the bulk response so a caller can see that N rows are stored-but-not-yet-searchable. Truthfulness over silence.
3. If inline bulk embedding is kept, use `Embed::embed_batch` (it already exists — `src/store/mod.rs:4031` uses it in the backfill sweep) rather than N single calls, so the vendor round trips are batched the way the INSERT already is.
4. Document the asymmetry in the meantime.

**Data-integrity note:** no text is lost either way — the durable memory text is written correctly by both paths, and embeddings are regenerable derived data. This is a *truthfulness* and *availability-of-recall* defect, not a corruption one.

## Reproduction

```
python3 .local-runs/wt-perf-lensb/bench/pgspy.py --listen 5434 --upstream 5433 --log spy.jsonl &
# daemon on tier=autonomous with a real embedder, pointed at the tap
python3 .local-runs/wt-perf-lensb/bench/e3_bulk.py <port> emb 3
# then: SELECT count(*) FILTER (WHERE embedding IS NULL) FROM memories WHERE namespace LIKE 'perfb/e3-%';
```

Filed by Claude Opus 5, LENS B (write path) of the 3x3 adversarial performance audit.

alphaonedev:
Closes #2550
Closes #2551
Closes #2552
Closes #2588
Closes #2594

Five filed issues, **one root cause**: `handlers::bulk_create` **re-implemented** the single-create funnel instead of calling it. Each issue is a different stage that re-implementation omitted, so this is one structural fix, not five patches.

| issue | omitted stage | measured pre-fix behaviour |
|---|---|---|
| #2550 | field resolution | caller `scope` and `kind_provenance` DROPPED |
| #2552 | error mapping | every row error → `"validation failed"`, no field, no index |
| #2594 | embedding | zero `app.embedder` references on either backend |
| #2551 | accounting | 7,912 sent / `created: 7912` / 7,855 landed / `errors: []` |
| #2588 | exit discipline | 31,000 rows dropped behind `200 {"created":0,"errors":["internal error"]}` |

`create::{resolve_create_metadata, build_create_memory}` are now **THE** projection for all four create-family write funnels (single sqlite + postgres, bulk sqlite + postgres). A field added there reaches every funnel by construction; it can no longer be honoured on one and silently dropped on another.

## R-203 — parent SHA `5449b6da`, observed failures

All 12 cells of `tests/bulk_write_truthfulness_2551.rs` FAIL at the parent commit, each with the exact value its issue reports. Verified by checking out `origin/release/v1.0.0 -- src/` (and removing the post-fix `src/handlers/bulk.rs`, which `git checkout` does not delete) and re-running: **`0 passed; 12 failed`**.

```
bulk_honours_caller_scope_2550                  metadata carries no `scope` at all
bulk_stamps_kind_provenance_2550                left: None                right: Some("declared")
bulk_matches_single_create_field_resolution_2550 left: String("collective") right: Null
bulk_created_counts_persisted_not_sent_2551     {"created":3,"errors":[]}  — 2 rows in the table
bulk_updated_is_distinct_from_created_2551       a cross-call upsert reports created: 1
bulk_row_errors_carry_index_and_code_2552       {"created":1,"errors":["validation failed"]}
bulk_total_rejection_is_not_200_2588            200 {"created":0,"errors":["internal error","internal error"]}
bulk_partial_fill_is_207_2588                   left: 200                 right: 207
bulk_refuses_disagreeing_agent_id_claim_2550    the claim was silently overwritten; the row landed
bulk_write_funnel_embeds_2594                   the funnel never references `app.embedder`
```

The `#2594` cell first passed against the parent tree because it read a fixed path that only exists post-fix. It now locates the funnel **by the handler it defines** and asserts on that file — a path-keyed assertion is exactly how a structural pin goes vacuous.

Two **CONTROL** cells (`bulk_all_success_is_still_200`, `bulk_empty_batch_is_200`) pin that the new refusals do not pass trivially — the #2449 lesson that a cell which never reaches its assertion also "fails". Counter assertions are **always** paired with an out-of-band `SELECT`, because a counter-only assertion is precisely how #2551 hid: the envelope was self-consistent and wrong.

## What changed

**#2550 — field resolution.** Caller `scope` was accepted by the DTO, never read, and the row landed on the `scope_idx` generated column's `'private'` default: a caller asking for a fleet-visible `collective` write got an owner-only row and **every other agent silently saw zero results**, reported as `200 {"errors":[]}`. `kind_provenance` was dropped the same way. Both now resolve through the shared funnel, pinned by a surface-parity cell posting the SAME body to both endpoints.

**#2550, security half.** The bulk branches stamped the resolved caller OVER a disagreeing `agent_id` / `metadata.agent_id` claim, silently. Single-create has refused that with 403 since #907. Bulk now refuses it too, typed `AGENT_ID_MISMATCH`.

**#2551 — accounting.** Envelope is now `{sent, created, updated, deduped, rejected, errors, pending}` with the reconciliation identity `created + updated + deduped + rejected + pending.len() == sent`. `created` counts rows INSERTED, `updated` rows upserted onto an existing `(title, namespace)`. **An in-batch collapse is a result the caller must be told about**: `deduped` counts them and `deduped_rows[]` makes each addressable by input index, so a loader can identify exactly which of its 1,000 source records was discarded.

Everything is derived from the persist stage's **returned ids** — `id` is absent from BOTH backends' `ON CONFLICT DO UPDATE SET`, so `returned == minted` is an exact, zero-round-trip insert-vs-update discriminator, and the grouping key is the returned id (never a recomputed `(title, namespace)`, which both funnels may rewrite via secret screening before the dedup key is taken). No trait signature change, no extra SQL. That property and the last-wins collapse order are now part of the `MemoryStore::store_batch` **trait contract**, because a first-wins adapter would make the handler name the WRONG row's content as durable.

**#2552 — error mapping.** `errors[]` entries are `{index, code, field?, error}`. #851 sanitization is preserved — `error` keeps the same allowlisted labels — and `field` passes a server-authored-slug **shape guard at the serializer**, so no future refactor can route caller-influenced text through the new channel.

**#2588 — exit discipline.** A quota arm in the classifier (matched on the slug, so a failed quota *read* still classifies internal — a broken substrate is not a signal to back off); a status that reflects reality; one structured summary WARN per call (the pre-fix path logged **nothing at any level**).

```
200  every row landed as its own distinct row
207  partial application
202  every row queued for governance approval
otherwise: the dominant cause, RETRYABLE first — 429 > 503 > 500 > 403 > 409 > 400
```

Dominant cause is a **precedence, never a count-majority**: 300 quota-rejected + 200 invalid rows returning `400` would tell the fleet "never retry" and permanently drop 300 writable rows. 207 is deliberately 2xx — the batch is not idempotent for rows whose title the substrate rewrote, so a 4xx on partial fill would arm retry middleware to replay writes that already landed.

The postgres branch additionally classified its own typed `StoreError::QuotaExceeded` as `internal error`, because its prose (`"quota exceeded for a in ns: memories_per_day 5/5"`) matches no allowlisted arm. **Typed SAL errors are now classified by `code()`, not by prose** — re-deriving from prose is a lossy inference over a value we already hold, and it made the two backends disagree about the same condition on the same request. Caught by a live-Postgres run, not by reading.

**#2594 — embedding.** Bulk now embeds **before the DB lock** (the #219 discipline) via `Embedder::embed_batch`, chunked so a 1,000-row ingest costs ~10 vendor round trips rather than 1,000; persists vectors on both backends; warms the HNSW index on sqlite; and reports `embed_status` truthfully when vectorisation degraded. A failure **degrades** — the durable memory TEXT still lands and vectors are regenerable derived data. An arity mismatch from the embedder refuses to guess which row each vector belongs to, because mis-attributing a vector produces WRONG recall results, which the North Star ranks strictly worse than fewer.

**Also closed, same root cause:** the sqlite bulk branch had **no namespace-governance consult at all** — it went straight from the quota check to `db::insert`, while the postgres branch has enforced it since #705 and single-create enforces it on both backends. That is the funnel's most security-relevant omitted stage, so it is fixed here rather than left as the one write surface where namespace governance does not apply.

## Crossroads — 3x3 adversarial vote (9 lenses)

Two coupled T1/T3 decisions: the wire shape of the per-row outcome accounting (D1) and the HTTP status disposition (D2).

- **D1 = A (6-2-1)** — typed `errors[]` objects + counters, over enriched strings (B) or a dense per-input-row `results[]` (C).
- **D2 = A (7-2)** — 200 / 207 / 202 / dominant-cause, over 200-on-partial (B) or 207-on-any-error (C).

Lenses: in-repo precedent, wire/back-compat, North Star literalist, HTTP/REST semantics, security disclosure, fleet orchestration, implementability, testability/R-203, blast radius.

**Objections that changed the design** (none of these were in the original plan):

1. **`errors[]` was not purely per-row.** Post-commit federation fanout and catchup failures were pushed into the same array, so `rejected` could not reconcile and a **fully durable** batch would have been 207'd by peer reachability. Those moved to a new `warnings[]` channel — which also collapsed the D2-A/D2-C disagreement, since "any error" and "partial application" then mean the same thing.
2. **"Dominant cause" as a count-majority loses data in both directions** → replaced with retryable-first precedence.
3. **Deriving a 403 identity refusal's class from its prose** classified it as a retryable 500 → `CreateFieldError::as_bulk_class` and `classify_store_err` project typed errors directly.
4. **Dedup must be per-index addressable**, not a scalar (lenses 3/6/7) → sparse `deduped_rows[]`, which answers that without option C's O(rows) tax on the 99.9% happy path.
5. **`field` will become an oracle** if validation is ever refactored to per-field attribution → shape guard at the serializer.
6. **`pending[]` was emitted on the postgres branch only**, so the same batch produced a different envelope shape per backend → emitted on both.
7. **The last-wins collapse order is a hand-maintained SQL convention**, not a trait contract → written into the `store_batch` trait doc and pinned cross-backend.
8. **Counters must not be re-derived from `allowed.len()`** or the fix ships #2551 unchanged behind prettier fields → every counter is derived from an observed per-row outcome, and every counter assertion is paired with an out-of-band `SELECT`.

The vote decision was recorded here rather than via `memory_store` — the memory MCP tools are not in this agent's tool set.

## NOT in scope

**#2589** (the per-row statement gate) is deliberately excluded. It is blocked on **#2595**: the "7 SQL round trips per row" **IS** the per-write governance resolution, and batching on top of an unfixed resolver is wasted work.

## Testing

- **sqlite**: `tests/bulk_write_truthfulness_2551.rs` (12 cells), `handler_error_sanitization`, `bulk_consolidate_quota_1788`, `http_bulk_pre_event_gate_2389`, `http_bulk_create_form4_form6_round_trips`, `http_bulk_write_sig_emit_2000` (`--include-ignored`), `l07_3_chunk_d_http_surface` (159), `integration` (216), `security_authz_isolation_cluster`, `handler_postgres_branches_fake_pg` (90), `recall_purity_caller_guard_p01`, `session_start_scope_private_1420`, `config_precedence`, `agent_attestation_integrity`, `cargo test --lib handlers::` (532).
- **live Postgres** (PG 18.4 + AGE 1.7.0 + pgvector 0.8.6, `--include-ignored --test-threads=1`, on a **scratch database** so the shared certified tier was never written): `g6_postgres_quota_enforcement_1795` (11 — now also the cross-backend proof that the batched `store_batch` branch and the per-row `db::insert` branch report identical status and counters for the same request), `issue_2059_2060_covenant_pg` (18), `serve_postgres_extended` (20), `qual_pg_write_funnel_parity_2393_2397`, `store_parity_gaps` (46).
- **SSOT / gates**: `qual_10_module_size_ceiling`, `qual_6_7_legacy_error_type_ceiling`, `budget_tokens`, `route_count_invariant`, `cli_subcommand_count_invariant`, plus all seven script gates (`check-docs-vs-ssot`, `check-hardcoded-literals`, `check-vendor-literals`, `qc-codegraph-precheck`, `check-l3-boundary`, `check-migration-ladder`, `check-required-contexts`).
- **CI parity**: `cargo fmt --all --check`, `cargo check --examples`, `cargo check --all-targets` and `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` on **all four feature legs** (default, `sal`, `sal-postgres`, `vectorlite`).

## Wire change

`errors[]` elements move from bare strings to objects, and a partial batch answers `207` instead of `200`. Both are pre-GA corrections to a contract that was reporting outcomes it had not achieved. `docs/API_REFERENCE.md` carries the full envelope; `sdk/python`'s `BulkCreateResponse` — which **already** declared `errors: list[dict]` (so it raised a pydantic `ValidationError` on any failed row against the shipped daemon) plus a phantom `ids` field the handler has never emitted — is corrected to match the real wire.

## Pre-existing findings surfaced, filed separately

- `store_parity_gaps::pg_parity_private_leak_and_bypass_a7_1720` fails on a **fresh** Postgres database. Confirmed **pre-existing**: it fails identically at parent `5449b6da` on the same database, and touches no bulk code.
- `sdk/typescript`'s `storeBulk` POSTs `{memories: [...]}` against a handler typed `Json<Vec<CreateMemory>>` and declares a `count` field the server never sends — broken at HEAD, no test covers it.
- The bulk `429` carries no `Retry-After`, matching the single-row 429. A per-agent quota-window-derived value with jitter would prevent a synchronized fleet retry at the reset boundary.

## AI involvement

Authored by Claude Opus 5 under operator authority. Design resolved by a 3x3 adversarial vote (9 distinct lenses, tallies and the eight design-changing objections recorded above). Every pre-fix behaviour cited was reproduced by **execution** against the parent tree, not adopted from the issue text.

---

## Addendum — one pre-existing assertion changed, and why that is legitimate

`tests/valid_from_write_surface_2258.rs::http_bulk_store_rejects_malformed_valid_from` asserted **`status == 200`** for a batch in which **nothing was persisted**. That assertion is now `400`. Calling this out explicitly because "I changed the test to match my code" and "the test was asserting a defect" are indistinguishable in a diff, and only one of them is legitimate.

**It was asserting the defect.** `200` for a batch that persisted nothing is precisely the contract #2588 was filed to remove — *"bulk write returns 200 OK … 31,000 rows silently dropped"*. The evidence that it was **incidental scaffolding** rather than a deliberate contract choice by that file's author:

- The file's subject is **#2258 claim-bitemporal `valid_from`/`valid_until`**, not the bulk envelope. Nothing in #2258 concerns envelope status.
- Its own comment justifies only `created == 0` and non-persistence — *"Per-row validate_create failure skips the row into `errors[]` (never a silent store)"*. The author's stated concern is the **row**, not the envelope. Nothing explains why the envelope should be `200`.
- The **sibling happy-path cell directly above** (`http_bulk_store_valid_from_round_trips`) uses the byte-identical `assert_eq!(status, StatusCode::OK, …)` boilerplate. It is **untouched and still green**, because a clean batch still returns `200`. That is the control: this was copy-paste, and I have not blanket-rewritten `200`s.
- Decisively: across **all three** D2 candidate designs in the 3x3 vote (A, B *and* C), a batch that persisted nothing **never** returns `200`. The options differed only on the **partial** case. So this is not the losing side of a contract disagreement — it is outside every design that was on the table.

**The test's subject is unchanged.** That a malformed `valid_from` is rejected and the row does not persist is still asserted, and still passes.

**`400` is where the vote said, not merely somewhere non-200.** The dominant cause is `VALIDATION_FAILED`, and `400` is the identical status the same body gets from single-create `POST /api/v1/memories` — the surface parity this PR exists to establish. Pinned as its own cell, `bulk_total_rejection_status_tracks_the_cause_2588`, deliberately placed beside the quota twin that drives the same nothing-persisted path to **429**: a single "not 200" assertion would pass against a handler that returned one blanket error status for every cause, which is exactly what the retryable-first precedence exists to prevent.

## Addendum — a gap in my own test sweep

I found this in CI, not locally, and the reason is a real methodology error worth recording. `valid_from_write_surface_2258` compiles **zero tests under default features**; my consumer sweep ran it and read `ok. 0 passed` as a pass. I had already caught that exact pattern once — I re-ran `http_bulk_write_sig_emit_2000` under `--features sal` when it reported 0 tests — and simply failed to apply the same rule here. Silence is not success.

Corrected by re-sweeping **every** file that references the bulk surface or any artifact I changed, under the feature matrix, treating any `0 passed` as a failure to investigate. That found a second gap of the same kind: **`issue_1481_store_batch`** — the suite that directly owns `store_batch`, the trait method whose returned-id semantics my entire accounting depends on — needs `--features sal-postgres` and had been silently reporting 0 tests. It passes **10/10**.


alphaonedev:
## Self-audit addendum (`e282d513`) — a sixth omitted stage

Auditing my own diff against "every reachable code path" surfaced one more instance of the same root cause, which I am flagging explicitly because it is a **data-integrity** fix, not a cleanup.

`db::insert` returns the **existing** row's id when a write lands on the `(title, namespace)` conflict arm. The bulk sqlite branch discarded it (`Ok(_) => created_mems.push(mem)`) and carried the **minted** uuid forward into two consumers that key on identity:

1. **The HNSW index was seeded with an id that resolves to nothing in the database.** Semantic recall could return a hit whose `get` then 404s — a **wrong** result, which the North Star ranks strictly worse than a missing one.
2. **The federation fanout published the minted id to peers**, creating a **divergent row** there for content that upserted here — id drift across the fleet, on the write surface built for corpus-scale ingest.

Single-create has adopted the persisted id since #866 (`fanout_and_assemble_create_response` sets `mem_echo.id = actual_id`). Bulk re-implemented the funnel and omitted it — the same shape as the five filed issues, which is why it belongs in this PR rather than a follow-up.

Also corrected in the same commit: the postgres embedding stamp now emits **one entry per persisted id, last-wins** — the rule the upsert itself used to pick the surviving content. Duplicate ids in a single multi-row `UPDATE` are the shape postgres refuses ("cannot affect row a second time"), and pairing the surviving **content** with an earlier row's **vector** would make semantic recall score a row against text it does not hold.

### Coverage

9 unit pins added for the accounting core (`src/handlers/bulk.rs::tests`), including the one that directly answers the implementability lens's killer objection — that the two branches are entirely different code paths whose merge semantics agree only by hand-maintained SQL convention:

- `classify_in_batch_collapse_postgres_shape` — `store_batch` drops all but the last duplicate before the INSERT, so the surviving id is the **last** row's minted id.
- `classify_in_batch_collapse_sqlite_shape_matches_postgres` — sequential upserts make the surviving id the **first** row's minted id.

Both must yield **identical** counters (`created 2, updated 0, deduped 1`, superseded index 0 by index 2) for the same request, or the backends would report different truth for the same batch.

Honest scope note: the HNSW / federation effects of the id adoption are **not** hermetically observable in a keyword-tier in-process router (no embedder, no peers), so the pin is the accounting path that consumes the same returned id, plus the unit matrix above. Re-verified after the change: all four clippy legs, `cargo test --lib handlers::` (541), the 12-cell regression suite, and the live-Postgres `g6_postgres_quota_enforcement_1795` (11).

alphaonedev:
# Code + security review — PR #2644 (bulk_create funnel truthfulness)

**Control:** bulk_create must reuse create funnel (no second semantics) and report truthful HTTP status / errors (not 200 with silent zeros).

**SHA:** e72944f8 (re-synced after #2643)

## Code
- bulk path routes through create funnel (2550/2552/2594 class)
- Status/error truthfulness for quota and total rejection (2551/2588)
- CHANGELOG + API_REFERENCE keep-both with claims train docs
- Tests: bulk_write_truthfulness_2551, bulk_consolidate_quota, etc.

## Security / data integrity
| Risk | Disposition |
| Bulk bypass of create gates | closed by funnel reuse |
| 200 success on total failure | status truthfulness |
| Quota silent drop | folded errors / proper codes |

alphaonedev:

**Control:** Capacity bulk write truthfulness — `bulk_create` reuses the single-create funnel and reports honest envelopes (#2550/#2551/#2552/#2588/#2594).

### Code
- New `src/handlers/bulk.rs` funnel; no more parallel re-implementation of create.
- Envelope: `created` = inserted, `updated` = upserted, `deduped` / `rejected` reconcilable; `created+updated+deduped+rejected+pending == sent`.
- Status: total success 200; total rejection not 200 (tracks cause); partial **207 Multi-Status**.
- Local R-203 suite `bulk_write_truthfulness_2551` **13/13 pass**.
- Rebased onto post-#2643 tip `8aa83e6f`.

### Security / integrity
- Closes false-success class: bulk no longer over-counts `created` or hides total rejection behind HTTP 200.
- Scope/agent_id claim honour (#2550) prevents bulk bypass of create-path authz stamps.

### Merge bar
- Await required checks green; merge commit only; then capacity fed #2662/#2663/#2673.

alphaonedev:
Superseded for merge by **#2689** (single SSH-signed re-land of the same tree). Original #2644 tip history included unsigned commits rejected by release verified-signature ruleset after CI green. Close this PR after #2689 merges.

alphaonedev:
Closed: merged via signed re-land #2689.

Issue #2032, opened 2026-07-14
🛡️ v1.0.0 adversarial security assessment (5-agent) + NSA CSI MCP compliance re-verification (580 tests green)


alphaonedev:
## Summary

Multi-agent **adversarial security assessment** of the ai-memory MCP daemon (HTTP/`axum` surface + SAL + federation/crypto), run as five parallel adversarial auditors over distinct domains, **plus a full re-verification of NSA CSI MCP compliance**. Motivated by the `7WaySecurity/ai_osint` exposure-recon threat model (Shodan/Censys discovery of exposed MCP servers, open vector DBs, leaked keys, 0.0.0.0 binds).

- **Provenance:** assessed at tree tip `6c97e644` (`v0.10.0-82-g…`, v1.0.0 dev line). Read-only analysis; all findings carry `file:line` evidence.
- **Method:** 5 domain agents — (1) AuthN/AuthZ & access control, (2) Injection/deserialization, (3) SSRF/outbound, (4) TLS/deployment exposure, (5) Federation-crypto/DoS.
- **Headline:** the cryptographic core is **sound** (macaroon verify, `verify_strict` everywhere, fail-closed federation defaults at the v1.0 tip, comprehensive SSRF guard, parameterized SQL, no command injection, no request decompression path, 2 MiB body cap, constant-time key compare). **No SQLi / command-injection / high-critical SSRF found.** The exploitable residual risk is **architectural**: the HTTP surface's *identity* is a self-asserted `X-Agent-Id` header that is cryptographically bound on the **create** path but **not** on the **read/mutate/admin** paths — the same root cause behind the two highest findings.

## ✅ NSA CSI MCP compliance — RE-VERIFIED GREEN

Re-ran every named test in the authoritative control→test matrix (`docs/compliance/nsa-control-test-matrix.html`) covering all **19 controls** (NSA concerns a–j + recommendations a–g + 2 meta) from *NSA CSI, MCP Security Design Considerations, U/OO/6030316-26, May 2026 v1.0*.

- **PART 1 — lib-side control tests** (tls / subscriptions SSRF-guard / signed_events / verify_signed_events / validate / mcp server_identity): **203 passed, 0 failed**.
[...]

Issue #1975, opened 2026-07-09
[v1.0][Gate-0][honesty] LongMemEval Gemma-4 refresh: re-run + republish (or retract) the stale gemma3:4b R@5 numbers


alphaonedev:
KEEP-v1.0 (Gate-0 honesty blocker). README:771/810/816 + docs/DEVELOPER_GUIDE.md:1096-1103 + ROADMAP §9.5 publish 97.8% R@5 measured with gemma3:4b LLM query-expansion, while prod now defaults to Gemma 4 (ROADMAP:673 admits it). A freeze+audit release cannot attest benchmark numbers for a model it no longer runs (§2.5). Discharge: re-run LongMemEval-S with the current default expansion model (gemma4) + republish across all surfaces, OR formally retract/caveat. Immediate caveat already applied to README pending the re-run.

Filed per the Sprint-0 W4 keep/cut ruling (2x5 vote wf_78b859e1-24e, memory 8c5e9f2a) under epic #1940 / Gate-0 #1938.

alphaonedev:
DISCHARGED per 2×5 adversarial vote wf_8ac90aca-6eb (C=6/B=4; ruling memory 176af6ed; commits 2b8a7c91 + 8c71124b pushed to main, e5cb7594..8c71124b).

**Disposition: formal retract + promote (the issue's second disjunct).** The prescribed local gemma4:e4b re-run is infeasible on the CPU-only reference host (~1 tok/s ≈ 79s/request vs 30s harness timeout; a first attempt produced 8 passes of silently-empty expansions — harness defect filed+fixed as #1983). Decisive adversarially-verified fact: the COMPILED default expansion model at HEAD is still gemma3:4b (src/config.rs::backend_default_model), so the historical 97.8% row is not counterfactual for the shipped product — only for Gemma-4 production deployments, which the already-measured OpenRouter leg covers.

[...]

Issue #1953, opened 2026-07-09
[v1.0][Gate-1prime] AI_MEMORY_RECALL_TOUCH_SYNC removal path


alphaonedev:
Completes P0-1 pure-recall permanence. One-cycle deprecation WARN in v0.10.0, removal in v1.0. Child of the v0.10.0 WARN-carrier epic. Pure recall shipped v0.9.0 (v77, #1869).

Filed under the v1.0.0 development epic #1940 (operator grant memory f9a0f397). Gate-item carrier from the 2026-07-09 kickoff audit (run wf_3c42dc07-761). P0/P1 substrate items are T1+T4 5-agent-vote crossroads; Rust skills (/rust-skills + /rust-microsoft) + the no-hardcoded-literals gates are mandatory on all src/ work.

alphaonedev:
Landed on main via PR #1997 (merge 533b80bf). `AI_MEMORY_RECALL_TOUCH_SYNC` removed: env const, `recall_touch_sync_enabled()`, the deprecation-warning const + emitter, and both sync-touch branches (postgres + sqlite) in `handlers/recall.rs` / `storage/mod.rs` deleted — recall is now unconditionally pure. Only a historical comment remains in postgres_schema.sql documenting the removed legacy flag. CLAUDE.md env row marked REMOVED. All 3 Check(os) + Per-Module Coverage green at merge. Closing as fixed.

Issue #1675, opened 2026-06-15
docs/cli-design-rationale.md missing the --features sal caveat for curator --reflect


alphaonedev:
**Source:** v0.7.1 adversarial audit. docs/cli-design-rationale.md describes `ai-memory curator --reflect` (lines ~14/22/35) with no note that it requires a `--features sal` build; the default build hard-bails (src/cli/curator.rs:548-560). ROADMAP §11.3.1 now tracks this doc update as a separate item.

**Proposed fix (~3 lines, docs):** add a note where `curator --reflect` is described: requires a binary built `--features sal` (SAL MemoryStore trait; still pure SQLite — `sal-postgres` is the Postgres backend) plus a configured LLM client; distinct from the ungated `ai-memory reflect` write primitive (#655).
Suggested labels: docs, honesty, sal

alphaonedev:
Fixed in v0.7.1 (released; tag v0.7.1, merged to main via PR #1702): docs/cli-design-rationale.md --features sal caveat for curator --reflect added (commit 54b48805). Gates green at release (default 5613 / sal-postgres 6034 tests, 0 failed; clippy -D pedantic; fmt; cargo audit). Verified per prime directive pm-v3 (verify-before-claiming + recompile-retest).

PR #2654, opened 2026-08-01
docs(readme): correct every false published claim against code at HEAD (CERT GATE 2)

### LINKED ISSUE #2450: [verification-integrity][100%-rust] the published 97.0% R@5 headline is produced by a 353-line PYTHON reimplementation of the ranking SQL that never invokes the binary — and the copy has already drifted from the shipped Rust
**Class:** verification integrity + 100%-Rust violation at the most load-bearing point. **Verified at `release/v1.0.0` @ `011dd8b9`.**

## The headline number does not measure the product

`docs/DEVELOPER_GUIDE.md:1098` attributes the published keyword-tier figure to `harness_99.py --no-expand`:

| harness | | R@5 | |
|---|---|---|---|
| `harness_99.py --no-expand` | Parallel FTS5, 10 cores | **97.0%** | 232 q/s (2.2s) |

That number is published as ai-memory's benchmark result in four places: `README.md:770`, `README.md:809`, `README.md:815` (in a table whose column header is **Tier**, so a reader parses it as ai-memory's keyword tier), and `ROADMAP.md:324`.

`harness_99.py` **never invokes the binary.** It opens its own database — `sqlite3.connect(":memory:")` (`:136`) — and runs its own copy of the ranking (`:61-70`):

```sql
SELECT m.id, m.tags,
       (fts.rank * -1) + (m.priority * 0.5) + (MIN(m.access_count,50)*0.1)
       + (m.confidence * 2.0)
       + (CASE m.tier WHEN 'long' THEN 3.0 WHEN 'mid' THEN 1.0 ELSE 0.0 END)
       + (1.0 / (1.0 + (julianday('now') - julianday(m.updated_at)) * 0.1))
FROM memories_fts fts JOIN memories m ON m.rowid = fts.rowid
WHERE memories_fts MATCH ?1 AND m.namespace = ?2
ORDER BY 3 DESC LIMIT ?3
```

That scoring expression is character-for-character `src/storage/mod.rs:15279-15282`. It also reimplements the FTS query sanitizer (`sanitize_fts_or`, `:72+`) that ships in Rust as `sanitize_fts_query`.

## The copy has ALREADY drifted — it measures a ranking no shipped build performs

Its entire `WHERE` clause is `memories_fts MATCH ?1 AND m.namespace = ?2`. Compared to the shipped Rust recall path it is missing:

- the **fail-closed `lifecycle_visible_clause`** allow-list (#1948, R19/A3) — `src/storage/mod.rs` interpolates `{lifecycle_vis}` into the real query
- the **visibility clause** (`visibility_clause(...)`, the scope=private / owner predicate)
- the **G7 soft-loser down-weight** (`src/storage/mod.rs:5939`, `:15920`) — a contradiction-conserved loser is penalized in the product and is not in the benchmark

So the published number is not merely "measured on a corpus that hides a defect" (#2437). It is measured by **a different, older, less-correct ranking function**, in a different language.

## Why this compounds every ranking issue currently in flight

A frozen Python copy cannot move when the Rust moves. **#2339, #2425 and #2436 can each be fixed correctly and the 97.0% headline will not change**, because the number does not come from the code being fixed. Conversely, a regression in `src/storage/mod.rs` cannot show up in the benchmark at all.

This is a second implementation of the ranking function, and it is the one that gets published.

## A Rust harness already exists

`src/bench.rs` is **1,739 lines** with `MAX_SCALE = 1_000_000`, `run_verified_recall`, `run_search_fts`, `run_recall_hot`, `run_recall_rerank_hot`. The Python harnesses are a parallel, divergent path alongside it.

## The repo's own benchmark doc already says this

`benchmarks/longmemeval/results.md:162-164` states the anti-goal verbatim — that the shadow-harness number is **not** to be presented as the binary-faithful number — and publishes the binary-faithful **96.4%** at `:40`. The honest document exists; two downstream summaries contradict it. Note also that `benchmarks/longmemeval/README.md:39` attributes the headline to `harness_fast.py` while `DEVELOPER_GUIDE.md:1098` attributes it to `harness_99.py` — **two incompatible provenances for one published number.**

## Suggested fix

1. **Re-measure the keyword tier through the binary**, or relabel the published figure to the binary-faithful **96.4%** with its provenance. `run_variants.sh` already scripts the run.
2. **Retire the four Python harnesses** (`harness.py` 392 · `harness_fast.py` 461 · `harness_blazing.py` 415 · `harness_99.py` 353 = **1,621 lines**) in favour of a LongMemEval mode on `src/bench.rs`. A benchmark that reimplements the thing it benchmarks cannot regression-test it.
3. **Add a gate** that fails when a scoring expression appears outside `src/`. The drift above was silent and would have stayed silent.
4. Reconcile the two contradictory harness attributions in `README.md:39` and `DEVELOPER_GUIDE.md:1098`.

## Relationship to #2437

#2437 found the harness corpus is uniform-priority so the additive prior cancels — true, and confirmed for **all four** harnesses. This issue is the deeper one: even with a heterogeneous corpus, the harness would still not be measuring the shipped ranking. **#2437's fix is necessary but not sufficient**; fixing the corpus without retiring the reimplementation yields a better measurement of the wrong artifact.

Context: #2440, #2437, #2425, #2436, #2339.

### LINKED ISSUE #2451: [100%-rust] 11 internal Python tooling files (3,142 lines) do Rust work — including a supply-chain CI gate and the CI baseline math
**Class:** language-boundary discipline. **Verified at `release/v1.0.0` @ `011dd8b9`.**

## The surface

37 tracked `.py` files, 7,005 lines total, splitting cleanly in two:

**26 files / 3,863 lines — legitimate ecosystem surface, NOT in scope here.** `sdk/python/**` (11), `clients/anthropic-shim-py/**` (7), `clients/openai-shim-py/**` (7), `clients/host-adapter-shim/python/` (1). A Python SDK and Python host shims are deliverables for other ecosystems; they cannot be written in Rust. (PyO3 bindings over the Rust core would be a stricter architecture, but that is a design choice, not a defect.)

**11 files / 3,142 lines — internal tooling doing work the Rust core should do:**

| File | Concern |
|---|---|
| `benchmarks/longmemeval/harness.py` (392) | **tracked separately — see the harness issue** |
| `benchmarks/longmemeval/harness_fast.py` (461) | ″ |
| `benchmarks/longmemeval/harness_blazing.py` (415) | ″ |
| `benchmarks/longmemeval/harness_99.py` (353) | ″ |
| `scripts/check-build-script-vetting.py` | **a supply-chain CI gate in Python**, while its sibling gates (`check-vendor-literals.sh`, `check-migration-ladder.sh`, `check-l3-boundary.sh`, `check-hardcoded-literals.sh`, `check-docs-vs-ssot.sh`) are shell. A build-script vetting gate is security tooling; it deserves the same language and test discipline as the thing it guards. |
| `.github/scripts/median_baseline.py` | CI **baseline math** — the statistics that decide whether a performance regression gate fires. Untested by the Rust suite. |
| `conformance/readers/reader.py` | conformance reader — the artifact that demonstrates the portability spec is implementable by a third party. Reasonable as a deliberate second-implementation proof; should be labelled as such if so. |
| `deploy/docker-1461/atlas/gen_batches.py` | deploy tooling |
| `deploy/docker-1461/bench/embed_bench.py` | bench tooling |
| `deploy/docker-1461/bench/recall_floor.py` | bench tooling |
| `deploy/docker-1461/bench/store_bench.py` | bench tooling |

## Why this is worth a tracker rather than a shrug

No document in the repo claims "100% Rust" — the `pure Rust` mentions at `docs/federation-identity.md:431` and `ROADMAP.md:1142` are correctly scoped to the federation-identity core and the HNSW fallback backend. So this is a **standard being eroded silently**, not a documented claim being contradicted. That is precisely the condition under which it keeps growing.

Two of the eleven are not incidental scripts: a **supply-chain gate** and the **regression-gate statistics**. Both are load-bearing for claims the project makes about itself, and neither is covered by `cargo test`, `cargo clippy`, or the coverage floor.

## Suggested action

1. **Decide and record the rule.** Either "the Rust core, its gates, and its benchmarks are 100% Rust; Python exists only as published SDK/client surface", or an explicit allowlist. Right now there is neither, which is why the count grew to eleven.
2. **Port the two load-bearing ones first** — `check-build-script-vetting.py` → a shell gate or a Rust `xtask`, and `median_baseline.py` → the existing `src/bench.rs` baseline path (`#1987`'s `--baseline` guard already lives there).
3. **Add a CI gate** that fails on a new `.py` outside the allowlisted SDK/client paths. Without it, any decision here decays.
4. Label `conformance/readers/reader.py` explicitly if it is an intentional independent-implementation proof — that is a legitimate reason to be non-Rust and should not read as drift.

Context: #2440. The four benchmark harnesses are tracked separately because they carry a much more serious defect than language choice.

alphaonedev:
## What

README claim correction. The input document is `docs/audit/3x7-claims-register-2026-08-01.md` §2 / §3.2 (carried onto this branch at `03bbd556`). README is a **binding claim surface**, so a wrong replacement would be worse than the original — **every number below was re-derived from source in this worktree**, not copied from the register.

Closes #2450
Closes #2451

## Corrections, with the file:line each was verified against

| Register ID | Old → New | Verified against |
|---|---|---|
| C-14 | `v0.9.0 — current release` → a written **v1.0.0** headline (v0.9.0/v0.8.0 relabelled prior-release, their surface numbers scoped to their own release) | `Cargo.toml:3` |
| C-14 | schema **v78 → v88** | `src/storage/migrations.rs:867`, `src/store/postgres.rs:1006` (lockstep) |
| C-14 / C-52 | MCP **101 → 103** at `--profile full` | `src/mcp/registry.rs` (103 `RegisteredTool::of`, pinned `== Profile::full().expected_tool_count()` at `:295`) |
| C-49 | routes **93 / 92 → 94**, unique paths **78 → 80**, at all 7 sites | `src/lib.rs:345`, `:375` |
| C-14 | CLI **89/87 → 91/89** | `src/lib.rs:415`, `:434` |
| C-14 | `Memory` **28 → 30** fields | `src/models/memory.rs:1097` |
| C-71 | 7-vs-8 core default reconciled in **one** direction — the page counts **advertised entries** everywhere, so core = **8** (7 Core-family + the always-on bootstrap) and full = **103** (102 callable + the same bootstrap). Badge updated to match | `src/profile.rs` `Self::Core` = 7 entries |
| C-01 | keyword **97.0% → 96.4%** | `benchmarks/longmemeval/results.md` (binary-faithful `harness.py`) |
| C-15 | semantic **97.4% → 96.8%**, and the **`autonomous` row RESTORED at 95.8%** — the only tier measured *worse* than baseline, previously dropped from the table | same |
| C-36 | **232 q/s** attributed in-cell to `harness_fast.py --parallel` (10-core in-process Python+SQLite, not the shipped path); binary-faithful path published at **~57 q/s** | `benchmarks/longmemeval/README.md:38-39` |
| — | The **97.2% Gemma-4 anchor and its #1975 retirement note are PRESERVED intact**, now labelled shadow-harness so the two harnesses can never be compared | #1975 ruling |
| C-40 | "recall mutates the database" / "each recall extends the TTL" → **recall is unconditionally pure**, stated as the stronger property (safe on a read replica, idempotent under retry), with the honest caveat that the fold job is a `serve`-daemon loop | `src/handlers/recall.rs` (#1953); `rg '\.touch_after_recall\(' src/` → all hits inside `#[cfg(test)] mod tests` |
| C-35 | "two production backends behind one identical API" → **73 of 94 route registrations Postgres-supported, 21 return a uniform 501**, plus **MCP stdio is structurally sqlite-only** | derived by evaluating `postgres_endpoint_supported()` (`src/handlers/postgres_gate.rs:97`) against every `.route(...)` registration in `src/lib.rs`; `CLAUDE.md` §Architecture (#1675/n24) |
| C-31 | "a CI gate that fails any PR…" → **advisory budget report**. The true half kept and named: `P95_TOLERANCE = 1.10` | `.github/workflows/bench.yml:18` verbatim "advisory (not in required-status-checks)"; `grep -ic bench scripts/qc-allowlists/required-contexts-release.txt` → **0**; `src/bench.rs:77` |
| C-45 | `*[advisory]*` marker **restored** on `memory_session_start`; section header `(v0.6.4)` dropped; `MACOS_BUDGET_MULT = 3.0` disclosed where budgets are published | no `Operation` variant in `src/bench.rs`; `PERFORMANCE.md:29`; `src/bench.rs:95` |
| C-21 | "≥30% AGE-over-CTE speedup, bench-gated in CI" → deleted; replaced with the fact that `benches/age_vs_cte.rs` self-skips without `AI_MEMORY_TEST_AGE_URL` and **no workflow runs it** | `grep -rni age .github/workflows/bench.yml` → 0; `grep -rn age_vs_cte .github/` → 0 |
| C-13 | **50 MB → 2 MiB**, noted as a **root-level layer** so it covers bulk/import/sync | `src/lib.rs:87`, applied `:1317` |
| C-63 | "Deep health check" → a `Health endpoint` section: constant-time liveness + **cached** verdict (default **21600 s / 6 h**), the `pending`/`stale`/`disabled`/`ok`/`failed` table, and the **503** condition | `src/handlers/transport.rs` (`health`, `health_status_code`); `src/background/fts_integrity.rs` (`DEFAULT_INTERVAL`, `STALE_INTERVAL_MULTIPLIER = 3`, `is_unhealthy` = `Failed` only) |
| C-68 | badge + prose **92% → 90%** | `coverage/thresholds.toml:70` `min_line_coverage = 90.0` |
| C-69 | SDK pair `0.9.0` → **1.0.0 / 1.0.0** | `sdk/typescript/package.json:3`, `sdk/python/ai_memory/_version.py:29` |
| C-70 | **re-measured, not copied**: 11,900 attributes — 7,634 under `src/` (6,438 + 1,196), 4,266 under `tests/` (2,577 + 1,689) — published with the four exact `rg` commands inline | measured at this commit |

Also corrected while in the file: `?api_key=` documented as **removed at v1.0.0** (#2032 L1) rather than "deprecated at v0.7.0"; the `45 q/s` / `12 q/s` cells deleted (**no producer exists anywhere in the tree**); the unsourced "10M+ memories" Postgres figure dropped.

## Register errors found

1. **C-35's "56 of 94" is an ARM count, not a route count.** 56 is the number of `=> true` arms in `postgres_endpoint_supported()`; several arms cover multiple registrations (`("GET" | "PUT" | "DELETE", memory_id)` alone covers 3). Evaluating the gate against all 94 registrations gives **73 supported / 21 unsupported**. Published 73/94.
2. **C-70's sub-figures under-count `#[tokio::test]`** (1,003 / 1,480 vs measured 1,196 / 1,689) — the register's regex appears to require the closing `]`, which drops every `#[tokio::test(flavor = "multi_thread")]`. Those are tests.
3. **C-36's throughput figures conflict with the benchmark's own README.** The register cites `docs/DEVELOPER_GUIDE.md:1098-1101` for "1.2 q/s shipped binary / 57 q/s single-process native", while `benchmarks/longmemeval/README.md:38` attributes **~57 q/s** to `harness.py` — the CLI-subprocess, binary-faithful harness. I published the benchmark README's attribution. **`docs/DEVELOPER_GUIDE.md` and `benchmarks/longmemeval/README.md` disagree and one of them is wrong** — outside this lane.

## Not fixed here (other lanes / other surfaces)

- **`docs/benchmark.svg` still renders the retired `harness_fast.py` figures (97.0% R@5 / 232 q/s).** The SVG is another surface's to regenerate; the README now discloses in-page that the image is stale and that the table/prose are the published numbers.
- `PERFORMANCE.md`, `ROADMAP.md`, `docs/**`, `sdk/*/README.md` untouched.

## CI

Docs-only change — `paths-ignore: ['docs/**','**/*.md']` on `bench.yml` (and the equivalent short-circuits elsewhere) will skip most of the matrix. Gates run locally: `cargo fmt --all --check`, `bash scripts/check-docs-vs-ssot.sh` (**PASS**: schema=88, full_tools=103, routes=94, paths=80, cli 89/91, mem_fields=30, release=1.0.0), `bash scripts/check-hardcoded-literals.sh` (**PASS**).

## AI involvement

Authored by Claude Opus 5 under operator direction (CERT GATE 2). Every published value re-derived from source at `03bbd556`; no claim carried over on the register's authority alone.


alphaonedev:
Follow-up on the badge (C-71), and a lane boundary worth recording.

`.github/workflows/tool-count-drift.yml:144` hard-codes the literal `MCP-7_default` in its README grep:

```
grep -qE "MCP-7_default.*${EXPECTED_FULL_COUNT}_full|MCP-7.*${EXPECTED_CALLABLE_COUNT}" README.md
```

So the badge message is **structurally forced to lead with `7_default`** — a badge reading `8_advertised` reds the gate, which is what the first push did. **That workflow pins the page's core-tool framing, and it is outside this lane.**

Resolved inside the lane by making the badge spell the arithmetic in place rather than assert one side of it: **`MCP | 7 default + 1 bootstrap • 103 full`**. It satisfies the gate, and because the bootstrap is named in the badge itself it cannot be read as contradicting the footnote's 8. The page still counts **advertised entries** everywhere (core = 8, full = 103), which is the single framing C-71 asked for.

If the operator wants the badge to lead with `8`, the workflow grep has to move first — that is a one-line change in `.github/workflows/tool-count-drift.yml`, deliberately not made here.

**Process note, self-flagged:** the badge fix was landed by `git commit --amend` + `git push --force-with-lease` on this topic branch, minutes after the initial push and before any review. That is still a force-push, which this repo's policy prohibits without explicit operator authorization. No further force-pushes on this branch; any further change lands as a normal commit.

alphaonedev:
Throughput attribution corrected — pushed as `c3dfd2b5` (normal commit, no history rewrite).

The `~57 q/s` I published for the shipped binary was wrong. `benchmarks/longmemeval/README.md:38-41` shifts every harness attribution **up one row** and carries **no elapsed column**, so nothing cross-checks it. `docs/DEVELOPER_GUIDE.md:1096-1102` is the auditable source — re-verified here, every row reconstructs the 500-question corpus from its own elapsed time:

```
harness_99.py --no-expand  232 q/s x   2.2 s = 510 q
harness_99.py              142 q/s x   3.5 s = 497 q
harness_fast.py             57 q/s x   8.8 s = 502 q
harness.py                 1.2 q/s x 414   s = 497 q
```

Independent check: `harness.py` spawns a fresh `ai-memory recall` subprocess per question. 57 q/s would require a 17 ms process spawn *plus* a full recall. 1.2 q/s (~830 ms/question) is what subprocess-per-operation actually costs.

README now publishes all three as a table **with elapsed times** — the thing the source I had been using lacked — and gives the mechanism rather than a hedge:

> **The 1.2 q/s is subprocess-spawn-dominated, not recall-dominated — and no real integrator pays it.** ~830 ms per question is the cost of forking a fresh `ai-memory` process, opening the database, resolving configuration and tearing the whole thing down again, **once per query**. That is a property of the benchmark harness's shape, not of the recall path. An MCP client holds one long-lived stdio process and an HTTP client holds one daemon; both issue N queries against a process that started once, and neither pays a per-query spawn.

Readers are pointed at the hot-path p95 budgets for the latency they will actually see. `232 q/s` is labelled as what it is: a 10-core parallel Python reimplementation of FTS5 scoring that never loads the Rust ranker.

Also fixed a misattribution in the stale-SVG note — the retired 97.0% / 232 q/s figures are **shadow-harness** numbers, not `harness_fast.py` numbers.

**One more cross-doc drift, for whoever owns `docs/DEVELOPER_GUIDE.md`:** its R@5 column gives **96.2%** for both `harness.py` and `harness_fast.py`, while `benchmarks/longmemeval/results.md` publishes the binary-faithful keyword R@5 as **96.4%**. README publishes 96.4% (results.md is the disclosure document). Not reconciled here — flagging, not fixing.

Gates re-run and PASS: `cargo fmt --all --check`, `scripts/check-docs-vs-ssot.sh`, `scripts/check-hardcoded-literals.sh`.

alphaonedev:
## Fable 5 — review + audit: APPROVED

Verified independently at `c3dfd2b5`. History clean: 3 commits, all signature-verified, scope is `README.md` + `CHANGELOG.md` only.

### I verified the cross-check, not the claim

The off-by-one-row shift is **confirmed**. `docs/DEVELOPER_GUIDE.md:1096-1102` reconstructs the 500-question corpus from every row:

| harness | q/s × elapsed | = |
|---|---|---|
| `harness_99.py --no-expand` | 232 × 2.2 | 510 |
| `harness_99.py` | 142 × 3.5 | 497 |
| `harness_fast.py` | 57 × 8.8 | 502 |
| `harness.py` | 1.2 × 414 | 497 |

`benchmarks/longmemeval/README.md:38-41` assigns **57 q/s to `harness.py`** and **232 to `harness_fast.py`** — every attribution shifted up one row — and carries **no elapsed column**, so nothing in it could ever have caught the error. Sourcing the table to the version whose rows cross-check is the right call, and stating *why* it is the auditable source is better than asserting it.

`benchmarks/longmemeval/results.md:40` independently confirms keyword-baseline **R@5 = 96.4%**.

### The judgement I want to single out

Publishing **1.2 q/s** for the shipped binary is the honest number, and publishing it **with the mechanism** is what makes it fair rather than merely honest: ~830 ms/question to fork `ai-memory`, open the DB, resolve config and tear down — once per query — which is a property of the harness shape, not of the recall path, and which **no real integrator pays**, because an MCP client holds one long-lived stdio process and an HTTP client holds one daemon.

Omitting that qualifier would understate the product as badly as 232 q/s overstated it. Under a bet-the-farm standard both failures are the same failure. Getting this right in *both* directions is harder than deleting a number, and it is the part of this PR I would point at as the template for the remaining claim surfaces.

Restoring the **`autonomous` row at 95.8%** — the only tier measured *worse* than baseline, previously dropped from the tier table — is the same discipline applied where it costs something.

### Also correct

- `/health` documented as what #2579 actually ships: a constant-time liveness probe plus a **cached** verdict on a 6-hour default interval. An operator wiring it as a corruption detector was getting evidence up to six hours old with no way to see its age.
- Coverage bar 92% → **90%**, the enforced floor. README:777 already said 90, so the page contradicted itself.
- Test counts **re-measured at this commit** and published with the four `rg` commands inline, so the next auditor re-derives rather than trusts.
- `?api_key=` corrected to **removed at v1.0.0**, not "deprecated at v0.7.0".
- The `45 q/s` / `12 q/s` cells deleted — no producer exists in the tree. Deleted rather than caveated, which is the rule.

### Carried forward, not blocking

A second instance of the same shift is flagged but not fixed: `docs/DEVELOPER_GUIDE.md` gives **96.2%** R@5 for both `harness.py` and `harness_fast.py` while `results.md` publishes **96.4%** binary-faithful — and those are exactly the two rows whose q/s were mis-assigned. Worth checking as the same defect in a different column.

`docs/benchmark.svg` still renders the retired figures and is disclosed as stale in-page; regenerating it belongs to another surface.

Merging.

alphaonedev:
Benchmark-image note reworked — pushed as `3fc0747c` (normal commit).

I verified the peer branch's SVG without merging it, as instructed:

```
$ git show origin/lane-e/claims-gate-2:docs/benchmark.svg | grep -oE ...
  >96.4%<   R@5 (482/500)   >1.2 q/s<   harness.py
$ (this branch)
  >97.0%<   R@5 (485/500)   >232 q/s<
```

482/500 = 96.4% exactly and 485/500 = 97.0% exactly, so both fractions resolve rather than approximate — the substitution is exact, not rounded.

**I did not write "the image now renders 96.4%".** That assertion is false until #2652 merges, and this branch cannot know the merge order. Asserting another branch's file contents from here would just move the staleness rather than remove it. Instead the note is now a **conditional that is true under every merge order**:

> **The table and prose in this section are the published numbers.** If the image above shows **97.0% R@5** or **232 q/s**, you are looking at a stale render of `docs/benchmark.svg` — neither figure was produced by the shipped binary; both came from Python shadow harnesses. The binary-faithful keyword row is **96.4% R@5** (482/500) at **1.2 q/s** (`harness.py`).

- SVG still stale → the conditional fires, the reader is warned, and the correct values are right there.
- SVG already fixed → the conditional does not fire, and every remaining sentence is still true.

Either PR can land first, in either order, with no coordination.

I considered deleting the note, which was offered as the safer option, and rejected it: an absent note cannot go stale, but it also cannot stop a reader from trusting a headline overclaim rendered **as pixels** — which is the exact failure this note exists to prevent, and the reason taking the SVG was the right call. A conditional gets the safety without the blind spot.

`docs/benchmark.svg` is untouched on this branch; the lanes stay independent.

**Merge state re-checked** per the CONFLICTING-stops-CI warning: `git log HEAD..origin/release/v1.0.0` is empty (no base drift past the merge already in this branch) and the merge-tree dry run reports **0 conflict hunks**. PR is `MERGEABLE`/`CLEAN`. I will keep watching `mergeStateStatus`, not just `gh pr checks` — noted that a `CONFLICTING` PR stops dispatching CI and that "no checks reported" reads like absence when it actually means CI cannot run.

Gates re-run and PASS: `cargo fmt --all --check`, `scripts/check-docs-vs-ssot.sh`, `scripts/check-hardcoded-literals.sh`.

alphaonedev:
**Heads-up: this PR was merged at `c3dfd2b5` while the benchmark-image-note fix was still being written, so `3fc0747c` did not land here.** It is now **PR #2661** (branch `lane-e/claims-readme-svg-note`, base `release/v1.0.0`, head `3fc0747cedb23d359b9c0f613bbdea5acf7c9840`) — two files, two lines.

I only caught it because `gh api .../pulls/2654` kept reporting the old head with `mergeable: null` after a clean push, while `git ls-remote` showed the branch at `3fc0747c`. That reads like the GitHub-lag failure mode described earlier, but it was the opposite and more dangerous shape: **the PR was showing green checks from a head that was no longer the branch tip.** A close/reopen to force a resync is what surfaced "already merged". Worth adding to the watch-list beside `CONFLICTING` — stale-green on a merged PR looks identical to healthy-green.

The unlanded change makes the `README.md:826` note **merge-order-independent**, so it cannot go stale in either direction once `docs/benchmark.svg` is corrected in #2652. Full rationale and the SVG verification (done via `git show origin/lane-e/claims-gate-2:docs/benchmark.svg`, no merge) are in #2661.

State at the time of writing: `release/v1.0.0` is at `60e9e4b0` and its `docs/benchmark.svg` still renders `97.0%` / `485/500` / `232 q/s`; #2652 is still OPEN. So the conditional note is currently firing correctly, and will stop firing on its own the moment #2652 lands — no coordination needed either way.

Issue #1622, opened 2026-06-11
context_tokens unreachable from HTTP recall — hard-coded None in both HTTP constructors (the #1098 class, one field left behind)


alphaonedev:
## Root cause (one-liner)
`src/models/recall_request.rs:263,309` hard-code `context_tokens: None` in `from_http_query`/`from_http_body`; `RecallQuery`/`RecallBody` (src/models/memory.rs:1182-1218, 1270+) carry no such field. Permissive serde silently ignores a caller-supplied `context_tokens`. MCP + CLI honor it (70/30 query-embedding bias, recall_request.rs:145-147). Exactly the class #1098 fixed for 4 other fields and #1257 fixed for session_id.

## Trigger
`POST /api/v1/recall {"context":"x","context_tokens":["alpha","beta"]}` → ignored, no error, different ranking than the identical MCP/CLI call.

## Proposed fix (size: S)
Add `context_tokens: Option<Vec<String>>` to `RecallBody` + wire through `from_http_body`; decide CSV shape for GET or document GET-exempt. Regression test mirroring the #1098 pins.

Found by the Class-D deep-audit reviewer. Severity: MED.

alphaonedev:
Fixed via commit 27b45dc2: context_tokens wired through from_http_body (POST array field) and from_http_query (GET CSV, kinds convention). Retest: from_http_body_wires_context_tokens_1622 + from_http_query_parses_csv_context_tokens_1622. Battery b4 green. Verified per pm-v3.3.

PR #2655, opened 2026-08-01
docs(api): correct 11 false API-contract claims + delete 4 SDK methods calling unregistered routes

alphaonedev:
CERT GATE 2 — API-contract surface. Input document: `docs/audit/3x7-claims-register-2026-08-01.md` §2 + §3.2.

The published claims are the binding surface, so **every replacement below was verified against the code at HEAD before it was written**, and the register's own file:line anchors were re-derived rather than trusted (the register warns they drift, and several had).

### C-19 — `grant()` / `revoke()` / `cluster()` DELETED (BREAKING)

`grep -oE '"/[^"]*"' src/handlers/routes.rs | sort -u` returns 80 paths. `/api/v1/cluster`, `/api/v1/memories/{id}/grant` and `/api/v1/memories/{id}/revoke` are **not among them, and never were** — every call 404'd in every release that shipped the methods. There is no server-side plan for them, so this is a deletion, not a caveat: a shipped method against a nonexistent route is a claim, not a feature.

Removed from `sdk/typescript/src/client.ts`, `sdk/python/ai_memory/client.py`, `sdk/python/ai_memory/async_client.py`, and both SDK READMEs. Replacements (both already exist) are named at each removal site: `metadata.scope` + namespace governance standards for per-memory access control; out-of-band peer config + `/api/v1/sync/{push,since}` for federation.

### C-20 — `unsubscribe(id)` FIXED on all three clients

`src/lib.rs` registers `delete` on the **collection** path `handlers::routes::SUBSCRIPTIONS` only; the id rides the query string (`UnsubscribeQuery`, `src/handlers/subscriptions.rs`). The SDKs sent `DELETE /api/v1/subscriptions/{id}`, which matches no route.

This is the worst of the four because it fails in the direction that *looks* safe: teardown appears attempted, the caller moves on, and the decommissioned endpoint keeps receiving HMAC-signed deliveries indefinitely with nothing in the operator's logs.

**The register named only the TypeScript client. Both Python clients had the identical defect** — `sdk/python/ai_memory/async_client.py` is not referenced anywhere in the register — and all three are fixed here.

### R-203 regression pins

Offline on both harnesses (injected `fetch` for TS via the client's second constructor arg; `httpx.MockTransport` for Python), plus a `hasattr` guard that the three removed methods are gone from both Python client classes.

**Verified FAILING at parent `03bbd556`:**

- jest — `Expected: "/api/v1/subscriptions"` / `Received: "/api/v1/subscriptions/sub-abc-123"`
- pytest — the same path assertion, plus `AssertionError: AiMemoryClient.grant must not exist`

## Docs — `docs/API_REFERENCE.md`

| ID | Was | Is |
|---|---|---|
| C-16 | "`?api_key=` still accepted at v0.7.0 … slated for rejection at v0.8" | BREAKING CHANGE at v1.0.0 (#2032 L1) with before/after `curl`, the `http::auth` WARN named as the only diagnostic channel, and a key-rotation note |
| C-17 | "opt-in … compiled default `0` = disabled … byte-identical to a build without admission control" | ON by default, CPU-scaled `clamp(parallelism×64, 256, 4096)`; only an explicit `0` disables; typed 503 + `Retry-After: 1`; `/health` + `/metrics` exempt |
| C-18 | `signature`, `attest_level`, `signed_at` | the actual 8-column projection; `signature` withheld **by design** (`memory_verify` owns it), `signed_at` is not a column at all; integrators repointed at `POST /api/v1/links/verify` |
| C-39 | `{created, errors}` with no status contract | explicit warning that **200 does not imply any row was created** — a wholly quota-rejected batch answers `200` / `created: 0`; callers MUST inspect `created`/`errors` |
| C-41 | "`format` MCP-only; HTTP responses are always JSON" | TOON **is** supported on HTTP recall + search; only `list` genuinely lacks it. The document no longer contradicts itself |
| C-50/C-51 | 80 in prose / 78 as "authoritative" / const says 80; "returns 101" vs 103 fourteen lines earlier | 80 paths, 94 registrations, 103 full-profile tools — prose, command output and const all agree, re-derived at HEAD. 7-vs-8 core ambiguity resolved in place |
| C-64 | `{status, service}` | full body + **the only way the probe fails** (503 on `!live \|\| verdict.is_unhealthy()`), the CACHED 6 h FTS verdict, and that `pending`/`stale`/`disabled` do NOT 503 — so a 200 is not proof of recent verification |
| C-65 | 429 absent; "No per-client rate limiting at the HTTP layer" | 429 documented with its nine emitting sites; the sentence deleted and replaced with what is true (daily quotas → 429, global admission → 503, no RPS throttle) |
| C-40 | recall "**Mutates the database** (touches, auto-promotes)" | recall is **PURE** (#1953) — a stronger property — with the honest no-daemon fold caveat. Same fix in `sdk/typescript/README.md` |
| C-11 | — | **no-op for this file**: the runbook lives in `docs/federation.md:397`, held by another lane. The underlying fact is published here anyway |

### Two false claims the register missed

Found by the mandated path-literal sweep and fixed:

- `memory_recall`/`context_tokens` was published as MCP-only. It is on **both** transports — `RecallQuery` (CSV, #1622) and `RecallBody` (array). The parity note under §Recall+search asserted the same falsehood.
- The `memory_store`/`ttl_secs` note claimed "the MCP tool exposes `expires_at`". MCP `StoreRequest` (`src/mcp/tools/store/mod.rs`) exposes **neither** — `expires_at` lives on `memory_update`.

## Path-literal sweep vs `src/handlers/routes.rs`

Every path literal from both SDK READMEs and both client sources, asserted for membership in the 80-path routes set. **Exactly four miss classes, all fixed:**

| Missing path | README (py) | README (ts) | client.py | client.ts |
|---|---|---|---|---|
| `/api/v1/cluster` | ✗ | ✗ | ✗ | ✗ |
| `/api/v1/memories/{id}/grant` | ✗ | ✗ | ✗ | ✗ |
| `/api/v1/memories/{id}/revoke` | ✗ | ✗ | ✗ | ✗ |
| `/api/v1/subscriptions/{id}` | — | ✗ | ✗ | ✗ |

(`async_client.py`, outside the register's scope, carried all four.) No other miss. This is the one-off run of the mechanical check being built in a sibling lane.

## Still required (not this PR)

**#2588 — the bulk status-code fix (207 Multi-Status / 400 on total rejection) is still needed.** Per the register's §3.1 split verdict that is another lane's call; this PR documents the actual contract truthfully so a retrying client is not misled in the meantime.

## Gates

- `cargo fmt --all --check` ✅
- `scripts/check-docs-vs-ssot.sh` ✅ PASS (`schema=88, full_tools=103, core_tools=7, routes=94, paths=80, …`)
- `scripts/check-hardcoded-literals.sh` ✅ PASS
- `pytest sdk/python` — 57 passed, 4 skipped
- `tsc --noEmit` clean; `jest` — 48 passed, 7 skipped
- **Zero `.rs` bytes in the diff** (`git diff --name-only | grep '\.rs$'` → empty), so the four clippy legs are identical to `origin/release/v1.0.0` by construction; CI runs them regardless.

## AI involvement

Authored by Claude Opus 5 under operator direction, per `docs/AI_DEVELOPER_WORKFLOW.md` §8.2. Every claim replaced was verified against the cited source at HEAD; register anchors were re-derived, not trusted.


alphaonedev:

**Control:** Gate 2 claims train item 1 — correct published API-contract falsehoods + remove SDK methods that 404'd on unregistered routes.

### Code
- Deletes `grant`/`revoke`/`cluster` from TS + Python clients (no routes ever existed — honest API surface).
- Fixes `unsubscribe` to `DELETE /api/v1/subscriptions?id=` matching `UnsubscribeQuery` registration (closes silent webhook-teardown failure).
- `docs/API_REFERENCE.md` corrections for C-16..C-65 class claims (admission default, bulk 200≠success, TOON coverage, health semantics, 429, recall purity).
- R-203 pins in SDK tests. Zero `.rs` in substantive commit.
- Rebased to `a9b77b24` (post Gate1 #2685) via signed merge `f35e8980`.

### Security
- Positive: removes false grant/revoke surface that could mislead integrators into thinking ACL endpoints exist.
- unsubscribe fix is security-relevant (decommissioned endpoints kept receiving signed deliveries).
- Residual claims: #2656 security prose, #2668 errata, #2659 CERT GATE 2 recurrence lock — merge LAST.

### Merge bar
- Await green required checks on tip; merge commit only; one CHANGELOG stream — do not land #2656 until #2655 lands.

alphaonedev:
# Code + security review — PR #2655 (claims train #1 / CERT GATE 2 API surface)

**Control:** published API/SDK claims must match registered routes and daemon behavior (no methods that 404; no false auth/admission/quota/recall contracts).

**SHA:** f35e898 (post-#2685 rebase)
**Files:** CHANGELOG, API_REFERENCE, SDK py/ts clients + READMEs + tests. **Zero .rs bytes.**

## Code review
1. **C-19** grant/revoke/cluster deleted from TS + both Python clients — routes never registered in routes.rs (spot-check tip).
2. **C-20** unsubscribe uses collection DELETE + id query (matches SUBSCRIPTIONS registration).
3. Regression pins (jest/pytest) assert path + hasattr absence.
4. API_REFERENCE corrections C-16..C-65 / C-40 etc. are documentation of existing daemon posture (admission default-on, 429 quotas, pure recall, TOON, etc.) — not new control code.
5. CHANGELOG documents BREAKING SDK deletions + unsubscribe fix honestly.

## Security of the control
| Risk | Disposition |
| Shipping SDK methods that 404 (false capability) | **Removed** grant/revoke/cluster |
| unsubscribe wrong path → continued HMAC deliveries | **Fixed** collection path |
| Docs claiming weak auth (`?api_key=`) still live | **Documented removed** at v1.0.0 |
| Docs claiming admission opt-in when default-on | **Corrected** |
| Docs claiming no 429 | **Corrected** with emit sites |

**Verdict:** APPROVE merge when required checks CLEAN. First car of claims train (#2655 → #2656 → #2668 → #2659 last).

Issue #1993, opened 2026-07-12
[ci][infra] 'Code Coverage' PR job times out on Generate step (~22-23m) — heavy full llvm-cov exceeds job wall limit


alphaonedev:
**Symptom.** The PR-only `Code Coverage` job (`.github/workflows/ci.yml`, posts the sticky coverage PR comment) fails on **every** substantial PR at ~22-23 min with the **`Generate coverage report`** step still `running` (never completes) and the job concluding `failure`. Observed on PR #1992 twice: run 29208193107 job 86691235243 (23m5s) and run 29209743986 job 86695354733 (22m5s). No step reports its own failure — the job is killed at its wall-clock/resource limit mid-generation.

**Root cause.** The full instrumented `cargo llvm-cov --features sal,sal-postgres --lib --tests --workspace` build+run is heavy (the same reason it's ~130G-of-target / disk-prohibitive locally). It exceeds this job's time (and possibly memory) budget. The coverage DATA itself is fine — the sibling `coverage.yml` `Per-Module Coverage Thresholds` job DOES complete (~34m) and reports GLOBAL 92.90% >= 90% + all per-module floors.

**Impact.** Non-required check (branch protection requires only the 3 `Check (os)` jobs), but it is a persistently-RED CI job that (a) violates the 'every CI workflow green' standard and (b) never posts its coverage comment. It is redundant with `coverage.yml`.

**Proposed fix (pick one, ~small).**
1. Raise the `Code Coverage` job `timeout-minutes` (and/or runner size) so Generate completes; OR
2. Split coverage generation from the floor/comment steps so a slow generate doesn't kill the whole job; OR
3. Drop the redundant `ci.yml` `Code Coverage` job and let `coverage.yml` own coverage (it already computes global + per-module + is the authoritative gate), moving the PR-comment post into `coverage.yml`.

Recommend (3) — deduplicate. Auto-filed-by-agent during PR #1992 base-fix. Verified per prime directive pm-v3 (memory cd8ede94).

alphaonedev:
Resolved by PR #2058 (merged to release/v1.0.0). Removed the redundant second full llvm-cov 'Code Coverage' job from ci.yml; the Per-Module Coverage Thresholds gate (coverage.yml) already enforces the global 90% floor + every per-module floor (a superset) and now also posts the sticky coverage comment. No floor lowered.

alphaonedev:
[...]

PR #2656, opened 2026-08-01
docs(security): correct four false security claims + reconcile the H1 contradiction (C-02/C-04/C-29/C-38/C-47/C-48/C-55)

alphaonedev:
## What this is

CERT GATE 2, claims-correction lane E — the **security and compliance document surface**, the pages a Fortune 500 procurement reviewer reads. Documentation only; no code change. Every replacement claim was verified against `release/v1.0.0` HEAD **before** it was written, because a wrong replacement is worse than the original: it carries an audit's authority.

Input: `docs/audit/3x7-claims-register-2026-08-01.md`, items **C-02, C-04, C-29, C-38, C-47, C-48, C-55** plus §3.2 C ("security and compliance properties that are not implemented").

The recurring shape being corrected: **the code states its own limits accurately while a document forty lines away asserts the opposite.** The fix is usually to publish what the code already says.

### C-02 — audit-log tail truncation (TIER-1, FALSE)
`audit-trail.md` published `CHECKPOINT.sig` markers as the tail-truncation mitigation. `src/audit.rs` says the marker "is **NOT** yet implemented: no emission code exists"; its only runtime effect is a one-shot `tracing::warn!`. Verified at HEAD: three `CHECKPOINT.sig` hits in `src/audit.rs`, all comment/WARN text, **zero emission sites**.

- Mitigation row **deleted**; `CHECKPOINT.sig` marked **RESERVED** in `audit-trail.md` + `audit-schema.md`, matching the source comment.
- **Not over-corrected.** Tail truncation of the in-DB `signed_events` chain *is* detected — by the #1850 off-table forensic-watermark high-water (`TruncationCheck`) — and a same-length whole-suffix rewrite by the #1873/#2202 head-hash anchor (`HeadHashCheck`). Both are now threat-model rows **with their honest residual bounds**: `canonical_chain_bytes` excludes `prev_hash`, so an interior/mid-suffix rewrite below the anchored row and the up-to-63 un-anchored rows above the last watermark are **not** caught in-DB; `AI_MEMORY_LOG_SINK=syslog` is the residual-closing control. A false control replaced by a true one, not by a blank.

### C-38 — SOC2 CC7.2 / FedRAMP AU-12 attestation cadence
[...]

PR #2659, opened 2026-08-01
ci(claims): CERT GATE 2 — the structural fix that stops the 71 corrected claims drifting back

### LINKED ISSUE #2492: [gate-gap] check-docs-vs-ssot.sh misses API_REFERENCE.md route-count drift (says 92, SSOT is 94) — plus PR #2354 shipped 9 fixes with no CHANGELOG entry
## Summary

Two SSOT-hygiene defects found while verifying closure claims across 23 issues. Both are the same class the campaign keeps hitting: **a control that reports success while missing the thing it exists to catch.**

## 1. `check-docs-vs-ssot.sh` misses a route-count drift in the count it was built to pin

- `docs/API_REFERENCE.md:974` states **"92 production"** routes.
- `src/lib.rs:327` declares `EXPECTED_PRODUCTION_ROUTES_COUNT = 94`.
- `scripts/check-docs-vs-ssot.sh` is **green**.

The gate's stated purpose (CLAUDE.md §"Lint gates", gate 4) is: *"parses the canonical Rust SSOT consts (`CURRENT_SCHEMA_VERSION`, `EXPECTED_PRODUCTION_ROUTES_COUNT`, …), walks the operator-facing `.md` files for known narrative-count patterns, and HARD-BLOCKS when any cited value drifts from the canonical."* `EXPECTED_PRODUCTION_ROUTES_COUNT` is named explicitly. The drift is in a file the gate walks. It still passes.

Most likely cause: the narrative-count pattern set does not match the phrasing at `:974`. That is exactly the failure mode of a pattern-matching gate — it catches the phrasings someone thought of, and reports success for the rest. The gate has a `--self-test` that stages a contrived stale claim; **that self-test evidently uses a phrasing the gate recognises**, so it proves the matcher works on its own example rather than on the corpus.

**Proposed fix:** invert the direction. Rather than hunting known phrasings, extract every integer adjacent to a pinned concept in the operator-facing docs and require each to match the SSOT or be explicitly allowlisted. Fail on *unrecognised* count-shaped claims near a pinned noun instead of passing on them. Then extend `--self-test` to plant the drift using the **actual phrasing found at `API_REFERENCE.md:974`**, so the self-test proves coverage of the real corpus, not of its own fixture.

## 2. PR #2354 claimed nine fixes and shipped no CHANGELOG entry

Issues **#2334 – #2339** (six of the nine) have **no CHANGELOG entry** despite the PR asserting the fixes landed. Several are user-visible: #2334 added `Stats.live` + `expired_pending_gc` to a public response shape; #2331 changed expiry semantics on tier promotion; #2339 changed the priority ladder's ceiling and added a decay arm.

Per the project's own commit policy, a user-facing change carries a CHANGELOG entry. Nothing blocked this PR for omitting them, which means the CHANGELOG's completeness rests entirely on author diligence.

Note this compounds #2483: `check-docs-vs-ssot.sh` **explicitly excludes** `CHANGELOG.md`, and the `changelog.d/` fragment convention has begun (one fragment, `2452.md`) with **no assembly script** — so the GA-tag assembly obligation is live and unmechanised. A CHANGELOG that is neither gated nor assembled is not a release artifact.

**Proposed fix:** land the `changelog.d/` assembly script, then gate on it — a PR touching `src/` with a user-visible change and no fragment fails. Backfill the six missing entries for #2334–#2339 (and #2331) in the same pass.

## Why file this rather than just fixing it

The route-count drift is one line. But a docs-vs-SSOT gate that misses a docs-vs-SSOT drift is a **gate correctness** defect, and the pattern-matching design will keep producing false greens on phrasings nobody enumerated. That deserves the design change, not just the one-line doc edit — otherwise the next drift lands the same way.

Both items should be settled before the GA cut, since the release narrative will cite these docs.

🤖 Filed by Claude Opus 5 during closure verification of 23 audit claims

### LINKED ISSUE #2629: docs-vs-SSOT gate pins values but not symbols: prose naming migrate_v87()/functions/paths can go stale silently
## Summary

`scripts/check-docs-vs-ssot.sh` pins canonical **values** and greps operator-facing docs for
patterns asserting those **numbers**. Prose that names a **symbol** — `migrate_v87()`, a type,
a function, a path — matches no pinned pattern and is **completely unguarded**.

So any documentation sentence naming a code symbol can go stale silently, indefinitely, with
every gate green.

## Concrete instance

`CLAUDE.md` claimed:

> postgres ladder ends at `migrate_v87()`

This went stale the moment the v88 arm landed. The *neighbouring* clause — "Current schema =
v88" — was correctly updated, because that one asserts a **number** the gate pins. The
sub-clause naming the tip **function** in the same paragraph was not, and nothing caught it.

Found by re-reading during a base merge, not by any gate.

## Why this is a real class, not a one-off

The docs-vs-SSOT gate exists precisely so operator-facing prose cannot drift from code. It
does that job for counts (schema version, route count, CLI subcommand count, tool count,
field count). But a large share of this repo's most load-bearing documentation names symbols
rather than counts — migration arms, trait methods, env-var resolver functions, gate script
paths. That entire surface has no guard.

A doc that names the wrong function is arguably worse than one citing a stale count: a reader
grepping for `migrate_v87` finds it and concludes the ladder is current.

## Proposed fix (needs no new SSOT)

The postgres ladder tip is **already derivable** — `scripts/check-migration-ladder.sh` computes
exactly that value for its rule (d) cross-adapter agreement check. Teaching the docs gate to
resolve a `migrate_v<N>()` mention against it is a cross-reference between two gates that both
already exist, not new machinery.

Suggested scope for a first pass, ordered by value-per-line:

1. `migrate_v<N>()` mentions → resolve against the ladder gate's computed tip.
2. Referenced script paths under `scripts/` → assert the file exists (catches renames).
3. Referenced `src/` paths in the module table → assert the file exists.

Each is a existence-or-equality check against a value the repo already computes.

## Related

- #2627 — the sibling discovery that a ceiling number is uninterpretable without its tree.
  Same underlying theme: the gate pins a value but not the value's provenance or referent.

Found during the v1.0.0 3x3 performance audit, by the #2585 lane during its base merge.

alphaonedev:
## What this is

The **structural fix** the 3x7 claims register calls for in §3.3
(`docs/audit/3x7-claims-register-2026-08-01.md`). A 265-claim audit found
**71 FALSE / OVERCLAIMED published claims** while
`scripts/check-docs-vs-ssot.sh` was **GREEN** — README.md alone carried
**five stale SSOT values** (94→92/93 routes, 88→78 schema, 30→28 `Memory`
fields, 103→101 tools, 91/89→89/87 CLI subcommands).

README is, and always was, inside that gate's `DOC_FILES`. **The gap was
never the file walk — it was the PATTERN SET.** Every original rule is a
regex pinned to one exact sentence shape, so a document that says the same
thing the seventh way is invisible. Per the register: *"a gate that greens
a page carrying five stale SSOT values is the #2444 'reports success while
doing nothing' shape"*, and *"the drift direction is consistently toward
MORE CLAIMED ENFORCEMENT THAN EXISTS. That is not random staleness; it is a
systematic bias that a gate must be built to counter."*

Four closures. All HARD-BLOCK. All wired **unconditionally** into
`c8-precheck.yml` — no `needs: classify`, no job-level `if:`, no `paths:`
filter, because a gate policing the docs-only short-circuit must not be
subject to it (gate 7 rules (b1)/(b4)/(c)/(d)/(e)).
`bash scripts/check-required-contexts.sh` was re-run against the edited
workflow and passes.

## Closure 1 — #2492: the pattern set missed values it nominally covered

`check_narrative_count_rule` fired only on hand-written regexes. It now
also runs a **generalised numeric-claim scanner**: for each SSOT const, a
small set of NOUN-PHRASE anchors (`HTTP route registrations`,
`unique URL paths`, `unique paths`, ``MCP tools at `--profile full` ``,
`-entry surface`, `CLI subcommands`, ``-field `Memory` ``, `schema **v`)
and ANY adjacent integer in bold / code / plain form.

**The historical guard is load-bearing and is preserved.** README carries
prior-release narrative paragraphs and ROADMAP §11.3.1 carries a
self-correcting frozen v0.7.1 baseline; those numbers are TRUE statements
about a PAST release, and re-pointing them would falsify the record — the
same reasoning that keeps CHANGELOG.md, the RFC files and the three frozen
v0.7 migration guides out of the walk entirely. **Rule N1** closes the hole
that guard would otherwise open: a paragraph labelled `— current release`
MUST attribute the `Cargo.toml` version, which is what catches the one
README paragraph carrying four of the five stale shapes.

**R-203, mechanically.** The pre-fix script is frozen **verbatim** at
`scripts/test/fixtures/docs-vs-ssot-prefix-2492.sh` (the
`ci-classify-prefix-2496.sh` / `required-contexts-prefix-2494.txt`
precedent). `--self-test` plants the exact pre-fix README phrasings and
asserts **both** directions:

```
PASS: self-test R-203 — frozen pre-fix gate ACCEPTS all five stale README shapes (the #2492 defect, reproduced)
PASS: self-test — widened gate REJECTS all 11 planted claims across the five register shapes
PASS: self-test — historical release-narrative / frozen-baseline / ladder mentions still PASS
```

A self-test that only proved the new gate works would be tautological — the
finding was precisely that the old gate greened them.

## Closure 2 — #2629: the gate pinned VALUES, not SYMBOLS

New **gate 8**, `scripts/check-doc-symbol-anchors.sh`. The audit sampled
six anchors and found **6/6 MISS at HEAD**, including `decorate_memory` —
gone since the recall decorator became `decorate_memory_many`. Four rules,
all on PATH-QUALIFIED grammar: **PATH** (a cited `src/<p>.rs` must exist),
**LINE** (a `src/<p>.rs:<N>` anchor must name a line the file has),
**QUAL** (every identifier in `src/<p>.rs::<sym>` / `::{a, B::c, d}` must
be defined in that file), **MDLINK** (a ``[`sym`](../src/<p>.rs)`` link
must resolve).

A bare backticked identifier sharing a line with a `src/` path is
**deliberately not checked**: measured against the tree, that grammar
yields **1,827 hits over 879 distinct tokens** (MCP tool names, DB columns,
env vars), and a gate nobody can leave on is worse than no gate.

**No new SSOT** (operator direction): the migration-ladder tip is
**extracted and reused** from
`scripts/check-migration-ladder.sh::read_current_schema_version` rather
than derived a third time — and that immediately caught
`docs/postgres-age-guide.md` naming a ten-versions-stale `migrate_vNN()` as
the ladder end, which is #2629's own title example.

## Closure 3 — SDK path literals vs `src/handlers/routes.rs`

New **gate 9**, `scripts/check-sdk-route-paths.sh` (register §3.3.2).

- **C-19** — `grant()` / `revoke()` / `cluster()` shipped in **both** SDKs
  against three paths with **zero hits** in `routes.rs`: three typed,
  documented methods that **404 at runtime**.
- **C-20** — TS `unsubscribe(id)` targets `DELETE /api/v1/subscriptions/:id`
  while `src/lib.rs` registers delete on the **collection** path only and
  the id rides the query string.

Path **parameters are normalised on both sides** (`{id}`, `:id`,
`${encodeURIComponent(id)}`, `<id>` → `{}`). That step is what makes the
rule catch C-20 at all: a raw-string membership test passes it (the
collection path *is* registered) and a parameter-blind test passes it too.
Near-miss controls in the self-test must PASS — a correctly-templated
member path in each SDK dialect, a collection call with a **query string**
(the shape C-20's fix must adopt), and the bare `/api/v1/` prefix.

## Closure 4 — named CI jobs must exist, and enforcement claims must be true

New **gate 10**, `scripts/check-ci-job-claims.sh` (register §3.3.3).

**EXISTENCE** catches **C-24** (`ci.yml`'s "Code Coverage" job, removed in
#1993, still cited as the live coverage gate in README + ROADMAP) and
**C-23** (the `postgres-age` nightly, deleted in `da3fb9cc`, still cited as
gating).

**ENFORCEMENT-TRUTHFULNESS** is the rule that counters the register's
systematic bias, and the only one that catches **C-31**: `bench.yml`
exists, its job exists, and the claim that it "gates every PR and trunk
push" is still false — `bench.yml:18` says of itself *"Bench is advisory
(not in required-status-checks)"*. Two calibrations aim it at the defect
rather than its neighbourhood: the enforcement verb must sit within **90
characters** of the citation, and `workflow_dispatch`-ONLY workflows are
exempt from enforcement (never from existence) because demanding
`release.yml` be a required PR context is incoherent — a `schedule:`-only
nightly is **not** exempt.

## Ledgers: two dispositions, deliberately different

| Ledger | Stale entry | Why |
|---|---|---|
| `doc-numeric-claims-pending.txt` (gate 4) | **NOTICE** | five doc lanes are correcting these concurrently |
| `sdk-route-paths-pending.txt` (gate 9) | **NOTICE** | `lane-e/claims-api` is deleting/fixing these now |
| `ci-job-claims-pending.txt` (gate 10) | **NOTICE** | same race |
| `doc-symbol-anchors-allow.txt` (gate 8) | **FAILS** | nothing is concurrently fixing these; a stale entry is pure rot |

The NOTICE disposition is the `dual-trigger-cancel-allow.txt` precedent
verbatim: a stale entry can only suppress a failure that no longer happens,
and failing on it would red whichever PR lost the race. The FAILS
disposition is the `required-contexts-joblevel-if-allow.txt` discipline. In
**all four**, a MALFORMED entry HARD-FAILS, so no ledger can rot into prose.

## MERGE ORDERING — this PR must land LAST

The ledgers hold every claim the new rules reject at `03bbd556`. As each
document-correction lane merges, its entries turn into NOTICEs and should
be deleted in that same merge. **Two entries sit in lane-owned files under
the stale-FAILS gate 8 allowlist** and must be deleted in the same merge
that repoints them: `PERFORMANCE.md src/mcp.rs` and
`docs/security/build-script-vetting.md src/inner.rs`.

## Verification

- `cargo fmt --all --check`; `cargo check --examples`
- clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` on
  **all four legs** — default / `sal` / `sal-postgres` / `vectorlite`: 0 errors each
- `cargo test --test qual_10_module_size_ceiling` +
  `qual_6_7_legacy_error_type_ceiling` — pass, no ceiling bump needed
- `( umask 022; cargo test --test doc_claims_integrity )` — 4/4 pass; **fails
  2/4 with the ledgers emptied** (R-203)
- All four new/widened gates + their `--self-test`s pass; so do
  `check-required-contexts.sh` (incl. its own self-test),
  `check-migration-ladder.sh`, `check-vendor-literals.sh`,
  `check-hardcoded-literals.sh`, `check-l3-boundary.sh`

## AI involvement

Authored by Claude Opus 5 under operator direction, per
`docs/AI_DEVELOPER_WORKFLOW.md` §8.2.

Closes #2629
Closes #2492


alphaonedev:
# Code + security review — PR #2659 LAST (CERT GATE 2 structural claims integrity)

**Control:** CI must fail when published docs/SDK claims drift from SSOT (prevents the 71 false claims from returning). This is the structural fix that must land **LAST** after claim-content PRs #2655→#2656→#2668.

**SHA:** 03c7274d (re-synced onto a2bc9066)

## Code review
- Expands `check-docs-vs-ssot` pattern set + new gates: doc-symbol-anchors, sdk-route-paths, ci-job-claims.
- Wiring via `.github/workflows/c8-precheck.yml` + allowlists for pending debt.
- `tests/doc_claims_integrity.rs` pins control behavior.
- CHANGELOG keep-both with prior train cars.

## Security of the control
| Risk | Disposition |
| Gate greens while docs lie (reports success while doing nothing) | Pattern set + new scripts close the #2444-shaped gap |
| Pending allowlists hide forever | Explicit qc-allowlists; ratchet not silent opt-out of whole gate |
| Order violation (#2659 before content cars) | Enforced by process: #2655/#2656/#2668 already merged |

**Verdict:** APPROVE when required checks CLEAN. Completes Gate2 claims train.

alphaonedev:
## Code + security review (AI NHI, pre-merge) — CERT GATE 2 LAST

**Control:** Structural recurrence lock so the 71 corrected claims cannot drift back. Hard-block gates in `c8-precheck.yml` (unconditional — not subject to docs-only short-circuit).

### Code
- Expanded `check-docs-vs-ssot.sh` pattern set (#2492 generalized numeric scanner + N1 current-release attribution).
- New: `check-doc-symbol-anchors.sh`, `check-sdk-route-paths.sh`, `check-ci-job-claims.sh` + allowlists + fixtures + `tests/doc_claims_integrity.rs`.
- Required-contexts allowlist updated for three new contexts.
- Rebased after claims train #2655→#2656→#2668 (tip base `a2bc9066`).

### Security / cert integrity
- Closes the “gate greens while README carries five stale SSOT values” false-success class.
- This is the LAST car on the claims train — correct merge order.

### Merge bar
- All required checks green including new #2629 contexts; merge commit only.

Issue #2341, opened 2026-07-23
federation W-of-N ack treats any HTTP 2xx as applied: post_once never reads the receiver's applied/skipped/unsupported_on_postgres report — quorum met + push-DLQ rows retired for writes the peer never applied


alphaonedev:
**Severity:** high. Findings Grok W1A2-01 + W1A2-02 (one issue — same sender-side ack-classifier root cause).

**Root cause (one-liner):** `src/federation/sync.rs::post_once` classifies ANY `status.is_success()` response as `AckOutcome::Ack` (unless a test-only `ids` array disagrees) and never parses the receiver's structured report (`applied` / `skipped` / `unsupported_on_postgres`), so W-of-N quorum counts peers that provably did NOT apply the write, and the push-DLQ replayer permanently retires (`mark_dlq_row_replayed`) rows the peer never landed.

**Evidence (at release/v1.0.0 HEAD 808b9740):**
- Sender: `src/federation/sync.rs:183-201` — 2xx ⇒ `Ack`; the only non-Ack success leg is the `ids` drift check, and no production receiver ever returns `ids` (only test mocks in `src/federation/mod.rs`). The envelope fields `skipped` / `unsupported_on_postgres` are referenced NOWHERE on the sender side (repo-wide).
- Postgres receiver: `src/handlers/federation_signing_check.rs:619` counts archives+restores+pendings+pending_decisions+namespace_meta+namespace_meta_clears+checkpoints into `unsupported_on_postgres` WITHOUT applying (FED-RQ-01 honest-count posture) and returns `StatusCode::OK` (envelope at :814-:823).
- SQLite receiver: `src/handlers/federation_receive.rs:1296-1299` (validate failure), :1413 (attestation refusal), #1934 namespace-scope refusals etc. all land `skipped += 1` inside a 200 envelope (:2428-2450). Idempotent no-ops correctly land `noop`, NOT `skipped`, so `skipped > 0` is a genuine non-apply signal.
- Consequence: `broadcast_archive_quorum` / `broadcast_restore_quorum` / `broadcast_pending_quorum` / `broadcast_checkpoint_resolution_quorum` (+ namespace_meta / decisions fanouts) record a peer ack on the false Ack — quorum "met" while a pg peer applied nothing; `src/federation/push_dlq.rs` replay marks the row replayed (permanent retirement) on the same false Ack. Silent replication loss with a success verdict in a heterogeneous mesh.
[...]

Issue #1821, opened 2026-06-28
🎯 Session orchestration — TRACT design + v0.8.0 assessment + corpus-completeness + v0.8.1 carve-out (2026-06-28)


alphaonedev:
# 🎯 Session Orchestration — TRACT design → v0.8.0 assessment → corpus-completeness → v0.8.1 carve-out

**Date:** 2026-06-28 · **Operator:** @alphaonedev · **Agent:** Claude Code · Opus 4.8 (1M) · **Branches:** `release/v0.8.0` (work) + `main` (mirror, signed)

This is the global orchestration record for a full day of adversarial design + assessment + remediation work. All commits are **SSH-signed** (`id_ed25519`, Verified). Predecessor tracker: #1820 (Red Queen / RQGM — now closed; its decision is canonical in ROADMAP §25).

---

### A. Definitive endpoint-memory design — **TRACT**
Multiple 3×7 (21-agent) adversarial councils designed, from first principles (no reference to ai-memory), the "absolute perfect endpoint AI memory" across the AI↔AGI↔ASI spectrum, 100% OSS:
- Two independent clean-slate 21-agent designs (Opus + Grok) → a 21-agent head-to-head → the definitive **[`TRACT-the-definitive-endpoint-ai-memory.md`](https://github.com/alphaonedev/ai-memory-mcp/blob/release/v0.8.0/docs/design/TRACT-the-definitive-endpoint-ai-memory.md)** (Tamper-evident Record of Attested Claims, Tiered). The council renamed itself away from "APEX∞ / Eternal Ledger" as a violation of its own banned-grandeur discipline.

### B. TRACT vs ai-memory v0.8.0 — **four assessment passes + adjudication**
- Grok first-party + Opus first-party 21-agent councils → two reconciliations → a final **21-agent adjudication council**, CodeGraph-anchored (846 files / 27,062 nodes).
- Reduced to **two canonical deliverables**: **[CORRECT-NOW](https://github.com/alphaonedev/ai-memory-mcp/blob/release/v0.8.0/docs/design/TRACT-v0.8.0-CORRECT-NOW-CANONICAL-opus.md)** (10 of 14 pillars CORRECT) + **[DEVELOPMENT-GAPS](https://github.com/alphaonedev/ai-memory-mcp/blob/release/v0.8.0/docs/design/TRACT-v0.8.0-DEVELOPMENT-GAPS-CANONICAL-opus.md)**.
[...]

Issue #933, opened 2026-05-20
[Track D] Federation push has no DLQ on quorum-not-met or peer-down (no replay on recovery)


alphaonedev:
**Track D finding** (Tier 2 Tracks C/D Docker-pivot run, 2026-05-20). Filed per prime directive. Surfaced by verdict memory `2f0151a3-374b-4c86-a518-2618f02231d1` in `ai-memory/v0.7.0-nhi-testing`.

## Summary

With W=2 (quorum) and bob stopped, alice correctly returns `HTTP 503 {"error":"quorum_not_met","got":1,"needed":2,"reason":"timeout"}` to a `POST /api/v1/memories` write. However:

- `signed_events_dlq` row count stays at 0 (this table is for the deferred-audit drainer, not federation -- per `src/store/postgres_schema.sql:579-587` comment)
- `subscription_dlq` row count stays at 0 (this is for subscription webhook failures, not federation)
- There is no federation-push-DLQ table at all in the schema (grep for `federation_dlq|push_dlq|fanout_dlq` returns zero hits in `src/store/`)
- On bob's recovery, alice does NOT replay the missed push. The catchup loop pulls bob's rows but does not push alice's. Cross-recall consistency after restart only works because both daemons share the same PG store (Track B finding #925) -- a regression that masks the DLQ gap.

## Reproduction

```bash
# Alice fed-ON (W=2 over peer bob).
docker stop ai-memory-lan-parity-bob

# Write to alice
curl -s -X POST http://127.0.0.1:19180/api/v1/memories \
  -H "X-API-Key: lan-parity-test-key" -H "X-Agent-Id: probe" \
  -H "Content-Type: application/json" \
  -d '{"title":"dlq-probe","content":"x","namespace":"track-d/dlq","tier":"long"}'
# HTTP 503 {"error":"quorum_not_met","got":1,"needed":2,"reason":"timeout"}

# Inspect DLQ tables
psql -c "SELECT COUNT(*) FROM signed_events_dlq; SELECT COUNT(*) FROM subscription_dlq;"
# Both 0

# Restart bob
docker start ai-memory-lan-parity-bob

# Alice's logs after restart:
docker logs --since 5m ai-memory-lan-parity-alice | grep -iE 'replay|dlq|catchup|push'
# Only "catchup: peer peer-0 returned HTTP 401 Unauthorized" lines.
```

## Proposed fix scope

[...]

PR #2662, opened 2026-08-01
fix(federation): land a push-DLQ row for every non-acking peer on the delete lane (#2498)

### LINKED ISSUE #2498: [federation][observability] a namespace-refused or unresolvable federated deletion is invisible to the sender — broadcast_delete_quorum never enqueues to the push DLQ
## Summary

The federated **delete** lane has no DLQ enqueue. A deletion refused by the namespace gate — or skipped because the target row's namespace was unresolvable — increments a `skipped` counter inside an **HTTP 200** and then vanishes. There is no retry, no quarantine cause on the wire, and nothing an operator can query.

Filed to give a real tracker to a residual that #2497 documents as "tracked separately." It was not, and an honesty-audit lens correctly flagged that an unverifiable tracking claim is itself the defect class this campaign exists to close. If #2497 closes #2491, this is the only remaining on-record home for the residual.

## Current state, verified

- `src/federation/push_dlq.rs:331-340` — `classify_quarantine_cause` **does** map the new `namespace_probe_unresolvable` cause token, with an in-code comment stating it is *"dormant until the delete lane gains a DLQ enqueue"*. The classifier half exists; the producer does not.
- `enqueue_push_failure` has exactly **one** caller in `src/federation/sync.rs` — line 663, the **store** lane. Deletions never enter the DLQ at all.
- `broadcast_delete_quorum` (`sync.rs:700-785`) handles an `AckOutcome::Fail` with `tracing::warn!` (`:752-756`) plus a post-quorum debug log (`:768-772`). Nothing else.
- The only retry is `post_and_classify`'s single 250 ms re-POST (`sync.rs:333-368`), which re-observes the identical deterministic policy refusal.
- `/sync/since` (`src/handlers/federation_sync_since.rs`) carries no deletion or tombstone content, so a missed deletion is never repaired by anti-entropy either.

## What the sender does and does not learn

Correcting a claim that has circulated in this campaign and is **stale**: since **#2341**, `success_report_non_ack_reason` (`src/federation/sync.rs:261-276`) inspects a 2xx report and treats `skipped > 0` as a **non-ack**, so the origin's quorum tracker records `Fail`. The sender is *not* told the erasure propagated.

What remains true, and is the actual defect:

1. **No DLQ, no retry.** A `Fail` on this lane is warn-only. A refusal caused by a fixable misconfiguration (peer scope too narrow) is never replayed after the operator fixes it.
2. **Quorum can absorb it.** `QuorumPolicy::majority` is `W = floor(N/2)+1` counting the local commit (`src/replication.rs:89-95`, `:205-212`). In a 3-node mesh, origin-local + one ack = W, so `finalise_quorum` succeeds and `DELETE /api/v1/memories/{id}` returns **200** while a third node keeps the row forever. Only in a 2-node mesh does the refusal surface as a 503.
3. **No operator surface.** `ai_memory_federation_push_dlq_quarantined_by_cause_total{cause}` cannot show a cause that is never enqueued, so the closed label set documented in CLAUDE.md env row 84 has a permanently-zero member.

## Why it matters beyond tidiness

On a GDPR-erasure or credential-rotation path, a deletion that reached 2 of 3 nodes and returned 200 is a compliance failure that is invisible by construction. And it compounds #2446: erasure does not fan out from MCP or CLI at all (`broadcast_delete_quorum` has one production caller, the HTTP DELETE handler), so the one path that *does* replicate erasure is also the one with no durable failure record.

## Proposed fix

1. Enqueue delete-lane `Fail` outcomes via `enqueue_push_failure` with a cause distinguishing **fixable** refusals (`namespace_scope_refused`, `namespace_probe_unresolvable`) from **permanent** ones, so replay after a config fix is possible and a poison-pill retry loop is not.
2. Keep the per-item `skipped` disposition on the receive side — a 409 refuse-whole-push would block co-batched in-scope `memories[]` entries and make *writes* diverge as collateral damage.
3. Surface refused ids or per-cause counts in the 200 body so the sender's sync client can annotate rather than infer. This is a wire-shape change (T1/T4) and needs its own vote.
4. Add the `namespace_probe_unresolvable` case to whatever asserts the closed cause set, so a dormant label cannot silently persist.

## Note on the sibling residual

`CAUSE_NAMESPACE_PROBE_UNRESOLVABLE`'s mapped string is currently exercised only by a fabricated test literal — no production path produces a `last_error` containing the token. The same is arguably true of the `unenrolled_author_strict` precedent it was modelled on. Worth checking both when this lands.

Refs #2488 #2491 #2446 #2497

🤖 Filed by Claude Opus 5 from a Fable 5 honesty-audit finding on PR #2497

alphaonedev:
Closes #2498

## The defect

`enqueue_push_failure` had exactly **one** production caller: the #933 "DLQ landing pass" at the bottom of `broadcast_store_quorum`. The DELETE lane — `broadcast_delete_quorum` (`src/federation/sync.rs:706`) — had **no DLQ enqueue at all**; its per-peer failure arm did nothing but `tracing::warn!`.

So a federated deletion that failed to reach a peer — because the peer was DOWN, because the peer's deadline elapsed, or because the receiver REFUSED it and reported `skipped > 0` inside an **HTTP 200** — was logged once and then vanished. Nothing retried it. The replica kept a row the origin erased, permanently, and could later LWW-resurrect it. On a GDPR-erasure or credential-rotation path that is a compliance failure invisible by construction, and under `QuorumPolicy::majority` a 3-node mesh still returns 200 to the caller while a third node keeps the row forever.

Already-correct and **not** re-litigated here: `sync::success_report_non_ack_reason` (#2341) converts a receiver 200 carrying `skipped > 0` / `unsupported_on_postgres > 0` into `AckOutcome::Fail`, so a namespace-refused deletion already reached the Fail arm. The missing piece was purely that nothing durably recorded it.

## What landed

A faithful mirror of the #933 landing pass onto the delete lane:

- `dispatched_peer_ids` snapshotted **before** the fanout.
- `explicit_failures: Vec<(String, String)>` recorded in the `Fail | Throttled` arm.
- After the tracker is finalised, every dispatched peer NOT in `tracker.acked_peer_ids()` gets `sink.enqueue_push_failure(id, peer_id, &body, &reason)` — `reason` = the explicit failure text, or `"deadline_exceeded"` for a peer whose task never reported (the bucket only the `dispatched ∖ acked` subtraction can see).
- `#[cfg(feature = "sal")]`-gated identically (the sink trait is a sal-only surface).
- Best-effort: a sink error warns under `push_dlq::PUSH_DLQ_TRACE_TARGET` and continues; it never propagates past a local erasure that already committed.

**No replay-side change was needed, and this was verified rather than assumed.** The enqueued `payload_json` is the verbatim delete body (`{sender_agent_id, memories: [], deletions: [id], dry_run: false}`). `push_dlq::replay_once` passes `row.memory_id` as both `expected_id` and the idempotency key to `post_once`, and `expected_id` is consulted **only** for the `ids` echo (`sync.rs:191-196`) — a delete response carries no `ids`, so the ack path is unaffected. Pinned behaviourally by `replay_redelivers_deletion_and_clears_row_2498`, which asserts the re-POSTed body is a correct deletion.

## Data-integrity analysis (in-code at the call site, and here)

The DLQ upsert key is `(memory_id, peer_id)` with `ON CONFLICT … DO UPDATE SET payload_json = excluded.payload_json`, so store-lane and delete-lane rows for the same memory+peer **collide, last-failure-wins**.

**1. store-then-delete — CORRECT, and the collision is a feature.** A pending STORE row is overwritten by the DELETE payload. The delete is the newer intent for that key; replaying the superseded store body would resurrect a row the origin erased; the peer converges to "absent", which is exactly the origin's state.

**2. delete-then-restore — REACHABLE. Shipped as a documented, bounded residual, not silently.**

Reachability, checked rather than assumed:
- `db::restore_archived` (`src/storage/mod.rs:12762`) restores **by the archived row's id**, verbatim — ids are reused, so a restored memory collides with the pending delete row's key.
- **No successful push clears a pending DLQ row today.** The store/restore lanes enqueue only on FAILURE, and `FederationDlqSink` has no "clear pending row for `(memory_id, peer_id)`" verb — only `mark_dlq_row_replayed(id, expected_attempt_count)`, which the replay worker owns and which needs a row id it does not have at push time.

So: delete fails to peer P → row pending; the id is restored locally and re-replicated to P; the pending delete then replays and **destroys the restored replica row**.

No cheap safe mitigation exists inside this lane, and I want to be explicit about what I rejected and why:
- **A distinct upsert key for the delete lane does NOT help** — it leaves case 2 exactly as reachable AND regresses case 1 into a transient resurrection (store row replays first by `failed_at ASC`, re-creating a deleted row on the peer before the delete row lands). Strictly worse.
- **A TTL on pending delete rows** would silently drop an erasure that has not propagated — trading a bounded loss for an unbounded compliance failure, and `FederationPushDlqRow` does not even surface `failed_at` (adding it is a public-struct change).
- **The real fix** is a supersede-on-success sink verb, which is a **public-trait change** and therefore a T1 crossroads needing its own 5-agent vote. Out of scope for #2498 — flagged for follow-up.

Bound on the residual, stated honestly: the replay drains a pending row on the first healthy peer tick (default 30 s cadence), the origin's **durable TEXT is untouched** (only a replica row is at risk), and the replica re-converges on a subsequent anti-entropy catchup. The widest window is a receiver REFUSAL that keeps the row pending indefinitely, followed by a restore, followed by an operator fixing the peer's scope. **The counterfactual it replaces is graver**: no delete DLQ at all means a permanent, silent replica divergence where a peer keeps GDPR-erased content forever and can LWW-resurrect it into the mesh.

**3. The #1821/G30 `forget_tombstones` guard does NOT already cover case 2.** Verified: `db::memory_is_tombstoned` is consulted at exactly two production sites — the federation `restores[]` receive chokepoint (`src/handlers/federation_receive.rs:2324`) and the LWW write funnel `db::insert_if_newer` (`src/storage/mod.rs:13568`). It gates inbound **WRITES / restores** against the **receiver's own** tombstones and never gates the inbound DELETE lane. In case 2 the peer never applied the delete, so it holds no tombstone for that id; and an operator-initiated `restore_archived` deliberately bypasses the tombstone gate as an authorized un-forget (documented in-code at `storage/mod.rs:12786-12793`).

## `namespace_probe_unresolvable` — verdict: STILL DORMANT

The lane lead's reading is **correct, and I verified it**. `push_dlq::classify_quarantine_cause` carried an in-code comment saying the label was "dormant until the delete lane gains a DLQ enqueue (#2498)". This PR gives the delete lane a DLQ enqueue — and the label is **still unreachable**.

The receiver emits the `namespace_probe_unresolvable` token **only as a `tracing` field**: `src/handlers/federation_receive.rs:892,2168,2494,2600` and `src/handlers/federation_signing_check.rs:671` (all five are `tracing::warn!(… cause = CAUSE_NAMESPACE_PROBE_UNRESOLVABLE …)`). The HTTP 200 body carries only the `skipped` counter, so the sender's `last_error` is #2341's `success_report_non_ack_reason` text — `"peer 2xx but N item(s) skipped (refused/not applied by receiver)"` — which can never contain the token.

Net: the row now **EXISTS** and the refusal is durably visible and retried; the specific cause **LABEL** still needs the receiver to echo the token in its 200 body, which lives in a file owned by another lane. The stale comment is corrected in place to state that real remaining precondition instead of pointing at #2498 as if this PR closed it.

## Scope survey (reported, not silently widened)

Thirteen `broadcast_*_quorum` lanes in `src/federation/sync.rs`. **Two** have a DLQ landing pass: `store` (#933) and, as of this PR, `delete` (#2498). The other **eleven still lose a per-peer failure to a single warn line** — listed for follow-up, deliberately **not** fixed here:

| Lane | file:line |
|---|---|
| `broadcast_archive_quorum` | `src/federation/sync.rs:923` |
| `broadcast_restore_quorum` | `src/federation/sync.rs:1023` |
| `broadcast_link_quorum` | `src/federation/sync.rs:1117` |
| `broadcast_consolidate_quorum` | `src/federation/sync.rs:1212` |
| `broadcast_pending_quorum` | `src/federation/sync.rs:1313` |
| `broadcast_pending_decision_quorum` | `src/federation/sync.rs:1412` |
| `broadcast_action_transition_quorum` | `src/federation/sync.rs:1566` |
| `broadcast_checkpoint_resolution_quorum` | `src/federation/sync.rs:1672` |
| `broadcast_signal_create_quorum` | `src/federation/sync.rs:1772` |
| `broadcast_namespace_meta_quorum` | `src/federation/sync.rs:1872` |
| `broadcast_namespace_meta_clear_quorum` | `src/federation/sync.rs:1975` |

Plus `bulk_catchup_push` (`src/federation/sync.rs:2101`), which is not a `*_quorum` lane but has the same no-durable-record property. Note `restore` and `namespace_meta*` are the lanes whose absence interacts with the residual above.

## R-203 evidence

Parent SHA **`03bbd556`**. `git stash push -- src/` → `( umask 022; cargo test --features sal --test federation_delete_dlq_2498 )` → `git stash pop`. **All five assertions fail at the parent**, each for the right reason (zero rows, because the lane enqueued nothing):

```
running 5 tests
test peer_200_with_skipped_leaves_pending_delete_dlq_row_2498 ... FAILED
test replay_redelivers_deletion_and_clears_row_2498 ... FAILED
test acking_peer_leaves_no_delete_dlq_row_2498 ... FAILED
test peer_down_leaves_pending_delete_dlq_row_2498 ... FAILED
test deadline_evicted_peer_leaves_pending_delete_dlq_row_2498 ... FAILED

---- peer_down_leaves_pending_delete_dlq_row_2498 stdout ----
assertion `left == right` failed: #2498: a failed federated DELETE must land
exactly one DLQ row (pre-fix the delete lane enqueued nothing at all)
  left: 0
 right: 1

---- peer_200_with_skipped_leaves_pending_delete_dlq_row_2498 stdout ----
assertion `left == right` failed: #2498: a receiver refusal reported as
`skipped` inside an HTTP 200 must ALSO land a DLQ row — pre-fix it was a warn
line and nothing else
  left: 0
 right: 1

---- deadline_evicted_peer_leaves_pending_delete_dlq_row_2498 stdout ----
assertion `left == right` failed: #2498: a deadline-evicted delete peer must
land a DLQ row
  left: 0
 right: 1

---- acking_peer_leaves_no_delete_dlq_row_2498 stdout ----
assertion `left == right` failed: #2498: exactly one row — the ACKing peer
converged and must NOT be enqueued; got []
  left: 0
 right: 1

---- replay_redelivers_deletion_and_clears_row_2498 stdout ----
assertion `left == right` failed: #2498: the failed erasure must be pending
before the replay tick
  left: 0
 right: 1

test result: FAILED. 0 passed; 5 failed
```

After the fix: `test result: ok. 5 passed; 0 failed`.

Note the ACK control (`acking_peer_leaves_no_delete_dlq_row_2498`) is deliberately written with **two** peers — one healthy, one down — asserting *exactly one* row and that it belongs to the DOWN peer. A bare "no row" assertion would have been vacuously green at the parent commit; this one is not.

## Testing

New `tests/federation_delete_dlq_2498.rs`, built on the existing `tests/federation_dlq_replay.rs` harness (mock Axum peer + `SqliteDlqSink` + `replay_once`; scratch DB under `.local-runs/` per the no-`/tmp` hard rule) rather than a new fixture:

- (a) peer DOWN → one pending row with the right `memory_id` / `peer_id` / verbatim delete payload / `attempt_count == 1`.
- (b) peer answers **200 with `skipped > 0`** (the namespace-refusal shape, the "invisible refusal" the issue is named for) → ALSO leaves a row, `last_error` carrying the #2341 non-ack reason.
- (a′) deadline-evicted peer → row with `last_error == "deadline_exceeded"`.
- (c) an acking peer leaves NO row (two-peer control, above).
- (d) `replay_once` against a recovered peer re-delivers `deletions:[id]` and clears the row, POSTing exactly once.
- `tests/federation_dlq_replay.rs` (the #933 store-lane suite) re-run unchanged: **10 passed**.

**Both backends.** The enqueue is `FederationDlqSink` **trait dispatch**, so the landing pass is backend-blind by construction — but that is asserted with evidence, not just claimed: `pg_delete_lane_dlq_parity_2498` (`sal-postgres`, `#[ignore]`) drives the **same** `broadcast_delete_quorum` code path with a `PostgresDlqSink` and asserts the identical row + payload, then clears it. Run green against the live PG 18.4 at `127.0.0.1:5433` on a **scratch database** (`ai_memory_2498_scratch`, created and dropped): `6 passed; 0 failed`. The shared tier was verified afterwards at **schema v88** with **zero residue** (`SELECT count(*) … WHERE memory_id LIKE 'del-2498-%'` → `0`).

## CI parity (run locally before push)

- `cargo check --all-targets` — default, `--features sal`, `--features sal-postgres`: green.
- Clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` on **all four legs** (default, `sal`, `sal-postgres`, `vectorlite`): green.
- `cargo fmt --all --check`, `cargo check --examples`: green.
- `cargo test --test qual_10_module_size_ceiling` (2 passed) and `--test qual_6_7_legacy_error_type_ceiling` (2 passed) — **no ceiling bump needed**.
- Shell gates: hardcoded-literal, vendor-literal, docs-vs-SSOT, L3-boundary — all PASS.
- No new MCP tool / HTTP route / CLI verb / schema version, so no SSOT pin moves. CHANGELOG `[Unreleased]` entry added.

## Crossroads protocol

No `T1`–`T6` condition fires. No public-contract change (no trait method, no route/tool/CLI verb, no wire or schema shape), no sync↔async boundary, no gate posture change, no hard-to-reverse representation, no spec deviation — and `T6` does not fire because the #933 store-lane landing pass is a **direct precedent being copied verbatim** (CLAUDE.md: "copying the precedent IS the decision"). The one genuine crossroads — closing the delete-then-restore residual via a new sink verb — is a T1 public-trait change and is explicitly deferred to its own vote rather than smuggled in here.

## AI involvement

Implemented end-to-end by Claude Opus 5 under Lane G (federation reliability), on operator authority. Codegraph used for navigation; every load-bearing claim in this description (the `expected_id` reach, the five `namespace_probe_unresolvable` emission sites, the two `memory_is_tombstoned` call sites, the id reuse in `restore_archived`, the absence of a clear-on-success verb) was verified against the source at `03bbd556`, not inferred.


alphaonedev:
# Code + security review — PR #2662 (#2498 delete-lane DLQ)

**Control:** federation delete-lane non-acking peers must land a push-DLQ row (no silent drop of delete fanout failures).

**SHA:** 758dd1c8

## Code
- push_dlq + sync path for delete lane
- Test federation_delete_dlq_2498
- CHANGELOG keep-both re-sync

## Security / data integrity
| Risk | Disposition |
| Delete fanout failure lost | DLQ row per non-acking peer |
| Replica divergence | durable queue for replay |

alphaonedev:

**Control:** Capacity federation — `broadcast_delete_quorum` lands push-DLQ rows for every non-acking peer (#2498), mirroring store-lane #933.

### Code
- Snapshot `dispatched_peer_ids` + `explicit_failures`; after quorum, enqueue `dispatched ∖ acked` via `FederationDlqSink` (`#[cfg(sal)]`).
- Payload is verbatim delete body; replay path unchanged (delete has no `ids` echo).
- Documented upsert collision: store-then-delete correct; delete-then-restore **bounded residual** (no new sink verb; out of scope).
- `namespace_probe_unresolvable` still dormant (receiver wire body) — comment corrected.
- All tip commits SSH-signed (G).

### Security / integrity
- Closes silent permanent replica divergence on failed federated erasure (GDPR/credential paths).
- Best-effort enqueue never undoes local commit.
- Residual: other 11 broadcast lanes still warn-only; delete-then-restore DLQ race documented.

### Merge bar
- Green required checks; merge commit only; close #2498 with evidence.

Issue #2383, opened 2026-07-24
[3x7-round23][N1] encrypt-at-rest upsert-merge seals content to the INCOMING writer key while keeping the existing agent_id — namespace-wide AEAD read poison


alphaonedev:
**Severity / category**: HIGH / data-integ (hard)

**Root cause**
Upsert-merge under `AI_MEMORY_ENCRYPT_AT_REST` seals the merged content to the INCOMING writer's key while keeping the EXISTING row's `agent_id` — key/identity mismatch on the durable row.

**Evidence**
- `src/storage/mod.rs:1176-1245` (upsert-merge seal path)
- `src/storage/mod.rs:959-982`
- federation twin: `insert_if_newer` `src/storage/mod.rs:12689-12763`

**Reproduction / impact**
Agent B upserts onto agent A's `(title, namespace)` row with encrypt-at-rest on: content is sealed under B's key but the row keeps A's `agent_id`. Subsequent `get`/`list`/`recall` AEAD-decrypt with A's key fail closed; since `list` collects `Result`, one poisoned row poisons namespace-wide reads. North-Star data-integrity class: mixed/incompatible state on the durable TEXT.

**Proposed fix**
Seal merged content consistently with the row's retained identity (existing writer key), or re-key + restamp `agent_id` atomically — at BOTH the upsert-merge path (`src/storage/mod.rs:1176-1245`) and the federation `insert_if_newer` twin (`:12689-12763`). Add regression tests covering cross-agent upsert + fed receive under encrypt-at-rest. Classified "hard" in triage — expect >50 LOC incl. tests.

---
Auto-filed by AI NHI per prime directive pm-v2/pm-v3 (3x7 round-2/3 net-new triage). Source: .local-runs/roadmap-review-3x7-round23-netnew.md

alphaonedev:
Fixed via PR #2421, squash-merged to `release/v1.0.0` as **e2a8017e** (GitHub-signed, `verified=true`).

**The defect.** Under `AI_MEMORY_ENCRYPT_AT_REST`, upsert-merge sealed content to the INCOMING writer's key while retaining the existing `agent_id` — so `get`/`list`/`recall` fail-closed on AEAD, and because `list` collected `Result`, ONE poisoned row denied the entire namespace. Namespace-wide read poison from a single cross-agent re-store.

[...]

Issue #1476, opened 2026-06-03
perf(federation): list_memories_updated_since OR-NULL is non-sargable + missing updated_at index → ~3× catchup latency


alphaonedev:
## Summary
The federation catchup query `list_memories_updated_since` (postgres `src/store/postgres.rs:8646`, sqlite mirror `src/storage/mod.rs:8210`) uses the non-sargable predicate `($1 IS NULL OR updated_at > $1)` and there is **no index on `updated_at`** on either backend. Both issues compound: the bind-param OR-NULL form cannot be pushed into an index range even when an index exists, and without an index the only plan is a (parallel) seq scan.

## Evidence (live PG16, 200k-row seeded probe, `EXPLAIN (GENERIC_PLAN)` — param-agnostic, matches the prepared/bind shape that actually executes)

| Variant | Plan | Predicate node | Full-scan cost |
|---|---|---|---|
| A. OR-NULL + **no** index (current prod) | Parallel Seq Scan + top-N sort | `Filter: updated_at > $1` | 2940 seq + sort |
| B. OR-NULL + index present | Index Scan | **`Filter: ($1 IS NULL) OR (updated_at > $1)`** | 7177 |
| C. **Sargable** `updated_at > $1` + index (the fix) | Index Scan | **`Index Cond: updated_at > $1`** | 2396 |

Key point: variant **B** proves the index *alone* is insufficient — the OR-NULL form degrades to an `Index Scan + Filter` (scans from the index start, filters each row) rather than a true `Index Cond` range scan. Both the index **and** the sargable rewrite are required to get the ~3× reduction (7177 → 2396) and early-stop under `LIMIT`.

## Fix
1. Schema migration **v55**: `CREATE INDEX IF NOT EXISTS idx_memories_updated_at ON memories (updated_at)` on both postgres and sqlite backends; bump `CURRENT_SCHEMA_VERSION` 54→55 in lockstep (`postgres.rs` i32 + `migrations.rs` i64), pin the prior v54 dispatch arm to a version literal.
2. Rewrite both `*_updated_since` functions to a sargable `None`/`Some` split:
   - `None`  → `SELECT … ORDER BY updated_at ASC LIMIT $1`
   - `Some(t)` → `SELECT … WHERE updated_at > $1 ORDER BY updated_at ASC LIMIT $2`
3. Regression tests (max line coverage): plan-shape / behavioral parity on both backends + migration idempotency + schema-version capability assertions.

## Backend parity
[...]

Issue #146, opened 2026-04-14
Package distribution: Homebrew, Ubuntu PPA, Fedora COPR/EPEL — v0.5.4.4 release


alphaonedev:
## Summary

All three package distribution channels are live for ai-memory v0.5.4.4 with Apache-2.0 licensing, dual-architecture support (x86_64 + ARM64).

---

## 1. Homebrew (macOS + Linux)

**Tap repo:** `alphaonedev/homebrew-tap`
**Formula:** `Formula/ai-memory.rb`
**Install:** `brew install alphaonedev/tap/ai-memory`

### Architecture Support
- `aarch64-apple-darwin` (Apple Silicon)
- `x86_64-apple-darwin` (Intel Mac)
- `aarch64-unknown-linux-gnu` (Linux ARM64)
- `x86_64-unknown-linux-gnu` (Linux x86_64)

### Update Process
1. Download all 4 platform tarballs from GitHub Release
2. Compute SHA256 hashes for each: `sha256sum ai-memory-*.tar.gz`
3. Update `Formula/ai-memory.rb`:
   - `version` field
   - `license` field
   - All 4 `sha256` values (macOS ARM, macOS Intel, Linux ARM, Linux x86)
4. Commit and push to `alphaonedev/homebrew-tap` main

### SHA256 Hashes (v0.5.4.4)
```
aarch64-apple-darwin:       e4294488f6c0196af01ecb51ff73e07f6f151a24f6e45fc51b796e5b900c317f
x86_64-apple-darwin:        351d4fc14eda2cccb24861b3f142e247603b4293ecec599ac7aa5cf04b8800b9
aarch64-unknown-linux-gnu:  7cbc91e2b7b2ed5d34814741622f8a02dc2042b2e71011b83f084312de06c3f3
x86_64-unknown-linux-gnu:   2c0abb33d7bd418c827a0742624e953723771ef5ac75385b16c1cb3821058a0a
```

---

## 2. Ubuntu PPA

**PPA:** `ppa:jbridger2021/ppa`
**Install:**
```bash
sudo add-apt-repository ppa:jbridger2021/ppa
sudo apt update
sudo apt install ai-memory
```

### Architecture Support
- `amd64` (x86_64)
- `arm64` (aarch64)

### Build Process (Manual)
1. Create working directory with pre-built binaries for both architectures
2. Create `debian/` packaging files:
   - `control` — Architecture: `amd64 arm64`
[...]

PR #2663, opened 2026-08-01
fix(#2441): advance the /sync/since cursor on rows EXAMINED, not applied

### LINKED ISSUE #2441: [federation][liveness] anti-entropy cursor can permanently stall: LIMIT applied BEFORE the per-peer namespace/visibility filters, watermark advances only on APPLIED rows
**Class:** convergence liveness / silent divergence. **Verified at `release/v1.0.0` @ `011dd8b9`, both backends.**

## The defect

`/sync/since` applies the page LIMIT in SQL, then applies the per-peer security filters **in memory, after the limit**:

```rust
// src/handlers/federation_sync_since.rs:93
let limit = q.limit.unwrap_or(500).min(app.max_page_size);

// :246 — LIMIT goes to SQL
let mems = app.store.list_memories_updated_since(q.since.as_deref(), limit).await
// :253 — per-peer namespace allowlist, AFTER the limit
let ns_filtered: Vec<Memory> = mems.into_iter().filter(|m| allowed(&m.namespace)).collect();
// :261 — #948 scope=private visibility, also AFTER
let filtered: Vec<Memory> = ns_filtered.into_iter().filter(visibility_ok).collect();
```

Both store impls are `WHERE updated_at > ? ORDER BY updated_at ASC LIMIT ?` — `src/storage/mod.rs:16486+`, `src/store/postgres.rs:17477-17497`.

The puller advances its watermark **only from rows it actually applied**:

```rust
// src/federation/receive.rs:323-325
if memories.is_empty() {
    continue;
}
```

`advance_catchup_watermark` is called only inside the applied-row arms (`src/federation/receive.rs:368`, `:403`, `:527`).

## Failure scenario

A peer scoped to one tenant's namespaces pulls against a shared corpus. A 500-row page lands whose rows are **all** filtered out by `allowed()` / `visibility_ok()`. The response carries zero rows → the puller `continue`s → the watermark does not move → **the identical window is re-requested every 30s forever**, and the visible rows behind it are never reached.

This is not slower convergence. It is a **hard liveness stop**, and it is **silent** — the log line reads `catchup: pull: peer-N ok (0 row(s) returned)` (`src/federation/receive.rs:35-37`), which is indistinguishable from "nothing new".

## Why this is a difference in KIND at fleet scale

Stall probability is **monotone in tenant count**. At 10² there is effectively one tenant, the filtered-out fraction is ~0, and the cursor always advances. Enterprise multi-tenancy is precisely the condition that drives the filtered-out fraction of a page toward 1. The documented deployment shape asks for exactly this — `docs/enterprise-deployment.md:1165-1167`: *"if subsets of peers should NOT see each other's data"*.

## Compounding: this is the ONLY multi-hop path

Push is strictly **one hop** — `broadcast_*_quorum` is invoked only from client-facing handlers, and the receive path (`src/handlers/federation_receive.rs:914`) never re-broadcasts. A non-adjacent node can therefore learn a write **only** through this catch-up pull. This is not a resilience nicety; it is the substrate's entire anti-entropy backbone.

## Related, same file, likely same fix

**Validation-failed rows are skipped past permanently.** `src/federation/receive.rs:352-358` and `:393-397` `continue` on an unparseable or `validate_memory`-failing row **without** setting `catchup_halted`, so a later row in the same page advances the watermark past it. It is never re-offered, and the `validate_memory` arm does not even log. Composed with #2384 (sqlite catch-up SELECTs 16 of ~31 columns while the pg twin does `SELECT *`), a **mixed-version fleet diverges silently and permanently** — and during a rolling upgrade of a large fleet, mixed-version is the steady state, not a transient.

## Suggested fix

Push `allowed()` into the SQL `WHERE`, **or** return a cursor = max `updated_at` **scanned** rather than max **applied**. Separately, do not advance the watermark past a row that failed validation.

## Caveat — not yet executed

Derived by reading LIMIT-before-filter against a watermark that only advances on applied rows. **A two-node test with a scoped per-peer allowlist should be written before anyone acts on the fix**, and would be worth more than the reading. Note also that federation is off by default (`--quorum-writes` defaults to 0, `src/daemon_runtime.rs:843`; `FederationConfig::build` returns `None` at `src/federation/peer.rs:69-71`), so this bites only configured-federated deployments — which is the stated enterprise target.

Found by Wave 2 of the v1.0.0 GA gap review; independently re-verified. Context: #2440, doc `docs/reviews/v1.0.0-GA-GAP-VOTE-AND-ROADMAP-CROSS-CORRELATION-OPUS-5.md`.

alphaonedev:
## The defect

`/sync/since` sends the page `LIMIT` to SQL, then runs the per-peer namespace allowlist and the `scope=private` visibility filter **in memory, after it** (`src/handlers/federation_sync_since.rs`). Every cursor-shaped value it produced — the `latest_updated_at` diagnostic **and** the server's own durable `db::sync_state_observe` clock entry — was derived from the POST-filter vector, and the puller (`daemon_runtime::sync_cycle_once`) derived its next `since` from `pulled.memories.last()`.

A 500-row window composed entirely of out-of-scope rows therefore answered `count: 0`, `latest_updated_at: null`, **HTTP 200**. The puller got no new cursor, re-requested the identical window forever, and never advanced. The stall was recorded on **both** sides, and every observable signal — status code, error rate, `scope_status` — read healthy.

Anti-entropy, the mechanism whose entire job is to converge diverged replicas, was the thing that could not converge. For the 500–1000 agent clusters this release certifies, that is a module that silently stops converging while its dashboards stay green.

## The fix — two halves, either alone is inert

- **Server**: a new **additive** `next_since` wire field carries the honest pull cursor, derived from the rows the server EXAMINED. The same value now feeds `sync_state_observe`, so a fully-filtered window advances the server's record of the peer's watermark too.
- **Client**: `sync_cycle_once` prefers `next_since` and falls back to the legacy `memories.last().updated_at` when the field is absent, so a mixed-version mesh keeps working exactly as before.

`latest_updated_at` / `earliest_updated_at` keep their existing meaning (they describe the PROJECTED page). No wire semantics are redefined; the field is purely additive. This follows the precedent already set in this handler by the S39 diagnostic-echo fields, so the crossroads protocol's "copying the precedent IS the decision" exemption applies.

## The cursor is also tie-group-safe, which `memories.last()` never was

The predicate is a strict `updated_at > since`, so advancing onto a timestamp that several rows share **SKIPS** every sharer that fell beyond the `LIMIT` — silent row loss on the replica. `next_since_watermark` therefore:

- advances to the last examined row when the page came back **short** (it cannot have been cut);
- drops the trailing tie group when the page came back **FULL**, landing on the newest strictly-older timestamp; those rows are re-examined next round, which is a free no-op (`insert_if_newer`);
- **holds** the cursor and WARNs when a FULL page is one single tie group. Stalling loudly is a DEGRADE; advancing would lose rows.

## Honest residuals (documented, not hidden)

- The postgres SAL adapter clamps its own `LIMIT` internally, so an operator who raises `AI_MEMORY_MAX_PAGE_SIZE` above that clamp can make a genuinely-cut page look short and re-open the tie hazard on that backend. That is the pre-fix behaviour of every puller, so it is not a regression — the guard simply does not extend to it.
- Both backends' catch-up SEND path skips rows whose at-rest envelope will not open (#2383, env #146), so a window composed entirely of undecryptable rows is invisible to this handler and still cannot advance. Same defect class one layer down; not addressed here.

## Both backends

The fix is applied to **both** dispatch branches in the handler — the sqlite `db::memories_updated_since` path and the `--features sal` postgres `MemoryStore::list_memories_updated_since` path. Both order `updated_at ASC` with a strict `> since` predicate (#1476 sargable split), which is what the watermark rests on.

## R-203

Parent commit `03bbd556`. `tests/federation_sync_since_watermark_2441.rs`: **5 of 7 fail at the parent, all 7 pass after.** Observed failures at the parent:

```
---- cursor_advances_when_every_row_is_out_of_scope_2441 ----
assertion `left == right` failed: #2441: a fully-filtered window MUST still
publish a cursor; got Null (count=0, excluded_for_scope=3) — the puller
re-requests this window forever
  left: None
 right: Some("2026-01-01T00:00:03+00:00")

---- pull_loop_converges_instead_of_stalling_2441 ----
assertion `left == right` failed: #2441: the cursor must reach the newest
examined row; a stalled cursor stops at None after 1 round trip(s)
  left: None
 right: Some("2026-01-01T00:00:06+00:00")

---- cursor_advances_when_every_row_is_visibility_filtered_2441 ----
  left: None
 right: Some("2026-02-01T00:00:02+00:00")
```

The two that pass at the parent are the tie-group guards — they pass there for the wrong reason (a null cursor never skips anything) and exist to block the naive `examined.last()` fix.

The file builds its **own** router because the shared `build_router_fixture` sets `AI_MEMORY_FED_SYNC_TRUST_PEER=1`, which resolves `allow_all_legacy` and makes both filters no-ops — over which the broken cursor is invisible. Same reasoning as `tests/federation_write_ns_scope_2447.rs`.

## CI parity run locally

`cargo check --all-targets`; clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` on **all four legs** (default / `sal` / `sal-postgres` / `vectorlite`) — all exit 0; `cargo fmt --all`; `cargo check --examples`; `qual_10_module_size_ceiling`; `check-docs-vs-ssot.sh` PASS.

## AI involvement

Authored by Claude Opus 5 under the Lane G federation-reliability dispatch. Design followed the existing additive-wire-field precedent in this handler rather than a new vote (crossroads-protocol precedent exemption); the tie-group safety property was added on data-integrity grounds after tracing the strict `>` cursor predicate.

Closes #2441


alphaonedev:

**Control:** Capacity federation — `/sync/since` cursor advances on rows **EXAMINED**, not only applied (#2441).

### Code
- Pull watermark no longer stalls when every row is visibility-filtered / out-of-scope.
- Tie-group hold + full-page trailing-tie drop documented in tests.
- Rebased onto `3bd01c32` (post-#2662); CHANGELOG keep-both.
- Local: `federation_sync_since_watermark_2441` **7/7 pass**.

### Security / integrity
- Closes permanent pull-stall / silent skip of future deltas when scope filters drop all rows.
- Default-deny publishes null cursor (no false progress).

### Merge bar
- Green required checks; merge commit only; close #2441 after merge.

Issue #2448, opened 2026-07-28
[SECURITY] federation outbound TLS accepts ANY server certificate by default, while federation replicates plaintext content


alphaonedev:
**Class:** confidentiality — active-adversary content disclosure on the federation wire. **Verified at `release/v1.0.0` @ `011dd8b9`.**

## The default

When `AI_MEMORY_FED_PEER_FINGERPRINTS` is unset — **the default** — the sync daemon's outbound client installs a verifier that accepts any server certificate (`src/tls.rs:1034`):

```rust
Arc::new(DangerousAnyServerVerifier)
```

The WARN emitted alongside it (once per process) states the position plainly:

> federation client TLS accepts ANY server certificate (server-cert verification is OFF); peer authenticity relies entirely on the peer fingerprint-pinning our client cert via `--mtls-allowlist`.

## Why the compensating control does not cover this direction

The documented mitigation (`src/tls.rs:1084-1099`) is that the **peer** pins **our client cert**. That protects the peer from an impostor client. It does not protect **us** from an impostor **server** — and that is the direction governing content confidentiality.

An adversary in DNS or BGP position presents any certificate, we accept it, we present our client cert, the handshake completes, and we **push plaintext memory content** to it.

## The compounding factor

Federation is not end-to-end encrypted. `src/encryption/mod.rs:11-16`, verbatim:

> NOT end-to-end across federation (#1809): federation catch-up decrypts `content` and the receiving peer re-seals under its own per-node key, so a federated peer holds plaintext transiently at apply time.

#1968 is OPEN with zero implementation; ROADMAP §11.6 rules E2E DEFERRED to v1.x. So the content on that wire is plaintext by design, and the wire's server identity is unverified by default.

## And the hardened profile does not pin it

`src/security_profile.rs:183-253` — the `asi-hard` `KNOBS` SSOT pins **14** knobs, all crypto/attestation/durability (`SECRET_SCREEN_MODE`, five `FED_REQUIRE_*_SIG`, `CID_ENFORCE`, witness/role-separation/cause-binding/rollback/lineage, `DB_SYNCHRONOUS=FULL`). It pins **zero network access-control knobs** — not `FED_PEER_FINGERPRINTS`, not TLS-required, not HTTP identity mode, not api-key-required.

An operator deploying `docs/deploy/asi-hard.env` verbatim — a file whose first line reads *"PROCUREMENT deployment ENV template"* — gets signed federation over a wire with no server authentication.

[...]

Issue #2442, opened 2026-07-27
[federation][data-integrity] DLQ routing key is a POSITIONAL peer index — decommissioning a peer misroutes queued writes to the wrong host and marks them replayed


alphaonedev:
**Class:** cross-domain content disclosure + silent unintended write loss. **Verified at `release/v1.0.0` @ `011dd8b9`.**

## The defect

Peer identity is the **enumerate index** of the `--quorum-peers` flag — deliberately positional, to keep the Prometheus label space small; the URL is dropped from the identity:

```rust
// src/federation/peer.rs:108
PeerEndpoint {
    id: format!("peer-{i}"),
    sync_push_url: format!("{trimmed}/api/v1/sync/push"),
}
```

Undelivered writes **persist durably** keyed by that id — `sink.enqueue_push_failure(&mem.id, peer_id, &body, &reason)` (`src/federation/sync.rs:467`, `:655-663`) into `federation_push_dlq(memory_id, peer_id, payload_json, ...)` (`src/storage/migrations.rs:551-560`, `src/store/postgres_schema.sql:927-936`).

Replay resolves that durable key against the **live** config:

```rust
// src/federation/push_dlq.rs:476
let Some(peer) = config.peers.iter().find(|p| p.id == row.peer_id) else {
```

then POSTs `row.payload_json` to `peer.sync_push_url` (`:490-497`).

**The DLQ is durable. The index is not.**

## Failure scenario

The code explicitly handles peer **removal** — it leaves the row pending with a warn (`:477-487`). It does **not** handle **reindexing**.

Decommission `peer-1` from a 3-peer list. Every id above it shifts down by one. `find()` now **succeeds against the wrong host**. The queued **full memory payload** — enqueued for the old peer-1 — is delivered to the peer that now occupies index 1. On that peer's ack the row is stamped `replayed_at` (`src/federation/push_dlq.rs:183`, `:516-532`), so:

1. an unintended peer receives content it was never authorized to hold, and
2. the intended peer **never receives the write** and diverges permanently, with no counter and no error.

Both halves of the north star — unintended disclosure and silent data loss — from an operation as ordinary as removing a decommissioned host from a flag.

## Why this is a difference in KIND at fleet scale

[...]

Issue #1968, opened 2026-07-09
[v1.0][11.6][F-53] Federation E2E content encryption


alphaonedev:
Real-work carrier replacing ROADMAP 25.3's stale citation of CLOSED #1809 (a LOW docs-drift fix, not a crypto epic). Peers receive plaintext at apply time; at-rest is per-node only. Per-scope envelope keys -> ciphertext-only replication (R11). 11.6 v1.0 maturity.

Filed under the v1.0.0 development epic #1940 (operator grant memory f9a0f397). Gate-item carrier from the 2026-07-09 kickoff audit (run wf_3c42dc07-761). P0/P1 substrate items are T1+T4 5-agent-vote crossroads; Rust skills (/rust-skills + /rust-microsoft) + the no-hardcoded-literals gates are mandatory on all src/ work.

alphaonedev:
## Disposition at the v1.0.0 GA freeze — docs-carrier slice FIXED; crypto epic DEFERRED to v1.x (kept OPEN)

Fresh Tier-D re-assessment. This issue is two distinct pieces, dispositioned separately:

### 1. The "real-work carrier" docs slice — ✅ FIXED

C5-verified the premise in-tree:
- **#1809 is a CLOSED `[SECURITY][LOW]` docs-drift fix** ("#228 at-rest encryption labeled 'E2E' but is per-node — federated peers receive plaintext at apply time"), **not** the crypto epic.
- **ROADMAP.md:1229** literally mis-cited `F-53/#1809 federation E2E` in the v1.0.0 line.

Fixed the stale citation → this issue (#1968) as the real-work carrier, with the honest v1.x-defer note. Commit `0a7ed906` on `feat/1875-curator-parity-disclosure`; docs-vs-SSOT drift gate **PASS**. (Prime directive: docs drift is a real defect — file AND fix.)

### 2. The E2E content-encryption crypto epic — DEFERRED to v1.x, kept OPEN

C5-verified the gap is real: no `encrypted_envelope`/ciphertext in `src/handlers/federation_receive.rs` or `src/federation/*` — **peers do receive plaintext at apply time**; per-node at-rest encryption exists (`#228` `encrypted_envelope` in `src/storage/mod.rs`).

The ask — **per-scope envelope keys → ciphertext-only replication (R11)** — is the largest new-crypto capability remaining, and it is **freeze-hostile at the GA cut**:
- **T1** — a new encrypted federation wire format on `/sync/push`.
[...]

Issue #1458, opened 2026-06-01
[SEC] Loopback-bind guard is host-string match; reverse-proxy/host-network re-exposes keyless write surface


alphaonedev:
Parent epic: #1450 · Severity: **MEDIUM** (deployment-conditional)

## Summary
With no `api_key` configured (documented default), `bootstrap_serve`
refuses to bind a non-loopback address — but the check string-matches the
configured host only. A reverse proxy, `--network=host` container, or
`socat` forward presents loopback to the daemon while exposing it
remotely, re-opening the unauthenticated write surface
(`POST /memories`, `/links`, `/agents`, `/subscriptions`, `/capture_turn`).

## Evidence
`src/daemon_runtime.rs:2770-2794` (loopback host-string check);
`src/handlers/transport.rs:747-750` (`api_key_auth` pass-through when no
key configured). CORS is restrictive and body/timeout limits are present
— this is specifically the keyless+proxy exposure.

## Fix (defense-in-depth / doc)
Require an api_key whenever the effective listener may be reachable
off-host, and/or prominently document the proxy hazard. Optionally bind
`X-Agent-Id` to the authenticated principal when an api_key is set.

alphaonedev:
Fixed in 148a77eef (pushed to release/v0.7.0). The keyless-bind check was refactored into the pure, testable api_key_bind_guard, the loopback warning now explicitly names the reverse-proxy / --network=host / socat re-exposure hazard, and a new AI_MEMORY_REQUIRE_API_KEY strict opt-in hard-refuses a keyless start on ANY bind (including loopback) for deployments fronted by a proxy. Documented in docs/CLI_REFERENCE.md. Five regression tests cover present-key, keyless-loopback-warns, keyless-non-loopback-refuses, strict-refuses-keyless-loopback, and env parsing. QC: all six gates green. (Co-located with #1455 in the same commit.)

PR #2664, opened 2026-08-01
fix(#2477): refuse plaintext federation peers on BOTH peer-URL doors

### LINKED ISSUE #2477: [SECURITY] federation peer URLs accept plaintext http:// with no flag, cert, or acknowledgement — strictly weaker than the accept-any TLS closed by #2448
**Class:** confidentiality — passive content disclosure on the federation wire. **Verified at `release/v1.0.0` @ `47939793`.** Surfaced by an adversarial-lens review during the #2448 fix (PR #2472); filed separately because the fix needs its own staged rollout.

## The gap

`#2448` / PR #2472 closes the *accept-any TLS* path: `ai-memory sync-daemon --insecure-skip-server-verify` now requires the flag **plus** `--client-cert` **plus** `--client-key` **plus** an explicit `AI_MEMORY_FED_REQUIRE_SERVER_VERIFY=0`, and is pinned shut under `asi-hard`.

**But a plaintext `http://` peer URL bypasses all of that with zero conditions.** There is no default refusal of `http://` anywhere on the sync or quorum path.

The only place the scheme is even inspected is `src/federation/sync.rs:87-91`, and it is inspected purely to feed the **opt-in** governance rule engine:

```rust
let scheme = reqwest::Url::parse(url).ok().map(|u| u.scheme().to_string()).unwrap_or_default();
let net_action = AgentAction::NetworkRequest { host, scheme };
if let Err(refusal) = governance::wire_check::check(&net_action) { … }
```

With no operator-authored governance rule — the default — that check is a no-op and the POST goes out in cleartext.

## Why this is worse than what #2448 closed

| | accept-any TLS (closed by #2448) | plaintext `http://` (this issue) |
|---|---|---|
| Conditions to reach it | 4 (flag + 2 certs + env), refused by default | **0** |
| Attacker position needed | active MITM (DNS/BGP) | **passive tap** |
| Refused under `asi-hard` | yes | **no** |

Federation is not end-to-end encrypted (`src/encryption/mod.rs:11-16`; #1968 open, ROADMAP §11.6 defers E2E to v1.x), so the memory `content` on that wire is plaintext by design. Over `http://` it is plaintext on the network too.

## Compounding: our own docs teach it

`docs/CLI_REFERENCE.md` demonstrates exactly this:

```
ai-memory sync-daemon --peers http://peer-a:9077,http://peer-b:9077 --interval 30
```

So `http://` is not a misconfiguration an operator stumbles into — it is the documented example.

## The migration hazard #2448 introduces

This is the reason to fix it in the same release train. Peers in this federation model are commonly self-signed, so under the #2448 secure default the operator menu is:

1. `--ca-cert` (safe), 2. `AI_MEMORY_FED_PEER_FINGERPRINTS` (best), 3. `AI_MEMORY_FED_REQUIRE_SERVER_VERIFY=0` (bad, now loud) — and now **4. `s/https/http/`**, which is silent, undocumented as a downgrade, and strictly worse than option 3.

Closing the accept-any path without an `https` floor can therefore *increase* plaintext exposure by making the cheapest workaround the worst one.

## Suggested fix (needs its own vote — this is NOT a drop-in of #2448's shape)

A bare fail-closed `https`-only flip would be a **synchronized, unpaced fleet-wide break** for every deployment federating over a trusted underlay (WireGuard / Tailscale / loopback / service-mesh sidecar), where plaintext `http://` to `127.0.0.1` or an encrypted overlay is a legitimate posture. That violates manageability-at-scale, so:

1. **v1.0.0 (this release):** a loud boot WARN on any non-loopback `http://` peer URL, naming the exposure and the acknowledgement knob. Mirrors the `tls_bind_guard` / `AI_MEMORY_ALLOW_PLAINTEXT_NONLOOPBACK` (#133) treatment of the *inbound* side — note the inbound plaintext bind already gets exactly this WARN, so the outbound side is the asymmetry.
2. **Acknowledgement knob** in the house grammar, e.g. `AI_MEMORY_FED_ALLOW_PLAINTEXT_PEERS`, defaulting to "warn"; loopback exempt.
3. **Pin it in `security_profile::KNOBS`** so `asi-hard` refuses plaintext peers outright — the same one-line closure #2448 used for the accept-any path.
4. **v1.1.0:** flip the non-loopback default to REFUSE, after the WARN cycle ships.
5. Fix the `docs/CLI_REFERENCE.md` example to use `https://`.

Also worth deciding in the same vote: whether `--ca-cert`'s single shared private CA should be considered adequate for a multi-peer mesh, since it gives **no per-host binding** — one compromised peer key impersonates every peer. Only `AI_MEMORY_FED_PEER_FINGERPRINTS` gives per-host identity, and it remains opt-in and unpinned by `asi-hard`.

## Note on CA validation vs. the stated threat model

Worth recording because #2448's issue text framed CA validation as the remedy: **CA validation does not stop a BGP-position adversary.** BGP hijack is precisely the ACME domain-validation threat model — the adversary mints a valid public DV cert for the peer hostname and passes `CaValidated` cleanly. Only fingerprint pinning defeats it. The #2448 fix is a real improvement (it closes the zero-work path), but `AI_MEMORY_FED_PEER_FINGERPRINTS` is the control that actually meets the stated threat model, and it is still opt-in.

Related: #2448 / PR #2472, #1968 (E2E deferred), #133/#134 (the inbound plaintext-bind guard this mirrors), #1678 (outbound pinning).

alphaonedev:
## The defect

`FederationConfig::build` (`src/federation/peer.rs`) formatted the raw operator `--quorum-peers` string straight into `PeerEndpoint::sync_push_url` with **no scheme validation whatsoever**. `src/federation/sync.rs` parses the scheme only to populate a governance `NetworkRequest` action — never to require TLS.

Federation replicates memory content that is **not** end-to-end encrypted (`src/encryption/mod.rs`; #1968 open), so `--quorum-peers http://peer:9077` shipped tenant memory in the clear to any on-path, DNS-position or BGP-position adversary. Meanwhile `docs/encryption.html` asserted unqualified that peer traffic "travels over mutual TLS … TLS 1.3 only (no fallback)".

**It was strictly weaker than a control already shipped.** #2448 built a **four-condition** opt-in ceremony for the *weaker* accept-ANY-server-cert case (`--insecure-skip-server-verify` **and** `--client-cert` **and** `--client-key` **and** an explicit falsy `AI_MEMORY_FED_REQUIRE_SERVER_VERIFY`, itself pinned by `asi-hard`). Plain `http://` needed **none** of them and disclosed strictly more. The mTLS control also cuts the other way: a peer pinning OUR client cert protects the PEER from an impostor client, not US from an impostor server.

## Enforcement completeness — both doors, one validator

`ai-memory sync-daemon --peers` (`cli::sync::build_sync_client`) is a **second, fully independent** peer-URL path: it wires the #2448 ceremony but validated no scheme either. A refusal scoped to `federation/peer.rs` alone would have been theatre, so both doors now call one shared `tls::validate_peer_url_scheme`.

Surveyed and confirmed complete: the one-shot `ai-memory sync` verb takes a local `remote_db: PathBuf`, not peers; every other `PeerEndpoint` construction site in the tree (`federation/mod.rs:478,1830,2740`, `push_dlq.rs:1158`) is `#[cfg(test)]`.

## Disposition

| peer | verdict |
|---|---|
| `https://…` | accepted |
| `http://` to a **literal** loopback host | accepted, **no hatch** |
| `http://` to anything else | **refused** unless the hatch is explicitly truthy |
| foreign scheme, or scheme-**less** `peer.example:9077` | **refused** |

Loopback is exempt unconditionally because the bytes never leave the kernel, dozens of this repo's own fixtures use `http://127.0.0.1:PORT`, and forcing a hatch-flip on every dev mesh would train reflexive hatch use — a worse long-run posture. The loopback set is now the SSOT `tls::host_is_loopback`, shared with the inbound `tls_bind_guard` (#1458/#2032 M2) so the two cannot drift.

A container-bridge hostname such as `http://alice:9077` is **not** loopback: it crosses an interceptable virtual NIC, and treating "feels local" as "is local" is the exact category error this control exists to prevent.

A scheme-less peer was previously accepted and then failed opaquely at request time with a relative-URL error; refusing at boot is strictly clearer.

## Refusal is whole-boot, never per-peer skip-and-continue

`n = 1 + peer_urls.len()` feeds `QuorumPolicy::new`, so silently dropping a peer would change the quorum guarantee without saying so — and `PeerEndpoint.id` is a positional index (#2442), so a shrunken list also re-keys every queued DLQ row above the gap.

## The hatch, and `asi-hard`

`AI_MEMORY_FED_ALLOW_PLAINTEXT_PEERS` (env #154): permissive-knob grammar, default OFF, opened **only** by an explicit truthy token — unset, empty, `0`, or an unrecognised word all KEEP the refusal (the #131/FBL-14 rule: an unrecognised token must never silently widen a security control).

**`asi-hard` pins it OFF** — the second permissive knob in that SSOT after `AI_MEMORY_ALLOW_SCHEMA_AHEAD`, and the second network access-control pin after `AI_MEMORY_FED_REQUIRE_SERVER_VERIFY`. Before this, `docs/deploy/asi-hard.env` verbatim still permitted `--quorum-peers http://peer:9077`.

The refusal message names the **secure remedies first** (`https://`, `--quorum-ca-cert`, `AI_MEMORY_FED_PEER_FINGERPRINTS`) and the escape hatch **last**, pinned by an ordering assertion — a control whose message leads with "here is how to turn me off" is a control that gets turned off.

## Operator-visible break, flagged deliberately

This repo's own `infra/plan-c` and `infra/lan-parity-test` compose fleets federate over **non-loopback** plaintext container hostnames (`http://ic-bob:19077`, routed via `entrypoint.plan-c.sh` into `--quorum-peers`). Both now carry an explicit acknowledgement line rather than being left to fail on upgrade. Any external deployment with a non-loopback `http://` peer must move it to `https://` or set the hatch.

`docs/encryption.html`'s unqualified mTLS claim is corrected in place so the shipped doc is true.

## R-203

Parent commit `03bbd556`. `tests/federation_plaintext_peer_2477.rs`: **7 of 9 fail at the parent, all 9 pass after.** Observed failures at the parent:

```
thread 'quorum_build_refuses_plaintext_non_loopback_peer_2477' panicked:
  #2477: a plaintext non-loopback peer must be REFUSED at boot; build()
  accepted it and would replicate memory content in the clear

thread 'sync_daemon_refuses_plaintext_non_loopback_peer_2477' panicked:
  #2477: the sync-daemon peer door must refuse plaintext non-loopback peers
  too — it is an independent path from FederationConfig::build

thread 'container_hostname_is_not_treated_as_loopback_2477' panicked:
  #2477: a container-bridge hostname must NOT be exempted as loopback

thread 'one_plaintext_peer_refuses_the_whole_build_2477' panicked:
  #2477: a mixed list must refuse the whole build, not silently drop the
  plaintext peer and shrink the quorum denominator

thread 'hatch_opens_only_on_an_explicit_truthy_token_2477' panicked:
  #2477: non-truthy hatch token "0" must KEEP the refusal (an unrecognised
  token must never widen a security control)

thread 'quorum_build_refuses_schemeless_and_foreign_scheme_peers_2477' panicked:
  #2477: a scheme-less peer must be refused at boot

thread 'asi_hard_pins_the_plaintext_hatch_off_2477' panicked:
  #2477: asi-hard must pin the plaintext-peer hatch
```

The two that pass at the parent are the loopback- and https-exemption guards, which exist to block an over-broad fix that would brick every dev mesh and CI fixture in this repo.

Existing lib tests that built configs from `http://peer.example` were migrated to `https://` (the URL is never dialed — those tests assert id/URL formatting and quorum policy). `config_build_rejects_duplicate_peers_differing_only_in_case` keeps its subject by using two `https://` peers.

## Both backends

The guard is backend-blind — it runs at `FederationConfig::build` / `build_sync_client`, before any store is involved. No sqlite/postgres divergence exists or is introduced.

## CI parity run locally

`cargo check --all-targets`; clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` on **all four legs** (default / `sal` / `sal-postgres` / `vectorlite`) — all exit 0; `cargo fmt --all`; `cargo check --examples`; `check-docs-vs-ssot.sh`, `check-vendor-literals.sh`, `check-hardcoded-literals.sh` all PASS; `qual_10_module_size_ceiling` + `qual_6_7_legacy_error_type_ceiling` pass; targeted `--lib federation` / `tls::` / `security_profile` / `cli::sync` and integration `cli_sync_coverage`, `fed_require_server_verify_2448`, `deploy_templates`, `tls_integration` all green.

## AI involvement

Authored by Claude Opus 5 under the Lane G federation-reliability dispatch. Disposition decided by a **3x3 adversarial vote** (9 distinct lenses — precedent, operator blast radius, threat-model literalism, test-fleet reality, docs-truth, North-Star data-integrity, reversibility, enforcement-completeness, simplicity), option D winning **6/9**, citing the 5-agent vote `4d3ea1c5`. Three amendments forced by losing lenses were adopted: the second peer-URL door (enforcement-completeness), the compose-fleet acknowledgement lines (operator blast radius), and the docs truth-fix (docs-truth).

Closes #2477


alphaonedev:
## Fable 5 — code review + security audit: APPROVED, merging

Audited the gate itself rather than the diff summary.

**`validate_peer_url_scheme` is correctly fail-closed on every branch:**

| input | disposition |
|---|---|
| unparseable URL | **Err** |
| `https://…` | Ok |
| `http://` + loopback | Ok (dev exemption) |
| `http://` + non-loopback, no hatch | **Err** |
| `http://` + non-loopback, hatch set | Ok + loud WARN naming the exposure |
| any other scheme (`file:`, `ftp:`, …) | **Err** |

The refusal message orders remedies **secure-first** — `https://`, `--quorum-ca-cert`, `AI_MEMORY_FED_PEER_FINGERPRINTS` — with the escape hatch last. That is the #2448 discipline and I checked it deliberately: a message that leads with the hatch teaches operators to disable the control.

**Both doors are gated** (the `args.peers` loop and the raw-peer parse), which is the point of the issue — a single-door fix would have been the #1989 twin-drift shape this campaign just spent a day on.

**`asi-hard` integration is right:** `hard_value: ""` with `plaintext_peers_hatch_meets_floor` returning false only for truthy values, so a hardened deployment that sets the hatch **refuses to boot**. Knob count assertion moved 16 → 17 in lockstep.

**Spoof resistance — the thing I most wanted to break.** `host_is_loopback` is exact equality across all five forms, so:

- `http://127.0.0.1.evil.com:9077` → no match → **refused** ✅
- `http://2130706433/` (decimal 127.0.0.1) → **refused** (fail-closed) ✅
- `http://127.0.0.2:9077` → **refused** (stricter than RFC loopback /8; safe direction) ✅

Every edge case I constructed fails in the safe direction.

**Reachable in the shipped artifact** — a new standing check after #2676: `src/tls.rs`, `src/cli/sync.rs` and `src/security_profile.rs` carry **zero** `cfg(feature = "sal")` gates, and the shipped v0.10.0 binary exposes `sync` / `sync-daemon`. This fix is live where it matters, not compiled out.

### One finding — not blocking, filed as a follow-up

**The exactness of `host_is_loopback` is load-bearing and untested.** The five test URLs cover the happy paths (`127.0.0.1`, `[::1]`, `localhost`, and two non-loopback refusals) but no spoof shape.

If a later refactor "tidies" that function into `host.starts_with("127.")` or `host.contains("localhost")`, then `http://127.0.0.1.evil.com` becomes an accepted plaintext peer — memory content replicated in the clear to an attacker-chosen host — and **every test in this file still passes.** The behaviour is correct today; the safety net around it is not.

I am merging rather than holding, because the fix is a genuine net improvement and blocking a correct security gate on test hardening would leave plaintext peers accepted for another cycle. The hardening follows immediately as its own issue with the specific vectors named.

Merging at `d89ddc52` — 35 green, 0 failures, CLEAN.

Issue #2665, opened 2026-08-01
[ci-integrity] a PR can display GREEN checks belonging to a superseded head — stale-green is indistinguishable from healthy-green, and no CI ran on the tip


alphaonedev:
## The trap

After a clean push, the PR API kept reporting the **previous** head with **green checks**, while `git ls-remote` showed the branch had moved. `gh pr checks` reported `23 pass` throughout, and **no workflows had dispatched for the new SHA**.

The PR was displaying green checks belonging to a head that was **no longer the branch tip**.

In the observed case the underlying cause was that the PR had already been merged, so the API was serving the last state it knew. A close/reopen to force a resync returned "already merged".

## Why this is worth a control rather than a note

**Stale-green on a superseded head is visually indistinguishable from healthy-green.** Both render as a column of ticks and both make `gh pr checks` print passing rows. There is no signal in the output that the checks describe a commit you are no longer working on.

That is materially worse than the two traps this repo already guards:

- `CONFLICTING` **suppresses** CI, so the absence of runs is at least visible as pending.
- A `cancelled` run **reads as pass** in `gh pr checks` but branch protection still refuses it, so the merge blocks and forces investigation.

Here, nothing blocks and nothing is pending. An agent or engineer polling for green gets green, concludes the work is verified, and proceeds — having verified a different commit.

## Detection

The check is cheap and exact:

```sh
# what the PR API believes
gh pr view <N> --json headRefOid --jq .headRefOid
# what the branch actually is
git ls-remote origin refs/heads/<branch>
# whether CI ever dispatched for that SHA
gh api "repos/<owner>/<repo>/actions/runs?head_sha=<sha>" -q .total_count
```

`total_count == 0` for the current tip means **no CI has run on the code you are about to merge**, regardless of how many ticks the PR shows.

## Suggested control

Any automated merge-readiness watcher — and any agent reporting a PR green — must **confirm head-sync before judging checks**:

1. resolve the PR's `headRefOid`
2. resolve the branch tip via `git ls-remote`
3. refuse to report green unless they match **and** `actions/runs?head_sha=<tip>` is non-zero

This is the same discipline as the R-203 rule and the census tests: prove the control is looking at the intended subject before trusting its verdict. A green check on the wrong commit is the [relocated-control class](https://github.com/alphaonedev/ai-memory-mcp/issues/2444) applied to CI itself.

## Related, and worth fixing together

[...]

Issue #2483, opened 2026-07-29
[campaign-throughput] every concurrent lane conflicts on CHANGELOG.md [Unreleased] — and a conflicted PR cannot run CI at all


alphaonedev:
## Summary

Three of three concurrent lanes hit the identical `CHANGELOG.md` `[Unreleased]` conflict today (#2470/#2449, #2471/#2447, #2476/#2444), each triggered simply by another lane merging first.

**The cost is not the resolution — it is that a conflicted PR cannot run CI.** GitHub cannot construct a merge commit for a conflicted PR, so `pull_request` workflows never fire and every required context is *absent*. The PR shows no failures, which reads as green and is not: it is unrun. So each lane stalls until someone resolves a one-hunk markdown conflict, and with ~100 issues queued across many lanes this serializes the entire campaign on a single file.

## Why fragments are safe here

- No gate requires `CHANGELOG.md` edits. `scripts/check-docs-vs-ssot.sh:155` **explicitly excludes** it (*"every entry is a historical … frozen-doc snapshot"*).
- The CLAUDE.md instruction to add an `[Unreleased]` entry is convention, not mechanism.
- No existing fragment tooling to conflict with (searched: `changelog.d`, `newsfragments`, towncrier — none).

## Proposed convention

One file per issue: `changelog.d/<issue-number>.md`, containing exactly the `### Fixed` / `### Added` block that would otherwise go into `[Unreleased]`.

Two lanes never touch the same path, so the conflict class disappears entirely rather than being resolved repeatedly.

## The obligation this creates — do not lose it

Fragments must be **assembled into `CHANGELOG.md` under the version heading before the GA tag is cut.** If that step is skipped, the release ships with an empty changelog — trading a throughput problem for a docs-completeness regression, which is a worse outcome than the conflicts.

That assembly belongs in Gate 3's documentation drive and should be an explicit pre-tag checklist line, not an assumption. Worth a tiny script (concatenate `changelog.d/*.md`, sort by issue number, splice under the heading, delete the fragments) so the step is mechanical rather than remembered.

## Scope note

In-flight lanes are **not** being retrofitted — their conflicts are already resolved. This applies to lanes dispatched from here on.

🤖 auto-filed-by-agent

alphaonedev:
Addressed — the per-issue `changelog.d/` fragment convention is adopted, eliminating the CHANGELOG [Unreleased] per-lane merge-conflict class.

[...]

PR #2668, opened 2026-08-01
docs(audit): §7 ERRATA — 22 errors found IN the claims register during remediation

alphaonedev:
## An audit that overclaims about claims fails its own standard

The register's binding standard is that **a correct system which overclaims fails the bet-the-farm bar.** That standard applies to the register itself. The seven CERT GATE 2 remediation lanes (#2651–#2661) were instructed to **verify every correction against the code rather than trust the register** — and doing so surfaced **22 defects in the register**.

They are recorded *in* the register rather than fixed silently, because §6.8 already warned this document would drift. This is that warning coming true, measured.

### Materiality

**None of these overturns a Tier-1 verdict. C-01 through C-09 stand.** The corrections shipped by the remediation lanes are unaffected — in every case the lane published the *verified* value, not the register's.

### The four substantive misreadings

| | Register | Verified |
|---|---|---|
| **E-01** | C-35: "**56** `=> true` arms out of 94 route registrations" | **56 is an ARM count, not a ROUTE count.** One arm can cover several routes. Real: **73 of 94** — the register understated Postgres coverage by 17 routes |
| **E-02** | C-23: AGE Cypher "could only be exercised on an operator's own machine … by design" | **False at HEAD.** `coverage.yml` sets `AI_MEMORY_TEST_AGE_URL` against an `apache/age` service container and runs those suites on every PR. True claim is narrower: **no CI job builds or version-asserts the CERTIFIED PG 18.4 + AGE 1.7.0 pins** |
| **E-11** | C-24: `.coverage-baseline` "hits prose only" | Nothing *reads* it — but **the file exists**, carrying `92.59`. Dead config, not absent |
| **E-16** | C-61: 2k fixture reachable "only via an **undocumented** env var" | It **is** documented — in the bench source. And the 43/56 µs figures **do** have a recorded producing run, so rule 2 did not apply |

### The undercounts — where the real defect was larger

- **E-03** — C-20 is a **three-client** defect. `sdk/python/ai_memory/async_client.py` **appears nowhere in the 429-line register** and carried all four defects.
- **E-04** — C-55 sampled 6 anchors; **13 checked, 13 miss**, including `tests/validate.rs` (a cited test file that **has never existed**) and one anchor the register called "pointing at nothing" that actually points at **unrelated code reading plausibly to a skimming reviewer** — worse than nothing.
- **E-05** — the HTML procurement page had **8** tally sites, not 5; two were `<meta>` tags carrying the retired "10 of 10" into **every shared link preview**. It also held an **independent instance of the C-04 false-mTLS claim** the register never located.
- **E-07** — C-47 missed that **`Export` and `Import` have zero production emitters** — a rule-2 no-producer claim sitting *inside* the enum being audited.
- Plus **E-06, E-08, E-09, E-10, E-17**.

### Errors in sources the register relied on

- **E-19** — `benchmarks/longmemeval/README.md` **shifts every harness throughput up one row.** Settled by internal consistency: every `DEVELOPER_GUIDE` row satisfies `published q/s == 500 / elapsed`; the benchmark README carries no elapsed times, so nothing cross-checked it. **Shipped-binary throughput is 1.2 q/s, not ~57.**
- **E-20** — `docs/benchmark.svg` rendered `97.0%` / `232 q/s` **as pixels** on the image the README embeds. **A claims gate that scans only text cannot see a claim that has been rendered.**

### Process findings

- **E-21** — a **`CONFLICTING` PR silently suppresses ALL CI dispatch**; `gh pr checks` reports "no checks reported," which reads as a pass. Four lanes were in this state at once. Filed as **#2665**.
- **E-22** — the parallel-lane bottleneck is **`CHANGELOG.md` `[Unreleased]` contention**, not branch-protection strictness. **#2483 is the real fix and remains unowned.**

### Gates

`scripts/check-docs-vs-ssot.sh` **PASS** · append-only to one file · no `src/` change.


alphaonedev:
# Code + security review — PR #2668 (claims train car 3 / register §7 ERRATA)

**Control:** the claims register itself must not embed false assertions that later remediations treat as ground truth (22 errata found during remediation).

**SHA:** 5ab72ca0 (clean merge onto tip after #2656)
**Files:** `docs/audit/3x7-claims-register-2026-08-01.md` only

## Review
- Docs-only register errata — honest self-correction of the audit artifact.
- No product control surface change; strengthens Gate2 integrity.
- Merge onto tip clean (no CHANGELOG contention).

**Verdict:** APPROVE when required checks CLEAN. Next: #2659 LAST.

Issue #1687, opened 2026-06-15
[HIGH] Federation catchup cursor advances past rows that failed to apply — silent permanent row loss (#1552 class)


alphaonedev:
**Source:** v0.7.1 AI-NHI audit. **Bucket:** v0.7.1 (#1683).

In `src/federation/receive.rs`, `latest_ts` is set at 268-273 **before** `apply_remote_memory` (274), and `sync_state_observe` runs gated only on `latest_ts.is_some()` (284-289) regardless of apply success — so the per-peer 'since' watermark moves past unpersisted rows → next delta excludes them → **permanent divergence** until a `since=None` re-snapshot. The SAME bug exists in **all three** catchup branches: 268-289 (SAL), 303-316 (legacy), 421-434 (legacy). ai-memory ref: MEMORY.md #1552 fanout class.

**Proposed fix (~10-20 lines):** move the `latest_ts` advance into the `Ok` arm of `apply_remote_memory` in **all three** branches, or skip `sync_state_observe` when any row in the batch failed. (The audit's first-pass cite was only 421-435 — fix must cover all three.)
Labels: bug, federation, v0.7.1, auto-filed-by-agent

alphaonedev:
Fixed in v0.7.1 (released; tag v0.7.1, merged to main via PR #1702): federation catchup watermark advances only on a successful, non-halted apply across all 3 branches (commit 2187af92). Gates green at release (default 5613 / sal-postgres 6034 tests, 0 failed; clippy -D pedantic; fmt; cargo audit). Verified per prime directive pm-v3 (verify-before-claiming + recompile-retest).

Issue #341, opened 2026-04-22
CRITICAL: replication.rs:188 — duplicate-ack allows same peer to count twice toward quorum


alphaonedev:
## Severity: CRITICAL (release-block for v0.6.2)

Source: ultrareview FINAL 2026-04-22 @ commit 635615a (v0.6.2 Patch 2 candidate).

## Finding

`src/replication.rs:188` — quorum accounting counts ACKs without enforcing peer uniqueness. The same peer fingerprint can contribute to quorum more than once if:

- The operator configures the same peer URL under two different labels (duplicate entry in peers list).
- Dynamic fingerprint regeneration across restarts makes a single peer appear twice during the quorum window.
- A misbehaving peer replays the ACK over multiple connections.

Effect: a write can be reported as \"quorum met\" when in reality only 1 distinct peer acknowledged, defeating the durability guarantee the API claims.

## Evidence

- `src/replication.rs` near line 188 — ACK loop increments a counter without checking peer identity into a seen-set.
- No unit test covers \"two ACKs from same peer → counted once\".

## Fix

1. `FederationConfig::build()` (startup): reject config where two peers have the same normalized URL or the same advertised fingerprint.
2. Replication quorum path: maintain `HashSet<PeerId>` of distinct ACK sources; count only first ACK per peer.
3. Require stable fingerprints (ed25519 public key hash, not random per-process nonce).
4. Add regression test: feed duplicate-peer ACKs, assert quorum not met.

## Related

- Ultrareview FINAL memory: `fe704ed6-b582-485a-b094-fecc510a63f9`.
- Epic: #14 (testbook v3.0.0 MAX-COVERAGE).
- Release freeze: no v0.6.2 tag until fix landed + tested.

alphaonedev:
> Authored by Claude Opus 4.7 on behalf of @alphaonedev.

Fixed in **PR #357** (merged to `release/v0.6.2` 2026-04-22; v0.6.2 tag cut 2026-04-24).

Verification: `federation.rs:86` carries explicit `Ultrareview #341` marker. `FederationConfig::build` rejects duplicate peer URLs at serve time.

---

*Closed by AI NHI (Claude Opus 4.7) on @alphaonedev's authorization. Note: `main` is behind `release/v0.6.2` until the sync lands — the fix is in the v0.6.2 release binary (crates.io, Homebrew, PPA, COPR, GHCR) regardless of main's HEAD.*

Issue #304, opened 2026-04-20
Code-review nits batch — 13 cosmetic findings for v0.6.x maintenance


alphaonedev:
## Summary

13 nit-level findings from the v0.6.0 code review. All cosmetic / readability. Batch into v0.6.x maintenance.

### Trident nits

- \`src/autonomy.rs:311\` — consolidated title uses \`cluster[0].title\` via \`map().next()\`; just use \`cluster[0].title.as_str()\` and drop the iterator ceremony.
- \`src/autonomy.rs:378-391\` — \`confirmed_contradictions\` comparison requires the superseder to be in the same candidate batch; cross-batch contradictions silently no-op. Add a comment or support cross-batch.
- \`src/migrate.rs:151-153\` — \`fresh.is_empty() → break\` ends migration when \`page.len() == batch_size\` but all in \`seen\`; should \`continue\` with cursor advance.
- \`src/federation.rs:93-97\` — \`PeerEndpoint.id = format!("peer-{i}:{trimmed}")\` leaks full URL into metrics labels (high cardinality). Use \`peer-{i}\` only; log URL separately.

### Security nits

- \`src/validate.rs:186-199\` — \`agent_id\` regex permits \`..\` (no path segment check). Mirror \`validate_namespace\`'s \`.\`/\`..\` rejection when id contains \`/\`.
- \`src/db.rs:200-208\` — SQLCipher passphrase logged via process env. Hardening: read from \`--db-passphrase-file\` fd directly + \`zeroize\` after \`PRAGMA key\`.
- \`src/subscriptions.rs:186\` — \`DispatchPayload\` not signed over canonical JSON; serde_json output order stable in practice but re-serialised input breaks on whitespace/ordering.

### SAL nits

- \`src/store/postgres.rs:405\` — \`_ctx\` leading-underscore on a used parameter.
- \`src/migrate.rs:80-82\` — \`sqlite://./foo\` URL parsing comment + \`strip_prefix('/')\` confusing. Use \`url::Url\` parse.
- \`src/store/postgres_schema.sql:26\` — hard-codes \`vector(384)\` (MiniLmL6V2 only). No failsafe for 768-dim embedder config. Add a \`connect()\` sanity check via \`pg_attribute\`.

### Test/harness nits

- \`src/main.rs:773,789\` — \`unsafe { std::env::set_var(...) }\` outside test cfg for runtime anonymize/passphrase. Could race test env.
[...]

PR #2669, opened 2026-08-01
fix(federation): derive the DLQ routing key from peer identity, not flag position (#2442)

### LINKED ISSUE #2442: [federation][data-integrity] DLQ routing key is a POSITIONAL peer index — decommissioning a peer misroutes queued writes to the wrong host and marks them replayed

alphaonedev:
## The defect

`src/federation/peer.rs` minted `PeerEndpoint { id: format!("peer-{i}") }` from the `--quorum-peers` `.enumerate()` index and discarded the URL. The in-code comment justified that as keeping a Prometheus label low-cardinality. It is not a label. It is the **durable routing key** of `federation_push_dlq.peer_id` (enqueued `src/federation/sync.rs:663`, resolved `src/federation/push_dlq.rs:488` via `config.peers.iter().find(|p| p.id == row.peer_id)`) **and** of the `sync_state` catch-up vector clock (`src/federation/receive.rs:249,381`).

Remove one peer and every higher index shifts down. That `find` then **succeeded against a different host**, POSTed another tenant's full memory payload there, and stamped the row `replayed_at`. Cross-tenant content disclosure and silent write loss, from an operation as ordinary as removing a decommissioned host from a flag. The same positional key also lets a surviving peer inherit a departed peer's catch-up watermark, which can skip rows outright — a second silent-divergence path the issue did not name.

## The fix

`id = peer-h1<32 hex>` where the digest is `SHA-256("ai-memory:peer-id:v1\0" || normalize_peer_url(raw))`. A pure, total function of peer identity, so adding, removing, or reordering peers never re-keys a survivor. The lookup can now only **fail** — it can no longer succeed against the wrong endpoint. That is the security property; the id string is incidental.

- **`normalize_peer_url` is shared with the `#341` duplicate-URL check** so the two cannot drift. Two spellings that check calls identical mint one id; two it calls distinct mint two.
- **Fail-closed collision guard** in `build()` — two peers minting one id refuses boot, naming both URLs. Mirrors the duplicate-URL refusal on the same loop. Extracted as a pure `first_peer_id_collision` so the branch is unit-testable rather than unreachable and untrusted.
- **Peer registration promoted debug -> INFO.** The id is opaque now, so that line is the operator's id->url decoder for DLQ triage. Once per peer at boot.

### Two departures from the brief, both driven by the vote

**1. 128 bits of digest, not the suggested 12 hex (48).** The brief's sizing argument is a birthday argument against fleet size (tens of peers), and it is correct as far as it goes. It is the wrong frame: the hash input is **operator-supplied**, so an adversary who can get one URL into `--quorum-peers` picks it freely. The profitable target is a peer that has been **removed** from the config but whose rows are still in `federation_push_dlq` — grind a URL whose id equals the departed peer's and `replay_once` hands you its queued payloads. That is a second-preimage search against one fixed digest, ~2^48 SHA-256 evaluations (hours of commodity GPU time), and **the boot-time collision guard cannot see it**, because the victim is not in the config to be compared against. At 128 bits the search is infeasible.

The width costs nothing, because the brief's premise is stale: **`PeerEndpoint.id` reaches ZERO Prometheus labels at v1.0.0.** I enumerated every `with_label_values` in `src/` — the federation lane's are all closed-set tokens (`cause` in `push_dlq.rs:466`, `outcome`/`reason` in `sync.rs`), `federation_partial_quorum_total` is a bare IntCounter, and the partial-quorum WARN logs `peer_urls`, not ids. Two vote lenses found this independently. The `#304` comment is corrected in place rather than left to be re-derived as a false constraint by the next agent. The legitimate half of `#304` still holds and is still honoured: the id is short and **fixed-width** (39 chars), never an unbounded URL.

**2. The derivation is domain-separated and version-tagged** (`ai-memory:peer-id:v1\0`, prefix `peer-h1`). Lens 4's killer objection: the shared `normalize_peer_url` welds two consumers with **opposing** long-run requirements — the duplicate check *should* normalise more over time (IDNA, default-port stripping, percent-decoding) to close the `#341` double-count gap, while the id must normalise identically **forever** or every persisted DLQ row and `sync_state` entry is orphaned. Whoever correctly hardens that normalisation would otherwise ship a silent fleet-wide identity rotation with no error and no failing test. `frozen_wire_format_golden_2442` is the tripwire that forces it to be a deliberate migration, and the versioned prefix makes the two generations structurally distinguishable in a table that will contain both.

The `h` is load-bearing too: 32 hex nibbles are all-decimal-digits often enough to matter, so it is what makes a minted id structurally unable to impersonate the legacy `peer-<digits>` shape.

## Legacy-row migration disposition (and what is lost)

**Skip, annotate, never auto-remap.** Rows written by a pre-fix binary carry `peer-0`, `peer-1`, … They are detected by the **shape** of `peer_id`, skipped, and left in the table with the payload intact.

Remapping `peer-N` -> `config.peers[N]` **is** the defect, and a binary upgrade is precisely when operators edit peer lists, so guessing would re-commit it at the worst possible moment.

**What is lost, stated plainly:** after `MAX_REPLAY_ATTEMPTS` (100) these rows quarantine and stop being retried automatically. The payload is retained, never deleted. Until an operator re-keys them, those writes reach the intended peer only if its own `/sync/since` catch-up carries them.

Three design points the vote forced, each of which I had initially gotten wrong:

- **The attempt bump is retained deliberately, for liveness.** `take_pending_dlq_rows` is `ORDER BY failed_at ASC`, and legacy rows are by construction the **oldest**. A row that never burned its budget would sit at the head of every batch forever and **starve live rows out of the replay window** — a new silent-loss vector introduced by the fix. Bumping lets them converge to take-exclusion and hands the queue back.
- **The annotation is applied exactly once and preserves the original enqueue reason verbatim.** `bump_dlq_attempt` overwrites `last_error`, and the DLQ row *is* the forensic record of why the write was undelivered; a constant reason would destroy that permanently.
- **No new closed-set cause label**, against my first draft and the issue's implied shape. Three lenses converged: `classify_quarantine_cause` is an **ordered substring matcher**, and a peer-supplied INTEGER already reaches that string: `sync::success_report_non_ack_reason` interpolates the receiver's own `skipped` count via `as_u64`, so a peer answering 200 with `{"skipped": 429}` mints a `last_error` containing `429` — matching that classifier's FIRST arm (`quota`) **and** `reset_throttled_quarantine`'s `LIKE '%429%'` on both backends, resetting `attempt_count = 0` forever and defeating the quarantine ceiling (now tracked as #2672). An arm keyed on a token in that string would widen the same laundering surface. It would also **override the row's real cause**: a legacy-keyed row already failing `http 400` is still `permanent`. The condition is decided from structured input and gets its own counter, `ai_memory_federation_push_dlq_legacy_positional_total`. (The pre-existing 2-cause drift in the `..._quarantined_by_cause` HELP string is re-synced in the same commit.)

**Remediation** is in `docs/TROUBLESHOOTING.md` §federation-push-DLQ: the operator supplies the old-index->URL mapping (nothing in the database can reconstruct it), computes the target id with a shell one-liner that needs no daemon, and re-keys — **including `attempt_count = 0`, without which the re-key is a silent no-op against the take-exclusion**. It also explicitly **forbids** mirroring the re-key onto `sync_state`: that clock is monotonic with no API that walks backwards, so one wrong row permanently advances a cursor past rows never pulled, converting a benign orphan into irreversible loss. The two have opposite correct answers.

**`sync_state` legacy entries self-heal** — a missing entry makes catch-up pull from the beginning, bounded to 500 rows/tick, oldest-first, idempotent via `insert_if_newer`. Verified: `receive.rs:246-252` yields `since=None`, the peer clamps to `limit.unwrap_or(500)`, `memories_updated_since` is `ORDER BY updated_at ASC`, and the `#1687` watermark advances only on a durable apply. Bounded, resumable, convergent.

## 3x3 adversarial vote — 9 lenses, both sub-decisions

Three independent agents, three lenses each, each returning VERDICT / CONFIDENCE / RATIONALE / TOP_RISK / KILLER_OBJECTION. Cites `4d3ea1c5`. Triggers: **T4** (hard-to-reverse persisted representation) and **T3** (security posture).

### (a) ID derivation — **ADOPT, 9/9** (5 ADOPT-WITH-CHANGES, all changes applied)

| Lens | Verdict | Conf | Killer objection -> disposition |
|---|---|---|---|
| 1 Data-integrity | ADOPT | 88 | Makes the key depend on a string operators edit routinely, so orphans get *more* common, not less -> accepted; the trade is misdelivery -> loud non-delivery, which is the North Star ordering |
| 2 Security | ADOPT-WITH-CHANGES | 85 | 48 bits is a second-preimage setting; grind against a REMOVED peer, boot guard blind -> **APPLIED: 128 bits** |
| 3 Operator | ADOPT-WITH-CHANGES | 74 | Opaque ids are the only decoder for a durable keyspace, living in ephemeral logs -> **APPLIED: INFO boot map + the derivation documented so operators can compute it from the URL alone** |
| 4 Crypto | ADOPT-WITH-CHANGES | 88 | Shared helper welds consumers with opposing requirements; a correct dup-check hardening silently rotates every id -> **APPLIED: domain separation + version tag + golden tripwire** |
| 5 Upgrade/rollback | ADOPT | 90 | A `peer_url` column is the real long-term answer; this ships the key without the binding -> accepted as a correct increment (a url column could not repair legacy rows either) |
| 6 Adversarial | ADOPT-WITH-CHANGES | 82 | Guard is scoped to the live config; the durable keyspace includes removed peers -> same fix as lens 2 |
| 7 Observability | ADOPT-WITH-CHANGES | 88 | Justifying a durable-key change with a cardinality argument that does not exist -> **APPLIED: stale `#304` comment corrected in place** |
| 8 Test | ADOPT-WITH-CHANGES | 90 | An id-only test passes for the wrong reason; the replay path is never entered -> **APPLIED: the load-bearing test drives `replay_once`** |
| 9 Simplicity | ADOPT-WITH-CHANGES | 82 | `id = normalized_url` is simpler and human-readable -> rejected: unbounded value in a durable key + log field, and it hands an attacker the keyspace in plaintext. `PeerId` newtype deferred with rationale |

### (b) Legacy-row disposition — **ADOPT, 9/9** (6 ADOPT-WITH-CHANGES, all applied)

Unanimous that auto-remap is refused. The changes that landed:

| Lens | Killer objection -> disposition |
|---|---|
| 1 | Burning the budget makes a recoverable condition need a human forever -> **APPLIED: remediation resets `attempt_count`**; no-bump rejected on the starvation finding below |
| 2 | Quarantine leaves plaintext third-party payloads at rest indefinitely -> acknowledged, unchanged: retention is the pre-existing DLQ posture, and destroying undelivered writes is the worse failure |
| 3 | The SQL re-key is a fig leaf at 1000 nodes -> partially applied: the runbook is executable and the id is computable without a daemon; a CLI verb is a follow-up |
| 4 | The cause label is derived from a string the upgrade itself overwrites, destroying the original reason -> **APPLIED: shape-based detection, annotate-once, original reason preserved** |
| 5 | The re-key is a no-op against take-exclusion without an `attempt_count` reset; and mirroring it onto `sync_state` is a data-loss instrument -> **APPLIED: both, explicitly** |
| 6 | A closed-set label is only as closed as the matcher producing it, and a peer can steer that matcher -> **APPLIED: no label added** |
| 7 | A permanent label naming a one-time event reports 0 forever -> same disposition |
| 8 | A negative assertion rots silently -> **APPLIED: positive control in the test** |
| 9 | A doc-comment removal horizon is not a mechanism -> accepted as permanent; the detector is ~6 lines at one site |

**Liveness finding that settled the design** (mine, from the vote's pressure): `take_pending_dlq_rows` is `ORDER BY failed_at ASC`. Never bumping legacy rows would starve live rows out of the replay window forever — so lens 1's "do not burn the budget" is right about forensics and wrong about liveness. Both are satisfied by bumping while preserving the original reason inside the annotation.

## R-203 evidence

Parent SHA **`03bbd556`**. Two probes, because the literal one is weaker than it looks.

**1. Literal `git checkout 03bbd556 -- src/`** — compile failure only, since the new symbol does not exist at the parent:

```
error[E0432]: unresolved import `ai_memory::federation::peer::is_legacy_positional_peer_id`
  --> tests/federation_stable_peer_id_2442.rs:45:5
error: could not compile `ai-memory` (test "federation_stable_peer_id_2442") due to 1 previous error
```

**2. Surgical parent-BEHAVIOUR probe** — revert only the minting expression to `format!("peer-{i}")`, keeping the surface, so the suite compiles and fails on **behaviour**. 7 of 10 failed. Verbatim:

```
---- replay::decommissioned_peer_dlq_row_is_never_delivered_to_a_survivor_2442 stdout ----
thread 'replay::decommissioned_peer_dlq_row_is_never_delivered_to_a_survivor_2442' panicked at tests/federation_stable_peer_id_2442.rs:407:9:
CROSS-TENANT DISCLOSURE: peer C received the payload queued for the decommissioned peer B (["for-B"]). The DLQ routing key must be derived from peer identity, not the --quorum-peers flag index (#2442)

---- decommissioning_a_peer_does_not_rekey_survivors_2442 stdout ----
assertion `left == right` failed: peer C must keep its routing key when peer B is decommissioned; positional ids gave C `peer-2` then `peer-1`, silently re-pointing every queued row for the old `peer-2` at a different host (#2442)
  left: "peer-2"
 right: "peer-1"

---- replay::legacy_positional_dlq_row_is_skipped_not_remapped_2442 stdout ----
assertion `left == right` failed: `peer-1` must NOT be remapped onto config.peers[1] — that mapping is only correct if the peer list never changed, and an upgrade is exactly when it does (#2442)
  left: 1
 right: 0

failures:
    decommissioning_a_peer_does_not_rekey_survivors_2442
    frozen_wire_format_golden_2442
    minted_id_never_matches_the_legacy_positional_shape_2442
    peer_id_is_short_bounded_and_label_safe_2442
    reordering_peers_does_not_rekey_2442
    replay::decommissioned_peer_dlq_row_is_never_delivered_to_a_survivor_2442
    replay::legacy_positional_dlq_row_is_skipped_not_remapped_2442

test result: FAILED. 3 passed; 7 failed
```

`["for-B"]` is the whole defect in one line: peer C received the payload queued for the decommissioned peer B, and its own row was skipped as unresolvable. Disclosure and silent loss in a single replay tick.

At HEAD: **10/10 pass** (11/11 with the postgres leg).

The load-bearing test deliberately is **not** an id comparison. It stands up three live mock peers, queues a row for B and one for C, decommissions B, drives a real `replay_once`, and asserts no payload queued for B reaches C **and** that C's own row still lands — the positive control, without which a refactor that made `replay_once` bail early would satisfy every other assertion while proving nothing.

## Both backends

**The id derivation is backend-blind** — a pure function of a string, no SQL. This PR adds **no sink method, no schema migration, and no SQL change**; `replay_once` is backend-agnostic over `&dyn FederationDlqSink`, and both `federation_push_dlq.peer_id` and `sync_state.peer_id` are bare `TEXT` on both backends. So the only thing worth proving on postgres is that the shared skip behaves identically against the postgres sink, which `pg_legacy_positional_dlq_row_is_skipped_not_remapped_2442` does.

Run against live PG 18.4 on a **scratch** database (`ai_memory_2442_scratch`, created and dropped): **11 passed; 0 failed**. Shared tier verified before and after — `max(schema_version) = 88`, `federation_push_dlq` 0 rows, `memories` 8759 rows, scratch database absent. **Zero residue.**

## Consumer survey for `PeerEndpoint.id`

None depends on the id being positional or human-ordered:

- `AckTracker::record_peer_ack` / `acked_peer_ids` (`src/replication.rs:191,275`) — an opaque `HashSet<String>` key, used only for dedup and set subtraction. Doc comment already says "typically the peer's mTLS fingerprint or agent id".
- `dispatched_peer_ids` (`src/federation/sync.rs:467`) — a per-write in-process `Vec<String>`, compared only against `acked` from the same config instance.
- Partial-quorum WARN (`sync.rs:611-630`) — logs `peer_urls = ?missing`, i.e. **URLs, not ids**.
- Metric labels — **none**. Verified by enumerating every `with_label_values` in `src/`.
- `push_dlq` log fields — opaque, now decodable via the boot INFO map.
- `peer_attestation::scope_for(peer_id)` — a **different namespace**: the receiver-side wire `X-Peer-Id`, never `PeerEndpoint.id`. Unaffected.

**Reported, not silently widened:**
1. `sync_state` is keyed by `peer.id` while `VectorClock`'s doc comment says the keys are peer **agent ids** — a pre-existing semantic mismatch, orthogonal to this fix.
2. `handlers/federation_receive.rs:3144` folds a **wire-supplied** vector clock into local `sync_state` via a monotonic pointwise max. Under the old scheme every node used the identical `peer-0..N` namespace, so that collision was free; identity-derived ids shrink the guessable keyspace but do not close deliberate poisoning. **Deserves its own issue** — it is not fixed here, and I am not claiming it is.
3. The catch-up loop (`receive.rs:200-208`) sleeps a fixed interval with **no jitter**, unlike the jittered background loops elsewhere. Orphaning every cursor at once on a fleet upgrade is a synchronized re-pull. Bounded (500 rows/tick) and idempotent, so not a correctness issue, but worth jitter.
4. `src/federation/mod.rs:476` — an in-src **test helper** still mints the reverted `peer-{i}:{url}` shape. Inert (it never calls `build()`), left alone to keep this diff minimal; noted so it is not mistaken for production.

## Hard-coded `peer-0` / `peer-1` shapes

- **Fixed:** `src/federation/mod.rs:1444-1445`, the only load-bearing in-src pin of the minted shape. Now asserts the shape and distinctness rather than a literal digest — pinning the literal there would turn a silent derivation change into a one-line test edit instead of the migration it actually is. The frozen-format contract is pinned deliberately, once, in the new test file.
- **Docs listing only (no code change needed):** `docs/ADMIN_GUIDE.md`, `docs/evidence/index.html`, `docs/v0.7.0/hive-1461-baseline/index.html`, `docs/v0.7.0/docker-1461-baseline/index.html`, `docs/v0.7.0/regression-runs/2026-06-02-do-swarm-t4/README.md` — historical run artifacts and prose, not contracts.
- **Incidental, correctly untouched:** ~16 test fixtures across `tests/` construct `PeerEndpoint` literally and never call `build()`; `src/replication.rs` and `src/federation/receive_auth.rs` use `"peer-1"` as an opaque map key or a wire `X-Peer-Id` in an unrelated namespace.

## CI parity

- `cargo check --all-targets` on default / `sal` / `sal-postgres` — clean; `cargo check --examples` clean
- Clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` on **all four** legs (default, `sal`, `sal-postgres`, `vectorlite`) — clean. Two pedantic lints in the new test file initially redded three legs simultaneously; fixed before push.
- `cargo fmt --all`
- `scripts/check-hardcoded-literals.sh`, `check-vendor-literals.sh`, `check-docs-vs-ssot.sh` — all PASS
- `cargo test --test qual_10_module_size_ceiling --test qual_6_7_legacy_error_type_ceiling` — 2+2 pass, **no ceiling bump needed**
- `--test federation_stable_peer_id_2442` 10/10, `--test federation_dlq_replay` 10/10 (no regression in the existing DLQ suite), `--lib federation::peer` 27/27
- CHANGELOG `[Unreleased]` entry added

Closes #2442


alphaonedev:
## Fable 5 — code review + security audit: APPROVED

Verified at `aac33034`. Head-synced (API head == branch tip, 6 CI runs dispatched, **0 failing check-runs** read from `commits/<sha>/check-runs`, not `gh run list`). Three commits, all signature-verified.

### The fix is correct and the derivation is properly frozen

`PeerEndpoint.id` now derives from **identity** — `peer-h1<32 hex>` = `SHA-256("ai-memory:peer-id:v1\0" || <normalised url>)` — instead of the `.enumerate()` index. Domain-separated and version-tagged in both the preimage and the visible prefix.

### What I audited hardest: the legacy-row disposition

This is where a fix like this usually goes wrong, and it does not.

**It refuses to remap `peer-N` → `config.peers[N]`.** The stated reason is exactly right: that mapping is only correct if the peer list has not changed since the row was written, and **a binary upgrade is precisely when operators do change it**. Guessing would reintroduce the misdelivery the issue exists to close. `DEGRADE (loud, actionable quarantine) beats CORRUPT (content to the wrong host)` is the North Star ordering applied where it costs something, and the cost is stated in the operator-facing message rather than buried.

**Two second-order effects it caught that most implementations would not:**

1. **`last_error` is annotated exactly once, preserving the original enqueue reason verbatim.** `bump_dlq_attempt` overwrites that field, so a naive constant marker would permanently destroy the forensic record of why the write was undelivered — and as the comment says, *the row IS that record*.
2. **The attempt bump is not skipped**, because pending rows drain `ORDER BY failed_at ASC` and legacy rows are by construction the **oldest**. A row that never burns its budget sits at the head of every batch forever and **starves live rows out of the replay window**. Quarantining without bumping would have converted a routing bug into a queue-starvation bug.

### The coupling it found is the most valuable part

`normalize_peer_url` is **shared** with the #341 duplicate-URL check, and the two consumers have **opposing** long-run requirements: the dup check *should* normalise more over time (IDNA, default-port stripping, percent-decoding), while the id must normalise **identically forever** or every persisted row is orphaned.

`frozen_wire_format_golden_2442` makes that a deliberate migration instead of a silent fleet-wide identity rotation with no failing test. Nobody asked for this; it was found by asking what else touches the function.

The `h` in `peer-h1` is load-bearing and correctly justified: 32 hex nibbles are all-decimal often enough to matter, so it is what makes a minted id structurally unable to impersonate the legacy `peer-<digits>` shape — which keeps `is_legacy_positional_peer_id` free of false positives.

### Severity confirmed

Both halves of the North Star broke at once on the old code: an **unintended peer received content it was never authorised to hold**, and the **intended peer never received the write and diverged permanently** — with no counter and no error. The same positional key also let a surviving peer inherit a decommissioned peer's catch-up watermark, which can skip rows outright.

Closes **#2442**, one of the 25 must-close issues. Merging.

Issue #2577, opened 2026-07-31
perf(recall): query-embedding remote call is unbounded on the read path — +156 ms floor per recall, one 39 s stall observed


alphaonedev:
## Summary

At `tier = semantic` / `autonomous`, **every `memory_recall` makes a synchronous remote HTTP call to the embeddings provider to embed the query**, on the read path, with no observed bound. One sampled call took **39,268 ms** (39 seconds).

## Evidence

Measured on the production config (`backend=openrouter`, `model=google/gemini-embedding-2`), MCP stdio, 7,855-row corpus:

| tier | p50 | min | p95 | p99 | max observed |
|---|---|---|---|---|---|
| keyword (no embed call) | 3.05 ms | 1.81 | 17.9 | 19.84 | 21.3 |
| semantic | 158.78 ms | 119.34 | 268.57 | 492.26 | 492.3 |
| autonomous, limit=1 | 289.47 ms | 188.98 | — | — | **39,268.39** |

The keyword tier — identical code path minus the embed call — never exceeded 35 ms across 300 samples. The 39 s outlier occurred only on an embed-bearing tier.

Even the healthy case is a **~156 ms floor added to every recall** (158.78 − 3.05), i.e. **50x the entire keyword recall cost**, paid before any local work begins.

## Why this matters (North Star: MANAGEABLE AT SCALE / DEGRADE, never stall)

- A recall is a **read**. A provider stall converts a read into a multi-second (here 39 s) hang.
- On the HTTP daemon this holds a request slot; under `AI_MEMORY_MAX_INFLIGHT_REQUESTS` admission control (default-on since #2032 M3), sustained provider latency will saturate the in-flight cap and shed *healthy* traffic with 503s.
- There is no degrade-to-keyword path on slowness — only on embedder *construction* failure (#1593). A provider that is up but slow degrades nothing; it just blocks.

## Reproduction
```bash
AI_MEMORY_NO_CONFIG=1 AI_MEMORY_DB=<corpus.db> HF_HUB_OFFLINE=1 \
AI_MEMORY_EMBED_BACKEND=openrouter AI_MEMORY_EMBED_MODEL=google/gemini-embedding-2 \
AI_MEMORY_EMBED_API_KEY=<key> AI_MEMORY_EMBED_BASE_URL=https://openrouter.ai/api/v1 \
ai-memory mcp --profile full --tier autonomous
[...]

Issue #1602, opened 2026-06-11
memory_forget dry_run returns only a count — no ids/titles, so the blast radius of a destructive bulk op is invisible before commit


alphaonedev:
**Parent:** #1588, dimension 6. Sibling of the forget OR-matching issue (distinct defect: preview fidelity vs. match semantics).

**Root cause one-liner:** `memory_forget {dry_run: true}` responds `{"dry_run": true, "would_delete": N}` — a bare count. The caller cannot see WHICH rows die without manually re-implementing the match (impossible faithfully, since the FTS pattern semantics are undocumented). A preview that can't be reviewed defeats the purpose of dry_run on the substrate's most destructive bulk surface.

**Evidence (live session):** `forget {pattern: "D6 scratch", namespace: "fable5/worklog", dry_run: true}` → `{would_delete: 3}` — which 3? Resolving required a separate `memory_list` of the whole namespace plus guessing the matcher's behavior (and the guess "title-substring" would have been WRONG — it's fuzzy OR over title/content/tags).

**Proposed fix (~20 lines + test):** dry_run returns `would_delete` plus a `rows: [{id, title, namespace, tier}]` array (capped at e.g. 50 with `truncated: true`) — same shape archive_list uses. Non-dry-run response should likewise return the deleted ids (it already archives them; returning ids makes recovery one call instead of an archive_list hunt).

alphaonedev:
Fixed via commit 07c7ee95 on release/v0.7.0. forget dry_run returns rows preview (cap 50 + truncated flag); live run returns deleted_ids; regression tests/issue_1602_forget_match_listing.rs. Full suite green (EXIT=0), fmt+clippy-pedantic+audit+3 QC lint gates clean. Verified per prime directive pm-v3.3; dogfood re-verification follows in the #1588 Phase-3 re-run on the rebuilt golden.

PR #2673, opened 2026-08-01
fix(federation): replicate MCP/CLI erasure to peers via a durable outbox (#2446)

### LINKED ISSUE #2446: [federation][gdpr] erasure does not replicate but writes do — `broadcast_delete_quorum` has ONE production caller (HTTP DELETE); MCP and CLI forget are purely local
**Class:** asymmetric fanout — GDPR Art.17 erasure defeated across replicas. **Verified at `release/v1.0.0` @ `011dd8b9`.**

## The asymmetry

`broadcast_delete_quorum` (`src/federation/sync.rs:700`) has exactly **one** production caller:

```
src/handlers/memories.rs:1215    crate::federation::broadcast_delete_quorum(fed, &target.id).await
```

Every other hit in `src/` is a doc comment or a unit test (`src/federation/mod.rs:874-3344`). Grepping `src/mcp/tools/forget.rs`, `src/mcp/tools/delete.rs` and all of `src/cli/` for any broadcast returns **nothing**.

The **write** path does broadcast — `src/mcp/tools/store/` fans out via `broadcast_store_quorum`, and `src/mcp/tools/store/transport.rs:14` states it *"closes the MCP-stdio-vs-federation gap"*. That gap was closed for writes and left open for deletes.

## Failure scenario

A GDPR erasure issued through **MCP** — the primary agent interface this entire product is built around — or through the **CLI** is purely **local**. Peers that were never told keep serving the erased content indefinitely. The v71 `forget_tombstones` row is only a local anti-resurrection guard on the node that holds it; it does not travel.

Composed with **#2312** (the same destructive `forget` wire call deletes different row sets per backend — FTS5 token-AND vs ILIKE substring) and **#2313** (postgres hard-forget still snapshots `memory_links` into `archived_memory_links`, which no code path ever deletes from), the erasure story is: **incomplete, non-deterministic across backends, and non-propagating.**

## Why this is not merely technical

This is legal and regulatory exposure. An enterprise DPIA rejects it on day one, and no amount of per-node correctness fixes it — the deletion simply does not reach the replicas. It is also the mirror of #2391: a `broadcast_*_quorum` that exists and is never invoked.

## Suggested fix

1. Fan out erasure from the MCP and CLI delete/forget paths, not only from the HTTP `DELETE` verb.
2. Add a gate: **every `broadcast_*_quorum` must have a production caller or be deleted.** This shape has now appeared twice (#2391 outbound checkpoint resolution, and this) — that is a pattern, not a coincidence, and a caller-reachability assertion beside the coverage gate would catch both. Note the coverage floor currently reports these GREEN because unit tests exercise them.

## Related

#1852 (mesh-wide un-forget, open), #2312, #2313, #2314, #2391.

Found by Wave 2 of the v1.0.0 GA gap review; independently re-verified. Context: #2440.

alphaonedev:
Closes #2446

## The defect

`sync::broadcast_delete_quorum` (`src/federation/sync.rs:700`) had exactly **ONE** production caller: the HTTP `DELETE /api/v1/memories/{id}` handler (`src/handlers/memories.rs:1215`). Every other erasure funnel was purely local:

- MCP `memory_delete` (`src/mcp/tools/delete.rs`) and `memory_forget` (`src/mcp/tools/forget.rs`)
- CLI `delete` (`src/cli/crud.rs`) and `forget` (`src/cli/forget.rs`)

An agent calling `memory_forget` over MCP erased locally and got `success`, while every federated replica kept the row, re-served it as substrate truth, and could later LWW-resurrect it. The STORE path meanwhile fanned out. That is a GDPR Article 17 erasure failure **and** a permanent replica divergence, invisible by construction.

**Anti-entropy catch-up cannot cover this.** It propagates writes by PULLING rows; an erased row is exactly the thing it can no longer see. Only an explicit tombstone push can carry an erasure.

## Approach — option B of the 9-lens vote (8 of 9), not re-litigated

A 9-lens 3x3 adversarial vote (citing `4d3ea1c5`) chose **a durable OUTBOX drained by the daemon's existing federation worker — 8 of 9 lenses**, over threading `FederationConfig` into MCP/CLI for an inline broadcast. Both reasons are load-bearing and were re-verified against the source, not assumed:

- **Inline is structurally blocked.** `FederationConfig::build` is called in exactly one place — `daemon_runtime::bootstrap_serve` — and `AppState.federation` is an HTTP-only struct. `ai-memory mcp` and the CLI never construct or receive one.
- **Inline is behaviourally wrong.** A blocking quorum broadcast inside the single-threaded JSON-RPC stdio loop (the #965 audit) would stall EVERY subsequent tool call including `memory_store` — precisely the #2577 failure mode whose fix had to introduce a hard budget.

## Mechanism validation — the three questions, answered before building

The vote assumed a new table; the lane lead's hypothesis was that `federation_push_dlq` (v48, both backends) already IS the outbox. **It holds. No migration.**

**1. Does the sentinel round-trip the partial-unique index without colliding with a real per-peer row for the same `memory_id`? YES.** `idx_federation_push_dlq_pending_uniq` is `ON federation_push_dlq(memory_id, peer_id) WHERE replayed_at IS NULL` — keyed on the PAIR. The reserved sentinel `peer_id` `__ai_memory_all_peers__` is a distinct value, so the sentinel and a real `peer-0` row for the same memory coexist as two pending rows (pinned by `sentinel_and_real_peer_row_coexist_for_the_same_memory_2446`, which seeds the exact shape PR #2662's delete-lane landing pass writes). `take_pending_dlq_rows` surfaces it like any other row. It is **not** left to `find(|p| p.id == row.peer_id)`: the sentinel branch runs BEFORE peer resolution in `replay_once`, because otherwise the row would be bumped every tick under the LIE `peer_removed` (a real closed-set cause label) until quarantine — an erasure silently converted into an operator-drain chore.

**2. Can the MCP/CLI erasure sites INSERT without touching `src/storage/mod.rs`? YES.** All SQL lives in the NEW `src/federation/erasure_outbox.rs`, raw `rusqlite` against the connection those funnels already hold. No existing storage function was edited. The module is deliberately NOT `--features sal`-gated (unlike `push_dlq`) so the write side compiles on every build; only the DRAIN needs the SAL sink trait.

**3. Is the sentinel reachable by `classify_quarantine_cause`'s closed label set without lying? YES, and the texts are asserted.** A sentinel is never POSTed, so the only route to quarantine is 100 consecutive EXPANSION failures (a broken DB), whose `last_error` is a backend error and classifies as the honest catch-all `other`. Both the queued text and the expanded per-peer text are asserted to contain none of `429` / `401` / `403` / `400` / `422` / `signature` / `schema` / `id_drift` / `no longer in FederationConfig` (`queued_last_error_does_not_lie_to_the_cause_classifier`), across all four surface tags. A repeat erasure of the same id coalesces WITHOUT bumping `attempt_count` — burning the quarantine budget on a re-request would be a fabricated failure.

## What landed

The four erasure funnels insert ONE sentinel row after the local erasure commits. `push_dlq::replay_once` expands it into per-peer rows on the first drain tick — the DAEMON owns the live peer set AND the federation identity, so it builds the canonical delete body (`{sender_agent_id, memories: [], deletions: [id], dry_run: false}`, shape-identical to `broadcast_delete_quorum`'s). After expansion, everything downstream is the existing machinery unchanged: attempt budget, quarantine, cause labels, depth gauge, `AI_MEMORY_FED_DLQ_REPLAY_MAX_BATCH` pacing, operator drain.

The expanded rows are **key-identical** to what PR #2662's delete-lane landing pass writes for the same `(memory_id, peer_id)`, so the two converge on the index rather than fighting over it. On conflict the delete body SUPERSEDES (the erasure is the newer intent; replaying a superseded store would resurrect a row the origin erased), `attempt_count` is not bumped (an erasure is not a delivery failure) and `failed_at` is not refreshed (so an older pending row keeps its place in the `ORDER BY failed_at ASC` queue instead of being starved).

## The accumulation bound is structural, not a sweep

A row is written **only when the local database carries a drainability marker**, stamped by `bootstrap_serve` on every boot where — and only where — **all three** hold: federation is configured, the build carries `--features sal` (the entire push-DLQ surface is gated on it, so a default-features binary has no replay worker at all), and the resolved sink is the SQLITE one (the worker drains the SAME file the MCP/CLI processes write to). **Every other boot CLEARS it.**

- A single-node / MCP-only deployment — the overwhelmingly common case — writes **ZERO** rows, forever.
- Turning federation off self-heals on the next `serve` boot; accumulation stops at the erasures performed between the two boots, bounded by operator action rather than by time.
- The marker rides in the same table with `replayed_at` pre-set, so it is invisible to every existing `WHERE replayed_at IS NULL` query, to `pending_dlq_count`, to the depth gauge, and to the partial unique index. Nothing in `src/` deletes from `federation_push_dlq`, so no retention sweep can silently un-stamp it.

## Data integrity

**The restore-after-delete race is CLOSED on this path, not merely documented.** A queued erasure replaying after the id was legitimately restored locally would DESTROY the peer's copy of a row this node currently holds — unintentional data loss, the highest-order harm under the North Star. Reachability was verified, not assumed: `db::restore_archived` restores by the archived row's id VERBATIM (`WHERE id = ?1`), so ids are reused. **This does NOT contradict PR #2662**, which analysed the same hazard on the inline broadcast path and shipped it as a bounded residual: that path has no queue delay to re-check in, and its residual is unchanged. The OUTBOX path has a drain delay, so it can and must re-check — the expansion probes `memories` for the id inside the SAME transaction and, when live, clears the sentinel WITHOUT fanning out, with a loud WARN naming the supersession. Pinned by `restore_after_delete_supersedes_the_queued_erasure_2446` (peer POST count asserted `== 0`).

**Erasure never starts failing.** Every outbox call is best-effort and infallible to the caller; a previously-local, always-succeeding operation does not begin returning errors because federation is unhealthy or the DLQ table is unwritable (`a_broken_outbox_never_fails_the_erasure` drops the table and asserts the erasure still succeeds). The local erasure is already committed when the outbox runs; the row is a PROPAGATION record, never the erasure itself.

**Deliberately NOT queued:** gc / size-gc eviction and archive purge. Those are LOCAL retention-policy decisions; replicating one node's TTL sweep to its peers would be data-loss amplification, not erasure. `memory_forget` queues the FULL matched id set (resolved before the delete commits, on the same connection, so it cannot drift) rather than the #1602 preview list — that one is capped at `DRY_RUN_PREVIEW_CAP` (50) and reusing it would have silently propagated only the first 50 erasures of a bulk forget. Beyond `MAX_ERASURE_IDS_PER_CALL` (100k) the sweep emits a LOUD `error!` naming the exact residual rather than under-propagating quietly.

## The `forget_tombstones` composition — confirmed, and the asymmetry reported

`grep -rn forget_tombstone src/federation/` returns **ZERO** hits: tombstones do NOT propagate. `db::memory_is_tombstoned` gates a node against its OWN tombstones on the receive funnel, refusing LWW resurrection. So the two controls compose in one direction only: this fix makes the peer APPLY the delete (after which the peer holds no tombstone of its own, because only `forget` writes one and the peer received a `deletions[]` push). Until the queued erasure lands, a peer that never applied the delete remains re-resurrectable by a third node.

Separately, and worth stating even though this PR does not close it: **`memory_delete` writes no `forget_tombstone` at all today — only `forget` does.** A deleted-but-not-forgotten id therefore has no local resurrection guard on ANY node.

## Postgres, stated plainly

The WRITE side is **sqlite-by-construction**: MCP stdio is structurally sqlite-only (CLAUDE.md #1675/n24) and the CLI erasure verbs open a local `rusqlite` connection. So no erasure sentinel reaches a postgres `federation_push_dlq` today. A postgres-backed `serve` drains the POSTGRES table and never the co-located sqlite file — **which is exactly why it does not stamp the marker**: promising propagation there would be a lie, and a co-located sqlite CLI and a postgres daemon do not share a database. On such a deployment MCP/CLI erasure also does not touch the postgres corpus in the first place, so there is no erasure to propagate.

The DRAIN side is nevertheless implemented and tested on **both** backends, because an invalid SQL string on an unreachable path ships undetected until the day it becomes reachable (a future pg CLI, or a relayed sentinel). `pg_expand_erasure_sentinel_parity_2446` (`sal-postgres`, `#[ignore]`) drives `PostgresDlqSink::expand_erasure_sentinel` against live PG 18.4 and asserts the sentinel clears and exactly the per-peer rows remain.

## R-203 evidence

Parent SHA **`03bbd556`**. The suite mirrors the three reserved tokens as LITERALS rather than importing them, deliberately, so it **compiles at the parent** and fails BEHAVIOURALLY (zero rows) instead of failing to build — a compile error proves an API is absent, not that a behaviour is. Token values are pinned against the module consts by `reserved_tokens_are_wire_stable`.

`git checkout 03bbd556 -- src/ tests/cov_federation_sync_quorum_arms.rs tests/federation_dlq_replay.rs` then `( umask 022; cargo test --features sal --test federation_erasure_replication_2446 )`. **All 9 fail**, each for the right reason:

```
running 9 tests
test peer_down_leaves_the_expanded_row_pending_2446 ... FAILED
test cli_delete_queues_an_erasure_outbox_row_2446 ... FAILED
test drain_expands_sentinel_and_delivers_deletion_to_peer_2446 ... FAILED
test restore_after_delete_supersedes_the_queued_erasure_2446 ... FAILED
test cli_forget_queues_erasure_outbox_rows_2446 ... FAILED
test sentinel_and_real_peer_row_coexist_for_the_same_memory_2446 ... FAILED
test unconfigured_deployment_erasure_succeeds_and_queues_nothing_2446 ... FAILED
test mcp_memory_delete_queues_an_erasure_outbox_row_2446 ... FAILED
test mcp_memory_forget_queues_erasure_outbox_rows_2446 ... FAILED

---- mcp_memory_delete_queues_an_erasure_outbox_row_2446 stdout ----
assertion `left == right` failed: #2446: MCP memory_delete must leave exactly ONE
pending erasure-outbox row (pre-fix it left NOTHING and the erasure never reached
any peer); got []
  left: 0
 right: 1

---- mcp_memory_forget_queues_erasure_outbox_rows_2446 stdout ----
assertion `left == right` failed: #2446: MCP memory_forget must queue an
erasure-outbox row for EVERY forgotten id (pre-fix it queued NOTHING)
  left: []
 right: ["mcp-fgt-1", "mcp-fgt-2", "mcp-fgt-3"]

---- cli_delete_queues_an_erasure_outbox_row_2446 stdout ----
assertion `left == right` failed: #2446: CLI delete must leave exactly ONE pending
erasure-outbox row (pre-fix it left NOTHING); got []
  left: 0
 right: 1

---- cli_forget_queues_erasure_outbox_rows_2446 stdout ----
assertion `left == right` failed: #2446: CLI forget must queue an erasure-outbox
row for EVERY forgotten id (pre-fix it queued NOTHING)
  left: []
 right: ["cli-fgt-1", "cli-fgt-2"]

---- drain_expands_sentinel_and_delivers_deletion_to_peer_2446 stdout ----
assertion `left == right` failed: queued before the drain
  left: 0
 right: 1

---- peer_down_leaves_the_expanded_row_pending_2446 stdout ----
assertion `left == right` failed: #2446: a DOWN peer must leave the erasure PENDING
for the next tick, never drop it; got []
  left: 0
 right: 1

---- unconfigured_deployment_erasure_succeeds_and_queues_nothing_2446 stdout ----
assertion `left == right` failed: #2446: the SAME erasure on a DRAINABLE database
must queue exactly one row — the drainability marker is the discriminator, not a
blanket never-queue
  left: 0
 right: 1

---- sentinel_and_real_peer_row_coexist_for_the_same_memory_2446 stdout ----
assertion `left == right` failed: #2446: the sentinel row must land despite a real
per-peer row for the same memory_id
  left: 0
 right: 1

test result: FAILED. 0 passed; 9 failed
```

After the fix: `test result: ok. 9 passed; 0 failed`.

Note the unconfigured-deployment control is deliberately **not** a bare "nothing is queued" assertion — that is vacuously green at the parent, where nothing is ever queued. It pairs an unmarked database with a DRAINABLE twin and asserts the SAME erasure queues exactly one row on the twin, so it can only pass when the bound is real AND the mechanism works.

## Testing

New `tests/federation_erasure_replication_2446.rs` (9 tests + 1 live-PG `#[ignore]`), built on the existing `tests/federation_dlq_replay.rs` harness shape (mock Axum peer + `SqliteDlqSink` + `replay_once`; scratch DBs under `.local-runs/` per the no-`/tmp` hard rule):

- **MCP** `memory_delete` / `memory_forget` driven through the REAL `ai-memory mcp` stdio binary via `CARGO_BIN_EXE_ai-memory` (the pm-v3.3 fresh-subprocess discipline — those handlers are `pub(super)` and unreachable in-process, and probing a stale daemon is not load-bearing evidence).
- **CLI** `delete` / `forget` driven in-process.
- **Drain**: tick 1 expands the sentinel into a per-peer row carrying `deletions:[id]`; tick 2 delivers exactly one POST whose body is that deletion and clears the row.
- **Peer DOWN**: the expanded row stays pending for the next tick — the erasure is never dropped.
- **Restore supersede**, **unconfigured bound**, **sentinel/real-row coexistence** as above.
- 8 unit tests in `src/federation/erasure_outbox.rs`, 11 in `src/federation/push_dlq.rs`.
- Neighbouring suites re-run unchanged: `federation_dlq_replay` **10 passed**, `cov_federation_sync_quorum_arms` **25 passed**.
- **Live postgres**: `pg_expand_erasure_sentinel_parity_2446` green on PG 18.4 against a SCRATCH database (`ai_memory_2446_scratch`, created and dropped). The shared tier was verified afterwards at schema **v88** with **zero residue**.

## CI parity (run locally before push)

- `cargo check --all-targets` — default, `--features sal`, `--features sal-postgres`: green.
- Clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` on **all four legs** (default, `sal`, `sal-postgres`, `vectorlite`): green.
- `cargo fmt --all --check`, `cargo check --examples`: green.
- `qual_10_module_size_ceiling` (2 passed) + `qual_6_7_legacy_error_type_ceiling` (2 passed) — **no ceiling bump**. The boot-time verdict was factored into `erasure_outbox::apply_drainability` so `daemon_runtime.rs` grows 24 lines rather than 56, which is what kept QUAL-10 under its ceiling without moving it.
- Shell gates: hardcoded-literal, vendor-literal, docs-vs-SSOT, migration-ladder, L3-boundary — all PASS.
- No new MCP tool / HTTP route / CLI verb / schema version, so **no SSOT pin moves**. CHANGELOG `[Unreleased]` entry added.

## Lane coordination

- **PR #2662** (`fix/2498-delete-lane-dlq`) — `broadcast_delete_quorum` is NOT edited here; this work sits upstream of it and the expanded rows are key- and shape-identical to what its landing pass writes. Its delete-then-restore residual analysis was read in full and is not contradicted (see Data integrity).
- **`fix/2442-stable-peer-id`** — nothing depends on `PeerEndpoint.id` being positional. The sentinel is a reserved token no derivation can produce, and the expansion re-verifies that against the LIVE peer set and refuses fail-closed if it ever collides.
- **PR #2664** (`fix/2477-plaintext-peer-refusal`) — untouched.
- **Forbidden files confirmed untouched**: `src/handlers/federation_receive.rs`, `src/storage/mod.rs`, `src/store/postgres.rs`. **No schema migration.**

## Reported, not silently absorbed

The entire federation push-DLQ surface — `#933`, `#2498`/#2662, and this — is `#[cfg(feature = "sal")]`, and `.github/workflows/release.yml:132` builds the shipped binary with **default features only**. So the drain half of this fix is compiled OUT of the released artifact; the write half is inert there by construction (a default-features `serve` never stamps the marker, so it queues nothing). That is a pre-existing, lane-wide condition this PR does not introduce and cannot close from here — federation quorum broadcast itself is not sal-gated, only its durability net is. Flagging it because "#2446 is closed" is true for a `--features sal` build and not yet true for the release artifact.

## Crossroads protocol

`T1` fires — `FederationDlqSink` gains a public trait method (`expand_erasure_sentinel`) and a public enum (`SentinelExpansion`). The design decision it implements was taken by the **9-lens 3x3 adversarial vote, option B, 8 of 9 lenses** (citing `4d3ea1c5`) and is recorded above; the trait shape is the mechanical consequence of it. No `T3` (no gate posture changes — every new path is best-effort and fail-safe in the never-destroy direction), no `T4` (no on-disk format change; no migration).

## AI involvement

Implemented end-to-end by Claude Opus 5 under Lane G (federation reliability), on operator authority. Codegraph used for navigation. Every load-bearing claim here was verified against the source at `03bbd556` rather than inferred: the single `broadcast_delete_quorum` caller, the `#[cfg(feature = "sal")]` gating of `push_dlq` and the release workflow's feature set, the zero `forget_tombstone` references under `src/federation/`, the verbatim id reuse in `restore_archived`, the `DRY_RUN_PREVIEW_CAP` truncation of `deleted_ids`, and the absence of any `DELETE FROM federation_push_dlq` in `src/`.


alphaonedev:

**Control:** Capacity federation — MCP/CLI erasure replicates via durable outbox (#2446), not only HTTP `/sync/push` delete lane.

### Code
- New `federation/erasure_outbox.rs` + DLQ hardening; MCP/CLI delete/forget enqueue.
- Drain expands sentinel → peer rows; restore supersedes queued erasure.
- Unfederated: erasure succeeds, queues nothing (read-only boot marker path).
- Rebased onto `b3096c15` (post-#2662/#2663); CHANGELOG keep-both.
- Local: `federation_erasure_replication_2446` **9/9 pass**.

### Security / integrity
- Closes silent non-replication of MCP/CLI erasures (GDPR/credential paths).
- Complements #2662 delete-lane DLQ (HTTP push path).

### Merge bar
- Green required checks; merge commit only; close #2446 after merge.

Issue #2500, opened 2026-07-30
[ci-reliability] tests/e2_post_ship_dry_run.rs runs a NESTED cargo build and false-reds the Postgres feature gate — the e1 prebuild fix (env row 117) was never applied to e2


alphaonedev:
## Summary

`tests/e2_post_ship_dry_run.rs` invokes a **nested `cargo build`** of the sibling workspace tool crate `ai-memory-post-ship-converge` from inside the test process. Under CI resource pressure that build fails, the test panics, and a **required** status check goes red on a PR that touched nothing related.

This is the same defect class as the documented e1 problem — and e1's fix was never ported to e2.

## Observed

`ci.yml` run `30518877707`, PR #2497 (a CWE-284 federation security fix), head `3c3878da`:

```
---- e2_dry_run_emits_well_formed_envelope stdout ----
thread 'e2_dry_run_emits_well_formed_envelope' panicked at tests/e2_post_ship_dry_run.rs
cargo build for ai-memory-post-ship-converge failed

---- e2_dry_run_supports_brew_install_method stdout ----
thread 'e2_dry_run_supports_brew_install_method' panicked at tests/e2_post_ship_dry_run.rs
cargo build for ai-memory-post-ship-converge failed

test result: FAILED. 1 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out
```

**`Postgres feature gate` → `failure`.** It is a required status check on `release/v1.0.0`, so this blocks the merge.

PR #2497's diff touches `src/federation/*`, `src/handlers/federation_*`, `src/store/*`, its own two test files, and three docs. **Zero** files under `tests/e2_post_ship*`, `tools/`, or the `ai-memory-post-ship-converge` crate.

Collateral damage worth noting: because the first failing binary aborts the invocation, **the run never reached `tests/federation_delete_ns_scope_2488_pg.rs`** — the binary that proves the security fix on postgres. A false red here doesn't merely annoy; it silently withholds the evidence the PR exists to produce.

## Root cause

`tests/e2_post_ship_dry_run.rs:100`:

```rust
.expect("invoke cargo build for ai-memory-post-ship-converge");
```

A nested `cargo build` inside a test, contending for the same CPU, disk, and cargo lock as the suite running it.

## The fix already exists for e1 and was not applied here

`tests/e1_orchestration_dry_run.rs:61-85` solves precisely this:

```rust
[...]

Issue #1989, opened 2026-07-12
[test-infra][flake] Intermittent full-suite wedge: lib suite spins (~130% CPU) with handlers::tests threads alive; ~1-in-4 on loaded/idle host; invisible to CI


alphaonedev:
Found during the v1.0.0 Gate-1 train (epic #1940, iterations 36-38), filed per the testing-loop discipline (an intermittent test is a real bug).

**Symptom:** `AI_MEMORY_NO_CONFIG=1 cargo test --lib` (parallel, default threads) intermittently never completes — the test process spins at ~130% CPU (56 min CPU observed over 42 min wall) with ~26 threads alive: 14 named `handlers::tests`, 3 `llm::perf9_asyn*`, 3 `cli::shell`, 1 `mcp::auto_tag`, 1 `llm::wiremock*`. The last completed log line was `subscriptions::tests::issue_1253_dlq_overflow_cap_refuses_past_max` in BOTH observed wedges (likely coincidental alphabetical adjacency, not the culprit — it passes serially).

**Evidence:**
- Wedge 1: merged stage2+quarantine tree (pre-stage-3), host running 3 concurrent full suites — so NOT introduced by stage 3.
- Wedge 2: stage-3 stack, host otherwise IDLE.
- Non-repro: `handlers::tests` serial = 405/405 in 52s; immediate parallel re-run of the identical tree = 6441+/0 green in ~406s. Frequency ≈ 1-in-4.
- CI never sees it (CI serializes postgres suites and runs on fresh runners; all pushes 8/8 green).

**Suspects to investigate:** a lock-order or async-runtime interaction among handlers HTTP tests + llm wiremock/perf9 tests under high thread interleave; possibly a bounded-channel/semaphore wait with a spinning poll. The deferred_audit supervisor ERROR ('drainer task panicked … receiver moved') from the mock-panic test appears nearby in one log and may poison a shared runtime.

**Repro recipe:** loop `timeout 700 env AI_MEMORY_NO_CONFIG=1 cargo test --lib -j 6` until exit=124; then inspect `/proc/<pid>/task/*/comm` for the alive set.

**Not tag-blocking by itself** (CI green, deterministic serial runs green) but must be root-caused or mitigated (e.g. libtest timeout wrapper) before the Gate-3 review rounds rely on repeated local full-suite runs.

alphaonedev:
[...]

PR #2674, opened 2026-08-01
ci(#2657): hoist the sal-postgres test compile out of the 2100s watchdog window

### LINKED ISSUE #2657: [ci-availability] Postgres feature gate is running close enough to the #1492 2100s watchdog cap to fail intermittently on unrelated PRs (exit 124, zero test failures in 30,604 lines)
## The observation

`Postgres feature gate` failed on an intermediate head of PR #2642 with **exit code 124** — the `#1492` watchdog killing the suite at its **2100 s** cap, cut off mid-`webhook_http_parity.rs`.

Across **30,604 log lines** there are **zero** occurrences of `test result: FAILED` or `failures:`. Every `test result:` line reads `0 failed`.

**Duration exhaustion, not a defect.** It passed on re-run at the next head with no code change to that suite.

## Why this is worth an issue rather than a shrug

The suite is now running **close enough to the 2100 s cap to fail intermittently**, and when it does it fails on PRs that contributed nothing to it. #2642 added zero Rust and still ate a red.

That has three costs, in increasing order of seriousness:

1. **Wasted cycles.** Under `strict: true` every red forces a re-run, and every merge invalidates the other open PRs.
2. **Misattributed failure.** A red square on a required context reads as "this PR broke postgres." Diagnosing it correctly took reading 30,604 log lines to establish a negative — that no test actually failed. Nobody does that twice; the second time it will be dismissed as flake.
3. **The failure mode this campaign keeps finding.** A signal that means something other than what it appears to mean. Here the red is real (the job did fail) but its *cause* is invisible unless you check the exit code and count the absent failures. An operator triaging by colour learns the wrong lesson.

## What makes it urgent rather than cosmetic

The margin is shrinking from both directions: the suite grows as coverage is added, and this campaign is adding postgres-executed regression tests deliberately — several lanes are under instruction to test against the live PG 18.4 + AGE + pgvector tier because sqlite-only correctness is not correctness. **Doing the right thing on test coverage makes this worse.**

Once it crosses the cap consistently, a required context becomes permanently red for reasons unrelated to any change, and the branch wedges.

## Suggested directions (not a decision)

- **Measure first.** Publish the current wall-clock and the per-test distribution; the cap was set at #1492 without the suite being near it.
- **Split the suite** so the watchdog bounds a smaller unit, or shard across jobs.
- **Distinguish the exit code in the UI.** A watchdog kill (124) is not a test failure and should not present identically — this is the same "a control's report does not match what happened" class as #2641.
- **Raise the cap only with evidence**, and only after the distribution is known. Raising a timeout to silence a signal is how a real slowdown gets hidden.

## Provenance

Found by the lane that fixed #2635/#2636 (PR #2642, merged `fc70789f`). It declined to classify the red as flake and produced the log evidence instead, which is why the diagnosis exists at all.

alphaonedev:
## What

`.github/workflows/ci.yml` carries two near-identical test-runner shell functions: `run_tests` in the `check` job and `pg_test` in the `Postgres feature gate` job. #1989 hoisted compilation OUT of the timed window in `run_tests` — citing exactly this reasoning in its comment — and never touched its twin. The postgres gate's 2100s cap has therefore been measuring `compile + run` rather than `run`.

This adds `cargo test --no-run "$@" || return "$?"` as the first statement of `pg_test`, mirroring `run_tests`.

## The measurement

Run `30718303500` attempt 2:

| | |
|---|---|
| timed invocation begins | `22:16:43` |
| first `Running tests/…` | `22:24:41` |
| **compile inside the window** | **478s — 23% of the 2100s cap, 316 crates** |
| tests then ran | 1605s before SIGTERM |
| total | 478 + 1605 = 2083s vs the 2100s cap |
| binaries that ran | **547 of 550** |

The kill landed 99% of the way through. Only `wire_check_sole_path_pin`, `wt_1_a_schema_migration`, `wt1c_mcp_atomise` and `zeroize_secret_holders_1258` never started. Nothing hung — and the watchdog reported it as a hang, so the misfire actively misdirects triage.

The failure mode is silent and **growing**: compile time rises with every test file added, so each new binary shrinks the runtime budget the cap actually grants.

## What is deliberately NOT changed

The 2100s cap is untouched. Raising it would mask the growth curve and need raising again; the cap must keep timing hang-detection only. `--no-fail-fast` (#2500) and the exit-124 `::error::` branch are also untouched. The `ci.yml` diff is a pure insertion — 33 added lines, 0 removed, 0 modified.

## Evidence

**1. YAML validity.** `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` → parses, 11 jobs. Structural assertions off the parsed tree: the `postgres-feature` step contains the prebuild; `2100` still occurs 8×; `--no-fail-fast` still occurs 4×.

**2. Shell-function proof.** The step body was extracted **verbatim from the parsed YAML** (not hand-copied) and driven over a stub `cargo` on `PATH` under `.local-runs/`:

```
CASE allpass (TEST_IMPACT=__ALL__)                    -> step exit 0
  cargo test --no-run      --features sal-postgres -- --test-threads=1
  cargo test --no-fail-fast --features sal-postgres -- --test-threads=1

CASE allfail (stub exits 101 on --no-run)             -> step exit 101
  cargo test --no-run      --features sal-postgres -- --test-threads=1
  (stub: exiting 101)
  << no second invocation — short-circuited >>

CASE impact (TEST_IMPACT=<two binaries>)              -> step exit 0
  cargo test --no-run      --features sal-postgres --lib -- --test-threads=1
  cargo test --no-fail-fast --features sal-postgres --lib -- --test-threads=1
  cargo test --no-run      --features sal-postgres --test A --test B -- --test-threads=1
  cargo test --no-fail-fast --features sal-postgres --test A --test B -- --test-threads=1
```

`--no-run` fires **first** in every branch, receives the identical `"$@"`, and a non-zero prebuild short-circuits before the timed invocation — so a compile failure still fails the job rather than falling through to a `timeout` run. A second pass had the stub log its parent process, confirming the prebuild's parent is the shell while the timed run's parent is `timeout --signal=TERM --kill-after=60 2100 cargo test --no-fail-fast …`: the wrapper is intact and now wraps only the run.

**3. `"$@"` expansion — verified, not assumed.** On cargo 1.96.0 against a throwaway 3-binary crate, all three pg call-site arg shapes:

| invocation | exit | tests executed |
|---|---|---|
| `--no-run --features sal-postgres -- --test-threads=1` | 0 | 0 |
| `--no-run --features sal-postgres --lib -- --test-threads=1` | 0 | 0 |
| `--no-run --features sal-postgres --test a --test b -- --test-threads=1` | 0 | 0 |
| same, with a deliberate compile error | **101** | — |

`cargo test --no-run` accepts and ignores the trailing harness args; `grep -cE '^(running\|test result)'` on its output is `0`, so the flag is genuinely inert rather than silently running anything.

**4. `scripts/check-required-contexts.sh`** — exact output:

```
check-required-contexts: OK (release/v1.0.0: every mirrored required context maps to a reporting job; no matrix+if wedge; no decider with a job-level if; no path-filtered carrier; every needs-classify step guarded; no dual-trigger cancelled-duplicate carrier; every job in ci.yml c8-precheck.yml coverage.yml declared required or dated-and-tracked as not-required)
```

Exit 0 — no job names, `if:` conditions, or concurrency keys were touched.

## The gate (Section D)

The defect class is **twin drift**: two copies of one discipline, one fixed and one silently left behind. Fixing `pg_test` alone leaves that class open the next time a runner is added, so the rule is stated over **every** watchdog-wrapped `cargo test` invocation rather than over the two functions that exist today — each must be preceded, *inside its own shell function*, by a short-circuiting `--no-run` prebuild.

Run against the **pre-fix** `ci.yml` it fails correctly and names the culprit:

```
FAIL  D: 1 of 2 watchdog-wrapped 'cargo test' invocations compile INSIDE the timed window
      no short-circuiting `cargo test --no-run "$@" || return "$?"` between the
      function opener (line 952) and the timed invocation (line 983)
```

An R-203 leg strips every prebuild from a throwaway copy and requires the scan to catch all of them, so a regex that silently stopped matching cannot leave the rule vacuously green. Suite: **31/31 PASS** (was 29/29).

## What this evidence does and does not establish

**Established.** The prebuild is invoked first, with identical arguments, outside the `timeout` wrapper; a failing prebuild aborts the step; `cargo test --no-run` tolerates the trailing harness args at every call site; the workflow still parses; no other control in `pg_test` changed.

**Not established.** That the postgres gate now completes within 2100s of *runtime*. That is a claim about a live CI run, and the run that would validate it **is the postgres gate on this PR** — which is exactly the gate being changed. The measurement above says the tests themselves ran 1605s of the 2100s budget with 3 binaries left, so recovering the 478s should leave real headroom; but until this PR's own Postgres gate reports green, that is a projection from one attempt's timings, not a verified result. It is also a projection from *one* run: compile time varies with cache state, and the 478s figure is a single sample, not a distribution.

**Deliberately out of scope.** This does not reduce total job wall-time — the compile still happens, just outside the timed window, still bounded by the job-level `timeout-minutes: 45`. If the suite's genuine *runtime* later approaches 2100s on its own, that is a real signal about test growth and should be met with test-level work, not a cap bump. Keeping the cap fixed is what preserves that signal.

Closes #2657

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -952,6 +952,39 @@ jobs:
           pg_test() {
             local label="$1"; shift
             local rc=0
+            # #2657: COMPILE the test binaries OUTSIDE the watchdog first, so the
+            # 2100s per-invocation cap times test RUNTIME (the hang-detection
+            # purpose) rather than compile+run wall-time. This is the same fix
+            # #1989 applied to `run_tests` in the `check` job above; that issue
+            # fixed one of the two near-identical runners and left this twin
+            # untouched, so the postgres gate kept timing compile+run.
+            #
+            # MEASURED on run 30718303500 attempt 2: the timed invocation began
+            # 22:16:43 and the first `Running tests/...` line did not appear
+            # until 22:24:41 — 478s (23% of the 2100s cap) spent compiling 316
+            # crates INSIDE the window. Tests then ran 1605s before the kill;
+            # 478 + 1605 = 2083s against the 2100s cap, and the SIGTERM landed
+            # 99% of the way through (547 of 550 binaries had run; only
+            # wire_check_sole_path_pin, wt_1_a_schema_migration, wt1c_mcp_atomise
+            # and zeroize_secret_holders_1258 never started). Nothing hung — the
+            # watchdog was measuring the build.
+            #
+            # The failure mode this closes is a silent, GROWING one: compile time
+            # rises with every test file added, so each new binary shrinks the
+            # runtime budget the cap actually grants. Hoisting the build is the
+            # correct fix rather than raising the cap, which would only reset the
+            # same curve and have to be raised again; the cap must keep timing
+            # hang-detection alone. The job-level `timeout-minutes` still bounds
+            # the build, and a genuine runtime hang still trips the 2100s below.
+            #
+            # `|| return "$?"` is load-bearing: a compile failure must fail the
+            # job here, not fall through into a `timeout` run that would report a
+            # confusing second error. `cargo test --no-run` accepts (and ignores)
+            # the trailing `-- --test-threads=1` that every call site passes —
+            # verified on cargo 1.96.0 for all three arg shapes used below
+            # (bare, `--lib`, and repeated `--test <name>`): exit 0, zero tests
+            # executed.
+            cargo test --no-run "$@" || return "$?"
             # `--no-fail-fast` (#2500) — DIAGNOSTIC COMPLETENESS. Without it,
             # cargo exits after the FIRST failing test executable and every
             # later binary is never run. That is not hypothetical: on PR
--- a/scripts/test/test-ci-workflow-invariants.sh
+++ b/scripts/test/test-ci-workflow-invariants.sh
@@ -558,6 +558,103 @@ else
     fi
 fi
 
+# ===========================================================================
+# SECTION D — #2657: the watchdog must time test RUNTIME, not compile+run.
+#
+# THE DEFECT. `ci.yml` carries TWO near-identical test-runner shell functions:
+# `run_tests` in the `check` job and `pg_test` in the `Postgres feature gate`
+# job. #1989 hoisted compilation OUT of the timed window in `run_tests` and
+# never touched its twin, so for the whole of the v1.0.0 epic the postgres
+# gate's 2100s cap measured compile+run.
+#
+# MEASURED on run 30718303500 attempt 2: the timed invocation began 22:16:43
+# and the first `Running tests/...` line appeared at 22:24:41 — 478s (23% of
+# the cap) spent compiling 316 crates INSIDE the window. Tests then ran 1605s
+# before the SIGTERM; 478 + 1605 = 2083s against a 2100s cap, and the kill
+# landed 99% of the way through (547 of 550 binaries had run). Nothing hung.
+#
+# WHY IT IS A SEPARATE INVARIANT FROM SECTION B. B is about diagnostic
+# completeness — a run that DID report should reach the binary carrying the
+# proof. This is about a MEASUREMENT being of the wrong quantity: the cap is
+# hang-detection, and a hang is a property of the RUN. Worse, the error it
+# produces is a `::error::[#1492 watchdog] ... exceeded the timeout` naming a
+# hang that did not occur, so the misfire actively misdirects triage.
+#
+# WHY IT MUST BE A GATE RATHER THAN A FIX. The failure mode is TWIN DRIFT: two
+# copies of one discipline, one fixed and one silently left behind. Fixing
+# `pg_test` alone would leave the class open the next time a runner is added.
+# So this rule is stated over EVERY watchdog-wrapped invocation, not over the
+# two functions that exist today.
+#
+# WHY THE SHORT-CIRCUIT IS PART OF THE RULE. The prebuild must be
+# `... || return "$?"`: a compile failure has to fail the job at the prebuild,
+# not fall through into a `timeout` run that reports a second, confusing error.
+# ===========================================================================
+echo "  --- Section D: #2657 watchdog measures RUNTIME, not compile+run ---"
+
+# d_scan <ci.yml-path> — prints "<bad>/<total>" over the watchdog-wrapped
+# `cargo test` invocations in that file. A wrapped invocation is COVERED when a
+# short-circuiting `--no-run` prebuild appears between the opener of its own
+# enclosing shell function and the invocation itself. Factored out so the R-203
+# regression leg below can run the IDENTICAL logic against a mutated copy.
+d_scan() {
+    local file="$1" total=0 bad=0 line w opener prebuild
+    while IFS= read -r line; do
+        [ -n "$line" ] || continue
+        total=$((total + 1))
+        w="${line%%:*}"
+        # Nearest enclosing `name() {` opener above the invocation.
+        opener="$(awk -v w="$w" '
+            NR < w && /^[[:space:]]+[A-Za-z_][A-Za-z0-9_]*\(\)[[:space:]]*\{[[:space:]]*$/ { n = NR }
+            END { print n + 0 }' "$file")"
+        if [ "$opener" -eq 0 ]; then
+            bad=$((bad + 1))
+            printf '         watchdog line %s is not inside a shell function\n' "$w" >&2
+            continue
+        fi
+        prebuild="$(awk -v a="$opener" -v b="$w" '
+            NR > a && NR < b && /cargo test --no-run "\$@" \|\| return "\$\?"/ { n = NR }
+            END { print n + 0 }' "$file")"
+        if [ "$prebuild" -eq 0 ]; then
+            bad=$((bad + 1))
+            printf '         no short-circuiting `cargo test --no-run "$@" || return "$?"` between the function opener (line %s) and the timed invocation (line %s)\n' \
+                "$opener" "$w" >&2
+        fi
+    done <<< "$(grep -nE '(TIMEOUT_BIN|--kill-after=60)[^|]*cargo test' "$file" || true)"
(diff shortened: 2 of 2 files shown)


alphaonedev:
## Fable 5 — code review + security audit

**Verdict: APPROVED pending the postgres gate.** I re-ran your central claims myself rather than accepting the report; all of them hold.

### What I verified independently

I extracted your `d_scan` and drove it against both files directly:

| file | bad/total |
|---|---|
| `release/v1.0.0` (pre-fix) | **1/2** |
| this branch | **0/2** |
| this branch, every prebuild stripped | **2/2** |

On the pre-fix file it names the culprit exactly — *"no short-circuiting `cargo test --no-run "$@" \|\| return "$?"` between the function opener (line 952) and the timed invocation (line 983)"*. The gate catches the defect, clears on the fix, and catches a full strip. It is load-bearing, not decorative.

I also checked placement in the raw file rather than the diff: `pg_test()` 952 → `local rc=0` 954 → prebuild 987 → timed invocation 1016, mirroring `run_tests` at 655 → 664 → 675. And `git diff` is **0 deletions across the entire PR** — the 2100s cap, `--no-fail-fast`, and the exit-124 branch are untouched, as claimed.

### The blind-scan hazard — you closed it, and I confirmed it the hard way

While constructing a counterfactual I fat-fingered a `sed` delimiter and produced a mutant with **zero** watchdog matches. `d_scan` returned `0/0` — which in a `bad/total` format reads as *clean*. That is precisely the #2444 shape this campaign keeps finding: a control reporting success over a subject it never examined.

Your line 630 already fails it explicitly — *"found no watchdog-wrapped 'cargo test' invocation … this guard has gone blind"* — and the R-203 leg is separately gated on `d_total > 0`. So the hazard is closed twice over. I mention it because the accident independently reproduced the exact failure mode your guard anticipates, which is about as good evidence as that guard can get.

### Security audit

No weakening. Zero deletions, no gate relaxed, no job name / `if:` / concurrency key touched (`check-required-contexts.sh` exit 0, which I take as the relevant jurisdiction). `cargo test --no-run` compiles the same crates and runs the same build scripts that already ran inside the window — the execution surface is identical, only the timer around it moves.

**One posture change your PR body does not state, and should:** compile is no longer bounded by 2100s; it is now bounded only by the job-level `timeout-minutes: 45`. A pathologically hung *build* would therefore run ~28 minutes longer before being killed than it would have pre-change. I judge this correct and accept it — the watchdog exists to detect a hung *test*, `timeout-minutes` is the right bound for a build, and this is exactly the posture `run_tests` has carried since #1989 — but it is a real difference and belongs in the record rather than being absorbed into "no other control changed."

### On your honest boundary

You were right to state it, and right about where it falls. Your evidence establishes the mechanism — prebuild fires first, outside `timeout`, same args, short-circuits on failure. It does not establish that the gate now finishes inside 2100s of *runtime*; only the postgres gate on this PR can. I am treating that run as the merge condition and have a monitor on it.

Your reasoning that a future approach to the cap should be met with test-level work rather than a cap bump is correct and is the standing position: the cap must keep measuring one quantity, or it stops being a signal at all.

### Merge condition

`Postgres feature gate` green on `7ad1a517`. Nothing else outstanding.

alphaonedev:
## Measured on this PR's own Postgres gate — the projection is now a measurement

The PR body said the one thing this evidence could **not** establish was that the gate completes within 2100s of *runtime*, because the run that would validate it is the postgres gate on this PR. That run has now completed: **`Postgres feature gate` — success** (job `91428625808`, run `30722557128`).

Timings read from the job log:

| event | timestamp | |
|---|---|---|
| step begins | `23:08:01` | |
| `Compiling ai-memory` (prebuild) | `23:08:23` | |
| **prebuild ends** — `Finished \`test\` profile in 6m 28s` | `23:14:29.69` | **388s hoisted OUT of the window** |
| timed invocation's own re-check — `Finished \`test\` profile in 2.91s` | `23:14:36.21` | nothing left to build |
| first `Running unittests` | `23:14:36.61` | |
| last `test result: ok` | `23:38:49.60` | |

So the two invocations are cleanly separated in the log: the 6m28s build lands under the **prebuild**, and the timed invocation's own cargo pass finds nothing to do and finishes in **2.91s** before the first test binary starts. That is the fix working, visible directly rather than inferred.

**What the watchdog now measures:**

| | seconds | % of the 2100s cap | headroom |
|---|---|---|---|
| timed window today (post-fix) | **1456s** | **69%** | 644s (31%) |
| same run, pre-fix (compile+run) | 1844s | 88% | 256s (12%) |

The change converted a 12% margin into a 31% margin on this run by removing the build from the measurement — it did not make anything faster, which is the intended behaviour.

**Suite outcome:** 554 test binaries started, 559 `test result: ok`, **0 `test result: FAILED`**, 0 `##[error]` annotations, no watchdog misfire. (The one grep hit for `exceeded the 2100s per-invocation timeout` at line 984 is the cyan workflow-source echo GitHub prints when rendering the `run:` block at `23:08:01`, not an emitted annotation.)

### Corrections and remaining boundaries

- **Compile time was 388s here, not the 478s** measured on run `30718303500`. Same defect, different cache state — which is exactly the variance I flagged. Treat 388-478s as the observed range from two samples, not a constant.
- **Still two samples.** One passing run does not establish a distribution. What it does establish is the mechanism: the build is provably outside the timed window in this log.
- **Unchanged:** total job wall-time (~31 min) is not reduced — the compile still happens, just outside the cap, bounded by `timeout-minutes: 45`. If genuine runtime later climbs toward 2100s, that is a real signal about test growth and should be met with test-level work, not a cap bump.

CI overall: **30 pass, 0 fail**, 2 long jobs still running (`Check (windows-latest)`, `Per-Module Coverage Thresholds`), 1 skipping (bench baseline). `Required-context + classify-base soundness gate` green with the new Section D live: `ci.yml invariants: 31/31 PASS`.

PR #2686, opened 2026-08-04
feat(#2676): ai-memory features self-report for Gate 3 packaging

alphaonedev:
## Summary

Gate 3 packaging / #2676: the binary can **self-report compiled cargo features** so provision/measure cannot false-succeed on `--version` alone against a wrong feature set.

## Control

- `src/build_features.rs` — compile-time sorted feature table (`cfg!`)
- `ai-memory features` — human (`version:` / `features:`) + `--json`
- `scripts/assert-compiled-features.sh` — harness pin (`--require sal-postgres` etc.)
- CLI SSOT lockstep: `EXPECTED_CLI_SUBCOMMANDS_DEFAULT` 89→90, `_SAL` 91→92 + docs

## Verification

- `cargo test --lib build_features` — 3 passed
- `bash scripts/check-docs-vs-ssot.sh` — PASS (cli_default=90, cli_sal=92)
- `./target/debug/ai-memory features` → `features: sqlite-bundled`
- `scripts/assert-compiled-features.sh ./target/debug/ai-memory` — OK

## Security of the control

Does not widen attack surface (read-only, no DB/network). Closes false-success packaging risk for Gate 3 measurement.

Refs: #2676 #2682

## AI involvement

Authored under AI NHI orchestration (Grok 4.5).

Issue #2682, opened 2026-08-03
EPIC: Graph Engineering × Grok Build — AI NHI enterprise certifies ai-memory v1.0.0 GA (ready-to-tag)


alphaonedev:
## North Star

**AI NHI enterprise-certifies ai-memory v1.0.0 GA as ready-to-tag** — as quickly as possible under full AI NHI engineering authority — using **Grok Build + Graph Engineering** as the production development process.

A Fortune 500 would bet the farm on published claims, real PostgreSQL + AGE + pgvector, no silent data loss, no false-success controls. Scale claim 500–1000 agents/module — not 1M+.

**Operator-gated only:** cutting the `v1.0.0` tag + publish (crates.io / GHCR / Homebrew / COPR).  
**Never** create/push tags. **Never** `workflow_dispatch` release.yml.

**Authority (2026-08-03):** YES — AI NHI 100% of engineering decisions, autonomous run, 100% decision-making authority. DO testing fully approved (tokens `~/.env`).

---

## Epic kickoff

| Field | Value |
|-------|--------|
| **Kicked off** | 2026-08-03 (Grok 4.5 NHI orchestrator) |
| **Tree** | `/home/fate_two/v07/v09-dev` |
| **Branch tip at kickoff** | `release/v1.0.0` @ `b766713b` |
| **Codegraph pin** | ALWAYS `projectPath="/home/fate_two/v07/v09-dev"` — never a worktree |
| **Identity** | AlphaOne \<Justin@alpha-one.mobi\>, SSH-signed commits (Good `G`) |
| **Merge policy** | Merge commits only; never squash/force-push; stage explicit paths; Co-Authored-By every commit |
| **PR gate** | Code review **and** security audit of the **control** before every merge to `release/v1.0.0` |

### Tier-0 already landed this campaign

- #2679 silent wrong-store → **#2680** merged `48d547aa` → issue CLOSED  
- #2678 federation DLQ compiled out → **#2681** merged `b766713b` → issue CLOSED  

---

### Layer A — Outer cycle (`/goal`)

Autonomous objective runs across rounds. Completes only after **independent evidence review** of ready-to-tag certification — **not** after cutting the tag.

### Layer B — Campaign graph (stateful nodes + edges)

```
SessionStart → [Gate1 structural confinement]
            → [Gate2 claims train: #2655→#2656→#2668→#2659 LAST]
            → [Security/data capacity: #2643 #2644 #2662 #2663 #2673]
[...]

Issue #2676, opened 2026-08-02
[release][data-integrity][claims] the shipped binary is built WITHOUT --features sal, so the entire federation push-DLQ is compiled out — every fanout failure is silently lost in the released artifact


alphaonedev:
Surfaced during #2446 (PR #2673) and **independently verified**. This is not a lane-G defect — it invalidates a shipped guarantee across the whole federation surface, and it has been true of every released artifact since #933 landed in v0.7.0.

## The chain, verified

1. **The shipped binary uses DEFAULT features.**
   - `.github/workflows/release.yml:132` — `cargo build --release --target ${{ matrix.target }}`. No `--features`.
   - `Dockerfile:34` — `RUN cargo build --release && strip target/release/ai-memory`. No `--features`.
   - (The `--no-default-features --features sqlite-bundled` builds at `release.yml:414`/`584` are the iOS/Android **static FFI libraries**, not the daemon.)

2. **`sal` is not a default feature.**
   - `Cargo.toml:284` — `default = ["sqlite-bundled"]`.

3. **The push-DLQ is `sal`-gated end to end.**
   - `src/federation/mod.rs:131` — `#[cfg(feature = "sal")] pub dlq_sink: Option<Arc<dyn push_dlq::FederationDlqSink>>`. Without `sal` **the field does not exist**.
   - `src/daemon_runtime.rs:5744` — the `spawn_replay_federation_push_dlq` call sits behind `#[cfg(feature = "sal")]`. The replay worker is never spawned.
   - `src/federation/sync.rs` — 4 `#[cfg(feature = "sal")]` gates, including the #933 store-lane landing pass and the #2498 delete-lane landing pass.

## The consequence

In the released `ai-memory` binary and the released Docker image:

- `broadcast_store_quorum` and `broadcast_delete_quorum` **never enqueue anything**.
- The replay worker **never runs**.
- The `federation_push_dlq` table still exists (migrations are not feature-gated), so an operator inspecting the schema sees the mechanism and finds it permanently empty.

**Every per-peer federation fanout failure is silently lost in the shipped product** — a peer that is down, deadline-evicted, or that refuses the item inside a 200. That is exactly the pre-#933 behaviour that #933 was filed to fix.

## Why this is a claims failure, not only a build-config one

`docs/TROUBLESHOOTING.md:339` states, unconditionally:

> Failed quorum pushes land in the `federation_push_dlq` table and a [replay worker retries them]

[...]

PR #2700, opened 2026-08-04
fix(#2676): ship --features sal on release and Docker binaries

alphaonedev:
## Summary
Packaging residual from Gate3 cert note: release/Docker channel no longer ships default-only (`sqlite-bundled`) binaries that omit `sal`.

### Control
1. **Dockerfile** — `cargo build --release --features sal` + `assert-compiled-features.sh --require sqlite-bundled --require sal`
2. **release.yml** — matrix binaries built with `--features sal` + same assert after build
3. **Cert note** — packaging residual updated; Gate packaging → PASS

### Not in this PR
- `sal-postgres` on every multi-OS release target (sqlx/pgvector weight) — remains plan-c / explicit asserted build (Gate3 measure tip)
- Tag cut / `workflow_dispatch` release.yml — **never** (operator only)

### Security / review
Fail-closed packaging: wrong feature set fails the image build and release job rather than shipping a silent false-success artifact.

Refs: #2676 #2682 #2686

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,8 +128,22 @@ jobs:
 
       - uses: Swatinem/rust-cache@v2
 
+      # #2676 packaging residual — release channel must not ship default-only
+      # binaries that omit `sal` (federation/SAL surface). Assert via the same
+      # pin Gate3 provision uses (`scripts/assert-compiled-features.sh`).
+      # Note: `sal-postgres` is NOT enabled here (sqlx/pgvector weight + native
+      # deps on every matrix OS); operators needing PG use an asserted
+      # sal-postgres build or the plan-c image. Agents never workflow_dispatch.
       - name: Build release binary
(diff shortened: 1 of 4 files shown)


alphaonedev:
## Code + security review (control) — AI NHI

**Control (#2676 packaging residual):** Release channel must not ship a binary whose compiled feature set is weaker than the federation/SAL surface Gate3 measured.

| Surface | Change |
|---------|--------|
| Dockerfile | `--features sal` + assert `sqlite-bundled,sal` |
| release.yml matrix | same build + assert |
| sal-postgres on all OS matrix | deliberately **not** (sqlx weight); PG uses plan-c / explicit build |

**Fail-closed:** assert script non-zero fails image/release job — no silent false-success artifact.

**Never** tag / never workflow_dispatch release.yml.

Merge when required checks green (Dockerfile build is load-bearing).

Refs: #2676 #2682

Issue #2677, opened 2026-08-02
host_is_loopback exactness is load-bearing for the plaintext-peer gate and untested


alphaonedev:
Follow-up to #2477 / PR #2664 (merged `9916165d`). **Not a live defect** — behaviour is correct at HEAD. This is about the safety net around it.

## The load-bearing property

`tls::validate_peer_url_scheme` exempts `http://` from the plaintext refusal when `host_is_loopback(host)` is true. That function is exact equality:

```rust
host == "127.0.0.1" || host == "::1" || host == "localhost"
    || host == "0:0:0:0:0:0:0:1" || host == "[::1]"
```

Exactness is what makes the exemption safe. Verified today:

| input | today |
|---|---|
| `http://127.0.0.1.evil.com:9077` | refused |
| `http://2130706433/` (decimal 127.0.0.1) | refused |
| `http://127.0.0.2:9077` | refused |

## The gap

`tests/federation_plaintext_peer_2477.rs` covers only happy paths — `127.0.0.1`, `[::1]`, `localhost`, and two non-loopback refusals. **No spoof shape is tested.**

A later refactor that "tidies" the chain into `host.starts_with("127.")` or `host.contains("localhost")` makes `http://127.0.0.1.evil.com` an **accepted plaintext peer** — replicated memory content in the clear to an attacker-chosen host, since federation is not end-to-end encrypted (#1968) — and **every existing test still passes**.

That is the same shape as the #1989/#2657 twin drift and the #2444 false-success class: the control is right, and nothing would notice if it stopped being right.

## Asked for

Regression tests pinning refusal for at least:

- `http://127.0.0.1.evil.com:9077` — suffix-spoof, the dangerous direction
- `http://localhost.evil.com:9077` — same, `contains` variant
- `http://evil.com/?x=127.0.0.1` — loopback string present, host is not
- `http://2130706433/` and `http://0x7f000001/` — encoded loopback; assert the CURRENT refusal so a future "fix" that accepts them is a deliberate, reviewed change rather than a silent widening

Cheap to add and it converts an invariant that currently lives in a reviewer's head into one the suite enforces.

alphaonedev:
## Closed — regression suite on `release/v1.0.0`

**PR:** #2702  
**Merge:** `b4512b8430afac1c9d45cb73c195396bab4fdf08`

Pins exactness of `host_is_loopback` for the #2477 plaintext-peer gate:
[...]

PR #2702, opened 2026-08-04
test(#2677): pin host_is_loopback exactness against spoof shapes

alphaonedev:
## Summary
Post-tag residual **#2677**: suite now enforces exactness of `host_is_loopback` used by the #2477 plaintext-peer gate.

### Control
- Unit: spoof hosts refused; literal loopback accepted
- Unit + integration: `127.0.0.1.evil.com`, `localhost.evil.com`, query-embedded loopback, `127.0.0.2` refused without hatch
- Decimal/hex IPv4 (`2130706433`, `0x7f000001`) normalise via URL parser to `127.0.0.1` — pin **accept** (true loopback)

### Security
False-success class: happy-path tests alone would not notice a weakened predicate. No behaviour change for production control.

Refs: #2677 #2477 #2682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed (plaintext peer gate: loopback exactness is now suite-enforced)
+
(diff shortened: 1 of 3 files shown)


alphaonedev:
## Code + security review (control) — AI NHI

**Control (#2677):** Pin exact equality of `host_is_loopback` so the #2477 plaintext exemption cannot silently widen.

| Case | Expected |
|------|----------|
| Suffix spoof `127.0.0.1.evil.com` / `localhost.evil.com` | refuse |
| Query-embedded loopback string | refuse (host is not loopback) |
| `127.0.0.2` | refuse |
| Decimal/hex IPv4 | accept after URL normalise to 127.0.0.1 (true loopback) |

No production behaviour change — safety net only. Merge when required checks green.

Refs: #2677 #2477

Issue #1824, opened 2026-06-28
[TRACT-gap G7] Contradiction auto-resolved, not conserved (curator hard-DELETEs the loser)


alphaonedev:
## Gap G7 — TRACT §26 canonical-gaps catalog (UNTRACKED → tracked per v0.8.1 W7)

**Priority tier:** P1  ·  **Milestone:** v0.9  ·  **TRACT ref:** L1 permanent-dissent / conserved contradiction

### Gap statement
When two memories contradict, the curator adjudicates and **destroys the loser**. A substrate that aspires to permanent dissent must conserve both sides; the winning Claim should never be silently decided by the curator.

### Anchor(s) (CodeGraph-verified on `release/v0.8.0`)
- `forget_if_superseded` hard-DELETEs the older contradicted memory — `src/autonomy.rs:483` (via `db::delete`, `src/autonomy.rs:538`)

### Fix shape (from the canonical doc)
- conserved `fork_set` + a signed tombstone for the superseded side; the curator never adjudicates the winning Claim (it records the fork, a human/court resolves)

### Provenance
Filed per the **v0.8.1 (Patch 1)** work order §W7 and the §24 prime directive (every gap is a tracked defect — no UNTRACKED gap may rot in a queue; the v0.9.0 epic starts on a clean tracker). Sources: `docs/design/TRACT-v0.8.0-DEVELOPMENT-GAPS-CANONICAL-opus.md`, `docs/reviews/corpus-completeness-21-agent-OPUS.md`, ROADMAP §26. Orchestration hub: #1821.

🤖 Filed by Claude Code (Opus 4.8, 1M context).

alphaonedev:
**Implemented + CI-GREEN on `release/v0.9.0`** (commit 1a606580, signed; full CI matrix success, 0 failures). The curator no longer hard-DELETEs the losing side of a contradiction — `forget_if_superseded` now CONSERVES: retains both memories, writes one canonical signed `contradicts` link, emits one identity-only SUPERSEDE revision leaf (flag-gated on the G6 spine), and marks the loser with a reversible node-local soft-down-weight. No new RecordKind/relation/migration/flag; §26.5-compliant (no silent adjudication — the winner preference is reversible). Design ACCEPTED by 2×5 vote (both waves); full local gate suite green pre-push (416 test suites, clippy --all-targets --features sal-postgres pedantic, hardcoded-literal, docs-vs-SSOT). Ships with v0.9.0.

PR #2730, opened 2026-08-04
fix(#2436): correct postgres soft-loser penalty predicate (dead code)

### LINKED ISSUE #2436: [cross-backend] contradiction soft-loser penalty is DEAD on postgres — writer stamps JSON true, pg predicate tests ->> = '1' which can never match; sqlite twin works
## The defect

The G7 contradiction soft-loser down-weight is **inert on postgres** and **live on sqlite**, because the writer and the postgres reader disagree about the JSON type.

**Writer** — `src/storage/mod.rs:3579,3583`:
```rust
serde_json::Value::Bool(true),
```
stamps `metadata.contradiction_soft_loser` as a JSON **boolean**.

**Postgres consumer** — `src/store/postgres.rs:5677`:
```sql
* (CASE WHEN (m.metadata->>'contradiction_soft_loser') = '1'
```
`->>` extracts a JSONB value **as text**. For JSON `true` that text is `'true'` — never `'1'`. **The predicate can never match.** The `CASE` always falls through and the penalty factor is never applied.

**SQLite consumer** — `src/storage/mod.rs:5939` and `:6544`:
```sql
* (CASE WHEN json_extract(m.metadata, '$.contradiction_soft_loser') = 1
```
works, because SQLite's `json_extract` returns JSON `true` as the integer `1`.

So the same stamped row is penalised on sqlite and served at **full rank** on postgres.

## Second, independent half: the pg penalty is only on ONE lane

Even where the predicate is present, it is on the postgres **search** lane only (`postgres.rs:5677`). The postgres **recall** lanes carry no soft-loser penalty at all. On sqlite the penalty appears on both search (`:5939`) and recall (`:6544`).

So on postgres the down-weight is absent from recall by construction, in addition to being unmatched on search.

## Why this matters more than an ordinary parity bug

The whole point of the conserve pass is that a contradiction's losing side is **retained but demoted** rather than deleted — conserving the record while keeping the stale claim out of the top of the result window. With the penalty inert, **the conserved loser is served at full score**, i.e. a claim the substrate has already determined to be contradicted ranks identically to the claim that superseded it.

That is the failure #2337/#2338 describe — and it is **live and unmitigated specifically on postgres**, which is the backend a multi-tenant / federated deployment must run (MCP stdio is structurally sqlite-only; the postgres path is the daemon/HTTP surface). The backend that scales is the backend where the mitigation does not work.

## Why no test caught it

This is precisely the class #2425 names: `tests/recall_scoring_parity.rs` is green **for the wrong reason**. The shared additive prior dominates both backends' relevance terms, so a penalty that is applied on one backend and silently skipped on the other does not move either ordering enough to break a parity assertion. A parity gate that cannot distinguish "both correct" from "both dominated by a constant" will not catch a divergence of this shape.

## Proposed fix

1. **Make the postgres predicate type-correct.** Either compare the extracted text against `'true'`, or use the JSONB containment/boolean form rather than `->>`. Prefer whichever matches the writer's type without relying on a text coercion — a text comparison against a JSON boolean is fragile in exactly this way.
2. **Normalise the stamp.** The more durable fix is for the writer and both readers to agree on one representation. If the marker stays a JSON boolean, both readers should test it as a boolean; if it becomes an integer, the sqlite reader already matches.
3. **Add the missing pg recall-lane penalty**, so the two backends apply the down-weight on the same set of lanes.
4. **Add a parity test that would actually fail** — assert that a stamped soft-loser row ranks strictly below an equivalent unstamped row, on **both** backends, with the additive prior held constant across the two candidates so the assertion is sensitive to the penalty rather than swamped by the prior.

Point 4 is the load-bearing one: without holding the prior constant, the new test inherits the same blindness as the existing parity gate.

## Verification

Confirmed at `release/v1.0.0` @ `011dd8b9` by direct inspection of all three sites (writer, pg consumer, sqlite consumer). Found during a 2×5 adversarial review of gap coverage for large-scale federated deployments; independently surfaced by a security-lens reviewer and re-verified before filing.

Related: #2337 (conserve pass reachability), #2338 (soft-loser ranking consumers), #2425 (parity gate green for the wrong reason — same root cause for the missed detection).

alphaonedev:
Closes #2436

## Root cause (CB-16 — confirmed postgres correctness defect)

The G7 (#1824) contradiction **soft-loser** recall down-weight was **dead code on postgres**.

`conserve_contradiction` (`src/storage/mod.rs:3771`) stamps the marker as a **JSON boolean `true`** (`serde_json::Value::Bool(true)`). Postgres' `->>` operator returns the JSON value **as TEXT**, so a JSON boolean `true` renders as the string `'true'`. The recall predicate in `PostgresStore::search_with_source_uri` compared:

```sql
-- before (src/store/postgres.rs)
* (CASE WHEN (m.metadata->>'contradiction_soft_loser') = '1'
        THEN {soft_loser_factor} ELSE 1.0 END) AS rank
```

`'true' = '1'` is **always false**, so the multiplicative penalty never applied — an LLM-confirmed-contradicted loser kept outranking its fresh winner at full score on any postgres-backed daemon. The sqlite twins (`src/storage/mod.rs:6202` and `:6807`) use `json_extract(m.metadata, '$.contradiction_soft_loser') = 1`, which matches correctly because SQLite's `json_extract` of a JSON `true` yields the **integer 1**. A cross-backend correctness divergence: sqlite penalized, postgres did not.

## Fix

```sql
-- after
* (CASE WHEN (m.metadata->>'contradiction_soft_loser') = 'true'
        THEN {soft_loser_factor} ELSE 1.0 END) AS rank
```

- Matches the exact value the writer stamps (JSON boolean → `->>` text `'true'`).
- **Robust to an absent marker**: `->>` returns `NULL`, `NULL = 'true'` is `NULL`, so the `CASE ELSE` arm applies factor **1.0** (no penalty, no error) — never a wrong result. (North Star: degrade, never corrupt.)
- The false in-code comment that described the (non-working) behavior is corrected to document the real JSON-boolean shape and the absent-marker semantics.

## Test evidence

New `tests/pg_soft_loser_penalty_2436.rs`:

- **`writer_stamps_soft_loser_as_json_bool`** (always-run, sqlite) — pins the load-bearing contract: `conserve_contradiction` writes the marker as a JSON boolean `true` (not `"1"`/`"true"`/`1`). If the writer ever drifts, this fails alongside the pg predicate so they can't silently diverge again.
- **`sqlite_soft_loser_ranks_below_winner`** (always-run, sqlite) — the reference ordering contract the postgres twin must match.
- **`pg_soft_loser_penalty_applies_2436`** (`#[ignore]` + `sal-postgres`, live-PG) — mirrors the repo's `AI_MEMORY_TEST_POSTGRES_URL` harness: proves the penalty is **applied** for a soft-loser row (ranks below its winner at equal `ts_rank`) and **not applied** for a normal row. Run with `cargo test --features sal-postgres --test pg_soft_loser_penalty_2436 -- --ignored`.

Local verification: `cargo fmt --all --check` clean; clippy green on all four CI legs (default / `sal` / `sal-postgres` / `vectorlite`, `-D warnings -D clippy::all -D clippy::pedantic`); the two always-run sqlite tests pass. `src/store/postgres.rs` stays under its `qual_10` module-size ceiling (33,013 ≤ 33,050). CHANGELOG `[Unreleased]` updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Fixed (HTTP get-standard no longer serves private standard bodies)
+### Fixed (postgres soft-loser recall penalty was dead code)
+
+- **`PostgresStore::search_with_source_uri` now matches the G7 soft-loser marker's real JSON shape** (refs [#2436](https://github.com/alphaonedev/ai-memory-mcp/issues/2436), CB-16). `conserve_contradiction` stamps `metadata.contradiction_soft_loser` as a JSON boolean `true` (`serde_json::Value::Bool`), so postgres' `->>` operator renders it as the TEXT `'true'`. The predicate compared `(m.metadata->>'contradiction_soft_loser') = '1'`, which is ALWAYS false — the multiplicative recall down-weight never applied on postgres, so an LLM-confirmed-contradicted loser kept outranking its fresh winner at full score (a cross-backend divergence: the sqlite twins `json_extract(...) = 1` worked, the postgres mirror did not). Predicate corrected to `= 'true'`; an absent marker still yields `NULL` → factor 1.0 (no penalty, no error). The false in-code comment claiming the penalty applied is corrected. Regression: `tests/pg_soft_loser_penalty_2436.rs` (always-run sqlite value-shape + penalty pins; `#[ignore]`+`sal-postgres` live-PG parity proof).
 
 - **`GET /api/v1/namespaces?namespace=` and path-form `…/{ns}/standard` gate the standard body through `visibility::is_visible_to_caller`** (refs [#2543](https://github.com/alphaonedev/ai-memory-mcp/issues/2543); #959 residual; CWE-284). Pre-fix the sqlite arm passed `caller=None` into `handle_namespace_get_standard` and the postgres arm used `CallerContext::for_admin`, so any caller who could name a namespace received title + content + governance of a default-private standard (the last unfiltered read of that body after #2537 closed the injection path). Both arms now resolve `X-Agent-Id` and share the MCP honesty shape: count-only `standards_withheld`, no id/owner leak. Shared-scope and owner reads unchanged. CLI/MCP stdio `None` trust-all posture unchanged.
 
--- a/src/store/postgres.rs
+++ b/src/store/postgres.rs
@@ -6269,7 +6269,16 @@ impl PostgresStore {
                     -- multiplicatively penalized so it cannot outrank its
                     -- winner at full score (reversible — clearing the
                     -- marker restores rank; sqlite twin in db::recall).
-                    * (CASE WHEN (m.metadata->>'contradiction_soft_loser') = '1'
+                    -- #2436 (CB-16) — `conserve_contradiction` stamps the
+                    -- marker as JSON boolean `true` (serde `Value::Bool`),
+                    -- so `->>` yields the TEXT 'true', NOT '1'. The prior
+                    -- `= '1'` never matched on postgres (dead penalty); the
+                    -- sqlite twins compare `json_extract(...) = 1` because
+                    -- SQLite's json_extract of a JSON true yields integer 1.
+                    -- An absent marker makes `->>` NULL, so `NULL = 'true'`
+                    -- is NULL → the CASE ELSE arm applies factor 1.0 (no
+                    -- penalty, no error).
+                    * (CASE WHEN (m.metadata->>'contradiction_soft_loser') = 'true'
                             THEN {soft_loser_factor} ELSE 1.0 END) AS rank
              FROM memories m
              WHERE m.tsv @@ plainto_tsquery('english', $1)
--- a/tests/pg_soft_loser_penalty_2436.rs
+++ b/tests/pg_soft_loser_penalty_2436.rs
@@ -1,0 +1,290 @@
+// Copyright 2026 AlphaOne LLC
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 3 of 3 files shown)


alphaonedev:

Independently verified the diff at `d2c76c39` (SSH sig good, ED25519).

**Correctness (CONFIRMED):** `conserve_contradiction` stamps the marker as `serde_json::Value::Bool(true)`; postgres `->>` renders a JSONB boolean as text `'true'`, so the prior `= '1'` never matched (dead penalty) while the sqlite twin `json_extract(...) = 1` worked. `= 'true'` is the right shape. Absent marker → `->>` NULL → `NULL = 'true'` is NULL → CASE ELSE → factor 1.0 (no penalty, no error) — correct fail-safe, degrade-not-corrupt.

**Security:** neutral. Recall-ranking down-weight only; the literal is static and `soft_loser_factor` is a compile-time constant interpolated via `format!` — no user input, no injection surface. No auth/visibility/data-integrity change.

**Test adequacy:** strong. `writer_stamps_soft_loser_as_json_bool` pins the writer's JSON-boolean shape (`.is_boolean()`), so the writer and the pg predicate can never silently drift again — this is the load-bearing regression guard. `sqlite_soft_loser_ranks_below_winner` asserts the ordering flips only after conservation. `pg_soft_loser_penalty_applies_2436` (`#[ignore]` + `sal-postgres`, live-PG) holds priority equal so the penalty is the sole differentiator and asserts applied-for-marked / not-for-normal — mirrors the repo `store_parity_gaps` harness.

**rust-skills:** field_names constant used for the JSON key in tests; comment corrected to document real behavior + absent-marker semantics; no unwrap on non-test paths.

Merging via SSH-signed local merge once CI is green (necessary-not-sufficient; review is the sufficient half and it passes).

Issue #2504, opened 2026-07-30
[SECURITY] one malformed character in AI_MEMORY_FED_PEER_ATTESTATION silently disables the entire federated-delete namespace gate — and the WARN says the fallback is default-deny, which is false for this lane


alphaonedev:
## Summary

`PeerAttestationConfig::from_env` falls back to an **empty** config on any parse error. `has_allowlist()` is `!peers.is_empty()`, and that boolean is the master switch for the entire #2447/#2488 federated-delete namespace gate. So a single typo in `AI_MEMORY_FED_PEER_ATTESTATION` turns the gate off completely — and the WARN emitted on the typo tells the operator the fallback is *default-deny*, which is true for the read lane and **false for the destructive one**.

Pre-existing, probed at default config.

## Probe results

Victim row in `secure/ops`; the operator intends a `public/*`-scoped allowlist:

```
X9a  malformed JSON, anonymous push   -> has_allowlist=false  deleted=1   secure/ops row DESTROYED
X9b  malformed JSON, named peer       -> has_allowlist=false  deleted=1   (the #1056 401 is disabled too)
X9c  wrong JSON type                  -> has_allowlist=false  deleted=1
X9d  allowlist == "{}"                -> has_allowlist=false  deleted=1
```

`X9b` is the compounding part: `has_allowlist()` also gates the #1056 TOFU enrollment guard (`src/handlers/federation_receive.rs:952`), so the typo disables **both** the namespace gate and the unenrolled-peer 401 at once. The two controls that were meant to be independent share a single point of failure.

## The two lanes take opposite postures from that one value

```
X11  empty allowlist, peer=None              /sync/since read secure/ops -> DENIED
                                             federated DELETE secure/ops -> ALLOWED
X11  empty allowlist, peer=Some(stranger)    /sync/since read secure/ops -> DENIED
                                             federated DELETE secure/ops -> ALLOWED
```

**"Trust nobody" on the read lane; "trust everybody" on the destructive lane.** From the same misconfiguration, in the same request. An operator who fat-fingers the allowlist gets a *tighter* read posture and a *wide-open* delete posture, which is the opposite of every intuition about which direction a config error should fail.

## The WARN is actively misleading

[...]

Issue #1056, opened 2026-05-21
Federation peer-id unauthenticated when no key enrolled — TOFU-window spoofing (Agent-2 #6)


alphaonedev:
## Severity: MEDIUM

src/handlers/federation_signing_check.rs:567-574 (None, None) arm (no sig header AND no enrolled key for peer-id) returns None (allow) with warn log. An attacker who knows a legitimate peer's id but has NOT yet been enrolled can post /sync/push with x-peer-id: legitimate-peer and no signature — accepted unsigned. The 'neither side enrolled' rationale assumes legitimate peer hasn't enrolled either; in heterogenous-rollout (operator enrols half mesh), unenrolled half can be impersonated. Once enrolled the rejection arms close window — but TOFU period is exploitable. Fix: default-deny on any push whose x-peer-id is not in operator-configured [federation].peers map; treat 'unknown peer' as 401 rather than allow-with-warn. Surfaced by Agent-2 #6.

🤖 Filed by orchestrator on behalf of 6-agent v0.7.0 code+security review.

alphaonedev:
Fixed via commit 8a808d098.

## Summary

Added a TOFU spoofing guard at the top of `sync_push` (BEFORE the signature verifier): when the operator has enrolled ANY peer via `AI_MEMORY_FED_PEER_ATTESTATION`, any `x-peer-id` NOT in the allowlist is refused with `401 x_peer_id_not_in_allowlist`. The zero-config posture (no env var set) remains permissive so existing deployments are unaffected — the security uplift fires only when an operator has explicitly enrolled peers.

This closes the heterogeneous-rollout window the issue describes: pre-#1056, when an operator enrolled half the mesh, the unenrolled half could be impersonated by anyone who knew the legitimate peer's id.

## Files touched

- `src/federation/peer_attestation.rs` — new `has_allowlist()` helper (returns true when peer map is non-empty).
- `src/handlers/federation_receive.rs` — TOFU gate inserted at top of `sync_push` before `verify_signature_or_reject`.
- `tests/federation_sync_push_tofu_1056.rs` — new integration test file.

## Retest evidence

- `cargo test --test federation_sync_push_tofu_1056` — **3 passed, 0 failed** (unknown-peer + allowlist → 401; enrolled-peer + allowlist → passes; zero-config → skips gate).
- `cargo test --lib federation::peer_attestation` — **20 passed** (incl. 2 new `has_allowlist_*_1056` tests).
[...]

PR #2731, opened 2026-08-04
fix(#2722): peer-attestation NotUnicode/empty env fails closed, not silent zero-config

### LINKED ISSUE #2722: [CB-20][security][federation] peer-attestation NotUnicode/empty env falls through to silent zero-config — fail-open (#2705, #2504 residual)
**Audit finding (CONFIRMED).** Tier-3 security, epic #2705 (#2504 residual).

\`src/federation/peer_attestation.rs:209-239\`: the match is \`Ok(s) if !s.trim().is_empty() => {parse}\` / \`_ => Self::default()\` (\`:237\`). \`std::env::var\` returns \`Err(VarError::NotUnicode(_))\` when the var **is present** but not valid UTF-8 — that lands in the \`_\` arm alongside \`NotPresent\` → \`env_present=false\` → \`has_allowlist()=false\` → the exact pre-#2504 posture (federated-delete namespace confinement OFF, #1056 TOFU 401 OFF, pendings/namespace_meta scope gates OFF, faith-based attribution ON), with **no WARN** — quieter than the parse-error case #2504 fixed.

**Failure scenario:** a Latin-1 \`EnvironmentFile=\`/\`.env\` with one non-ASCII byte in an \`allowed_namespaces\` pattern or peer id. On Linux env values are arbitrary bytes, so this reaches the daemon; the allowlist evaporates and the node silently accepts unbounded federated delete-by-id from any peer.

**Fix:** match \`Err(VarError::NotUnicode(_))\` explicitly into the \`env_present:true, peers:empty\` branch with the CONFIGURED-BROKEN WARN. Add a one-shot WARN for the present-but-empty case (F-2, realistic producer: \`AI_MEMORY_FED_PEER_ATTESTATION="${FED_PEERS}"\` with \`FED_PEERS\` unset). Making \`VarError\` exhaustive here turns this into a compile-time question.

coder tier.

alphaonedev:
Closes #2722

## Cert-blocker CB-20 — confirmed fail-open security residual (#2504 class, CWE-636)

`PeerAttestationConfig::from_env` (`src/federation/peer_attestation.rs`) matched the result of `std::env::var(AI_MEMORY_FED_PEER_ATTESTATION)` with a catch-all `_` for the error case:

```rust
match std::env::var(PEER_ATTESTATION_ENV) {
    Ok(s) if !s.trim().is_empty() => { /* parse; Err => CONFIGURED-BROKEN + WARN */ }
    _ => Self::default(),   // zero-config
}
```

`std::env::var` returns `Err(VarError::NotUnicode(_))` when the variable **IS PRESENT** but not valid UTF-8. That landed in the `_` arm **alongside** `NotPresent` → `env_present=false` → `has_allowlist()=false` → the pre-#2504 zero-config posture (federated-delete namespace confinement OFF, #1056 TOFU OFF, pendings/namespace_meta scope gates OFF, faith-based attribution ON), with **NO WARN** — quieter than the parse-error case #2504 was written to fix.

On Linux env values are arbitrary bytes, so a mis-encoded (e.g. Latin-1) `EnvironmentFile` with a single non-ASCII byte in a pattern or peer id reaches the daemon and **silently disables the allowlist** — the exact fail-open the #2504 CONFIGURED-BROKEN posture exists to prevent, but through a byte-encoding door #2504 never closed.

## Fix

The `Err` arms are now matched **exhaustively by explicit `VarError` variant** (rust-skills `pat-exhaustive-enum`; a future `VarError` variant is a compile error, not a silent fall-through to zero-config):

- `Err(VarError::NotUnicode(_))` → routed into the **SAME** CONFIGURED-BROKEN branch as a JSON parse error: `env_present=true`, empty peer map, `has_allowlist()==true`, `is_configured_empty()==true`, and a per-lane CONFIGURED-BROKEN WARN. Destructive inbound lanes fail closed.
- `Err(VarError::NotPresent)` → the **only** zero-config path.
- **F-2** — the present-but-empty / whitespace-only value (`Ok(s)` where `s.trim().is_empty()`) is documented-intentional zero-config (realistic producer: an unexpanded `AI_MEMORY_FED_PEER_ATTESTATION="${FED_PEERS}"` with `FED_PEERS` unset). Its disposition is unchanged, but it now emits a **`Once`-deduped** informational WARN distinguishing it from a genuine unset. `from_env` is called per inbound push (`federation_receive.rs`), so a naive `warn!` would spam one per request; the security CONFIGURED-BROKEN WARNs stay per-call by design (a loud misconfiguration signal), the benign empty-case notice is deduped.

The non-empty parse path was factored into `from_present_value`, and the shared configured-broken construction into `configured_broken()`, so the two fail-closed callers (parse error + NotUnicode) build identical state.

## Tests (all env-lock-serialised via the existing `ENV_GUARD` idiom)

- `from_env_non_utf8_is_configured_broken_2722` — `#[cfg(unix)]`; builds a present-but-non-UTF-8 value via `OsStr::from_bytes(&[b'{', 0xFF, b'}'])`, asserts the fixture really is `VarError::NotUnicode`, then asserts `env_present`/`has_allowlist()==true` + `is_configured_empty()` (the delete lane is refused, NOT zero-config).
- `non_utf8_env_refuses_out_of_map_peer_2722` — `#[cfg(unix)]`; the security consequence — `namespace_allowed(Some("stranger"), …)` is refused under the non-UTF-8 posture.
- `from_env_present_but_empty_is_zero_config_f2` — asserts the present-but-empty value stays `has_allowlist()==false` (F-2 zero-config, not configured-broken).

Follows the 4 existing adversarial fixtures (`{}`, parse error, typo key, namespace_allowed-under-parse-error).

## CI-parity run locally (default + `sal` + `sal-postgres` + `vectorlite`)

- `cargo check --all-targets` — green
- `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` — green (default / `--features sal` / `--features sal-postgres` / `--features vectorlite`, all four legs)
- `cargo test --lib peer_attestation` — 26 passed (incl. 3 new)
- `cargo test --test qual_10_module_size_ceiling` / `qual_6_7_legacy_error_type_ceiling` — green (no ceiling bump needed)

No MCP tool / HTTP route / CLI / config-knob surface changed, so no SSOT pins move; CHANGELOG `[Unreleased]` entry added.

## AI involvement

Implemented by Claude Opus (Opus 5 main coder tier) under operator authority. Design was a single-correct-answer fail-closed fix mirroring the existing #2504 CONFIGURED-BROKEN precedent (no 5-agent vote required — copying an established security precedent).


alphaonedev:

Verified the diff at `aeb0f5a4` (SSH sig good).

**Correctness (CONFIRMED):** the `from_env` match is now exhaustive over `VarError` with no catch-all `_`, so the fall-through that hid the bug is gone: `Err(VarError::NotUnicode(_))` → `configured_broken()` (fail-closed), `Ok(_)` present-but-empty → one-shot WARN + zero-config, `Err(VarError::NotPresent)` → zero-config unchanged. The **only** behavioral change vs pre-fix is NotUnicode (now fail-closed) and empty (now warns) — NotPresent and the valid/parse-error paths are untouched. A future `VarError` variant is now a compile error (rust-skills `pat-exhaustive-enum`).

**Security (fail-closed direction, the point of the fix):** a present-but-non-UTF-8 env value (mis-encoded EnvironmentFile — arbitrary bytes on Linux) previously fell into the zero-config `_` arm, silently disabling the allowlist (delete-confinement OFF, #1056 TOFU OFF, faith attribution ON). It now keeps `has_allowlist()=true` with an empty peer map, so unenrolled peers are refused. Present-but-empty stays documented-intentional zero-config (per the existing contract at `peer_attestation.rs:74-80`) — the correct non-breaking choice — but the dangerous unexpanded-`${FED_PEERS}` producer now logs. The two CONFIGURED-BROKEN security WARNs stay per-call (loud); only the benign empty notice is `Once`-deduped. Correct.

**Test adequacy (strong):** `from_env_non_utf8_is_configured_broken_2722` builds a genuine non-UTF-8 `OsStr` (0xFF) and first *guards* that `env::var` really yields `NotUnicode` before asserting `has_allowlist()=true`; `non_utf8_env_refuses_out_of_map_peer_2722` asserts the security consequence (a stranger peer refused for `secure/ops`), not just the config flag; `from_env_present_but_empty_is_zero_config_f2` pins F-2. All use `lock_env()` for the repo env-serialization discipline.

Clean helper extraction (`from_present_value`/`configured_broken`), no unwrap on non-test paths. Merging via SSH-signed local merge on green CI.

Issue #1555, opened 2026-06-09
HTTP + MCP family loaders (load_family/smart_load) miss scope=private visibility filter (sqlite path + MCP dispatch)


alphaonedev:
Security review (final pre-GA, 6-agent). **HIGH (HTTP sqlite) + MEDIUM (MCP).**

`handle_load_family` (src/mcp/tools/load_family.rs:238) runs a raw SELECT filtered only by namespace+metadata.family+expiry, NO scope/owner predicate. Three callers omit visibility: (1) HTTP POST /api/v1/memory_load_family **sqlite branch** (src/handlers/power_consolidation.rs:880-892) — the QC-P1 fix (lines 842-849) landed ONLY on the postgres branch; sqlite is the DEFAULT backend; (2) MCP `memory_load_family` dispatch (mod.rs:1223); (3) MCP `memory_smart_load` (load_family.rs:332 → forward_to_load_family) dispatch (mod.rs:1227) — the always-on intent loader. Sibling MCP read tools all resolve resolve_read_visibility_caller()+is_visible_to_caller; these don't.

**Exploit:** shared namespace; Bob POSTs {family,namespace,k} or MCP memory_smart_load{intent,namespace} → Alice's scope=private family-tagged rows verbatim (content+metadata.agent_id).

**Proposed fix:** single gated implementation — change `handle_load_family(conn, params, caller: Option<&str>)`, post-filter `memories` via is_visible_to_caller when Some, recompute count; thread caller through handle_smart_load/forward_to_load_family; all 3 call sites (HTTP sqlite branch + both MCP dispatch arms) resolve+pass caller. None=trust-all preserved. Regression tests for HTTP-sqlite + both MCP tools.

alphaonedev:
Fixed via commit 5f9b73b9 (branch feat/1548-1549-postgres-recursive-coverage).

**Retest evidence:** 15 new regression tests (bound fixtures, no scattered literals) covering all 6 fixes; full lib suite **5249 passed, 0 failed**; fmt + clippy (default + sal-postgres --tests) + cargo audit + vendor-literal gate + the new hardcoded-literal gate all clean. The fix is gated behind the multi-tenant (AI_MEMORY_AGENT_ID-pinned) / FED-P4 hierarchical-CA posture; single-tenant trust-all is preserved unchanged (verified by None-caller tests). Verified per prime directive pm-v3.3. Live-hive revalidation (run.sh) folded into GA-prep redeploy.

PR #2736, opened 2026-08-04
fix(#2600,#2601,#2544): load_family lifecycle + pre-LIMIT visibility cross-backend parity

### LINKED ISSUE #2544: an expired / archived / tombstoned memory is still served as a live namespace standard, and its tokens are never counted against the recall budget
Two small residuals surfaced by the #2537 5-agent adversarial vote (`4d3ea1c5`, data-integrity lens). Neither is a disclosure bug and neither is fixed by #2540; both are honesty defects on the same code path.

## 1 — no lifecycle / expiry predicate on the standard read

`mcp::lookup_namespace_standard` resolves the bound memory via `db::get` (`src/storage/mod.rs`), which is a bare `SELECT ... WHERE id = ?` with **no expiry, lifecycle, or archive predicate**. So a memory that has expired, been tombstoned, or been quarantined is still injected as the namespace's **live governing policy** into every recall / session_start response, and still returned by `memory_namespace_get_standard`.

Every other read surface applies `crate::models::lifecycle_visible_clause` (the fail-closed allow-list added with #1948) or an expiry filter. This one does not, so the substrate can present a rule as current that its own lifecycle machinery considers dead.

Note the neighbouring `warning: "standard memory not found — may have been deleted"` arm already establishes that a dangling binding is a state worth reporting; an *expired* binding is the same class and is reported as healthy.

**Proposed fix.** Apply the same lifecycle predicate the rest of the read paths use. A standard that fails it should be treated as absent, and — following the #2540 precedent — disclosed rather than silently dropped (a distinct marker from `standards_withheld`, since the cause is different: this is "your policy is stale", not "you may not read it", and the operator needs to tell them apart).

## 2 — the injected standard's tokens are invisible to the recall budget

`decorate_budget(&mut resp, &outcome)` and the `tokens_used` accounting run **before** `inject_namespace_standard` at all three `handle_recall` call sites (`src/mcp/tools/recall.rs`). The standard's body — a full serialized `Memory`, frequently a long policy document — is appended afterwards and never counted.

So `budget_tokens` under-reports on every recall in a namespace that has a standard, and a caller who sets a budget to protect a context window can be handed materially more than they asked for. #2540's H2-b change (the standard is no longer retained out of `memories`) makes the undercount marginally larger in the one case where the caller's own query matched the policy, which is what surfaced this.

**Proposed fix.** Either move the budget decoration after injection, or account for the standard explicitly. Prefer the latter so the `standard` contribution is separable in the response — an operator debugging a blown budget should be able to see that governance, not results, consumed it.

Files: `src/mcp/mod.rs` (`lookup_namespace_standard`, `inject_namespace_standard`), `src/storage/mod.rs` (`db::get`), `src/mcp/tools/recall.rs` (`decorate_budget` ordering, all three branches), `src/models/` (`lifecycle_visible_clause`).

Related: #2537 / #2540, #1948 (the lifecycle allow-list), #1596 (expiry-floor semantics).

### LINKED ISSUE #2600: [data-integrity] memory_load_family / memory_smart_load bypass the #1948 fail-closed lifecycle allow-list on sqlite — quarantined + tombstoned rows are readable through an always-on core-profile tool

### LINKED ISSUE #2601: [correctness] sqlite memory_load_family applies scope=private visibility AFTER the SQL LIMIT, so it silently under-returns; postgres applies it before

alphaonedev:
Closes #2600
Closes #2601
Closes #2544

## CB-15 — cross-backend data-integrity + correctness on an always-on core-profile read path

`memory_load_family` / `memory_smart_load` are always-on (core profile). Two cross-backend divergences plus a same-class fold-in.

### #2600 — missing lifecycle allow-list (sqlite)
The sqlite `handle_load_family` query carried **NO** `lifecycle_visible_clause`, so quarantined + tombstoned rows were readable through this always-on tool, bypassing the #1948 fail-closed lifecycle allow-list. Postgres (`PostgresStore::list`) already filtered them → a data-integrity divergence. The query now threads `crate::models::lifecycle_visible_clause("")` (`lifecycle_state IS NULL OR lifecycle_state IN ('open','active','blocked','done','abandoned')`).

### #2601 — post-LIMIT visibility under-return (sqlite)
The #1555 `scope=private` visibility filter (`is_visible_to_caller`, which carries the #1921 team/unit/org subtree carve-out that has **no SQL twin**) ran in Rust **AFTER** the SQL `LIMIT ?4`, so it silently **under-returned** — dropping visible rows the caller should see whenever hidden rows sorted above the `k` boundary. The path now mirrors the postgres `load_family_handler` escalation: filter, and when the visible set is short of `k`, re-ask at the `LIST_MAX_LIMIT` (1000) window and re-filter, so both backends return the same rows up to `k`.

### #2544 (folded in — same class) — ns-standard lookup
`mcp::lookup_namespace_standard` resolved the bound standard via a bare `db::get` with **NO** lifecycle/expiry predicate, so an expired / tombstoned / quarantined memory was served as the live governing namespace standard into every recall / session_start (tokens uncounted). It now gates the bound memory with the same fail-closed lifecycle allow-list (`LifecycleState::is_recall_visible`) + expiry guard (`storage::lifecycle::is_expired`), returning `StandardLookup::Absent` for a non-live standard.

## Tests
- `tests/memory_load_family.rs::load_family_excludes_quarantined_and_tombstoned_rows_2600`
- `tests/memory_load_family.rs::load_family_visibility_applies_before_limit_no_under_return_2601`
- `src/mcp/mod.rs::tests::inject_namespace_standard_skips_non_live_standard_2544`

## SSOT lockstep
- `tests/qual_10_module_size_ceiling.rs`: `src/mcp/mod.rs` ceiling 16_300 → 16_400 (dated).

## Verification run locally
- `cargo build` (default) ✅ · `cargo build --features sal` ✅
- `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` (default leg) ✅
- `cargo fmt --all` ✅
- `cargo test --test memory_load_family` → 7/7 ✅ · `--lib inject_namespace_standard_skips_non_live_standard_2544` ✅
- `cargo test --test qual_10_module_size_ceiling --test qual_6_7_legacy_error_type_ceiling` ✅

## AI involvement
Implemented by Claude Opus 5 (MAIN coder). Postgres twin (`PostgresStore::list` + `load_family_handler`) used as the reference for lifecycle+visibility ordering. No public wire/tool/route/CLI surface change; no schema change.


alphaonedev:

Verified the diff at `71898702` (SSH sig good). Traced all three folded fixes.

**#2600 (lifecycle, sqlite):** `handle_load_family` now threads `lifecycle_visible_clause("")` into the SQL `WHERE` — identical predicate to the postgres twin (`PostgresStore::list` / `load_family_handler`, which use the `"m"`-aliased form). The passing `memory_load_family` suite proves the generated SQL is well-formed. Quarantined + tombstoned rows are now excluded before `LIMIT` on both backends (the #1948 fail-closed allow-list was previously sqlite-blind on this always-on core-profile path).

**#2601 (post-LIMIT under-return, sqlite):** the `is_visible_to_caller` filter (which carries the #1921 subtree carve-out with no SQL twin) ran after a bare `LIMIT k`, silently under-returning when hidden rows sat before the boundary. The fix mirrors the postgres escalation: `run(k)`, and when `caller.is_some() && visible < k`, re-query at `LIST_MAX_LIMIT` and truncate to `k`. `caller == None` keeps the exact-`LIMIT` trust-all path. Same `LIST_MAX_LIMIT` ceiling as postgres → parity.

**#2544 (fold-in):** `lookup_namespace_standard` gated the bare `db::get` with `is_recall_visible()` (the Rust twin of the SQL `lifecycle_visible_clause`) + `storage::lifecycle::is_expired`, returning `Absent` for an expired/tombstoned/quarantined standard so it is no longer served as the governing namespace standard into recall/session_start.

All three are fail-closed (exclude non-live/non-visible). Tests: `..._excludes_quarantined_and_tombstoned_..._2600`, `..._visibility_applies_before_limit_no_under_return_2601`, `inject_namespace_standard_skips_non_live_standard_2544`. The `qual_10` ceiling bump (mcp/mod.rs 16300→16400, dated) is a legitimate size adjustment. CHANGELOG entry present.

Merging via `gh pr merge --merge` on green CI.

Issue #2711, opened 2026-08-04
[CB-7][security][federation][B-2] postgres funnel leaves links/signals/action_transitions unconfined — namespace-scope fix landed sqlite-only (#2705 B-2)


alphaonedev:
**Confirmed by two independent audits (B-set triage + Gate1 review).** Tier-1 security, epic #2705 (B-2).

\`sync_push\` dispatches to the postgres funnel before any sqlite lane loop (\`src/handlers/federation_receive.rs:1410-1413\`). In \`src/handlers/federation_signing_check.rs\`: \`links[]\` loop \`:714-773\` → \`apply_remote_link\`, **no namespace check**; \`signals[]\` \`:849-901\` → \`signal_author_authorized\` (authorship only, #1843) then \`apply_remote_signal\`, **no namespace check**; \`action_transitions[]\` \`:913+\` → crypto+nonce, **no namespace check**. Only memories (\`:247\`) and deletions (\`:686\`) are gated. #2683/#2684 (issues #2489/#2649/#2650) fixed the sqlite twin only. \`MemoryStore::namespace_by_id\` exists (\`src/store/mod.rs:1849\`) so this is an omission, not a capability limit.

**Failure scenario:** a \`public/*\`-scoped enrolled peer pushes links whose endpoints are \`secure/ops\` rows, or signals/transitions in \`secure/ops\`, to a **postgres-backed** receiver; all apply. Precedent parity tests exist for other lanes (\`tests/federation_write_ns_scope_2447{,_pg}.rs\`, \`_2478/_2479/_2488\`).

**Fix:** add the deletions-lane pattern (\`LANE_*\` + \`inbound_by_id_namespace_authorized\` / stored-namespace probe) to all three postgres lanes; add \`_pg\` adversarial fixtures. Note #2682 "Gate1 = PASS" claim is scoped to Gate-1's own issue list, not this.

hard-coder tier.

alphaonedev:
Fixed on release/v1.0.0 by PR #2734 (CB-7, B-2). The postgres /sync/push funnel now namespace-scope-authorizes links[]/signals[]/action_transitions[] through the same receive_auth verdicts as the sqlite twin: links gate BOTH endpoints' stored namespaces (scalar namespace_by_id probe), signals gate the claimed namespace, action_transitions gate the stored action namespace; fail-closed on unresolvable/missing endpoint; zero-config byte-identical. Fable 5 review + security audit: APPROVE. Closed manually (non-default branch).

PR #2737, opened 2026-08-04
fix(#2723): derive Gate-1 lane census from SyncPushBody; register sender_clock

### LINKED ISSUE #2723: [CB-21][ci][federation][B-3] Gate-1 lane census is a hand-written literal, not reflection over SyncPushBody; ConfinementKind wired to no enforcement site (#2705 B-3)
**Audit finding (CONFIRMED).** Tier-3, epic #2705 (B-3). This is the control that certifies CB-7's confinement.

\`src/federation/push_lanes.rs:10-11\` claims "Adding a new wire field without registering it here fails the unit test." It does not: \`:125-130\` asserts \`ALL_SYNC_PUSH_WRITE_LANES.len() == 13\` against a literal, and \`:133-156\` compares the array against a hand-written expected list in the same file. Adding \`pub foo: Vec<X>\` to \`SyncPushBody\` compiles and leaves all four tests green (no serde reflection, no macro, no field-count derivation). \`ConfinementKind\` and \`ALL_SYNC_PUSH_WRITE_LANES\` have zero consumers outside their own \`#[cfg(test)]\` module — the "one structural choke" is a documentation table not linked to any enforcement site. (The \`match\` arms in \`wire_field\`/\`confinement_kind\` are exhaustive with no \`_\`, so adding a *variant* is compiler-enforced — the gap is field→variant, not variant→arm.) Also excluded from the census: \`sender_clock\` (a peer-controlled \`SyncPushBody\` field that mutates local state, CB-14).

**Fix:** derive the census from \`SyncPushBody\` (serde field walk or a proc-macro), assert each declared \`ConfinementKind\` is actually invoked at both funnels, and register \`sender_clock\`. Write CB-21's assertions to be red before CB-7 and green after.

coder tier. Order: land before/with CB-7.

alphaonedev:
Closes #2723

## The defect (CB-3 / CB-21)

The Gate-1 `/sync/push` write-lane census in `src/federation/push_lanes.rs`
advertised (`:~10-11`) that "adding a new wire field without registering it here
**fails the unit test**." It did not. The two guards were both hand-written
literals that live in the same file as the thing they check:

- `:~125` asserted `ALL_SYNC_PUSH_WRITE_LANES.len() == 13` against a **literal**.
- `:~133-156` compared the array against a **hand-written expected list in the
  SAME file**.

So adding `pub foo: Vec<X>` to `SyncPushBody` compiled and left every test green —
the census was decoupled from the struct it claims to reflect. Separately,
`sender_clock` (a peer-controlled `SyncPushBody` field that mutates local
`sync_state` via `VectorClock::merge` inside `sync_state_merge`) was **absent from
the census entirely** — neither a write lane nor documented as excluded.

## The fix — census DERIVED from `SyncPushBody`

1. **Compile-time exhaustiveness pin** (`reflect_sync_push_fields`): destructures
   EVERY field of `SyncPushBody` with **no `..` wildcard**. Rust requires a struct
   pattern without `..` to name every field, so a newly-added field fails to
   **compile** here until its author classifies it (write lane vs sync metadata).
2. **Reflection test** (`gate1_census_is_derived_from_sync_push_body`): asserts
   `ALL_SYNC_PUSH_WRITE_LANES` equals the destructured write-lane serde field names
   (via `stringify!`, no string literals) **in order**. A new `Vec<_>`
   subcollection classified as a lane must also appear in the census or this fails.
3. **`sender_clock` registered**: it plus the 5 other non-lane fields
   (`sender_agent_id`, `sender_wall_clock`, `sender_policy_seq`,
   `sender_policy_digest_hex`, `dry_run`) now live in the new
   `SyncPushMetadataField` inventory (`ALL_SYNC_PUSH_METADATA_FIELDS`), each
   carrying the new `ConfinementKind::SyncMetadata` variant. This documents WHY
   they are not namespace-confined (namespace-agnostic transport metadata) rather
   than silently omitting them — the census is now COMPLETE and honest for every
   `SyncPushBody` field.
4. **Funnel structural pin** (`gate1_every_lane_is_handled_by_the_receive_funnel`):
   `include_str!` over `federation_receive.rs` asserts each declared lane's wire
   field is read as `body.<field>` by the receive funnel, so a census entry cannot
   be orphaned from the CB-7 confinement/apply sites.

### Red-before / green-after mechanism

The load-bearing guard is mechanism (1): a struct pattern `let SyncPushBody { … } = body;`
with **no `..`** is a hard compile error the moment a field is added and not named
in the pattern (`error[E0027]: pattern does not mention field 'foo'`). That points
the author directly at the census. Once they classify the new field:
- as a **write lane** → they must add a `SyncPushWriteLane` variant + an
  `ALL_SYNC_PUSH_WRITE_LANES` entry, or mechanism (2)'s ordered-equality assertion
  fails at test time; or
- as **sync metadata** → they add a `SyncPushMetadataField` variant, a conscious,
  reviewable decision (a `Vec<_>` placed in the metadata arm is a review flag),
  and mechanism (2)'s metadata assertion enforces it too.

Either way a new subcollection **cannot land green without a census entry** — which
is exactly what the old literal-vs-literal tests allowed.

## Scope

Census / reflection integrity ONLY. The CB-7 funnel confinement this census
certifies (#2711, already merged at the base of this branch) is untouched — the
funnels are not rewritten; the new structural pin only asserts each declared lane
is wired to a funnel apply site.

## Verified locally (foreground)

- `cargo build --features sal` — green
- `cargo clippy --all-targets --features sal -- -D warnings -D clippy::all -D clippy::pedantic` — clean
- `cargo test --features sal --lib federation::push_lanes` — 7 passed (3 new)
- `scripts/check-hardcoded-literals.sh` — PASS

Files changed: `src/federation/push_lanes.rs`, `CHANGELOG.md`.


alphaonedev:

Verified the diff at `6347b68e` (SSH sig good).

**Correctness (CONFIRMED):** the census is now genuinely derived, not two literals compared:
- `reflect_sync_push_fields` destructures `SyncPushBody` with **no `..`**, so a new field is a hard `E0027` compile error that forces the author into this exact function (the red-before-green forcing function the pre-fix literal `len()==13` lacked entirely).
- `gate1_census_is_derived_from_sync_push_body` asserts `ALL_SYNC_PUSH_WRITE_LANES.map(wire_field)` equals the `stringify!`-reflected write-lane idents **in order**, and the metadata inventory likewise — the consts are pinned to the struct's serde field names, not to a hand-written twin.
- `sender_clock` (peer-controlled, mutates local `sync_state`) is now registered under the new `ConfinementKind::SyncMetadata` / `SyncPushMetadataField` inventory — honestly classified as namespace-agnostic transport metadata rather than silently omitted.
- `gate1_every_lane_is_handled_by_the_receive_funnel` `include_str!`s `federation_receive.rs` and asserts each lane's `body.<field>` is read (no orphaned census entry). CB-7 funnels untouched — scope is census integrity only, correct.

**Residual (non-blocking, noted for honesty):** the disjointness check (`seen.len() == lanes.len()+metadata.len()`) proves no field is double-classified, but nothing mechanically forces a *destructured* field into one of the `stringify!` vecs — an author could bind `newfield: _` and omit it from both. Full mechanical coverage would need a field-enumerating proc-macro. This is a review-catchable omission, not the silent literal-drift #2723 was about (a new field now cannot land with ZERO signal — it breaks compilation), so it clears the issue's bar. Worth a small follow-up if you want the last increment of enforcement; I would not hold the PR for it.

**rust-skills / gates:** clippy sal leg clean, fmt clean, hardcoded-literals gate PASS, 7/7 push_lanes tests. CHANGELOG entry present. No cfg-specific code, so the default/sal-postgres/vectorlite clippy legs should not diverge.

Merging via `gh pr merge --merge` on green CI.

Issue #929, opened 2026-05-20
SECURITY-high: namespace_set_standard has no ownership gate — any caller can hijack any namespace policy


alphaonedev:
## RCA

The HTTP endpoint `POST /api/v1/namespaces/{ns}/standard` and the MCP tool `memory_namespace_set_standard` have NO ownership gate. Any authenticated caller can overwrite any namespace's standard with an arbitrarily hostile policy.

### Code path

`src/handlers/hook_subscribers.rs:266 set_namespace_standard_inner`:
1. Resolves the caller via `resolve_http_agent_id` (line 278-280).
2. Emits a #913 audit row marked `"allow"` (line 281-292) — but this is a logging facility, NOT an enforcement gate.
3. Stamps the placeholder standard memory with `metadata.agent_id = "system"` (line 511, SQLite path; line 347, Postgres SAL path). The placeholder owner is "system" regardless of who called the API, so a subsequent `namespace_owner()` lookup returns "system" (or fails to match the caller, depending on path).
4. Dispatches to `mcp::handle_namespace_set_standard` (src/mcp/tools/namespace.rs:9) which records a SOC2 audit row + merges governance + calls `db::set_namespace_standard`. No `enforce_governance` / `check_caller` invocation anywhere.

### Why the substrate-side gate doesn't fire

The `enforce_governance` family in `src/storage/mod.rs:7270 evaluate_level` IS capable of enforcing owner-level checks, BUT:
- The set-standard handler never calls it.
- The standard's `metadata.agent_id = "system"` means even if a downstream caller invoked `namespace_owner(ns)`, it would resolve to "system" — not the actual creator. The substrate has no record of who created the first standard.

## Impact

Cross-tenant policy takeover:
1. Alice sets `track-a/p6-alice/ns` standard with `write=owner, delete=owner, approver=human, inherit=true`.
2. Bob (any X-Agent-Id) POSTs to the same URL with `write=any, delete=any, approver={agent:bob}, inherit=false`.
3. Bob's policy is now the namespace standard. Every subsequent write/delete in the namespace evaluates under Bob's policy. Bob can then write/delete in Alice's namespace at will.

Severity: **SECURITY-high** — this is a privilege-escalation surface, NOT just a leak. Effectively allows any HTTP caller to hijack any namespace's governance policy.

[...]

Issue #913, opened 2026-05-19
v0.7.0 docs: CLAUDE.md schema version stale (says v43, code constant is v47)


alphaonedev:
### Claim
`CLAUDE.md:217`:
> SQLite with WAL mode, FTS5 virtual table for full-text search. **Current schema = v43**

### Probe
- `src/storage/migrations.rs:478`: `const CURRENT_SCHEMA_VERSION: i64 = 47;`
- Fresh sqlite DB created by the release binary (`14fb8a781`, AI_MEMORY_NO_CONFIG=1) → `SELECT MAX(version) FROM schema_version;` → **47**.

### Verdict
DRIFT (doc undercount). Claim says v43; code + actual DB = v47.

### Proposed doc-fix
`CLAUDE.md:217`: change `Current schema = v43` → `Current schema = v47`.

The migration ladder constant is the canonical source. Recent PG-parity fix (#902) bumped postgres to v47 to match sqlite.

---
Filed by Truthfulness-Audit Section 1 (numeric architecture). Base SHA: `14fb8a7813121469899aa117b6c9a78df4048310` on `local/install-815-816`.

alphaonedev:
Fixed at HEAD `5788fbe96` (this batch) building on the earlier landed work — CLAUDE.md line 217 narrative now states 'Current schema = v47' (was v43). Code constant in `src/storage/migrations.rs` is `CURRENT_SCHEMA_VERSION = 47` (verified at line 478 of that file). Postgres ladder description updated to reference the actual ladder end at `migrations/postgres/0029_v07_links_temporal_columns.sql`. Module table row for `storage/` also updated to cite v47.

Per pm-v3 (memory cd8ede94) maximum-truthfulness — landed inline with the other 7 truthfulness-audit issues; no deferral.

Retest: `grep -n 'schema = v47' CLAUDE.md` returns line 217 (one match, no v43 residue).
Verified per prime directive pm-v3.

PR #2739, opened 2026-08-04
fix(#2634): governance audit records the true verdict, not a pre-gate allow

### LINKED ISSUE #2634: governance audit records verdict 'allow' BEFORE the owner gate that can refuse — refused set_standard attempts are logged as allowed
**Not covered by any existing issue.** Confirmed by direct read at HEAD `5449b6da`.

## The defect

`src/mcp/tools/namespace.rs:148` writes:

```rust
record_decision(&caller, "allow", "namespace_set_standard", ...)
```

**before** the #929 owner gate at `:186`, which can `return Err` at `:194`.

So a **refused** `set_standard` attempt is recorded in the signed forensic chain as **`allow`**. The audit verdict field does not mean what it says.

## The internal inconsistency that makes it a defect rather than a design choice

The sibling operation in the *same file* — `clear_namespace_standard` at `:565` — records **after** its gate. Two halves of one file disagree on ordering, so the chain contains a mix of pre-gate and post-gate verdicts with nothing distinguishing them.

## Why the stated justification is insufficient

The doc comment justifies pre-write ordering as "capture intent even on failure." That is defensible for a **storage** failure — you want evidence of an attempt that crashed. It is not defensible for an **authorization refusal**, because the record actively asserts the opposite of what happened, on the surface an auditor consults to find refused attempts.

An attacker probing namespace ownership therefore generates a chain of `allow` records.

## Proposed fix

Two-phase record — `intent` then `outcome` — or move the verdict to post-gate and record a separate `attempt` entry pre-gate. Either preserves the "capture intent" property without lying about the verdict. Whichever is chosen, make both halves of the file agree.

Found by the 3x7 issue audit, security/confinement lens. Relates to the false-success class (#2444): a control that reports a disposition it did not reach.

alphaonedev:
Closes #2634

## Defect (CB-24 / the #2634 governance-audit-verdict class; CWE-778)

The tamper-evident governance audit chain called
`record_decision(actor, "allow", …)` **before** the gate that can refuse a
governed action ran. So a **refused** governed action was chained as
**allowed** — or, for `clear_namespace_standard`, **not chained at all**. No
surface could record a DENY; the forensic chain lied about the outcome.

**Root cause: record-before-gate.** The #913 audit-emit contract (comment
near `namespace.rs:~183`) requires the row to precede the *storage* write so
intent survives a failed write. But the pre-gate row also preceded the
*authorization* gate, so it asserted `"allow"` before the verdict existed.

## Intent-vs-outcome design choice

I evaluated the three shapes in the task:
- (a) two rows (intent before + outcome after),
- (b) move the single record after the gate,
- (c) capture the verdict and record **once** after the gate.

**Chosen: (c) — a single row per governed decision carrying the TRUE
verdict, recorded AT the authorization gate but still BEFORE the downstream
execution/storage write.** This satisfies both requirements without a second
row:
- **Intent preserved (#913):** on the allowed path the `"allow"` row is
  emitted the instant the gate authorizes and *before* the durable execution
  write, so a downstream execution error (the exact case #913 cites) still
  leaves the approver's identity + pending_id in the chain.
- **True outcome:** a refused action records `"refuse"` (the token the
  verify-audit-trail chain already understands — I grepped the accepted
  vocabulary; `"refuse"` is the established governance-refusal token, e.g.
  `pending_reject`; no new token invented, and `verify_audit_trail` verifies
  the hash chain, not the decision semantics).

The intent-preservation requirement does **not** force the two-row shape:
because the verdict is known at the authorization point and recorded before
the durable write, one truthful row per decision is sufficient. An
operational error that prevents a decision from being reached (NotFound 404 /
consensus 500) is not a governance verdict and correctly records no verdict
row.

## Surfaces fixed

| Surface | File | Pre-fix | Fix |
|---|---|---|---|
| `set_namespace_standard` | `src/mcp/tools/namespace.rs` | `"allow"` fired before the #929/#2541 ownership gate | gate runs first; refused bind → `"refuse"` (before `db::update`) |
| `clear_namespace_standard` | same | refusing gate ran BEFORE the record → refused clear had **no row** | refused clears now record `"refuse"` at the gate; allowed keeps `"allow"` |
| `approve_pending` (HTTP) | `src/handlers/governance.rs` | `"allow"` before `approve_with_approver_type` / consensus | verdict at each outcome arm (allow/vote-accepted → `allow`; #2643 rejection → `refuse`) |
| `approval_decide` (HTTP K10) | `src/handlers/approvals.rs` | same | same |
| `handle_pending_approve` (MCP) | `src/mcp/tools/pending.rs` | same, plus the R40 signed-approval gate | same; rejected signature quorum → `refuse` |

The **MCP `pending.rs`** surface was discovered to carry the identical defect
and is fixed under the same class (prime directive: fix the whole class, not
just the three named files). `handle_namespace_clear_standard` was widened
`pub(crate)` → `pub` for integration-test reach (mirrors the already-`pub`
`set_standard` sibling).

## Not weakened

- The #913 intent-capture guarantee is preserved on every allowed path.
- `verify_audit_trail`'s parsing contract is unchanged (no new decision
  token; the chain verifies hashes, not decision strings).

## Tests

`tests/governance_audit_verdict_2634.rs` (8 tests): for each surface a
refused governed action chains `refuse` and **never** `allow`; an allowed
action chains `allow`; `clear_standard`'s refused path now produces a row at
all; both HTTP self-approval surfaces (via router + HMAC) and the MCP surface
are pinned. `tests/admin_audit_chain_913.rs` still green.

## Verification run (foreground)

- `cargo build` (default) + `cargo build --features sal`
- clippy all four legs: default / `sal` / `sal-postgres` / `vectorlite`, each `-D warnings -D clippy::all -D clippy::pedantic` — green
- `cargo fmt --all --check`, `cargo check --examples`
- `cargo test --test governance_audit_verdict_2634` (8 passed), `admin_audit_chain_913` (4 passed)
- `qual_10_module_size_ceiling`, `qual_6_7_legacy_error_type_ceiling` — green (no ceiling bump needed)

No MCP-tool / HTTP-route / CLI-subcommand count change (no SSOT reconciliation
needed). CHANGELOG `[Unreleased]` entry added.


alphaonedev:
## Fable 5 review + security audit — APPROVE (queued for rebase + CI)

Verified the diff at `f8f45971` (SSH sig good) against CB-24's true base `09f250b6` — the `load_family.rs` churn in the raw `--stat` is a stale-base artifact (this branch predates CB-15/#2736); the REAL change set is governance-audit-only.

**Correctness (CONFIRMED across all four surfaces):**
- **governance.rs `approve_pending`** (pg + sqlite branches): the unconditional pre-gate `record_decision("allow")` is REMOVED; each outcome arm now records via `audit_pending_verdict` — Approved to `allow`, vote-accepted-pending to `allow` (the vote call WAS authorized), Rejected(reason) to `refuse`. The `allow` is still emitted BEFORE the execute write, so the #913 approver-identity intent capture survives a downstream execution error.
- **approvals.rs `approval_decide` (+ postgres)**: same via `audit_decide_verdict`.
- **namespace.rs `set_standard`**: records the true verdict (`refuse` when `authorize_namespace_standard_bind` would refuse, else `allow`) before the bind write. `clear_standard`: `record_clear_refusal` added at BOTH refuse-return points (previously refused clears were UN-audited), `allow` before the clear write.
- **pending.rs MCP `handle_pending_approve`** (sibling you found): identical pre-gate `allow` defect, fixed under the same class. Correctly closed per the fix-the-whole-class directive — leaving it would have kept a surface that could not record a DENY.

Operational errors (NotFound/storage) reach no governance verdict and record no verdict row — correct. The chain can now record a DENY and never asserts a false ALLOW; `verify_audit_trail`'s hash-chain parsing is untouched (reuses the existing allow/refuse tokens, no new vocabulary). Design option c — single row, true verdict at the gate before the durable write — is the right call and is correctly reasoned; flagged T3 in the commit, no vote needed for a single-correct outcome.

**Tests:** 8 in `governance_audit_verdict_2634.rs` (set/clear/approve refused to refuse never allow; allowed to allow; clear-refused now produces a row; both HTTP self-approval surfaces + MCP). `admin_audit_chain_913` still green (intent capture intact). CHANGELOG entry present.

Queued behind #2737/#2738; I will rebase onto the final tip (governance files are untouched by the ahead-of-queue PRs, so only a CHANGELOG resolve) and merge on green CI.

Issue #2724, opened 2026-08-04
[CB-22][data-integrity][bulk] bulk_create omits post-commit stages — no audit, no subscription dispatch, no postgres federation fanout (#2705)


alphaonedev:
**Audit finding (CONFIRMED).** Tier-3 data-integrity/observability, epic #2705. The #2689 headline "bulk reuses the single-create funnel" is an overclaim (reuses field-projection, re-implements the rest); these three post-commit stages are absent on one or both backends and none is disclosed in the omitted-stage table/CHANGELOG/API_REFERENCE.

- **F1 audit:** \`src/handlers/bulk.rs\` has zero \`crate::audit::emit\` calls (both backends). Single-create emits \`AuditAction::Store\` per row (\`create.rs:913\` sqlite, \`:1322\` pg). A 32k-row load leaves zero Store audit records.
- **F2 dispatch:** zero \`subscriptions::dispatch_event\`. Single-create fires it (\`create.rs:968\`/\`:1348\`). A \`memory_store\` subscriber receives nothing from a bulk ingest (#932 class re-appears).
- **F3 pg federation:** \`bulk_create_postgres\` (\`bulk.rs:937-1157\`) ends at \`classify_persisted\` with no \`broadcast_store_quorum*\`/\`bulk_catchup_push\` — the sqlite branch has both (\`:885-930\`), pg single-create broadcasts (\`create.rs:1282\`). On pg + \`--quorum-peers\`, N bulk rows are durable-local and never pushed, and the envelope returns 200 with empty \`warnings[]\` affirmatively asserting no replication problem.

**Fix:** add the three stages per bulk row (or per batch where semantically correct), matching single-create on both backends; update the omitted-stage disclosure.

coder tier.

alphaonedev:
Fixed on release/v1.0.0 by PR #2738 (CB-22). bulk_create now runs the three post-commit stages single-create runs — audit emit (AuditAction::Store per persisted row), subscription dispatch (memory_store event per row), and postgres federation fanout (broadcast_store_quorum + bulk_catchup_push, previously sqlite-only) — driven by distinct_committed_rows (dedup-by-persisted-id, last-wins) so each fires exactly once per committed row and never for rejected/pending rows, on both backends. Fable 5 review + security audit: APPROVE. Closed manually (non-default branch).

Issue #1773, opened 2026-06-23
Federation merge_inbound (overwrite_full_row_by_id) skips seal_content + encrypted_envelope (encryption desync) and takes no pre-merge snapshot


alphaonedev:
**Auto-filed by the 2026-06-22 destructive-ops audit** (Agent 3; verified by direct read of `overwrite_full_row_by_id` + reachability).

## Defect (peer-reachable via /sync/push same-id divergence — `src/handlers/federation_receive.rs:935` → `db::merge_inbound` → `overwrite_full_row_by_id` `src/storage/mod.rs:8306`)
1. **Encryption desync (HIGH when encryption-at-rest is ON).** `overwrite_full_row_by_id`'s UPDATE sets 27 columns incl. `content = mem.content` (plaintext) but **omits `encrypted_envelope`** and never calls `seal_content` (verified: columns ?1–?27, no envelope; contrast `insert`/`insert_if_newer`/`update_with_expected_version` which seal+envelope as a matched pair). Result: merged plaintext lands in `content` while the row's OLD non-NULL `encrypted_envelope` remains; on next read `row_to_memory` decrypts the **stale** envelope → silently returns pre-merge content → the merged value is effectively lost.
2. **No pre-merge snapshot (MEDIUM).** Unlike `update_with_expected_version` (#1725), this peer-driven content overwrite archives nothing — a federation merge where the remote wins the LWW tiebreak permanently discards prior local content with no recoverable copy.

## Proposed fix
- Route `overwrite_full_row_by_id` through `seal_content(&mem.content, agent_id)` + add `encrypted_envelope = ?N` to the UPDATE (mirror `update_with_expected_version` ~1851-1856), `src/storage/mod.rs:8306`.
- Add an `in_place_edit` snapshot before the merge-write (reuse `archive_memory_insert_only`), mirroring #1725.
- Add an encryption-on federation-merge round-trip test (`overwrite_full_row_by_id`/`merge_memory` currently have no covering tests).

Severity: HIGH (encryption-on) / MEDIUM (no-snapshot). Evidence: audit mem 88e8050d.

alphaonedev:
[...]

Issue #1725, opened 2026-06-18
🧬 `#884` in-place content overwrite is lossy — unify under archive-before-overwrite (Fix A) [§2.5 + §2.2]


alphaonedev:
# 🧬 `#884` in-place content overwrite is lossy — unify under archive-before-overwrite (Fix A) [§2.5 + §2.2]

**Severity:** `correctness-low` (provenance / anti-confabulation integrity gap — **NOT** a cross-tenant leak; that was #1720). **Parent:** v0.8.0 EPIC #1709, **Phase 4 (Typed Cognition / Maintenance / Merge)** — same archive/supersede cluster as #888 + `lifecycle_state` (v64). **Not a release blocker**, but carries a §9/§17 tag-cut gate condition (below).

**§2-property (§3 scope test — passes on two):** strengthens **§2.5 (attested)** — makes the shipped "audit chain forensically reconstructible at any past timestamp" claim true on this path — and **§2.2 (coherent)** — collapses two divergent content-durability contracts into one.

> **Provenance.** Two adversarial tribunals (2026-06-17, codegraph-verified, index re-synced 6,012 nodes): a 7-seat memory-mutability ruling (converged 2 rounds) surfaced the defect; a 7-seat placement tribunal (converged 3 rounds) ruled severity/EPIC/phase/blocker/fix. All anchors re-verified at HEAD via `codegraph_explore`/`codegraph_callers` before filing.

---

## 🧨 The defect (codegraph-verified, both backends)

There are two content-write paths with **different durability contracts**, and the **only reachable one is lossy**:

| Path | Location | Behavior | Reachable? |
|---|---|---|---|
| `#884 update_with_expected_version` | sqlite `src/storage/mod.rs:1468` · postgres `src/store/postgres.rs:3244` | **In-place** `UPDATE memories SET content=…, version=version+1 WHERE id=… AND version=…`. Reads existing, governance-gates the post-merge row (#1451), stamps `edit_source`, bumps `version` — **but does NOT archive the prior content.** | ✅ **YES** — trait `update()`, MCP `memory_update`, HTTP `PUT /memories/{id}` (If-Match **and** last-write-wins) |
[...]

PR #2741, opened 2026-08-04
fix(#2725): bulk_create honors on_conflict (default error), adds updated_rows[]

### LINKED ISSUE #2725: [CB-23][data-integrity][bulk] bulk_create has no on_conflict default — silently and unrecoverably overwrites pre-existing rows, no archive, no updated_rows[] (#2705)
**Audit finding (CONFIRMED — most material bulk data-integrity item).** Tier-3, epic #2705.

Bulk never resolves \`body.on_conflict\`; every row goes to \`db::insert\`, whose \`ON CONFLICT(title, namespace) DO UPDATE SET content = excluded.content\` (\`src/storage/mod.rs:1603-1610\`) replaces stored content. Three compounding facts:
- sqlite single-create defaults \`on_conflict="error"\` → 409 + \`existing_id\` (\`create.rs:501-544\`); bulk has no such default.
- \`insert_inner\` takes **no** pre-overwrite snapshot. \`db::update\` archives prior content under \`archive_reason='in_place_edit'\` (#1725), the federation merge under \`'federation_merge'\` (#1773); the upsert arm archives nothing — prior content is gone and \`ai-memory undo-edit\` has no snapshot.
- The envelope reports only the scalar \`updated: N\`; there is no \`updated_rows[]\` (input index → superseded id), even though the strictly *less* destructive in-batch collapse gets \`deduped_rows[]\`.

**Failure scenario:** a fleet loader replays a manifest whose titles collide with hand-curated \`long\`-tier memories → all are content-replaced, \`version\` bumped, no archive row, 200 OK \`updated: N\`, and the caller cannot determine which records did it.

**Fix:** honor \`on_conflict\` in bulk with the same \`error\` default as sqlite single-create; add \`updated_rows[]\` disclosure. Also fold F6/F7 (embed skip-guard parity for oversize/empty; the "~10 round trips" claim is false on the ollama default — per-text loop, \`llm.rs:2270\`).

coder tier.

alphaonedev:
## Summary

Fixes the CONFIRMED data-integrity defect **CB-23 / #2725**: `POST /api/v1/memories/bulk` silently and unrecoverably OVERWROTE pre-existing rows.

`bulk_create` never resolved `body.on_conflict`, so every row went straight to `db::insert`, whose `ON CONFLICT(title, namespace) DO UPDATE SET content = excluded.content` replaced the stored content of any pre-existing `(title, namespace)` row. `insert_inner` takes **no** pre-overwrite snapshot (unlike `db::update`'s `archive_reason='in_place_edit'` #1725, or the federation-merge `'federation_merge'` #1773), so the prior content was **gone** — `ai-memory undo-edit` had nothing to restore. sqlite single-create has defaulted `on_conflict=error` (409 + `existing_id`) since v0.6.3.1; bulk had no such default.

## Root cause + the in-batch-vs-pre-existing distinction

The fix resolves the **same** single-create vocabulary (`error` | `merge` | `version`, `crate::mcp::tools::OnConflictMode`, default `error`) per row, on **both** backends:

- **`error`** (default): a collision with a **PRE-EXISTING stored** row is REJECTED into `errors[]` as a `CONFLICT` carrying `existing_id`; the stored content is left **untouched**.
- **`merge`**: the explicit opt-in to the content-replacing upsert.
- **`version`**: the title is rewritten to the next free suffix (lands as a NEW row).

The critical distinction the fix preserves: an **IN-BATCH** `(title, namespace)` collapse (two rows in the SAME batch) is the existing `deduped_rows[]` path and must NOT be conflict-rejected — only a collision with a row that existed **before this batch** is. sqlite snapshots the pre-existing ids BEFORE the sequential-upsert loop; postgres probes before the deferred batched `store_batch` INSERT. So in-batch dedup is unchanged.

New envelope field **`updated_rows[]`** (`{index, superseded_by: <stored id>}`, mirroring `deduped_rows[]`) discloses, by input index, which rows overwrote existing content under `merge`. A conflict-rejected row now counts in `rejected`, never `updated`, so the reconciliation identity (`created + updated + deduped + rejected + pending == sent`) holds; `updated_rows[]` is a disclosure array, not a counter.

Orthogonal to and built on top of CB-22 (#2724, post-commit stages) at the current tip.

## Tests

- `tests/bulk_on_conflict_2725.rs` — (a) default collision REJECTED + stored content unchanged (out-of-band `SELECT`); (b) explicit `merge` updates + discloses the superseded id; (c) in-batch dedup unchanged; (d) mixed batch reconciles.
- Unit cells in `src/handlers/bulk.rs` (`classify_persisted` `updated_rows`, `reject_conflict`).
- Updated the #2551 cross-call-upsert cell to request the upsert explicitly (`on_conflict=merge`) — the default now correctly refuses it.

## Verification run locally (foreground)

- `cargo build` (default) + `cargo clippy --features sal --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` + `cargo clippy --all-targets ...` (default leg) — all clean
- `cargo test --test bulk_on_conflict_2725 --features sal` — 4 passed
- `cargo test --test bulk_write_truthfulness_2551 --features sal` — 13 passed

Closes #2725


alphaonedev:
## Fable 5 review + security audit — APPROVE (CI re-running after gate fix-forward)

Verified the diff at `248c72ac` (SSH sig good) + my fix-forward `35dfbbfa`. CB-23 is a confirmed data-integrity fix: pre-#2725 `bulk_create` never read `body.on_conflict`, so every row hit `db::insert`'s `ON CONFLICT(title, namespace) DO UPDATE` and silently OVERWROTE any pre-existing stored content — and `insert_inner` takes no pre-overwrite snapshot (unlike `db::update` #1725), so that content was unrecoverable.

**Correctness — the load-bearing point is the in-batch vs pre-existing distinction, and it is right on BOTH backends:**
- **sqlite**: a `pre_existing: HashMap<(title,namespace), Option<id>>` is snapshotted BEFORE the sequential-upsert loop, consulted ONLY by `error`-mode rows. A row colliding with an in-batch sibling (created DURING the loop) is therefore NOT in the snapshot → it is NOT conflict-rejected; it flows through the normal upsert and the collapse is disclosed by `deduped_rows[]`. A row colliding with a genuinely PRE-existing stored row → `reject_conflict` (carries `existing_id`), content untouched. `None` caches a confirmed miss (one probe per key).
- **postgres**: the batched INSERT (`store_batch`) runs only AFTER the loop, so a per-row `find_by_title_namespace` hit at resolution time is NECESSARILY pre-existing (in-batch collapse stays `store_batch`'s last-wins upsert → `deduped_rows[]`). Same three-way disposition. `error` default, `version` renames, `merge` opts into the upsert.

**Resolution happens BEFORE any content-replacing write and BEFORE the quota charge**, so a rejected row needs no refund. An unknown `on_conflict` token rejects the row (single-create returns 400 for the same input) rather than silently defaulting.

**Reconciliation identity holds.** `reject_conflict` → `reject_class` → `self.errors`, and `rejected() = errors.len()`, so conflict-rejected rows count in `rejected`; the identity `created + updated + deduped + rejected + pending == sent` is preserved. `updated_rows[]` is a DISCLOSURE array (`{index, superseded_by}`, `merge` overwrites only), never a counter, so it does not enter the identity.

**The `bulk_write_truthfulness_2551.rs` test change is honest, not regression-masking.** The `bulk_updated_is_distinct_from_created_2551` cell used to rely on the SILENT-default-upsert — which WAS the CB-23 defect. Now that the default (`error`) correctly refuses a pre-existing collision, the cell explicitly requests `on_conflict=merge` to keep pinning the surviving #2551 concern (an upsert reports `updated`, never `created`); its assertions (`created:0`, one row) are unchanged. A cell that asserted the buggy behavior is updated to assert the correct behavior with an explicit opt-in.

No `if let Ok`/`.is_ok()` as a security predicate. Per-row `#1919` attestation gate + namespace governance are preserved. CHANGELOG entry present.

**My fix-forward (`35dfbbfa`, SSH-signed):** the coder's commit tripped the pm-v3.1 Hardcoded-literal ratchet on three literals that crossed the ≥3-site threshold (`"existing_id"` +3, `"on_conflict"` +5, `"superseded_by"` +3). Per the operator directive I hoisted each to ONE named const referenced at every production site — reusing the existing `crate::models::field_names` SSOT (new `EXISTING_ID`/`SUPERSEDED_BY`, alongside `SIMILARITY`/`SUPERSEDED_ID`, incl. the single-create 409 in `create.rs` and the skill supersession response in `skill_get.rs`, via the parenthesized-key `(field_names::…): x` form `create.rs` already uses) and the existing `crate::mcp::param_names::ON_CONFLICT`. Const values are byte-identical; no behavior change. Gate now PASS; build/clippy(pedantic)/fmt clean; `bulk_on_conflict_2725` 4/4, `bulk_write_truthfulness_2551` 13/13, `skill_get` 20/20.

**Follow-up filed (prime directive — every gap is a defect):** the coder correctly flagged a PRE-EXISTING asymmetry — postgres SINGLE-create (`create_memory_postgres`) does not honor `on_conflict` at all (always upserts via SAL `store()`), so bulk-pg is now STRICTER than single-create-pg. Filed as a separate issue; out of scope for this PR (which makes bulk correct on both backends, the North-Star choice).

Queued behind #2739 (CB-24) + #2740 (CB-6/18); I will rebase onto the final tip (CHANGELOG resolve) and merge on green CI.

Issue #2670, opened 2026-08-01
[federation][data-integrity] an enrolled peer can permanently poison the receiver's sync cursor for a DIFFERENT peer — sender_clock is folded by monotonic max with no per-entry authorization


alphaonedev:
Found while reviewing #2442 (PR #2669). **Not fixed there** — the site is in a file held by another lane. Filing so it is tracked rather than lost in a PR body.

## The defect

`src/handlers/federation_receive.rs:3144` (sqlite receive funnel):

```rust
if !body.dry_run
    && let Err(e) = db::sync_state_merge(&lock.0, &local_agent_id, &body.sender_clock)
```

`body.sender_clock` is **wire-supplied** and its `entries` map is a `peer_id -> timestamp` dictionary whose **keys are chosen by the sender**. `db::sync_state_merge` folds it into the receiver's persisted `sync_state` by pointwise max.

The in-code comment presents monotonicity as the safety property — "an older incoming timestamp never regresses a newer stored entry". That is exactly backwards as a security argument: **monotonic max makes a poisoned entry unrecoverable.** There is no path that lowers a `sync_state` entry, so a single bad fold is permanent.

## Impact

Peer A pushes a `sender_clock` containing an entry for **peer B** with a far-future timestamp. The receiver's cursor for B jumps forward. Every subsequent `/sync/since?since=<poisoned>` to B returns nothing, so the receiver silently stops converging with B — permanently, behind HTTP 200s, with no operator-visible error.

This is the #2441 harm class (a cursor that reads healthy while the replica diverges), except:
- **irreversible** — monotonic max has no walk-back;
- **lateral** — A poisons the receiver's relationship with B, a confused-deputy step, not merely A's own lane;
- it survives the fix in #2441, which corrects how the cursor is *derived*, not what may *overwrite* it.

## Scope — stated accurately

`/sync/push` is signature-verified (#791) and sender-attested against `x-peer-id` (#238), so the sender must be an **enrolled** peer. This is therefore a compromised / buggy / malicious-insider **enrolled peer**, not an anonymous attacker. That is squarely inside the threat model this release certifies for: a 500–1000 agent cluster is expected to survive one bad module, and the whole point of the per-peer namespace allowlist is that enrollment is not unlimited trust.

[...]

PR #2740, opened 2026-08-04
fix(#2719): HTTP+sqlite clear_standard enforces ownership; pg unowned-vs-unresolvable

alphaonedev:
## Summary

Two cert-blocker fixes in the `clear_namespace_standard` path (both CWE-284), folded together because they live in the same surface.

### DEFECT A — CB-6 / 2704-F1 (HIGH, security; #2719) — HTTP+sqlite clear-standard gate was ABSENT

`clear_namespace_standard_inner`'s sqlite arm (`src/handlers/hook_subscribers.rs`) built `let params = json!({"namespace": ns})` — it did **not** thread the resolved caller. So `handle_namespace_clear_standard`'s `identity_claimed` probe (`params.agent_id` present + non-empty) was **FALSE**, and the entire #1777 owner gate + #2545 unresolvable-refusal block was **SKIPPED**.

Consequence: on a sqlite daemon, `DELETE /api/v1/namespaces/{ns}/standard` from **any** api-key/keyless caller deleted the `namespace_meta` row → dropped the #2503 severed floor → re-opened the #2545 attack on the network surface. (Postgres already refused everyone here, so the parity was inverted.)

The sibling SET path (`set_namespace_standard_inner`) already threads the resolved caller into its MCP params (`obj.insert("agent_id", json!(caller))`). The fix **mirrors that exactly** for CLEAR, so the same #1777 + #2545 gates run:
- a foreign named caller (`ai:mallory`) is refused against a named owner;
- a keyless caller resolves to a non-empty anonymous id → `identity_claimed` true → refused;
- an unowned/`system` standard still passes (unowned-pass);
- the owner clear still succeeds.

### DEFECT B — CB-18 / 2704-F2 (MEDIUM, correctness, postgres) — unowned vs. unresolvable conflated

`PostgresStore::clear_namespace_standard` (`src/store/postgres.rs`) decoded its `metadata->>'agent_id'` JOIN with `.flatten()`, collapsing two states into `None`:
- `Some(None)` — row exists, `agent_id` SQL NULL → **UNOWNED** (should ALLOW, the documented unowned-pass);
- `None` — no joined row → **unresolvable** (severed/dangling → refuse #2545).

So a legacy pre-#929 / federated / non-object-metadata pg standard was misclassified as unresolvable and **refused with a wrong "severed or dangling" error**, even though the sqlite twin (`recorded_owner == ""` → `is_unowned`) ALLOWS it.

[...]

PR #2742, opened 2026-08-04
fix(#2718): validate next_since + gate sender_clock per-key + raw-count tie-group guard

### LINKED ISSUE #2718: [CB-14][data-integrity][federation][B-5] next_since pull-cursor poisoning + sender_clock unauthorized fold (#2705 B-5, #2670)
**Audit finding (CONFIRMED).** Tier-2, epic #2705 (B-5, relates #2670).

Two peer-controlled cursor vectors with no per-key authorization:

1. **next_since (pull cursor).** \`src/daemon_runtime.rs:7299\` (\`#[serde(default)] next_since: Option<String>\`) → \`:6985\` → written straight into \`sync_state\` at \`:6998\` through an upsert that refuses to regress. A peer answering \`next_since: "9999-12-31T23:59:59+00:00"\` permanently pins this node's cursor: every later pull returns zero rows, nothing updates, and the poisoned value can never be lowered without a manual DB edit. Distinct from #2670 if that is scoped to \`/sync/push\` \`sender_clock\`.

2. **sender_clock (push).** \`src/handlers/federation_receive.rs:3342\` folds wire \`body.sender_clock\` via monotone-max \`db::sync_state_merge\` (\`src/storage/mod.rs:16840\`) with no per-key authorization (#2670). Postgres does not merge at all → also a parity question.

Also **F7 (federation_sync_since.rs:69):** the tie-group guard treats \`examined.len() < limit\` as "page not truncated", but \`memories_updated_since\` drops undecryptable rows (\`storage/mod.rs:17029\` \`flatten\`, #2383); one such row in a full page makes it look short, the guard advances over a straddling tie group → silent row loss. Fix: compare the raw SQL row count.

**Fix:** validate \`next_since\` (parseable RFC3339, ≤ now+skew, > since); add per-key authorization to the \`sender_clock\` fold + pg parity; raw-count the tie-group guard.

hard-coder tier.

alphaonedev:
Closes #2718 (CB-14, B-5). Relates #2670 (sender_clock per-key authorization).

A **confirmed data-integrity defect**: three peer-controlled vectors could permanently PIN this node's per-peer federation catch-up cursor (`sync_state`) or silently skip undelivered rows — each behind an HTTP 200 that reads as "in sync". This is exactly the North Star's "DEGRADE (fewer results), never lose data / never wrong results" line being crossed. All three are closed fail-closed.

## Vector 1 — `next_since` pull cursor (far-future poison)
The `/sync/since` response `next_since` field is peer-controlled and was written straight into `sync_state` through a **monotonic (refuse-to-REGRESS) upsert**, so a peer answering `next_since:"9999-12-31T23:59:59+00:00"` permanently pinned the puller's cursor: every later pull returned zero rows and nothing could lower it without a manual DB edit.

**Fix** (`src/daemon_runtime.rs::sync_cycle_once`): validate the candidate via the new `validate_pull_cursor` BEFORE honoring it — REFUSED if it fails RFC 3339 parse, LEADS the local wall-clock by more than `PULL_CURSOR_FUTURE_SKEW_SECS` (300 s; generous for NTP drift, fatal to far-future poison), or is not STRICTLY greater than the current cursor (lexical, matching the downstream `sync_state_observe` / server `updated_at > since` semantics). On rejection the cursor is left UNCHANGED and a WARN fires — the pulled rows still apply; the node re-pulls the same window next cycle rather than skipping forward. The legacy `memories.last()` fallback is held to the same validation. No `unwrap`/`.is_ok()` as a predicate on peer data — explicit `match` with the error arm refusing.

## Vector 2 — `sender_clock` push fold (the #2670 class)
The `/sync/push` receive funnel folded the sender's ENTIRE `sender_clock` into `sync_state` via monotone-max (`db::sync_state_merge`) with **no per-key authorization**, so peer A could inject an arbitrarily-high timestamp for peer **B's** key and permanently pin B's cursor.

**Fix**: the fold is now per-key authorized — `db::sync_state_merge_authorized` accepts ONLY the entry keyed by `body.sender_agent_id` (the sending peer's OWN key). Peer A can advance its own clock entry and nothing else; it can never touch peer B's entry. Monotonic behavior on the authorized key is preserved.

## Vector 3 — F7 tie-group guard (raw-count truncation)
`next_since_watermark` treated `examined.len() < limit` as "page not truncated", but `db::memories_updated_since` drops undecryptable rows (#2383), so `examined.len()` is POST-drop. One undecryptable row in an otherwise-full page made it look short and advanced the watermark over a **straddling tie group**, silently dropping the rows sharing the boundary timestamp beyond the `LIMIT`.

**Fix**: truncation is now measured against the RAW pre-drop SQL row count — `db::memories_updated_since_counted` / `MemoryStore::list_memories_updated_since_counted`, overridden on BOTH the sqlite and postgres adapters (**both** drop undecryptable rows). The walk-back still runs over the delivered rows (safe regardless of drops). `warn_if_cursor_pinned` also keys off the raw count so a single-timestamp full page still WARNs when drops shrank the delivered set.

## Postgres-parity decision
- **Vector 2 (`sender_clock`)** is **sqlite-only** and left as-is: the postgres `/sync/push` receive path **never merged `sender_clock`** into any `sync_state` (the pg backend has no such fold), so it has **no poisoning vector to gate**. Adding a gated pg merge would ENLARGE the surface, not close a hole — so it is deliberately not added.
- **Vectors 1 and 3** are backend-blind: the pull cursor rides the CLI sync client (`sync_cycle_once`), and the raw-count tie-group fix ships on **both** adapters.

## Tests
- `tests/federation_sync_state_merge_1709.rs` — a peer folds only its OWN key; peer A's push cannot advance peer B's established key (poisoning blocked); monotonic-non-regress retained.
- `src/daemon_runtime.rs` unit tests — far-future / unparseable / non-advancing cursors REFUSED; valid forward step + within-skew ACCEPTED.
- `src/handlers/federation_sync_since.rs` unit tests — a drop-shortened FULL page does not advance over its tie group (the exact F7 bug); single-timestamp / all-dropped full pages hold the cursor.
- `src/storage/mod.rs` unit tests — gated fold drops the foreign key; counted variant reports the raw pre-drop count.

## Verification run
`cargo build --features sal`, `cargo clippy --features sal --all-targets -- -D warnings -D clippy::all -D clippy::pedantic`, `cargo fmt --all --check`, and the targeted tests above — all green locally. Module-size ceilings for `daemon_runtime.rs` and `storage/mod.rs` bumped in lockstep with dated comments.

## AI involvement
Authored by Claude Opus 5 (hard-coder) under operator authority. Design followed the task's prescribed approach; the crossroads 5-agent vote did not fire (the fix shape was directed, single-correct-answer per vector).


alphaonedev:
## Fable 5 review + security audit — APPROVE

Verified the diff at `0eb66de1` (SSH sig good). CB-14 closes three independent peer-controlled `sync_state` cursor-poisoning vectors; each can silently pin this node's per-peer catch-up cursor or skip undelivered rows behind an HTTP 200 — a data-availability class.

**Vector 1 — `next_since` pull cursor (`daemon_runtime::sync_cycle_once`) — CONFIRMED fixed.** `validate_pull_cursor` refuses a peer-advertised cursor that is non-RFC3339, exceeds `now + PULL_CURSOR_FUTURE_SKEW_SECS` (300s) — **the load-bearing far-future poison check, done on the PARSED instant, not the string** — or is not strictly greater than the current watermark. On refusal the watermark is left unchanged + WARN; the pulled rows were already merged, so availability is preserved. Applied to BOTH the `next_since` path AND the legacy `memories.last()` fallback (both peer-controlled). The strictly-greater check uses a raw-string compare, but its only imprecision (differing RFC3339 renderings of near-equal instants) is availability-safe in both directions — a string-≤ candidate is refused (no regress), a string-> but instant-earlier candidate advances to an EARLIER `since` → a superset re-pull, never a skip; the instant-correct future-skew ceiling is the real anti-poison guard. No `unwrap`/`.is_ok()` predicate on peer data.

**Vector 2 — `sender_clock` push fold (`federation_receive::sync_push`, #2670 class) — CONFIRMED fixed, this is the core CWE.** Pre-fix the fold called the un-gated `db::sync_state_merge`, folding EVERY entry in `body.sender_clock` — so peer A could inject an arbitrarily-high timestamp for peer B's key and permanently PIN B's cursor (every later pull from B returns zero rows). The new `db::sync_state_merge_authorized` folds ONLY `incoming.entries.get(authorized_peer)`, and the call site passes `&body.sender_agent_id` as `authorized_peer` — the SAME sender identity the receive path already attests: under the enrolled posture an unenrolled `x-peer-id` is refused at line ~1315 (#1056 TOFU) and `sender_agent_id` is attested against `x-peer-id` at ~1375, so the fold advances only the real peer's own key; in zero-config the whole path is faith-based and folding only the claimed key is strictly better than folding all keys. This reduces the attack from "advance ANY peer's key" to "advance only your own attested key" — the same anchor the memory/signal/transition lanes already use. Peer A's push can never touch peer B's entry.

**Vector 3 — F7 raw-count tie-group guard (`federation_sync_since`) — CONFIRMED fixed, both backends.** Root cause: a FULL page (SQL `LIMIT` rows) with one undecryptable row dropped by the #2383 mapper has `examined.len() < raw_count == limit`; keying truncation off `examined.len()` made that full page look SHORT → "not truncated" → the watermark advanced to `last`, silently skipping the boundary timestamp's other members that fell beyond the `LIMIT`. `next_since_watermark` + `warn_if_cursor_pinned` now key truncation off `raw_count` (the pre-drop SQL count) via the new `db::memories_updated_since_counted` / SAL `list_memories_updated_since_counted` (overridden on sqlite + postgres; both `flatten`-drop undecryptable rows but return the pre-drop count). A straddling tie group is now dropped and re-pulled next cycle — degrade (re-pull), never skip.

**Pg-parity decision — sound.** Vector 2 is sqlite-only because the postgres receive funnel never merged `sender_clock` into any `sync_state` (adding a pg fold would enlarge the surface, not close a hole). Vectors 1 (backend-blind, in `sync_cycle_once`) and 3 (shipped on both adapters via the counted SAL method) are backend-covered. Documented in the PR body + CHANGELOG.

**Minor observations (non-blocking):** (a) `list_memories_updated_since_counted` is a new SAL trait method (borderline T1), but its shape is dictated by the need (return the raw count alongside the rows) and mirrors the existing `list_memories_updated_since` precedent — copying the precedent is the decision, T6 does not fire. (b) Module-size ceiling bumps (`daemon_runtime.rs` 11720→11880, `storage/mod.rs` 28120→28200) are lockstep with the additions and dated `#2718`; these are size ceilings, not behavior. CHANGELOG entry present; regression set (a/b/c) present incl. the exact drop-shortened-full-page bug and the own-key-only / peer-A-cannot-advance-peer-B fold tests.

Enters the queue behind #2739 / #2740 / #2741; touches `federation_receive.rs` + `storage/mod.rs` (orthogonal to CB-7's namespace gates and CB-24's governance audit — different lines), so I expect a CHANGELOG-only rebase. Merge on green CI.

Issue #2438, opened 2026-07-27
[architecture] stated 1M+ agent target is ~3 orders of magnitude beyond the documented topology envelope (T6 = 1000+ agents / mesh ceiling ~50 peers) — no shard, placement, or cross-mesh membership model exists


alphaonedev:
## The gap

The product target under discussion is **1,000,000+ federated AI agent deployments, enterprise-class secure configuration.**

The repository's own deployment documentation states a materially smaller envelope:

| source | statement |
|---|---|
| `docs/enterprise-deployment.md:70-79` | **T6 — Multi-region / global**, the TOP production tier: *"Est. agents **1000+**, Est. ai-memory instances **50-500**"* |
| `docs/enterprise-deployment.md` §8.8 | *"**>50 peers.** The peer-to-peer mesh model is **the wrong shape** — use a gossip layer or a proper consensus coordinator"* |
| `docs/enterprise-deployment.md` §9.5 | *"A hive in v0.7.0 is a **pilot**, not a production deployment for >100 nodes."* |

### The arithmetic

T6 density is ~2–20 agents per instance. So 10⁶ agents implies **5×10⁴ – 5×10⁵ instances**, against a documented mesh ceiling of ~50 peers. That is **10³ – 10⁴ disjoint meshes**, and there is no cross-mesh membership, routing, placement, or rebalancing protocol — federation peers are a static `--quorum-peers` list.

**The documented envelope is ~10³ agents. The stated target is 10⁶. The gap is three orders of magnitude, and it is a missing architecture rather than a set of missing features.**

## Why this needs its own issue

A review was run to decide whether three identified gaps (retrieval precision, external-usage signal, collection ingest) needed covering for the 1M target. All five reviewers independently converged on a prior observation: **closing those three gaps does not move the system toward 10⁶ agents**, because the constraint is topological, not featural.

Stating it plainly matters because the alternative is a roadmap that reads as though feature work closes the target. It does not. Two honest paths exist and the choice is an operator decision:

1. **Publish a shard/placement architecture** — cross-mesh membership, namespace/tenant sharding, a coordinator or gossip layer above the mesh, and a placement model that maps agents → instances → meshes. This is the work §8.8 already points at.
2. **Amend the stated target** to the documented T6 envelope and describe 10⁶ as an aggregate across independently-operated deployments rather than one federated system.

[...]

Issue #1539, opened 2026-06-08
No HTTP/admin route to bind an agent attestation pubkey — attesting clients need out-of-band DB write under REQUIRE_AGENT_ATTESTATION (#1535)


alphaonedev:
## Finding: no HTTP/admin route to bind an agent's attestation public key (postgres-backed daemon)

Under `AI_MEMORY_REQUIRE_AGENT_ATTESTATION=1`, a write must be Ed25519-signed by an agent whose **public key is bound** (`agent_pubkey()` reads `metadata.agent_pubkey` on the `_agents` row). `POST /api/v1/agents` registers an agent, but there is **no HTTP/admin route to BIND the agent's pubkey** — `PostgresStore::bind_agent_pubkey` is internal. To enroll an attesting client against a live postgres-backed daemon, the reference-architecture harness must bind the pubkey by a direct DB write (`psql jsonb_set` of `metadata.agent_pubkey` on the regional pg, mirroring `bind_agent_pubkey`).

### Impact
A pure HTTP client cannot self-enroll an attesting identity; out-of-band DB access is required. For an enterprise multi-tenant deployment this is a usability/operability gap in the maximum-secure (attestation-required) path.

### Proposed
Add an admin-gated HTTP route (e.g. `POST /api/v1/admin/agents/{id}/pubkey`) that calls `MemoryStore::bind_agent_pubkey`, so attesting agents can be enrolled over the wire without direct DB access. Surfaced by the #1535 reference-architecture harness.

alphaonedev:
Fixed on GA golden `c7c0bddaee3b601c8b18e07ec0a16d5d1e13a937a62e8d0827a147d66ee5fcac`. `PUT /api/v1/agents/{id}/pubkey` HTTP route now exists (src/lib.rs:901 → handlers::bind_agent_pubkey, src/handlers/admin.rs:232, route const handlers/routes.rs::AGENTS_ID_PUBKEY), routing through `MemoryStore::bind_agent_pubkey` on both adapters. Attesting clients no longer need an out-of-band DB write. Covered by tests/cov_ga2_pg_handlers_1.rs::pg_bind_agent_pubkey_postgres_arm (+ register/unregistered arms).

Issue #132, opened 2026-04-12
User Guide — What's New in ai-memory v0.5.4


alphaonedev:
## What's New in ai-memory v0.5.4

**Date:** 2026-04-12  
**Audience:** End Users

---

## What is ai-memory?

ai-memory gives your AI assistant a **permanent memory** that persists across every conversation. Without it, your AI starts fresh each time — forgetting your preferences, past decisions, project context, and everything you've discussed. With ai-memory, your AI remembers what matters and recalls it exactly when it's needed.

Unlike built-in memory features that silently load hundreds of lines into every conversation (costing you tokens whether you need them or not), ai-memory uses **zero resources until you actually need something recalled**. You only pay for what you use.

---

## Works With Your AI — Whatever You Use

ai-memory isn't locked to one platform. It works with:

- **Claude Code** (Anthropic)
- **Codex CLI** (OpenAI)
- **Gemini CLI** (Google)
- **Grok** (xAI) — both CLI and API
- **Cursor** IDE
- **Windsurf** (Codeium)
- **Continue.dev**
- **Llama Stack** (Meta)
- **OpenClaw**
- And any other tool that supports the MCP standard

Switch between AI tools and your memories come with you.

---

### Three Levels of Importance

Your memories are organized by how long they should stick around:

| Level | Lasts | Best For | Examples |
|-------|-------|----------|----------|
| **Short** | 6 hours | What you're working on right now | Current debugging session, temp notes |
| **Mid** | 7 days | This week's context | Sprint goals, recent decisions, active tasks |
| **Long** | Forever | Things that should never be forgotten | Architecture choices, your preferences, lessons learned |

### It Manages Itself

You don't have to babysit your memories:

- **Memories you use often get promoted automatically.** A mid-term memory you keep recalling gets upgraded to permanent after 5 accesses.
- **Expiring memories get extended when you use them.** Recall a short-term memory and it gets another hour. Mid-term gets another day.
- **Important memories rise in priority.** The more you access something, the higher it ranks in future results.
- **No duplicates.** Store the same thing twice and it updates the existing memory instead of creating a copy.
- **Expired memories are cleaned up automatically.** Background garbage collection runs every 30 minutes.

---

### Store What Matters
Your AI stores memories with a title, content, importance level, and tags. It can mark where the memory came from (you, the AI, a hook, an import) and assign a priority from 1-10.

### Recall With Context
[...]

PR #2813, opened 2026-08-09
infra(do-hive): provision N federated substrate droplets so Track D has a real DO harness

### LINKED ISSUE #2812: [cert] Track D cannot be run: infra/do-hive provisions exactly ONE substrate droplet, so REAL federated multi-node is unprovable on DO
## The gap

`docs/v1.0.0/test-campaign-2026-08-08-enterprise-cert/PLAN.md` **Track D** requires
"a W-of-N quorum write commits locally AND replicates to the peer over the
mutually-authenticated channel; peer-enrollment + write-sig + nonce +
policy-current gates all default-ON and satisfied by REAL enrollment", and §3
states the DO round exists precisely "for the parts local cannot prove: real
federated multi-node reachability across droplets".

`infra/do-hive/main.tf` provisions exactly ONE `digitalocean_droplet.memory`
(`name = "ai-memory-hive-substrate"`, no `count`). There is no way to spawn a
second substrate node, so **Track D has no DO harness at all** — the campaign
can only re-run the LOCAL 2-daemon legs (`infra/do-hive/crypto/test-federation-mtls.sh`,
`.../test-fed-write-sig-attestation.sh`), whose peers share one loopback
interface and one process table and therefore cannot demonstrate cross-host
reachability, cross-host mTLS, or cross-host quorum replication.

Campaign gap #1 (PLAN.md §4) already records that "the enrolled happy-path had
not been exercised end-to-end and recorded as cert evidence". Without a
multi-node DO topology that gap cannot be closed on DO.

## Why terraform alone cannot close it

A DO droplet's private IPv4 is allocated by the provider at create time and
`digitalocean_droplet` exposes no input for it, so node N's `user_data` cannot
reference node M's `ipv4_address_private`. With `count`, terraform models the
whole resource as ONE graph node, so any self-reference (including
`count.index` arithmetic) is a hard `Cycle: digitalocean_droplet.memory` at plan
time. Peer wiring is necessarily a post-create step.

Separately, anything rendered into `user_data` lives verbatim in
`terraform.tfstate` — which `spawn.sh` COPIES into
`.local-runs/do-hive-runs/<ts>/` on every apply — and is readable from the
droplet's own metadata service, so private crypto material must not ride
templatefile vars.

## Proposed fix

Additive, default-off, ~155 lines of terraform + ~230 lines of guarded
cloud-init + one orchestrator script:

- `infra/do-hive/main.tf` — new `memory_count` variable (default `1`, capped
  at 8) + `quorum_writes` (default 2); `digitalocean_droplet.memory` gains
  `count`; `:9077` east-west opened between memory nodes only when
  `memory_count > 1`; new `memory_nodes` / `memory_public_ips` /
  `memory_private_ips` / `federation_enabled` outputs; cost estimate scales.
- `infra/do-hive/cloud-init-memory.yaml.tpl` — a `%{ if federation_enabled }`
  guarded systemd drop-in (`ai-memory.service.d/10-federation.conf`, mTLS +
  `--quorum-*`), a one-shot `ai-memory-fed-bootstrap.service`, and
  `/opt/ai-memory/fed-bootstrap.sh` (mint node identity, mint api key, wait for
  the operator bundle, local #2477 peer-scheme pre-check, cross-enroll peer
  pubkeys, start serve, wait for each peer over the mutually authenticated
  channel, bind the author pubkey over the #1539 admin route). At
  `memory_count = 1` the rendered user_data must stay BYTE-IDENTICAL.
- `infra/do-hive/crypto/gen-certs.sh` — additive `HIVE_NODE_IPS` mode minting
  `hive-node-1..N` + `hive-operator` leaves and per-node allowlists; unset
  keeps the legacy output byte-identical.
- `infra/do-hive/federate.sh` — orchestrator-side post-apply wiring + the Track
  D assertions. Never calls terraform, never spawns a droplet.
- `infra/do-hive/README-measurement.md` — the 2-node run recipe, what to verify
  post-boot, and the secrets-in-state tradeoff stated honestly.

Money gate (`AI_MEMORY_OPERATOR_DO_SPEND_APPROVED=1`), VPC, SSH firewall rule,
audit dump and `teardown.sh` all UNCHANGED.

alphaonedev:
## What

`infra/do-hive` provisioned exactly ONE `digitalocean_droplet.memory`, so **Track D of the v1.0.0 enterprise-cert campaign had no DO harness at all**. PLAN.md Track D asks for "a W-of-N quorum write [that] commits locally AND replicates to the peer over the mutually-authenticated channel; peer-enrollment + write-sig + nonce + policy-current gates all default-ON and satisfied by REAL enrollment", and §3 says the DO round exists for "the parts local cannot prove: real federated multi-node reachability across droplets". The only runnable evidence was the LOCAL 2-daemon legs (`crypto/test-federation-mtls.sh`, `crypto/test-fed-write-sig-attestation.sh`), whose peers share one loopback interface and one process table — they prove the CONFIG, never cross-host reachability, cross-host mTLS, or cross-host quorum replication.

This adds `memory_count` (default `1`) and the wiring a real 2-node mesh needs.

**Additive and default-off.** At `memory_count = 1` the rendered `cloud-init-memory.yaml.tpl` user_data is **byte-identical** to the parent commit (rendered both through `templatefile()` and `cmp`-clean, `sha256 30e3f752…751ee78`); the firewall rule set is unchanged; `spawn.sh` / `teardown.sh` / the `AI_MEMORY_OPERATOR_DO_SPEND_APPROVED=1` money gate / the audit dump are UNTOUCHED. **BUILD-ONLY: no `terraform apply`, no droplets, no spend.**

### Why there is a second command (`federate.sh`) instead of pure terraform

Two constraints, both load-bearing:

1. **Terraform cannot express peer IPs.** A DO droplet's private IPv4 is allocated by the provider at create time and `digitalocean_droplet` exposes no input for it. With `count`, terraform models the whole resource as ONE graph node, so any self-reference — including `count.index` arithmetic like `memory[1 - count.index]` — is a hard `Cycle: digitalocean_droplet.memory` at plan time. Peer wiring is necessarily post-create. What DOES ride templatefile vars is everything knowable at plan time: `node_index`, `node_count`, `fed_identity`, `quorum_writes`.
2. **Secrets must stay out of terraform state.** Anything rendered into `user_data` lives verbatim in `terraform.tfstate` — which `spawn.sh` **copies into `.local-runs/do-hive-runs/<ts>/` on every apply** — and is also readable from the droplet's own metadata service.

### Cert-material tradeoff (the brief's explicit ask)

**Chosen:** private crypto material never touches terraform. The CA + leaf keys are minted on the OPERATOR host by `federate.sh` (reusing `crypto/gen-certs.sh` via a new additive `HIVE_NODE_IPS` mode) and pushed over SSH. Each node's Ed25519 federation signing key is generated **on the droplet** and its private half never moves at all; only `.pub` files cross between nodes — the same public-material-only discipline `infra/lan-parity-test/provision-peer-keys.sh` follows for the docker mesh (#1803). Net: **zero secrets in `terraform.tfstate`**, and none in the spawn.sh audit dump.

**Rejected (a) pre-generate locally + inject via templatefile vars:** every peer private key and the CA key would land in both state and droplet metadata. Directly contradicts "keep secrets out of terraform state".

**Rejected (b) generate on droplet 1, distribute to the others:** there is no authenticated channel between two fresh droplets *before* the mTLS material exists (that material *is* the channel). Any bootstrap over the VPC would be trust-on-first-use over plaintext — which would make the cert round's Leg-2 "unauthorised + plaintext peers are refused" assertion rest on a plaintext, unauthenticated key exchange. A federation-encryption certification cannot be built on that.

**Honest cost of the chosen approach:** `terraform apply` alone does not yield a running mesh. It yields two nodes parked in a **fail-closed wait**, and one operator command (`federate.sh`) completes them. That is a real extra step — and it is the step `crypto/KNOWN-DO-STAGING.md` §1 already prescribes ("scp/rsync the whole `crypto/` dir onto each droplet").

### Fail-closed, not silently-degraded

The federation drop-in's `EnvironmentFile` has **no leading `-`**, so systemd refuses to start `serve` until `/etc/ai-memory/fed/peers.conf` exists. A "federated" node that booted with an empty peer list would accept writes, satisfy `W=1` locally, and report healthy while replicating to nobody. `fed-bootstrap.sh` additionally refuses a non-`https` peer URL locally, mirroring the daemon's own #2477 boot refusal, so the cause is a named log line rather than an opaque unit failure.

### `AI_MEMORY_FED_*` are deliberately left UNSET

At v1.0.0 write-sig (#94), signal-sig (#96), transition-sig (#87), checkpoint-sig (#125), nonce (#30), peer-enrollment (#43) and policy-current (#132) are fail-closed by **compiled default**. Setting them to `1` would prove "the flag works", not "the shipped default is secure" — the same reasoning `crypto/test-fed-write-sig-attestation.sh` states for peerB ("env left unset on purpose to prove the DEFAULT-ON posture").

### Data-integrity notice

Adding `count` moves the state address `digitalocean_droplet.memory` → `…memory[0]` and renames the droplet (`name` is ForceNew). A LIVE pre-change hive must `terraform state mv 'digitalocean_droplet.memory' 'digitalocean_droplet.memory[0]'` **before the next plan**, or terraform destroys and recreates the substrate droplet along with its postgres volume. This is documented in the `main.tf` header and in the CHANGELOG entry.

## Files changed

| file | change |
|---|---|
| `infra/do-hive/main.tf` | `memory_count` (default 1, capped 8) + `quorum_writes` (default 2, `>= 2`); `count` on the memory droplet, named `ai-memory-hive-memory-1..N`; `lifecycle.precondition` refusing `quorum_writes > memory_count`; `:9077` east-west between memory nodes only when `memory_count > 1`; `memory_nodes` / `memory_public_ips` / `memory_private_ips` / `federation_enabled` outputs (singular `memory_public_ip` / `memory_private_ip` preserved as node 1); scaled cost estimate; one pre-existing `terraform fmt` alignment nit folded in |
| `infra/do-hive/cloud-init-memory.yaml.tpl` | `%{ if federation_enabled }`-guarded systemd drop-in (`ai-memory.service.d/10-federation.conf` — overrides ONLY `ExecStart`, so the base unit and the postgres/AGE/pgvector half cannot drift), one-shot `ai-memory-fed-bootstrap.service`, and `/opt/ai-memory/fed-bootstrap.sh` |
| `infra/do-hive/crypto/gen-certs.sh` | additive `HIVE_NODE_IPS` mode: `hive-node-1..N` leaves + per-node allowlists pinning every other node and the node itself, and nothing else (no operator/bastion cert — an unused trust anchor is a liability). Unset ⇒ legacy output byte-identical |
| `infra/do-hive/federate.sh` | **new.** Post-apply wiring + the Track D assertions. Never calls terraform, never spawns a droplet |
| `infra/do-hive/README-measurement.md` | Track D section: `-var memory_count=2` recipe, verification table, triage, unchanged teardown, the tradeoff |
| `CHANGELOG.md` | `[Unreleased]` entry |

No Rust changed; no MCP tool / HTTP route / CLI verb / schema version added, so no SSOT pin moves.

## On-droplet bootstrap (`fed-bootstrap.sh`) — a resumable state machine

A mint node Ed25519 identity (private half never leaves) → B mint local `api_key` + `config.toml` (the daemon correctly REFUSES a non-loopback bind without one — `KNOWN-DO-STAGING.md` §3) → C wait for the operator bundle → D local #2477 peer-scheme pre-check → E cross-enroll peer pubkeys into the key dir (transport lane) → F start `serve`, wait for local `/health` over mTLS → G **retry until every peer answers `/health` over the mutually authenticated channel** (the retry loop the brief asks for, using the terraform-derived peer addresses) → H register + bind the cert-round author pubkey over the admin-gated `PUT /api/v1/agents/{id}/pubkey` (#1539) because CLI `agents bind-key` speaks sqlite only, so the content write-sig lane can reach `agent_attested` on this postgres substrate.

Re-run after fixing a cause with `systemctl restart ai-memory-fed-bootstrap`. It never deletes or rewrites durable memory text.

## Verification (`./federate.sh verify`)

**Every assertion runs ON a droplet over SSH, never from the orchestrator host** — the peer URLs are PRIVATE VPC addresses and the `:9077` rule admits only hive droplet ids. (This was a real defect in the first revision; see `a5e9d535`.)

| assertion | expected |
|---|---|
| each node `/api/v1/health` over mTLS (its own peer cert, via loopback) | `200` |
| **cross-host**: node 1 -> node 2 at `https://<node-2 private ip>:9077/api/v1/health` with node 1's peer cert | `200` — the assertion the local 2-daemon legs structurally cannot make |
| each node with **no** client cert | refused at the rustls allowlist |
| `W=2` quorum write to node 1 | `201 quorum_met` (or `202` locally-durable on a late ack) |
| that row read back from **node 2** | present — it actually replicated |
| **signed** write authored on node 1, read at node 2 | `metadata.attest_level == "agent_attested"` |

## Evidence

- `terraform init -backend=false` + `terraform validate` → **Success! The configuration is valid.** (Terraform v1.14.9, run in the `.local-runs/` worktree copy; the real `terraform.tfstate` was never touched — it does not exist in the worktree, and no apply/plan against a provider was run)
- `terraform fmt -check` → clean
- Byte-identity at `memory_count=1`: rendered the parent-commit template and the new template through `templatefile()` and `cmp`-compared → identical, `sha256 30e3f752…751ee78`
- Rendered federated cloud-config parses as YAML; `bash -n` clean on the rendered `fed-bootstrap.sh` and `provision.sh`; `bash -n` clean on `federate.sh` + `gen-certs.sh`
- `gen-certs.sh` exercised in BOTH modes: unset ⇒ identical legacy file set; `HIVE_NODE_IPS="10.20.0.5 10.20.0.6"` ⇒ correct SANs (`DNS:localhost, IP:127.0.0.1, IP:10.20.0.5`) and 3-entry per-node allowlists
- Gates green: `check-cloud-init-ascii` (+ `--self-test`), `check-migration-ladder`, `check-l3-boundary`, `check-vendor-literals`, `check-hardcoded-literals`, `check-doc-symbol-anchors` (+ `--self-test`), `check-ci-job-claims`, `check-docs-vs-ssot`, `check-sdk-route-paths`, `check-required-contexts`, `check-conformance-readers`, `check-git-dependency-sources`

**No droplets were spawned and no spend was incurred.**

## AI involvement

Authored end-to-end by Claude Opus 5 under operator-delegated authority per `docs/AI_DEVELOPER_WORKFLOW.md` §8.2. Prior art extended (not forked) per the standing operator rule: `crypto/gen-certs.sh`, `crypto/test-federation-mtls.sh`, `crypto/test-fed-write-sig-attestation.sh`, `crypto/KNOWN-DO-STAGING.md` §§1-3, `infra/lan-parity-test/provision-peer-keys.sh` (#1803), and the #2438 additive-toggle pattern in `main.tf`.

Closes #2812


alphaonedev:
### Self-review hardening pushed (`a2c15954`)

Three narrower blast radii on the on-droplet bootstrap; no change to mesh behaviour.

1. **Custody split.** `/etc/ai-memory`, `$FED_DIR`, `$FED_DIR/peers` and `$XDG_DIR` are now **root-owned, group `aimemory`, 0750** — the service user READs its trust anchors (`ca.crt` 0644, `node.crt` 0644, `node.key` 0640, `peers.allowlist`, `peers.conf`) but cannot REWRITE them, so a compromised daemon cannot swap the CA it validates peers against. Only `$KEY_DIR` stays daemon-writable (0700 `aimemory`), because `identity generate` runs as `aimemory` and the daemon legitimately owns its own signing keys. Same separable-custody principle the substrate already applies to `AI_MEMORY_WITNESS_KEY_DIR` ("a mount the daemon can READ but a compromised daemon process cannot overwrite"). The first commit chown'd the whole tree to `aimemory:aimemory`.
2. **`config.toml` carries the `api_key`**, so it is written inside a `( umask 077 … )` subshell rather than created at the inherited 0644 and chmod'd afterwards — it never exists world-readable even momentarily. Final mode `root:aimemory 0640`; `/etc/ai-memory/api-key` the same.
3. **`peers.conf` is PARSED, not `.`-sourced.** systemd reads it as a plain `KEY=value` `EnvironmentFile` and grants it no shell semantics, so sourcing handed an over-the-wire-delivered file arbitrary-code-execution reach the script never needed. Parse verified against both the plain and the quoted+CRLF forms.

Re-verified after the change: `memory_count=1` render still **byte-identical** to the parent-commit template (`cmp`-clean); the federated render still parses as YAML; `bash -n` clean on the rendered `fed-bootstrap.sh`; `check-cloud-init-ascii` + `check-hardcoded-literals` green.

alphaonedev:
### Self-review found a real defect in my own harness (`a5e9d535`) — fixed

Not a style nit. `federate.sh verify` curl'd `${PEER_URLS[i]}` — `https://<private VPC ip>:9077` — **from the orchestrator host**. Those addresses are not routable outside the DO VPC, and the `:9077` firewall rule admits only hive droplet ids, so **every assertion would have failed with a connection error regardless of whether the mesh was healthy**. A verification harness that cannot pass when its subject is correct is worse than no harness: it would have been read as "Track D is red".

**Fix — every assertion now runs ON a droplet over SSH:**
- `node_sh <idx>` pipes a script to `ssh root@<public ip> bash -s`; `node_get` / `node_post` hit `https://127.0.0.1:9077` with that node's own `/etc/ai-memory/fed` material and its own api key. JSON bodies ride base64 so no quoting has to survive the ssh command line (expansion verified by substituting `cat` for `ssh` and inspecting the emitted remote script).
- **New assertion**: cross-host mTLS — node 1 → node 2 at node 2's *private* address with node 1's peer cert. That is exactly the property the local 2-daemon legs structurally cannot demonstrate, and the first revision could not have demonstrated it either.

**Consequence:** the `hive-operator` client cert is **removed** from `gen-certs.sh`'s `HIVE_NODE_IPS` mode. It existed solely so the orchestrator host could probe the nodes; nothing needs it now, and an unused trust anchor sitting on every node's mTLS allowlist is a liability, not a convenience. Each `hive-node-N.allowlist` is back to exactly two classes — every OTHER node (the quorum channel) and the node ITSELF (the only client cert it holds, used by the on-droplet bootstrap probe and these assertions). Verified: a 2-node mint yields 2-line allowlists.

README + CHANGELOG + this PR body updated to match. `bash -n` clean; `gen-certs.sh` exercised in `HIVE_NODE_IPS` mode; `terraform validate` Success, `fmt` clean; the eight doc/infra gates re-run green.

Issue #2836, opened 2026-08-09
docs(CONFIG_SCHEMA): sample config ships max_inflight_requests = 0 labelled 'opt-in' — silently disables default-ON admission control (#2032 M3); contradicts API_REFERENCE


alphaonedev:
**3x7 lane 3, 2026-08-09.** Base `release/v1.0.0` @ `f7399cfb`. **Routed to the config-docs lane (lane 1) — lane 3 deliberately did NOT edit these files to avoid a same-line collision.**

## Finding — the canonical config reference ships a sample that disables DoS admission control

`docs/CONFIG_SCHEMA.md:136-139`:

```toml
max_inflight_requests = 0          # #1733 HTTP admission cap; 0 = disabled
                                   # (opt-in). Positive n sheds >n concurrent
                                   # in-flight requests with a typed 503.
```

and `:127-129`:

> "All fall back to the compiled default when absent, non-positive, or unparseable."

Both statements are false since **#2032 M3**.

## Truth

`src/config.rs:4276-4330` — admission control is **ON by default**, resolving TRI-STATE to the CPU-scaled `clamp(available_parallelism × 64, 256, 4096)`. An **explicit `0` is a sentinel that means DISABLED and does NOT fall back** — that distinction is the entire point of the tri-state resolver (pre-#2032 `unset == 0 == disabled` because `env_pos_usize`'s `>0` filter collapsed the two).

## Impact

This is the canonical, copy-pasteable config reference. An operator who copies the sample block **explicitly disables** admission control fleet-wide — re-opening the single-authenticated-caller DoS against the one `Arc<Mutex<Connection>>` sqlite handle that #2032 M3 closed. It is a manageability-at-scale regression introduced by documentation alone, and the failure is silent: the daemon starts normally.

It also **directly contradicts** `docs/API_REFERENCE.md:234-255`, which states correctly: *"Admission control is ON by default … only an explicit `0` disables it."* Two tier-1 references disagree on a security default.

## Second instance

`ROADMAP.md:684` — "`AI_MEMORY_MAX_INFLIGHT_REQUESTS` (`[limits].max_inflight_requests`, default `0` = opt-in)", inside a v0.8.0 shipped-state blockquote that is not marked superseded.

## Proposed fix

1. `docs/CONFIG_SCHEMA.md` — remove `max_inflight_requests = 0` from the sample or comment it out; state the CPU-scaled default and that an explicit `0` is the only way to disable.
[...]

Issue #2835, opened 2026-08-09
docs(README): token-budget published as a 'required status check' + 'hard CI gate' with the wrong ceiling — and check-ci-job-claims.sh was structurally blind to the citation shape


alphaonedev:
**3x7 lane 3, 2026-08-09.** Base `release/v1.0.0` @ `f7399cfb`.

## The false claims

`README.md:794-795` (pre-fix):

> ### Token-Budget Gate (**hard CI gate**, v0.7 C5)
> The `token-budget` workflow is a **required status check**. …
> …the trimmed `tools/list` payload under `--profile full` may not exceed **11,000** cl100k_base tokens

Three defects:

1. **Not required.** `grep -ci token scripts/qc-allowlists/required-contexts-release.txt` → **0**. `token-budget` has no required context on `release/v1.0.0`; a red run does not block merge.
2. **Not a hard gate.** Same reason; the heading asserts it twice.
3. **Ceiling conflation.** The workflow enforces **6750** (`.github/workflows/token-budget.yml:115,131,141`, two dated bumps 6500→6650→6750). **11,000** is the deliberately-looser `cargo test` backstop (`tests/token_budget_guard.rs:87`). The README attributes the test's number to the workflow, so a contributor sizing a new tool budgets ~4,250 tokens of headroom that does not exist.

This is precisely the class #2629 built gate 11 to counter — *"a control was removed and the prose was not"*, and the register's stated systematic bias toward **more claimed enforcement than exists**.

## Why gate 11 could not see it

`scripts/check-ci-job-claims.sh` **did** scan `README.md`. It missed the sentence on citation shape:

- `WF_TICK = re.compile(r"`([A-Za-z0-9_.-]+\.ya?ml)`")` — requires a `.yml` suffix. README wrote `` `token-budget` ``.
- `JOB_CITE` — requires the literal word "job". README wrote "workflow".

So no citation was registered on that line, and the ENFORCEMENT rule never evaluated it. The most-quoted enforcement sentence in the tier-1 doc was structurally invisible to the gate built for exactly that claim. Running the gate at HEAD pre-fix: `✅ PASS (17 workflows parsed)`.

[...]

Issue #2832, opened 2026-08-09
docs: 6 surfaces publish 25/27 hook events (SSOT 22) and advertise REMOVED events; the docs-vs-SSOT HookEvent rule was a 2-phrasing hand regex over an incomplete scan set


alphaonedev:
**3x7 lane 3, 2026-08-09.** Base `release/v1.0.0` @ `f7399cfb`.

## The drift

`HOOK_EVENTS_COUNT = 22` (`src/config.rs:1045`). Six live surfaces published 25 or 27:

| File:line | Published |
|---|---|
| `docs/audience/developer.html:211,212` | **27** — and names **`pre_recall`** as a live example |
| `docs/compliance/nsa-csi-mcp.html:276,389,689` | **27** ("Track G 27-event hook pipeline") |
| `README.md:138` | **25** — and its enumeration names FOUR removed events |
| `docs/production-deployment.md:167` | **27** |
| `docs/strategy/coala-mapping.md:34` | **27** |
| `docs/essays/brass-tacks-3-why.html:99` | **25** |
| `ROADMAP.md:799` and `:1263` | **27**, contradicting `ROADMAP.md:431`'s correct **22** in the same file |

`pre_archive` was removed by #2637; `pre_recall` / `pre_search` / `pre_transcript_store` / `post_transcript_store` by #2758. **#2780** corrected the compliance count in one place and missed three cells on the same page.

The naming is worse than the count. #2758's own stated principle is *"a hook the substrate advertises must actually fire, or it must not be advertised"* — yet the developer-facing page still advertises `pre_recall`, and a `[hooks].required_events` entry naming it now **fails to parse**.

## Why the gate was green through all of it

`scripts/check-docs-vs-ssot.sh` reported `hooks=22` and **PASSED**. Two independent halves:

1. **Phrasing.** The rule was a two-alternative hand regex:
   ```
   '\*\*([0-9]+) hook lifecycle events\*\*|([0-9]+) lifecycle events\) — A programmable'
   ```
   None of `27-event hook pipeline`, `fires on 27 named substrate events`, `27 lifecycle events, see`, or unbolded `27 hook lifecycle events` matches. This is exactly the defect #2492 diagnosed and fixed for the *other* SSOT consts — *"a document that says the same thing in the seventh way nobody enumerated is invisible"* — the HookEvent rule was simply left behind as a legacy hand-regex.
[...]

Issue #2829, opened 2026-08-09
docs(encryption.html): claims CA-chain validation that does not exist, omits that federation is not E2E encrypted (#1968), calls a default-ON knob opt-in


alphaonedev:
**3x7 lane 3, 2026-08-09.** Base `release/v1.0.0` @ `f7399cfb`.

`docs/encryption.html` is the procurement-facing crypto page. It opens: *"Four crypto surfaces, every one named, every one with the algorithm and the threat model."* Four defects against that standard.

## 1. Claims a CA-chain validation that does not happen — CERT-BLOCKER

`docs/encryption.html:351` (pre-fix):

> "The receiving daemon **validates the peer's cert chain (using the configured CA bundle)** AND checks its SHA-256 fingerprint against the allowlist."

`src/tls.rs:1056-1072`:

```rust
fn verify_client_cert(
    &self,
    end_entity: &rustls::pki_types::CertificateDer<'_>,
    _intermediates: &[rustls::pki_types::CertificateDer<'_>],
    _now: rustls::pki_types::UnixTime,
) -> Result<ClientCertVerified, rustls::Error> {
    let fp: [u8; 32] = Sha256::digest(end_entity.as_ref()).into();
    if allowlist_contains_ct(&self.allowlist, &fp) { Ok(...) } else { Err(...) }
}
```

`_intermediates` and `_now` are **ignored**. No chain construction, no CA bundle, no expiry check. It is PIN-ONLY — the SSH `known_hosts` model, exactly as CLAUDE.md env row #85 documents for the OUTBOUND mirror ("the pin is the trust anchor; the CA chain is not consulted").

The second clause is true and is the real control. The first claims a defence-in-depth layer that is absent.

## 2. Never states federation is NOT end-to-end encrypted — CERT-BLOCKER (overclaim by omission)

Surface 1 is the #228 per-memory ChaCha20-Poly1305 envelope; Surface 2 is "In transit (federation)", presented as solved. Nowhere does the page say that **federation replicates plaintext memory content**.

That is not a nitpick — it is the stated rationale of two shipped v1.0.0 controls. #2448 and #2477 both say verbatim: *"Federation replicates plaintext memory content (NOT end-to-end encrypted — `src/encryption/mod.rs`, #1968 open)."* **#1968 is OPEN.** Every peer and every TLS-terminating hop sees cleartext memory text.

[...]

PR #2840, opened 2026-08-09
docs(v1.0.0): 3x7 lane-3 — behavioral/enforcement claims, Pages overclaims, compliance, CI-job truthfulness

alphaonedev:
Lane 3 of the operator-mandated 3x7 documentation-drift sweep of `release/v1.0.0`.

**Scope:** behavioral / enforcement claims, GitHub Pages + marketing HTML, compliance docs, cross-doc + CHANGELOG consistency, CI-job-claims truthfulness. Lane 1 owns config/env-var docs and lane 2 owns API-surface counts — overlaps were routed to those owners rather than edited here, and the one shared file (`feature-matrix.html`) got a **sibling** callout so a concurrent count fix cannot conflict.

Every claim was verified against code or an SSOT test at this base. Where a sweep proposed a finding the code did not support, it was dropped and recorded as such.

## Highest severity — a document that weakened a security control

`docs/integrations/claude-code.md` prescribed, in a copy-pasteable `~/.claude.json` block:

```jsonc
"env": { "AI_MEMORY_REQUIRE_AGENT_ATTESTATION": "0" }
```

on the premise that "as of v0.9.0, ai-memory **rejects unsigned writes**". False since #1985 — `resolve_require_agent_attestation` (`src/identity/attest.rs`) resolves an unset knob to `matches!(surface, WriteSurface::HttpDirect)`, so MCP `memory_store` is permissive and a stock Claude Code setup needs no override at all.

Because `=0` is **global**, following the guide turned OFF the fail-closed `403 ATTESTATION_FAILED` on the HTTP direct-write surface — the network surface, the only place the gate is load-bearing. The call-out now **explicitly retracts** the instruction rather than silently dropping it, so an operator who already applied it learns to remove it. Same unscoped claim corrected on 13 further live surfaces.

## Pages / marketing overclaims

Pre-fix the strings `59`, `21` and "59 of 80" appeared in **zero** HTML files: #2811 corrected the markdown and never reached the rendered pages, leaving ~60 current-facing sites implying the backends are one API. Fixed the load-bearing six:

| Surface | Was |
|---|---|
| `performance.html:365` | "SQLite operations and PG operations stay **behaviorally identical**" — in the paragraph recommending Postgres for >1M scale |
| `architectures-t5.html:514` | "**identical surface area** from one agent to the global hive" |
| `whats-new-v09.html:248` | "**one API**" — contradicting `:446` on the same page |
| `index.html:436` | "**Identical semantics**" |
| `evidence.html:256` | "**full SAL-parity suite**" under a Production-ready badge |
| `feature-matrix.html:630` | the canonical route table — "the same operations as MCP", **no backend column at all** |

`docs/encryption.html` carried three separate crypto defects: it claimed the daemon "validates the peer's cert chain (using the configured CA bundle)" when `FingerprintAllowlistVerifier::verify_client_cert` takes `_intermediates` and `_now` as **ignored** parameters (pin-only, SSH `known_hosts` model); it never stated that **federation is not end-to-end encrypted** (#1968 **OPEN** — the stated rationale for both #2448 and #2477); and it called `AI_MEMORY_FED_REQUIRE_WRITE_SIG` "opt-in strict mode" when `FED_REQUIRE_WRITE_SIG_DEFAULT = true`.

## Hook events + compliance

Six surfaces published 25 or 27 against `HOOK_EVENTS_COUNT = 22`, two of them advertising events #2637/#2758 **removed** — including `docs/audience/developer.html` naming `pre_recall`, whose removal rationale was that advertising a hook that never fires is a false enforcement claim. #2780 fixed one compliance cell and missed three on the same page. `ROADMAP.md` contradicted itself (22 at `:431`, 27 at `:799` and `:1263`).

## Two gates widened — both proven load-bearing

**`check-docs-vs-ssot.sh`** reported `hooks=22` and PASSED through all of the above. The HookEvent rule was the last legacy hand-regex #2492 left un-generalised, over a scan set that could not reach four of the six surfaces. Generalising it surfaced two things worth noting:

- this per-rule function had **no historical guard** (only the #2492 scanner did), which was harmless only while every regex was narrow enough to miss history by accident — so a mirrored `is_historical` was added to protect README's prior-release paragraph and ROADMAP's frozen v0.7.1 baseline;
- `(?<!ships )` alone is insufficient on the bare `lifecycle events` anchor — without `(?<![0-9])` the engine re-anchors one character right and matches `5` out of `25`.

On its first run the widened rule caught 7 real sites.

**`check-ci-job-claims.sh`** could not see ``The `token-budget` workflow is a **required status check**``: `WF_TICK` needs a `.yml` suffix and `JOB_CITE` needs the literal word "job", so no citation registered and the ENFORCEMENT rule — built for exactly that false claim — never evaluated the line. `token-budget` has no required context, and the cited 11,000 ceiling is the `cargo test` backstop; the workflow enforces 6750. Added the `WF_WORD` citation shape and widened the scan set.

Both widenings carry new `--self-test` legs asserting the pre-fix shapes are REJECTED (naming the rule) **and** that the near-misses still PASS.

## Preserved, not "corrected"

True historical snapshots (`whats-new-v0*`, `v070-*`, frozen v0.7.1 baselines) are untouched. `docs/v1.0.0/perfect-endpoint-assessment/` gets a FROZEN-ARTEFACT banner with a five-row reversal table instead of a rewrite — those 49 ballots drove the v1.0.0 decisions, and re-pointing their numbers would falsify the record.

## Verified and deliberately NOT changed

- `essays/brass-tacks-3-why.html:86` "the same Cypher-shaped query surface in both" — **accurate**; the whole `kg_*` family is pg-supported, only the legacy `/api/v1/find_paths` alias is 501.
- `PostgresStore::link_signed` is **not** a no-op at this base (it calls `link_internal`), and #1536/#1541/#1542/#1543 are all **CLOSED** — the reference-architecture annex publishing them as open PG gaps is drift in the *under*-claiming direction, routed as a separate finding.

## Findings filed

#2827 (attestation class incl. the security-weakening remediation) · #2828 (Pages pg-parity) · #2829 (encryption.html) · #2832 (hook events + gate gap) · #2835 (CI-job claims + gate blind spot) · #2836 (routed to lane 1 — `CONFIG_SCHEMA.md` ships `max_inflight_requests = 0` labelled "opt-in", silently disabling default-ON admission control and contradicting `API_REFERENCE.md`)

Full register: `.local-runs/3x7-2026-08-09/lane3-claims-pages-register.md`.

Gates run locally: `check-docs-vs-ssot.sh`, `check-ci-job-claims.sh`, `check-doc-symbol-anchors.sh` — all PASS, plus both `--self-test` suites.

Not merging — Fable 5 is sole gatekeeper.


alphaonedev:
Superseded by the consolidated 3x7 docs-drift integration PR #2844 (merged fa73a7bf). All of this PR's fixes landed there verbatim after Fable resolved the 13-file cross-lane collisions (validate.rs=const-name form, docs-vs-ssot gate=coherent union of all three widenings with self-test proven load-bearing, every security-truthfulness fix preserved). Branch deleted on merge.

Issue #2942, opened 2026-08-15
asi-hard posture cannot cold-boot a fresh node: REQUIRE_ROLLBACK_CHECK refuse-open with no off-table head anchor (exit 75)


alphaonedev:
## Summary
A fresh (empty-DB) node started under the certified `asi-hard` security posture (`AI_MEMORY_SECURITY_PROFILE=asi-hard`, e.g. via `docs/deploy/asi-hard.env`) REFUSES to boot:

```
fatal: rollback-check refuse-open (AI_MEMORY_REQUIRE_ROLLBACK_CHECK): no pinnable
  off-table head anchor available to check the DB head against (exit 75)
```

Root cause (verified): asi-hard PINS `AI_MEMORY_REQUIRE_ROLLBACK_CHECK=1` (env #124, `src/security_profile.rs::KNOBS`) with a no-disable contract. In require-mode, `enforce_rollback_check_at_open` (`src/storage/connection.rs::open`) treats an ABSENT/unpinnable off-table head anchor (`RollbackCheck::Missing`) as a refuse-to-open. That anchor (`head-anchor.log` on the witness key dir) is emitted ONLY by the witness watermark cadence over the `signed_events` audit chain (`src/signed_events.rs::try_emit_audit_head_witness`), which is EMPTY on a fresh node — ordinary `memory_store` writes do not append to `signed_events`, so `head_seq==0` and `try_force_emit` no-ops. There is no ordinary path to seed the anchor before the first open, so a fresh asi-hard node is un-bootable (chicken-and-egg).

## Why this matters (enterprise-ready)
The certified enterprise-federation posture IS asi-hard. An operator performing a FIRST/COLD deployment of a certified node (fresh DB, no pre-seeded anchor) hits `exit 75` and cannot start. This is directly load-bearing for the "an enterprise can bet the business on it" standard: the certified posture must be cold-deployable, or the exact bootstrap procedure must be documented + validated.

## Repro
```
source docs/deploy/asi-hard.env          # sets AI_MEMORY_SECURITY_PROFILE=asi-hard + the 17 pins
# enroll operator pubkey + role keys so the OTHER pins are satisfiable
ai-memory serve --db <fresh.db> --host 127.0.0.1 --port 9077 --tls-cert ... --tls-key ... --mtls-allowlist ...
# => fatal: rollback-check refuse-open ... (exit 75)
```
[...]

Issue #2941, opened 2026-08-15
agents bind-key: intermittent silent no-op (register row created but metadata.agent_pubkey unset → signed store 403s)


alphaonedev:
## Summary
During the MCP-over-HTTP attested-proxy verification (v1.0.0 federation-config assessment), `ai-memory agents bind-key` intermittently did NOT persist the agent's public key: in 1 of 4 fresh-DB enrollment runs, `agents register` created the reserved agents-namespace registry memory-row but `metadata.agent_pubkey` was left UNSET. The subsequent signed `POST /api/v1/memories` (WriteSurface::HttpDirect, `AI_MEMORY_REQUIRE_AGENT_ATTESTATION=1`) then failed closed with `403 ATTESTATION_FAILED` because `prepare_signed_store` had no bound pubkey to verify against.

An intermittent enrollment failure that leaves attestation unsatisfiable is a real defect (per project memory: "an intermittent test = a real bug"), and it is directly load-bearing for the certified enterprise-federation posture where every direct write must be attested.

## Evidence
- Observed 1/4 fresh-DB runs of the proxy-shim `verify.py` harness (`.local-runs/cert-federation/mcp-proxy/verify.py`).
- Failure signature: `agents list` shows the agent, but the reserved agents-namespace registry row's `metadata.agent_pubkey` is absent; signed store 403s.
- Note: `agents list` does NOT expose the pubkey, so the silent no-op is invisible to the standard enrollment check — the harness had to read the registry row's `metadata.agent_pubkey` directly to detect it.

## Repro (intermittent)
Fresh DB → `ai-memory identity generate` → `ai-memory agents register` → `ai-memory agents bind-key` → read the reserved agents-namespace registry row's `metadata.agent_pubkey`. Occasionally unset after a nominally-successful bind.

## Proposed fix (concrete)
1. `agents bind-key` must FAIL LOUD (non-zero exit + typed error) if, after the bind, `metadata.agent_pubkey` is not readable on the registry row — never a silent success. (`src/identity/` + the `agents` CLI dispatch in `src/daemon_runtime.rs` / `src/cli/`.)
2. Make register+bind atomic (single transaction) so a partial enrollment cannot leave a pubkey-less registry row.
3. Add a regression test that enrolls, then asserts `metadata.agent_pubkey` is present and that a signed store lands `attest_level=agent_attested`; run it under a loop/`--test-threads=1` to catch the intermittency.
[...]

Issue #2852, opened 2026-08-10
infra(do-hive): Track D serve fail-closes on 0.0.0.0 bind — api_key written to XDG path the daemon never reads (config_path ignores XDG_CONFIG_HOME)


alphaonedev:
## Root cause
The Track D federation serve drop-in sets `Environment=XDG_CONFIG_HOME=/etc/ai-memory/xdg` and `Environment=HOME=/opt/ai-memory`, and `fed-bootstrap.sh` **stage B** writes the minted `api_key` config to `$XDG_CONFIG_HOME/ai-memory/config.toml` (`/etc/ai-memory/xdg/ai-memory/config.toml`). But the daemon's config resolver **ignores `XDG_CONFIG_HOME`**: `AppConfig::config_path()` (`src/config.rs:7072-7076`) reads `$HOME` only and joins the hardcoded `const CONFIG_DIR = ".config/ai-memory"`. So the daemon loads `/opt/ai-memory/.config/ai-memory/config.toml` (which has NO api_key), never the XDG file stage B wrote.

## Evidence
```
journalctl -u ai-memory:
  ai-memory: loaded config from /opt/ai-memory/.config/ai-memory/config.toml   # NOT the XDG path
  ERROR daemon bootstrap failed; error=refusing to bind to non-loopback address "0.0.0.0"
        without an API key: the daemon's api_key is unset ... (exit 75)
# but the api_key IS present in the XDG file:
grep api_key /etc/ai-memory/xdg/ai-memory/config.toml
  api_key = "84300896..."   # written by fed-bootstrap.sh stage B, never read
```
Source: `src/config.rs:7072`
```rust
pub fn config_path() -> Option<PathBuf> {
    let home = std::env::var("HOME").ok()?;          // XDG_CONFIG_HOME is never consulted
    Some(Path::new(&home).join(CONFIG_DIR).join(CONFIG_FILE))
}
```

## Impact
Every Track D memory node's `serve` crash-loops (exit 75, restart counter 114+) and never binds :9077; `federate.sh` blocks forever at "waiting for MESH READY". The whole federation round cannot start. Silent — the api_key IS minted and IS in a config file, just not the one the daemon reads.

## Reproduction
`memory_count>=2` spawn -> federate.sh; serve never comes up; `journalctl -u ai-memory` shows the api_key error while `/etc/ai-memory/xdg/ai-memory/config.toml` carries the key.

## Fix applied this round (validated live)
[...]

PR #2943, opened 2026-08-15
infra(federation-lab): laptop reproducibility kit — one-command mTLS+attested federation lab

alphaonedev:
## What this does

Adds **`infra/federation-lab/`** — a laptop-first reproducibility kit that lets a stranger stand up a hardened, two-node **ai-memory v1.0.0 federation** with one command and get a pass/fail verdict:

```bash
cargo build --release --bin ai-memory --example attest_sign
infra/federation-lab/run.sh
```

`run.sh` mints crypto material, cross-enrolls two federation identities, binds an agent key, seeds a committed 300-row synthetic corpus, boots two `serve` daemons on loopback under mutual TLS with fingerprint-pinned peers and `--quorum-writes 2`, then asserts **18** positives and negatives and exits non-zero if any fails.

No cloud account, no Docker, no network egress, no `sudo`. It writes only inside its own directory — **no `/tmp`, no `$HOME`** — is idempotent (wipes and rebuilds `run/` every invocation), and self-cleans on `EXIT`/`INT`/`TERM`.

### Files

| File | Role |
| --- | --- |
| `infra/federation-lab/run.sh` | single entry point, 9 steps, PASS/FAIL ledger + exit code |
| `infra/federation-lab/README.md` | prereqs, per-step explanation, expected output, honest scope section, troubleshooting |
| `infra/federation-lab/HERO.md` | copy-pasteable release-page hero section (kept in-tree so it cannot drift from the kit) |
| `infra/federation-lab/lib/common.sh` | logging, PASS/FAIL ledger, daemon lifecycle, port probe |
| `infra/federation-lab/lib/posture.sh` | the asi-hard 16/17 knob set + the SSOT drift guard |
| `infra/federation-lab/sample/lab-corpus.json` | committed 300-row **synthetic** corpus (`lab-corpus`) |
| `infra/federation-lab/tools/make-synthetic-corpus.sh` | deterministic generator for that fixture |
| `infra/federation-lab/tools/make-local-slice.sh` | operator-side tool: slice YOUR real corpus into a **gitignored** file |
| `infra/federation-lab/.gitignore` | ignores the disposable `run/` work dir and `sample/local/` |
| `CHANGELOG.md` | `[Unreleased] → Added` entry |

## Extends prior art, does not fork it

Every certificate comes from **`infra/do-hive/crypto/gen-certs.sh`**, which the kit *calls* (`OUT_DIR=run/crypto`). Topology, flag layout, health-poll window and pos/neg shapes follow `test-federation-mtls.sh`, `test-fed-write-sig-attestation.sh` and `test-attestation.sh`. Their comments encode hard-won fixes that survive by reuse rather than rediscovery — #1842, #2293, the **RSA-not-Ed25519** chain requirement (an Ed25519 leaf carries no separate digest OID, so libpq aborts `verify-full` + SCRAM channel binding with `could not find digest for NID UNDEF`), and the `KNOWN-DO-STAGING.md` §2 boot race that the generous poll window exists for.

## Real-run proof

Executed end-to-end on this machine against a binary **built from this worktree** (`cargo build --release --bin ai-memory --example attest_sign`, `ai-memory 1.0.0`, base `5bb800b5`), with no `--bin` override and no `--keep`, so cleanup is part of the proof.

```
$ infra/federation-lab/run.sh ; echo "RUN_EXIT=$?"

══ SUMMARY
   PASS asi-hard posture list matches src/security_profile.rs::KNOBS — ok: lab posture covers all 17 SSOT knobs (16/17 at hard floor, 1 omitted per #2942)
   PASS gen-certs.sh minted the CA + peer/client leaves into run/crypto
   PASS cross-peer federation identities enrolled (ai:lab-node-a ↔ ai:lab-node-b)
   PASS node-a: author pubkey bound AND read back from the _agents registry row (#2941 guard, attempt 1)
   PASS node-b: author pubkey bound AND read back from the _agents registry row (#2941 guard, attempt 1)
   PASS node-a seeded with 300 rows in namespace 'atlas-corpus', all unexpired (from the committed text-only sample atlas-sample.json; file declares 300)
   PASS caveat demonstrated: full 17-knob asi-hard cold boot on a fresh DB exited 75 naming the rollback check (issue #2942) — evidence in run/evidence/caveat-asi-hard-coldboot.txt
   PASS both nodes answer /api/v1/health over mutual TLS with a PINNED client cert
   PASS node-a loaded its private config (tier=keyword) — no embedder, no network
   PASS N1 unpinned client cert refused at node-b's TLS layer (same CA, absent from the allowlist)
   PASS N2 plaintext http refused at node-b's mTLS port
   PASS N3 unsigned write refused 403 ATTESTATION_FAILED under AI_MEMORY_REQUIRE_AGENT_ATTESTATION=1
   PASS N4 asi-hard REFUSED to boot with a loosened pin (exit 1, names AI_MEMORY_SECRET_SCREEN_MODE) — the no-disable contract holds
   PASS P1 attested write accepted at node-a (HTTP 201, id=30df58d5-3a3c-46ba-ae38-1c862703b19c)
   PASS P2 node-a stored it at attest_level=agent_attested (signature verified against the bound key)
   PASS P3 the write REPLICATED to node-b over the mTLS quorum channel and arrived agent_attested
   PASS P4 federated recall: node-b returns the memory that was written to node-a
   PASS P5 corpus recall on node-a returned 5 result(s) from 'atlas-corpus' (lexical — tier=keyword)

   18 PASS / 0 FAIL

   federation lab GREEN — 18 assertions passed.

RUN_EXIT=0
```

Also verified:

- **Self-cleanup** — `run/` is absent after exit (`ls -d run` → `No such file or directory`), no stray `ai-memory` processes remain.
- **Idempotency** — an immediate second invocation returned `18 PASS / 0 FAIL`, exit `0`.
- **Fixture determinism** — `tools/make-synthetic-corpus.sh` reproduces `sample/lab-corpus.json` byte-for-byte across runs (`sha256 09da1a6c2eb372eaed740176a332a4336b827833fda3ac7edd0d831b2417d81e`), so a reviewer can regenerate and `diff` rather than trust the artifact.
- **`--corpus-db` lane** — green end-to-end against a real local corpus (400 rows, derived recall query), with its slice written to a gitignored path and deleted on exit.
- **Repo gates** — `check-doc-symbol-anchors.sh`, `check-hardcoded-literals.sh`, `check-vendor-literals.sh`, `check-docs-vs-ssot.sh`, `check-cloud-init-ascii.sh`, `check-ci-job-claims.sh`, `check-sdk-route-paths.sh` all PASS. No Rust changed, so no SSOT pin (tool count, route count, param census, token budget, schema version) moves.

## The asi-hard 16/17 caveat, stated honestly

The certified posture is `asi-hard`, which pins **seventeen** knobs and refuses to boot if any is set below its hard floor. **The lab runs sixteen and does not set `AI_MEMORY_SECURITY_PROFILE`.**

The seventeenth, `AI_MEMORY_REQUIRE_ROLLBACK_CHECK`, **cannot cold-boot a fresh node**: in require-mode the open-time rollback-evidence check treats an absent off-table head anchor as refuse-to-open, and that anchor is only emitted by the witness watermark cadence over `signed_events`, which is empty on a new database. Fresh DB → no anchor → **exit 75**. Tracked as **#2942** (open).

Because the profile's contract is *pin and refuse*, "asi-hard with rollback-check off" does not exist — that is exactly the case the profile refuses. So the lab sets the sixteen satisfiable knobs to their hard floor directly and leaves the seventeenth at its safe default. This is the same 16/17 shape a persistent hive node runs today.

The kit does not ask anyone to take that on faith:

- **Step 0** re-derives the pinned set from `src/security_profile.rs::KNOBS` and goes **red on drift** (self-tested: injecting a bogus knob into the lab list produces `DRIFT`, rc=1). If the code grows an eighteenth knob, the kit fails rather than demonstrating a posture that no longer exists.
- **Step 5** cold-boots a throwaway node under the *full* 17-knob profile and captures the real exit code and message. If that boot ever succeeds — i.e. #2942 gets fixed — the kit says so **loudly** and flags its own README as stale.
- **Step 7 N4** proves the no-disable contract by loosening a pin under the profile and asserting the refusal.

Both negative lanes require the failure to **name its cause** (`rollback` / `SECRET_SCREEN_MODE`) and pre-check the port, so an unrelated fault — a bind collision, a bad cert path — cannot pass as proof that the posture enforced itself.

Two of the seventeen are *permissive* hatches whose hard floor is "unset" (`AI_MEMORY_ALLOW_SCHEMA_AHEAD` #2445, `AI_MEMORY_FED_ALLOW_PLAINTEXT_PEERS` #2477). The lab actively **unsets** them rather than inheriting whatever the operator's shell exported.

## Known-flake guard (#2941)

`agents bind-key` has been observed to silently no-op on a fresh DB (~1 run in 4 in the reported repro): the registry row is created but `metadata.agent_pubkey` is never set, and every subsequent signed write then `403 ATTESTATION_FAILED` — which *looks* like broken attestation but is broken enrollment. `agents list` does not expose the pubkey, so the standard check cannot see it.

The kit binds, then **reads the key back out of the `_agents` registry row**, retries up to three times, and fails loud naming #2941 with a pointer to the logs. A silent enrollment no-op can no longer masquerade as an attestation defect.

## Findings surfaced while building this

1. **Corpus fixtures silently expire on import.** The sample rows are `tier = mid`, which carries a **write-time** TTL, so a slice whose `created_at` is two weeks old lands 300 rows that are *already expired*. `COUNT(*)` says 300, recall returns nothing (it filters `expires_at IS NULL OR expires_at > now`), and the next gc tick archives them mid-run. This cost a red P5 on the first run and presented as "recall is broken". Fixed by stamping the slice **long-tier** (permanent, no TTL, independent of when it was generated), preserving the original tier in `metadata.sample_source_tier`; step 4 now asserts every seeded row is **unexpired**, not merely present. Also affects export: without clearing `expires_at`, `db::export_all` drops expired rows and a 300-row slice exports **7**.
2. **`XDG_CONFIG_HOME` is inert for the daemon config** — `AppConfig::config_path` reads `$HOME/.config/ai-memory/config.toml` only (the #2852 shape the DO cloud-init already documents). Three sibling scripts in `infra/do-hive/crypto/` set `XDG_CONFIG_HOME="$CFG"` with `CFG="$WORK/xdg"` and write their `tier = "keyword"` override there, so **that override is never loaded** and those legs boot at the compiled default tier. Benign for what they assert, but the comment above it claims otherwise. Not touched in this PR (out of scope); flagging for a follow-up. The lab uses a private `HOME` and *asserts* the daemon logged `loaded config from …` rather than assuming.
3. **A self-inflicted bug worth recording**: an unquoted heredoc in the first draft of `tools/make-atlas-sample.sh` ran a backticked phrase inside a SQL *comment* as a shell command, leaving a stray file in the lab directory. Fixed structurally (quoted heredoc + validated placeholders), and `--namespace` / `--corpus-ns` now pass a charset guard before reaching SQL.

## The committed corpus is synthetic

`sample/lab-corpus.json` is a **300-row synthetic corpus** — deterministically generated, obviously fictional prose about an invented freight cooperative, every row stamped `metadata.synthetic = true`. It is **not** a slice of any real corpus.

This repository is **public**, and a verbatim slice of a real third-party corpus is not committable here: it would redistribute that corpus's text attribution-free under this project's name — the size of the slice and whether the embedding vectors ride along are irrelevant to the rights question — and it would carry along whatever identifiers the source rows happened to hold, such as an internal `metadata.agent_id` or a `version_vector` keyed by the generating machine's hostname. Synthetic fixtures are this repo's house standard for committed data for exactly these reasons (PR #2926 re-minted the golden/conformance vectors from synthetic identities).

The generator is fully deterministic — ids are UUIDv5 over a fixed namespace and the row index, timestamps are a constant, vocabulary is drawn with a seeded PRNG — so a reviewer regenerates and `diff`s rather than trusting the artifact. No hostname, no machine identity, no wall-clock appears in the output.

To run against a **real** corpus, `tools/make-local-slice.sh` + `run.sh --corpus-db` build a slice into a **gitignored** path (`sample/local/`, or `run/` which is deleted on exit), so it stays on your machine and out of git. That lane is exercised and green (see the proof section). Its expiry-normalisation logic — the finding below — is preserved there.

**Branch history was rewritten, not patched.** A follow-up deletion commit would leave the corpus text permanently readable in the published history, so the branch was reset to its base and recommitted as a single commit; `git rev-list origin/release/v1.0.0..HEAD` is one SHA and `git ls-tree -r` over it contains no such file. Honest residual: the previously-pushed objects may remain reachable **by SHA** on GitHub until its GC runs, and the PR's timeline retains references to the superseded SHAs — force-pushing removes them from every ref, which is the most a contributor can do from this side.

## Scope — what this does and does not prove

**Established** (asserted by the kit, red if it stops being true): two v1.0.0 nodes federate over mutual TLS with fingerprint-pinned peers; a same-CA-but-unpinned cert and plaintext are refused; an unsigned write is refused and a signed one is stored `agent_attested`; that write replicates and is recallable at the peer; the no-disable contract refuses a loosened boot; 16/17 asi-hard knobs cold-boot cleanly and the 17th does not.

**Plausible, not measured here**: that the same configuration behaves this way beyond two loopback nodes. The topology matches the production hive's, which is why the extrapolation is reasonable — the kit does not test it.

**Unverified / out of scope**: **performance, throughput, latency or capacity of any kind** (the kit deliberately reports no timings); retrieval relevance or semantic quality — the default corpus is **synthetic**, and embedding vectors are absent from both corpus paths (caveat **F-L8a**), so the recall lane is lexical and proves the pipeline works, not that it works well on your data; Postgres-backed nodes; at-rest encryption (these are not sqlcipher builds, so `doctor --posture enterprise-federation` would not pass and the kit does not claim it).

**Certification scope is unchanged**: v1.0.0 enterprise federation is certified for **500–1,000 agents and ≤50 peers**. Nothing in this kit extends that, and nothing here should be quoted as if it did.

## Data-tier pins (for the optional Postgres leg)

**PostgreSQL 18.6 / Apache AGE 1.8.0 / pgvector 0.8.6.** On AGE, both surfaces stated together because they disagree: `github.com/apache/age/releases` carries **PG18 / v1.8.0-rc0 (2026-07-09)**, the newest released AGE for PG18; Apache tags **every** release `X.Y.Z-rc0` as a release-vote convention, so pgdg reads `1.8.0~rc0-…` while `extversion` reports plain `1.8.0`; and the download page `age.apache.org/download` lags at 1.7.0. The `~rc0` suffix is **not** a Debian packaging revision — that claim is false and was purged from #2940.

## Hero snippet (for the release page)

Lives in [`infra/federation-lab/HERO.md`](infra/federation-lab/HERO.md), copy-pasteable between its markers. Preview:

> ### Run a hardened ai-memory federation on your laptop — one command
>
> ```bash
> git clone https://github.com/alphaonedev/ai-memory-mcp
> cd ai-memory-mcp
> cargo build --release --bin ai-memory --example attest_sign
> infra/federation-lab/run.sh
> ```
>
> Two v1.0.0 nodes federating over mutual TLS on loopback, a real 300-row corpus, and nine proofs: an **agent-attested** write replicates across the quorum mesh and is recallable at the peer, while an unpinned cert, plaintext, an unsigned write, and a loosened `asi-hard` pin are all **refused**.
>
> ```
>    18 PASS / 0 FAIL
>    federation lab GREEN — 18 assertions passed.
> ```

## AI involvement

Authored by Claude Opus 5 under operator delegation. Every claim above is either machine-asserted by `run.sh` on this host or explicitly labelled plausible/unverified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,66 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added (laptop federation reproducibility kit — `infra/federation-lab/`)
+
+- **A stranger can now stand up a hardened two-node ai-memory federation on
+  their own laptop with one command** — `infra/federation-lab/run.sh`. It mints
+  crypto material, brings up two `serve` daemons on loopback under mutual TLS
+  with fingerprint-pinned peers and cross-enrolled federation identities, seeds
+  a committed 300-row corpus sample, and then asserts — 18 PASS/FAIL rows, exit
+  code as the contract — that an agent-attested write on node A replicates
+  across the quorum mesh, arrives `attest_level=agent_attested` at node B, and
+  is recallable there; and that an unpinned client cert, a plaintext client and
+  an unsigned write are all refused. Idempotent (it wipes and rebuilds `run/`
+  every invocation), self-cleaning on `EXIT`/`INT`/`TERM`, and it writes nothing
+  outside its own directory — **no `/tmp`**, no `$HOME`, no network egress, no
+  Docker, no `sudo`.
+- **Extends the DigitalOcean hive crypto suite rather than forking it.** All
+  certificates come from `infra/do-hive/crypto/gen-certs.sh`, which the kit
+  CALLS; the topology, health-poll window and pos/neg shapes follow
+  `test-federation-mtls.sh`, `test-fed-write-sig-attestation.sh` and
+  `test-attestation.sh`. Their comments encode fixes (#1842, #2293, the
+  RSA-not-Ed25519 chain requirement, the `KNOWN-DO-STAGING.md` §2 boot race)
+  that survive by reuse instead of being rediscovered.
+- **Honest about the `asi-hard` 16/17 posture, and proves the gap rather than
+  asserting it.** The lab runs 16 of the 17 pinned knobs at their hard floor and
+  does not set `AI_MEMORY_SECURITY_PROFILE`, because the seventeenth
+  (`AI_MEMORY_REQUIRE_ROLLBACK_CHECK`) cannot cold-boot a fresh node — no
+  off-table head anchor exists yet, exit 75, tracked as
+  [#2942](https://github.com/alphaonedev/ai-memory-mcp/issues/2942). Step 5
+  cold-boots a throwaway node under the full 17-knob profile and captures the
+  real exit code as evidence; step 7 N4 proves the no-disable contract by trying
+  to loosen a pin and asserting the refusal; and step 0 re-derives the pinned set
+  from `src/security_profile.rs::KNOBS` and goes RED if the kit and the code
+  ever disagree.
+- **Guards the known `agents bind-key` flake**
+  ([#2941](https://github.com/alphaonedev/ai-memory-mcp/issues/2941)): the kit
+  binds the author key, then READS IT BACK from the `_agents` registry row
+  (`agents list` does not expose it), retries, and fails loud naming the issue —
+  so a silent enrollment no-op can never masquerade as broken attestation.
+- **The committed corpus is SYNTHETIC** — `sample/lab-corpus.json` (300 rows,
+  namespace `lab-corpus`, every row stamped `metadata.synthetic = true`), built
+  deterministically by `tools/make-synthetic-corpus.sh` from a seeded generator
+  over invented, obviously fictional prose. This repository is public, so a
+  verbatim slice of a real third-party corpus is not committable: it would
+  redistribute that corpus's text attribution-free under this project's name
+  (size and stripped vectors are irrelevant to the rights question) and would
+  carry along whatever identifiers its rows happened to hold — an internal
+  `metadata.agent_id`, a `version_vector` keyed by the generating machine's
+  hostname. Synthetic fixtures are the house standard for committed data here
+  (PR #2926 re-minted the golden/conformance vectors the same way). Ids are
+  UUIDv5 over a fixed namespace, timestamps are a constant, and there is no
+  hostname or wall-clock anywhere in the output, so a reviewer can regenerate
+  and `diff` rather than trust the artifact.
+- To run against a REAL corpus, `tools/make-local-slice.sh` + `run.sh
+  --corpus-db` build a slice into a gitignored path (`sample/local/`, or `run/`
+  which is deleted on exit) — it stays on your machine and out of git. Embedding
+  vectors are stripped there too (caveat F-L8a: a vector is only meaningful
+  against the embedder that produced it, and the lab's nodes run
+  `tier = "keyword"`), so the recall lane is lexical and the kit claims nothing
+  about semantic ranking quality — nor about performance, capacity, or any scope
+  beyond the certified 500–1,000 agents / ≤50 peers.
+
 ### Fixed (consolidation no longer launders derived confidence — min(sources), not hardcoded 1.0; #2935)
 
 - **A consolidated (derived) memory no longer claims maximal `confidence = 1.0`
--- a/infra/federation-lab/.gitignore
+++ b/infra/federation-lab/.gitignore
@@ -1,0 +1,13 @@
+# Everything the lab creates at runtime lives under run/ and is disposable:
+# certs, private keys, node databases, daemon logs, evidence captures.
+# `run.sh` recreates it from scratch on every invocation and removes it on the
+# way out (unless --keep). Nothing here is durable truth.
+run/
+
+# Slices of a REAL local corpus, built by tools/make-local-slice.sh. These are
+# verbatim extracts of whatever corpus you pointed it at, and this repository
+# is PUBLIC — committing one would redistribute that corpus's text under this
+# project's name and carry along whatever identifiers its rows happen to hold.
+# The committed fixture is SYNTHETIC (sample/lab-corpus.json); a real slice
+# stays on your machine. This ignore rule is the mechanical half of that rule.
+sample/local/
--- a/infra/federation-lab/HERO.md
+++ b/infra/federation-lab/HERO.md
@@ -1,0 +1,62 @@
+# Hero section — for the GitHub release page
+
+Copy-pasteable markdown for the v1.0.0 release notes / GitHub release body.
+Kept here so the release page and the kit cannot drift apart: if the lab
+changes, this file changes with it in the same commit.
+
+---
+<!-- ── copy from here ────────────────────────────────────────────────── -->
+
+## Run a hardened ai-memory federation on your laptop — one command
+
+```bash
+git clone https://github.com/alphaonedev/ai-memory-mcp
+cd ai-memory-mcp
+cargo build --release --bin ai-memory --example attest_sign
+infra/federation-lab/run.sh
+```
+
+That stands up **two v1.0.0 nodes federating over mutual TLS** on loopback,
+loads a 300-row synthetic corpus, and then proves the thing works with
+assertions instead of adjectives:
+
+| | |
+| --- | --- |
+| ✅ | Both nodes answer over **mutual TLS with a fingerprint-pinned client cert** |
+| ✅ | An **agent-attested** write is accepted on node A and stored `attest_level=agent_attested` |
+| ✅ | That write **replicates across the quorum mesh** and arrives at node B still `agent_attested` |
+| ✅ | **Federated recall** — node B, which never saw the request, returns the memory |
+| ✅ | Corpus recall returns hits from the seeded namespace |
+| ⛔ | A certificate signed by the **same CA** but absent from the allowlist is **refused** |
+| ⛔ | **Plaintext** HTTP against the mTLS port is **refused** |
+| ⛔ | An **unsigned** write is **refused** `403` under required attestation |
+| ⛔ | `asi-hard` **refuses to boot** when a pinned knob is loosened — the no-disable contract, demonstrated |
+
+```
+   18 PASS / 0 FAIL
+
+   federation lab GREEN — 18 assertions passed.
+```
+
+No cloud account, no Docker, no network egress, no `sudo`. Prereqs are
+`openssl`, `curl`, `jq`, `sqlite3`. The kit writes only inside its own
+directory and cleans up after itself — including on Ctrl-C.
+
+**Honest by construction.** The lab runs **16 of the 17** `asi-hard` posture
+knobs at their hard floor and tells you exactly why the seventeenth is missing:
+`AI_MEMORY_REQUIRE_ROLLBACK_CHECK` cannot cold-boot a fresh node (no off-table
+head anchor exists yet — [#2942](https://github.com/alphaonedev/ai-memory-mcp/issues/2942)),
+so the kit *demonstrates* that limitation with a captured exit code rather than
+papering over it. It also re-derives the pinned-knob set from
+`src/security_profile.rs` on every run and goes red if the kit and the code
+ever disagree.
+
+Equally honest about scope: this is a **functional demonstration**, not a
+benchmark. It reports no timings and measures no capacity. The v1.0.0
+enterprise-federation certification is scoped to **500–1,000 agents and ≤50
+peers**, and nothing here extends that.
+
+📖 Full walkthrough, per-step explanation and troubleshooting:
+[`infra/federation-lab/README.md`](infra/federation-lab/README.md)
+
+<!-- ── copy to here ──────────────────────────────────────────────────── -->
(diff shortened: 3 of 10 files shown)


alphaonedev:
Follow-up filed while building this kit: **#2944** — the `tier = "keyword"` override in `infra/do-hive/crypto/test-attestation.sh` and `test-fed-write-sig-attestation.sh` is written to `XDG_CONFIG_HOME`, which `AppConfig::config_path` ignores (it reads `$HOME/.config/ai-memory/config.toml` only — the #2852 class). Those legs therefore boot at the compiled default tier, not keyword. Deliberately **not** fixed in this PR to keep it scoped to the new kit; `infra/federation-lab/run.sh` uses a private `HOME` and asserts the daemon logged `loaded config from …`, so the same mistake cannot recur here silently.

alphaonedev:
Second commit `4d27d777` — two cleanup-safety fixes from self-review:

1. The `EXIT` trap was armed **before** step 1 took ownership of `run/`, so a *preflight* abort (busy port, missing binary, posture drift) would `rm -rf` a `run/` that a previous `--keep` invocation was preserving — destroying the exact evidence the operator asked to keep. Cleanup is now gated on a `RUN_OWNED` flag set only after this invocation creates the directory.
2. The preflight port probe open-coded the same `/dev/tcp` test `lab_port_free` already provides for the negative lanes; it now calls the helper, so there is one definition rather than two that can drift.

Re-validated end-to-end after the change: **18 PASS / 0 FAIL, exit 0**, `run/` removed on exit. The proof block in the PR description is from this head.

alphaonedev:
**Blocker addressed — the committed corpus is now synthetic, and the branch history was rewritten.**

Head is now a **single** commit `3e59246e` (was `c5d8c83f`); `git rev-list origin/release/v1.0.0..HEAD` returns one SHA and `git ls-tree -r` over it contains no `atlas-sample.json`. A follow-up deletion commit would have left the corpus text permanently readable in published history, so the branch was reset to its base and recommitted, then force-pushed. Honest residual: the previously-pushed objects may stay reachable **by SHA** on GitHub until its GC runs, and this PR's timeline still references the superseded SHAs — force-push removes them from every ref, which is the most I can do from this side.

**What changed**

- **Removed** `sample/atlas-sample.json`. **Added** `sample/lab-corpus.json` — 300 rows, namespace `lab-corpus`, every row stamped `metadata.synthetic = true`, generated by the new `tools/make-synthetic-corpus.sh` from a seeded template over invented, obviously fictional prose (an imaginary freight cooperative). Fully deterministic: UUIDv5 ids over a fixed namespace, constant timestamps, seeded PRNG — `sha256 09da1a6c2eb372eaed740176a332a4336b827833fda3ac7edd0d831b2417d81e`, reproduced byte-for-byte on a fresh run, so you can regenerate and `diff` rather than trust the artifact. **No hostname, no `agent_id`, no `version_vector`, no wall-clock** anywhere in the output — which also closes the identity-leak half of your point 4. (Worth noting the old fixture leaked *two* identities, not one: `metadata.agent_id="perf-opus5"` **and** `metadata.version_vector.entries."host:pop-os"`.)
- `make-atlas-sample.sh` → **`tools/make-local-slice.sh`**, retitled as the operator-side "build your own local slice" tool. Its expiry-normalisation logic is preserved intact. Default output is now `sample/local/`, and both that path and `run/` are gitignored; `run.sh --corpus-db` writes its slice under `run/`, which is deleted on exit. Its header states plainly why the output must not be committed.
- `run.sh` defaults to the synthetic fixture (`--corpus-ns` default `lab-corpus`). Since P5's query can no longer be hardcoded to fixture vocabulary, I added **`--recall-query`**: explicit wins; the fixture gets an authored phrase; a `--corpus-db` run with no query **derives** one from a seeded title. A derived query is a weaker proof — it shows the FTS index, query path and visibility filter work end to end, but not that the corpus answers a question posed independently of it — so the run **labels which kind it used** (`query fixture-authored` / `query derived from the seeded corpus`) rather than presenting them as equivalent.
- README/HERO/CHANGELOG rewritten: the committed fixture is synthetic, the real-corpus lane is local-only via `--corpus-db`, and F-L8a is re-framed (neither path carries vectors — the synthetic corpus never had any).
- **Nit fixed**: `lib/posture.sh` now says `run.sh --no-caveat-probe`, matching the real flag.

**Re-run proof** (binary built from this worktree, no `--bin`, no `--keep`):

```
   PASS node-a seeded with 300 rows in namespace 'lab-corpus', all unexpired (from the committed SYNTHETIC fixture lab-corpus.json; file declares 300)
   PASS P5 corpus recall on node-a returned 5 result(s) from 'lab-corpus' (lexical — tier=keyword, query fixture-authored)

   18 PASS / 0 FAIL

   federation lab GREEN — 18 assertions passed.
RUN_EXIT=0
```

Also verified: `run/` removed on exit; an immediate second invocation identical (idempotency); and the **`--corpus-db` lane green end-to-end** against a real local corpus (400 rows, `query derived from the seeded corpus`) with its slice landing in a gitignored path and deleted on exit. All seven shell gates still PASS locally.

Not merging — over to you for the final pass.

Issue #3089, opened 2026-08-19
🎯 v1.0.0 GA burn-down orchestrator — enterprise-federation → F500 touchdown (2026-08-19 hive triage)


alphaonedev:
# 🎯 v1.0.0 GA burn-down orchestrator — enterprise-federation → F500 touchdown

**Source:** 2026-08-19 full-spectrum hive triage of all 228 open issues (15 evidence-required assessment agents + synthesis). Companion epics: #1940 (global dev), #2705 (cert), #2440 (GA-review). Master remediation tracker for the pg-audit sub-wave: #3076.

> **Tag-cut stays operator-gated.** This issue is the live orchestrator: burn the waves top-to-bottom, tick each row, close on merge (`Closes #N` does NOT auto-fire on `release/v1.0.0`).

## Brass-tacks verdict
[...]

Issue #2424, opened 2026-07-24
[GA-BLOCKER] postgres v83→v84 upgrade is BRICKED — bootstrap schema creates idx_memories_embedding_space before migrate_v84 adds the column (#1861 lesson recurring)


alphaonedev:
## Severity: **v1.0.0 GA upgrade-path blocker.** Any existing postgres deployment on schema < v84 cannot start v1.0.0. It fails at `connect`.

Found live (not theorised) while standing up a postgres+AGE test database during the #2393/#2397 lane. Observed failure:

```
connect → init schema: column "embedding_space" does not exist
```

## Root cause

Two things create the same index, and the wrong one runs first.

1. **Ladder-owned (CORRECT)** — `migrations/postgres/0041_v84_embedding_space.sql:16` creates `idx_memories_embedding_space` inside the v84 arm, after the `ADD COLUMN`. This is right, and is exactly the #1861 "ladder-owned index" discipline.
2. **Bootstrap-inline (THE BUG)** — `src/store/postgres_schema.sql:284-285` creates the same index at file scope:
   ```sql
   CREATE INDEX IF NOT EXISTS idx_memories_embedding_space
       ON memories (embedding_space) WHERE embedding_space IS NOT NULL;
   ```

The bootstrap schema runs **before** the migration ladder. On a fresh database this is harmless: `CREATE TABLE memories (...)` at `postgres_schema.sql:231` declares `embedding_space TEXT` inline, so the index finds its column.

On an **existing pre-v84 database** the ordering is fatal:
- `CREATE TABLE IF NOT EXISTS memories` → **no-op** (the table exists, without `embedding_space`)
- `CREATE INDEX IF NOT EXISTS idx_memories_embedding_space ON memories (embedding_space)` → **ERROR: column "embedding_space" does not exist**
- `migrate_v84`, which would have added the column, is never reached.

`IF NOT EXISTS` does not save this: the *index* does not exist, so postgres proceeds to resolve the column, and fails.

## Why this is a v1.0.0 blocker specifically

v84 is recent (#2167). Deployments in the field are overwhelmingly past v74 but **not** past v84 — so this defect is dormant for fresh installs and universal for upgrades, which is the worst possible distribution for a GA cut. It will not appear in any CI that provisions a fresh database.

## The class is broader — this is not a one-column fix

[...]

Issue #1381, opened 2026-05-28
Postgres test-isolation: 4 tests assume empty schema, fail on shared lan-parity container with accumulated state


alphaonedev:
## Finding

Full postgres+AGE regression on 2026-05-28 against rebuilt lan-parity stack (\`infra/lan-parity-test/\`, HEAD \`be3347d70\`): **8028 passed | 9 failed**. After isolating cargo-target races (5 transients confirmed PASS on clean rerun), **4 deterministic failures remain**, all caused by tests assuming a known/empty postgres schema state but finding accumulated state from prior tests in the shared container:

### Failure 1: \`auto_migrate_converts_384_schema_to_768_on_daemon_bootstrap\`

\`\`\`
thread 'auto_migrate_converts_384_schema_to_768_on_daemon_bootstrap' panicked at tests/embedding_dim_migration.rs:139:5:
assertion \`left == right\` failed: step-1: fresh bootstrap must land vector(384)
  left: Some(768)
 right: Some(384)
\`\`\`

Test expects \`public\` schema starts at vector(384); shared container already has vector(768) from prior dim-migration tests.

### Failure 2 + 3: \`issue_1213_unscoped_probe_demonstrates_root_cause\` + \`issue_1213_atttypmod_probe_scopes_to_public_schema\`

\`\`\`
assertion \`left == right\` failed: test stage requires BOTH public.memories AND ai_memory_issue_1213.memories — got [("ai_memory_issue_1213", 768), ("ai_memory_issue_1213_demo", 768), ("ic_alice", 768), ("ic_bob", 768), ("public", 384)]
  left: 5
 right: 2
\`\`\`

Tests expect 2 schemas (\`public\` + \`ai_memory_issue_1213\`); shared container has 5 schemas including \`ic_alice\` + \`ic_bob\` (the lan-parity IronClaw daemon namespaces) and \`ai_memory_issue_1213_demo\` (a sibling test's residual).

### Failure 4: \`migrate_links_sqlite_to_postgres_to_sqlite_roundtrip\`

\`\`\`
thread 'migrate_links_sqlite_to_postgres_to_sqlite_roundtrip' panicked at tests/migrate_links_roundtrip.rs:278:52:
connect pg: BackendUnavailable { backend: "postgres", detail: "init schema: error returned from database: column \"title\" does not exist" }
\`\`\`

\`public.memories\` table left in a partially-migrated state by prior test; schema-init reuses it instead of fresh-creating.

[...]

PR #3103, opened 2026-08-20
fix(pg): land app tables in public, not ag_catalog, + safe in-place migration (#3055)

### LINKED ISSUE #3055: postgres+AGE: ai-memory app tables land in ag_catalog instead of public (fragile; cascade-drop/pg_dump risk)
## Summary
On PostgreSQL + Apache AGE deployments, ALL ai-memory relational tables (memories, memory_links, schema_version, signed_events, agent_lineage, recall_observations, … 39 tables) are created in the **`ag_catalog`** schema (AGE's extension catalog) instead of **`public`**. App tables squatting in the extension catalog are fragile: a `DROP EXTENSION age` / AGE upgrade can cascade-drop them, and `pg_dump` extension-scoping / backup tooling misbehave.

## Evidence
- `src/store/postgres_schema.sql:56` uses UNQUALIFIED `CREATE TABLE IF NOT EXISTS memories (...)` with no `SET search_path=public` and no `public.` qualification, so INIT_SCHEMA inherits the connection/db search_path.
- The postgres+AGE deploy substrate sets a db-level `ALTER DATABASE … SET search_path = ag_catalog, "$user", public` (ag_catalog FIRST + writable to the db owner), so every unqualified CREATE lands in ag_catalog.
- Confirmed GENERAL: the standard test substrate pg-main (:5433) has the identical placement (`ag_catalog.memories`), so the whole ai-memory postgres suite has always run against tables-in-ag_catalog. Functionally works (adapter reads via search_path; #2397 sets tx-scoped ag_catalog-first on the AGE unprojection path) — but it is an enterprise-correctness/durability defect.

## Proposed fix (candidates — a placement decision is a T4 back-compat crossroads → 5-agent vote before implementing)
- F-B (code hardening, GENERAL): `PostgresStore::connect` / `postgres_schema.sql` must DETERMINISTICALLY place the relational schema in `public` regardless of connection search_path — either `SET search_path = public, ag_catalog` at the top of the bootstrap, or schema-qualify `public.<table>`. This protects EVERY AGE deployment.
- F-A (deploy): certified initdb should not put ag_catalog first at the DB level.
- Migration for existing deployments (tables already in ag_catalog): `ALTER TABLE … SET SCHEMA public` under an idempotent guard, or a documented reprovision. MUST be adjudicated by the vote (existing dbs have data in ag_catalog).

Discovered during the enterprise-federation pg-tier standup (2026-08-18). Related ai-memory memory: 817bdf13.

alphaonedev:
Closes #3055

## The defect
On postgres+AGE, the recommended AGE setup puts `ag_catalog` FIRST on the search_path, so an unqualified `CREATE TABLE memories ...` landed the ai-memory application tables in `ag_catalog` — entangling durable app data with the AGE extension catalog. A `DROP EXTENSION age CASCADE` would then take the app tables with it (silent data loss), and placement was non-deterministic. Ratified approach: 5-agent vote #3089 cluster `C-fed-integrity-representation`, per_issue #3055 — SHIP BOTH HALVES.

## Half F-B — deterministic placement (ships unconditionally)
Every `CREATE` in `src/store/postgres_schema.sql` is now `public.`-qualified (tables, indexes, `REFERENCES`, views, functions), plus the three ladder-only creates that are not in the bootstrap (`transcript_line_dedup` in `postgres.rs`, `offloaded_blobs`, `memory_revisions`). A schema-qualified `CREATE` cannot be defeated by any search_path/GUC — stronger than any `SET search_path`. `ag_catalog` stays on the runtime search_path for AGE graph ops; only the CREATE targets changed. `CREATE EXTENSION vector` and function bodies are deliberately untouched.

## Half migration — in-place relocation (repairs the installed base)
A new pre-bootstrap, advisory-locked, idempotent self-heal `relocate_app_tables_to_public` moves the **37 app tables (+ 6 owned `BIGSERIAL` sequences)** from `ag_catalog` to `public` via `ALTER TABLE ... SET SCHEMA public` over an **EXPLICIT HARD-CODED allowlist** (`src/store/postgres_relocate_ag_catalog_3055.sql`) — never a schema wildcard, never an AGE-owned object (`ag_graph`, `ag_label`, `agtype`, `_ag_label_*`).

- Each move guarded by `to_regclass('ag_catalog.<t>') IS NOT NULL AND to_regclass('public.<t>') IS NULL` — this IS the crash-resume / idempotency mechanism (a re-run skips already-moved tables).
- The whole allowlist moves in ONE transaction (same-tx atomicity is load-bearing); a crash rolls back cleanly, and the move is reversible (`ALTER TABLE public.<t> SET SCHEMA ag_catalog`).
- `ALTER TABLE ... SET SCHEMA` is non-destructive and carries indexes/constraints/owned sequences with the table.
- The fully-qualified `ag_catalog.<t>` ALTERs and the `pg_namespace`-based probe are search_path-independent, so relocation is robust regardless of the connection's search_path.
- **pg_dump safety step**: before taking the transaction, a loud operator WARN fires naming the exact capture command — `pg_dump --format=custom --file=ai-memory-pre-3055.dump <STORE_URL>` — and only when the probe finds app tables actually mis-placed in `ag_catalog` (silent on fresh / already-healed installs).

### Ordering — why PRE-bootstrap, not a post-bootstrap numbered ladder arm
The vote described the relocation as a numbered `migrate_locked` arm. That is **internally inconsistent with F-B**: the qualified bootstrap (`INIT_SCHEMA`) runs BEFORE `migrate_locked`, so a post-bootstrap move would let `CREATE TABLE IF NOT EXISTS public.memories` create an EMPTY shadow over data still in `ag_catalog.memories`, and the relocation's `public.<t> IS NULL` guard would then SKIP the move — orphaning the real rows (silent data loss). The relocation therefore runs pre-bootstrap, under the same bootstrap `MIGRATION_ADVISORY_LOCK_KEY` held across `connect()`, as an idempotent self-heal (the #2578 precedent). Placement is **not** a schema-SHAPE change, so `CURRENT_SCHEMA_VERSION` is unchanged and both backends stay in lockstep at **schema=89** (the migration-ladder-uniqueness gate stays green). This is the one deviation from the literal vote text, forced by the north-star (never orphan durable data); flagged here for review.

## The full 37-table allowlist (confirm zero AGE objects)
`action_edges, actions, agent_api_keys, agent_lineage, agent_quotas, agent_subkey_certs, archived_memories, archived_memory_links, audit_log, checkpoints, confidence_shadow_observations, entity_aliases, federation_push_dlq, forget_tombstones, kg_projection_outbox, leases, memories, memory_links, memory_revisions, memory_transcript_links, memory_transcripts, model_attestations, namespace_meta, offloaded_blobs, pending_actions, recall_observations, routine_runs, routines, schema_version, signals, signed_events, signed_events_dlq, subscription_dlq, subscription_events, subscriptions, sync_state, transcript_line_dedup`

Sequences (6, owned; move with their table under SET SCHEMA — guarded as defense-in-depth): `confidence_shadow_observations_id_seq, signed_events_dlq_dlq_id_seq, federation_push_dlq_id_seq, kg_projection_outbox_id_seq, subscription_events_id_seq, subscription_dlq_id_seq`.

The vote's "39" was an estimate; the exact enumeration from the schema (bootstrap + migration-file + inline `postgres.rs` creates) is **37 app tables**. `federation_nonce_cache`, `governance_rules`, `skills`, `skill_resources` are **sqlite-only** (postgres ships explicit no-op DDL for v51/v66 etc.) and are correctly EXCLUDED — zero AGE catalog objects are included.

## R-203 evidence
- **`tests/pg_age_schema_placement_3055.rs`** (sal-postgres, `#[ignore]` + `AI_MEMORY_TEST_POSTGRES_URL`/`AI_MEMORY_TEST_AGE_URL`, Postgres CI cell):
  - `bootstrap_lands_app_tables_in_public_3055` — fresh connect lands `memories`/`schema_version`/`memory_links`/`namespace_meta` in `public` (placement test).
  - `relocation_moves_ag_catalog_table_to_public_3055` — seeds an allowlisted table in `ag_catalog` with data, reconnects → moved to `public` with row count preserved, no `ag_catalog` shadow; a second connect is a no-op (crash-resume).
  - `drop_extension_age_cascade_spares_public_app_data_3055` — **removal proof**: `DROP EXTENSION age CASCADE` SPARES `public.memories` (+ a public probe) while TAKING an `ag_catalog`-resident negative control. Destructive-then-restoring (recreates the extension + `memory_graph`).
- Existing gates updated to the qualified shape: `pg_projection_column_fidelity_2585` (parses `public.memories`), `migration_ladder_integrity` (`strip_ident` reduces `schema.table` so the #2424 forward-reference guard keeps firing instead of going vacuous — all 8 pass), and the `postgres.rs` `init_schema_*` bootstrap-shape unit-test needles now assert `public.<t>`.

## CI status (local, all green)
- `cargo fmt --all --check` clean; `cargo check --all-targets` (default + `sal-postgres`) + `cargo check --examples` clean.
- `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` clean on all four legs: default, `sal`, `sal-postgres`, `vectorlite`.
- `scripts/check-migration-ladder.sh`, `scripts/check-hardcoded-literals.sh`, `scripts/check-docs-vs-ssot.sh` (schema=89 unchanged) PASS.
- `qual_10_module_size_ceiling` (postgres.rs ceiling bumped 34_060→34_200 with dated #3055 comment), `qual_6_7_legacy_error_type_ceiling`, `migration_ladder_integrity` (8/8), `pg_projection_column_fidelity_2585` (5/5) PASS.
- The three `pg_age_schema_placement_3055` tests are `#[ignore]`+sal-postgres and run in the Postgres CI cell (not runnable in this offline environment); the DROP-EXTENSION test is destructive-then-restoring and best run serially in that cell.


alphaonedev:
## Merge-blocker fix pushed (bc507dcf)

Adversarial review found a silent data-loss regression: F-B qualified the bootstrap + 3 non-bootstrap creates but left ~27 **bare** `CREATE TABLE IF NOT EXISTS <t>` ladder arms. Confirmed empirically (PG16 + the live enterprise tier): `CREATE TABLE IF NOT EXISTS <bare>` checks the **creation namespace** (`ag_catalog`, first on the AGE search_path), NOT search_path visibility — so a bare ladder-arm create makes an `ag_catalog` **shadow even when a `public` twin already exists**, and the pre-bootstrap relocation can never heal it (its `public.<t> IS NULL` guard is then false). On a v0.8/v0.9 → v1.0.0 upgrade the bare `migrate_v70/76/78/79/83` arms would shadow the identity/crypto tables into `ag_catalog` → data loss on `DROP EXTENSION age CASCADE`. Worse than pre-PR.

**Fix:** `public.`-qualified every executable `CREATE TABLE` / `CREATE [UNIQUE] INDEX ... ON <t>` / `REFERENCES <t>` for an app table across **all** inline `migrate_vN` arms in `src/store/postgres.rs` and **24** `migrations/postgres/*.sql`. Index names stay bare (qualified index name = syntax error); only table/ON/REFERENCES targets change.

**New regression guards** (`tests/pg_age_schema_placement_3055.rs`, sal-postgres):
- `bare_create_shadows_but_qualified_create_does_not_3055` — pins the empirical PG semantics (bare shadows, qualified doesn't).
- `no_ag_catalog_shadows_after_connect_3055` — after a full connect (bootstrap + relocation + whole ladder), ZERO of the 37 app tables remain in `ag_catalog`; fails closed if any arm ever re-introduces a bare create.
- `relocation_guard_moves_and_is_idempotent_3055` — rewritten to RANDOMIZED throwaway tables (never fights a shared hive).
- `drop_extension_age_cascade_*` — now gated behind `AI_MEMORY_TEST_ALLOW_DROP_EXTENSION=1` so it can't nuke a shared hive.

**Verification:** all 5 tests GREEN against the live enterprise pg18.6+age+vector tier (mTLS verify-full, `--include-ignored --test-threads=1`). Local: fmt clean; clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` clean on all four legs; migration-ladder / hardcoded-literal / docs-vs-SSOT (schema=89) / create-extension gates PASS; `migration_ladder_integrity` 8/8, `pg_projection_column_fidelity_2585` 5/5, `qual_10`/`qual_6_7` PASS.

alphaonedev:
## Approach pivot pushed (15ca64c6, squashed) — search_path demotion, not `public.` qualification

Second review found that hard-coding `public.` on the DDL **broke the per-test-SCHEMA isolation harness** (#1381, `tests/common/postgres_env.rs`): it connects with `?options=-c search_path=<test_schema>,public` and relies on unqualified app CREATE/DML honouring that search_path, so forcing `public.` lost isolation (`perf_1579_postgres_backfill_tsv` went 2/13).

**Reconciled at the search_path layer instead.** Empirically confirmed on the live AGE 1.8.0 tier: the `ag_catalog`-first order comes from the **database default** `search_path`, and `LOAD 'age'` does **not** re-prepend it.

- **Reverted ALL `public.` qualification** (bootstrap schema, every ladder arm, all 24 migration files, the pg_projection / migration_ladder_integrity test edits) — net diff no longer touches those files at all.
- **`after_connect` demotion** (`demote_leading_ag_catalog`): when `ag_catalog` LEADS the session search_path, move it to the END. Unqualified creates then land in `public` in prod (like a non-AGE postgres) and in the per-test schema in the harness (whose path carries no `ag_catalog`, so it's left untouched). `ag_catalog` stays on the path for type resolution; AGE ops still `SET LOCAL` it per-tx.
- **Pre-bootstrap relocation** for the installed base (unchanged in spirit): guarded, advisory-locked, allowlisted `ALTER TABLE ... SET SCHEMA public`, crash-resume, reversible, `pg_dump` WARN.
- `CURRENT_SCHEMA_VERSION` unchanged (schema=89).

**Verified GREEN on the live enterprise pg18.6+age1.8.0+vector tier** (mTLS verify-full, `--include-ignored --test-threads=1`):
- `pg_age_schema_placement_3055` 5/5 (demotion proof + forced-`ag_catalog`-first negative control, no-shadows-after-connect, relocation move/idempotent/guard-skip, DROP-EXTENSION survival gated).
- **All 8 per-test-schema harness files**: `perf_1579_postgres_backfill_tsv` 13/0, `recall_purity_p01_postgres` 11/0, `issue_1213_atttypmod_age_schema_scope` 10/0, `decorrelation_probe_postgres_parity` 11/0, `embedding_dim_migration` 16/0, `migrate_links_roundtrip` 12/0, `reflect_pg_caller_identity_2857` 9/0.
- `demote_ag_catalog_tests` unit 3/3.

Local: fmt clean; clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` clean on all four legs; migration-ladder / hardcoded-literal / docs-vs-SSOT (schema=89) / create-extension gates PASS; `qual_10` bumped 34_200→34_280 (dated comment).

alphaonedev:
## Removal-proof hardened (amended into e16b053d)

While verifying the destructive `DROP EXTENSION age CASCADE` proof on the live tier I found two issues and fixed both:

1. **On AGE 1.8.0, `DROP EXTENSION age CASCADE` LEAVES the `ag_catalog` schema in place**, so a plain ag_catalog-resident table is NOT taken by the cascade — my old "ag_catalog victim is dropped" negative control was version-dependent and flaked. Removed it. The acute full-teardown vector is `DROP SCHEMA ag_catalog CASCADE`, which the fix defeats the same way (app tables never live in `ag_catalog` — covered by `no_ag_catalog_shadows_after_connect_3055`). The removal proof now asserts the deterministic, load-bearing fact: `public.memories` (no dependency on the `age` extension) SURVIVES `DROP EXTENSION age CASCADE`.
2. **The test now restores the extension + `memory_graph` BEFORE any assertion** (capture-then-restore-then-assert), so a failed assertion can never leave a shared database without `age`. (An earlier run had left AGE dropped on the tier; I restored it immediately and re-verified — `age` 1.8.0 + `memory_graph` healthy.)

Full `pg_age_schema_placement_3055` suite (incl. the destructive test with `AI_MEMORY_TEST_ALLOW_DROP_EXTENSION=1`) is **5/5 GREEN** on the live tier, and `age`/`memory_graph` are confirmed healthy after the run. fmt + clippy (all four legs, `--all-targets`) clean.

alphaonedev:
## `$user`-precedence blocker fixed + rebased onto release tip (5c6845b9)

Rebased onto the current release tip (#3101 merged; resolved the postgres.rs `qual_10` ceiling conflict by keeping the higher value **34_500** with both dated notes).

**BLOCKER (data-integrity):** the old `demote_leading_ag_catalog` turned `ag_catalog, "$user", public` into `"$user", public, ag_catalog` — leaving the ROLE schema `$user` AHEAD of `public`. Reproduced live: with `CREATE SCHEMA ai_memory AUTHORIZATION ai_memory` present, an unqualified `CREATE TABLE memories` landed in `ai_memory`, not `public` → on a CVE-2018-1058-hardened per-role-schema deploy all 37 app tables bootstrap into `$user` → split-brain.

**Fix — `normalize_app_search_path`:** reorder so the first REAL app schema wins the unqualified-CREATE target — drop BOTH `ag_catalog` AND `"$user"` from the front, keep remaining real schemas in order, re-append one `ag_catalog` at the END. So:
- `ag_catalog, "$user", public` → `public, ag_catalog`
- `"$user", ag_catalog, public` → `public, ag_catalog` (closes the minor non-leading-`ag_catalog` gap too)
- `<test_schema>, public` → **unchanged** (no `$user`/`ag_catalog` → harness isolation intact)

Never hard-codes `public`. `ag_catalog` kept last for AGE type resolution; AGE ops still `SET LOCAL` it per-tx.

**`$user` probe coverage:** the relocation probe stays `ag_catalog`-only — justified in the code + PR: no released binary ever landed tables in `$user` (every shipped binary resolves the AGE default with `ag_catalog` FIRST → lands in `ag_catalog`), and the normalization now structurally prevents any future `$user` landing. No `$user` installed base to heal.

**Verified GREEN on the live tier** (mTLS verify-full, `--include-ignored --test-threads=1`):
- pg_age **6/6** incl. the NEW `creates_land_in_public_not_user_schema_3055` (creates the role-named schema, asserts `memories`/unqualified-create land in `public` not `$user`, drops the schema — capture-then-cleanup-then-assert so the tier stays clean) + the destructive DROP-EXTENSION proof.
- All harness files: `perf_1579_postgres_backfill_tsv` **13/0**, `recall_purity_p01_postgres` 11/0, `issue_1213_atttypmod_age_schema_scope` 10/0, `decorrelation_probe_postgres_parity` 11/0, `embedding_dim_migration` 16/0, `migrate_links_roundtrip` 12/0, `reflect_pg_caller_identity_2857` 9/0.
- `normalize_app_search_path_tests` unit **3/3**.
- **Tier left clean**: 0 leftover throwaway tables, no role-named schema, 0 app tables in `ag_catalog`, `memories` in `public`, AGE 1.8.0 + `memory_graph` healthy.

Local: fmt clean; clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` clean on all four legs; migration-ladder / hardcoded-literal / docs-vs-SSOT / create-extension gates PASS; `qual_10` 34_500, `qual_6_7` pass.

Issue #2709, opened 2026-08-04
[CB-4][security][namespace-standard] postgres HTTP set_standard ownership gate skipped for non-visible rows — foreign private memory installable as governance standard (#2705 B-6)


alphaonedev:
**Fable review finding.** Tier-1 security, epic #2705 (B-6 core).

\`src/handlers/hook_subscribers.rs:451\`: \`if let Ok(resolved_mem) = app.store.get(&ctx, &standard_id).await { ... authorize_namespace_standard_bind(...) }\`. \`PostgresStore::get\` folds a visibility denial into \`Err(StoreError::NotFound)\` (\`src/store/postgres.rs:17455-17476\`); \`ctx = http_caller_ctx(h, None)\` has \`bypass_visibility=false\`; \`is_visible_to_caller\` treats absent \`metadata.scope\` as private. So whenever \`body.id\` names another agent's non-shared memory, \`get\` returns Err, the \`if let Ok\` arm is skipped, **no authorization runs**, and \`PostgresStore::set_namespace_standard\` (\`:19897-19936\`) binds it via unfiltered \`SQL_SELECT_MEMORY_ID_BY_ID\`.

**Failure scenario:** postgres daemon. \`POST /api/v1/namespaces/alice-ns/standard {"id":"<alice private memory id>"}\` with \`X-Agent-Id: bob\` → 201; alice's private memory is now alice-ns's governance standard, replacing alice's policy (typically → allow-on-silence). Integrity/authz defect (body still withheld by #2537/#2707, so not disclosure). The sqlite twin at \`:662\` uses storage-level \`db::get\` and is correct → **parity break in the security direction**. #2706 CHANGELOG "always enforces ownership" is false on this leg.

**Fix:** \`match\` on \`get\`'s error rather than \`if let Ok\`, and refuse on \`NotFound\` for a caller-supplied \`body.id\`.

hard-coder tier.

alphaonedev:
Fixed on release/v1.0.0 by PR #2735 (CB-4). The postgres HTTP set_standard ownership probe now runs under a bypass-visibility context so a foreign-owned scope=private row is resolved and authorize_namespace_standard_bind runs against the real request principal — Bob binding Alice's private memory as alice-ns's governance standard now 403s. sqlite parity restored; read-path withholding (#2537/#2707) untouched. Fable 5 review + security audit: APPROVE. Closed manually (non-default branch).

Issue #2479, opened 2026-07-29
[SECURITY][CWE-284] federated namespace_meta[] can re-parent a foreign namespace and rewrite its governance policy


alphaonedev:
Found while closing #2447 (3x3 adversarial red-team lens). **Not fixed there.** Verified at `release/v1.0.0`.

## The hole

The `/sync/push` `namespace_meta[]` and `namespace_meta_clears[]` loops (`src/handlers/federation_receive.rs`) validate only `validate_namespace` before calling `db::set_namespace_standard(conn, ns, standard_id, parent.as_deref())` / `db::clear_namespace_standard`. Neither consults the peer's `allowed_namespaces`.

The `parent` argument is the sharp edge: `set_namespace_standard` writes the explicit parent that `build_namespace_chain` walks, and that chain is what `resolve_governance_policy` consumes. So an enrolled peer can **re-parent `secure/ops` under a namespace it controls** and thereby rewrite that namespace's effective governance policy — approval level, reflection-depth cap, delete gating.

That is an authz-CONFIG takeover, not just a data write: it changes the rules every subsequent local write to that namespace is evaluated against. `namespace_meta_clears[]` is the same reach in the destructive direction (wipe a namespace's standard + parent).

## Why it is not folded into #2447

#2447 confines the lanes that write/destroy MEMORY ROWS by id (`memories[]`, `deletions[]`, `archives[]`, `restores[]`). Namespace-standard rows are a different object with different semantics — a peer legitimately propagating a standard for a namespace it DOES own is a supported flow, so the fix is a scope check on the affected namespace(s), including the `parent`, not a blanket refusal.

## Suggested fix

Gate both loops on the peer's `allowed_namespaces` for the target namespace AND the declared `parent` (a peer must not be able to attach a namespace it owns as the parent of one it does not, nor vice versa), reusing the `federation::receive_auth` verdict added by #2447. Both backends — note the postgres twin reports these subcollections as `unsupported_on_postgres` today, so the sqlite funnel is the live surface.

alphaonedev:
## Independent corroboration — two separate audits reached this finding without seeing each other

A Fable 5 audit of merge `458a8b43` (#2447) surfaced this same hole independently, rating it **HIGH** and the sharpest of the residual federation gaps. Agreeing evidence at `4f28f1d8`:

[...]

PR #3104, opened 2026-08-21
fix(cert): authorize namespace-standard parent graft + exclude inferred ancestors from governance layering (#2542)

### LINKED ISSUE #2542: namespace-standard chain grafting: caller-supplied `parent` and `-`-prefix auto_detect_parent let one namespace pull another's standards into its own inheritance chain
Found by the #2537 5-agent adversarial vote (`4d3ea1c5`, precedent + attacker lenses) while evaluating — and rejecting — a namespace-chain-membership admit arm. **Not fixed in #2540.** Verified at `release/v1.0.0`.

`build_namespace_chain` (`src/storage/mod.rs`) resolves a namespace's ancestor path from `namespace_meta.parent_namespace`, walking up to `MAX_EXPLICIT_DEPTH = 8`. Two independent routes let that chain be constructed by someone other than the namespaces involved:

1. **Caller-supplied `parent`.** `handle_namespace_set_standard` reads `params["parent"]` (`src/mcp/tools/namespace.rs:136`) and passes it to `set_namespace_standard` (`src/storage/mod.rs`). It is validated only as a well-formed namespace string. The #929 ownership gate checks ownership of *the standard memory being bound* — **never** of the declared parent. So Mallory can set a standard on her own namespace with `parent: "victim-ns"` and the victim namespace becomes a link in her resolved chain.

2. **`-`-prefix auto-detection.** `auto_detect_parent` truncates at each `-` and adopts the first ancestor that has a standard, so `acme-corp-attacker` inherits `acme-corp` into its chain **by naming coincidence**, with no action by `acme-corp`'s owner. Conversely, whoever sets a standard on the short prefix silently becomes governance ancestor of every later `acme-*` namespace.

## Two directions of impact

- **Read direction.** Any read that resolves standards through the chain now includes the grafted namespace's entries. #2540's visibility filter contains the *disclosure* consequence (a grafted standard the caller cannot see is withheld and counted), which is exactly why #2540 chose the canonical predicate over a chain-membership carve-out — a carve-out would have been defeated by this. But the grafting itself is untouched.
- **Integrity direction, NOT covered by #2540.** `resolve_governance_policy` layers `metadata.governance` along the chain. Combined with route 1 (or with the ungated bind in the sibling issue), an attacker can install their standard as an **ancestor of a victim namespace**, so their `write` / `promote` / `delete` / `approver` policy layers over the victim's writes. That is a governance-poisoning primitive, and no read-path filter addresses it.

## Proposed fix (needs its own vote — the two routes have different answers)

- **Route 1:** require the setter to be entitled to the declared `parent` — minimally, that the parent namespace either has no standard or is owned by the same principal. Refuse loudly rather than silently accepting an arbitrary parent.
- **Route 2:** `auto_detect_parent` should not be an implicit cross-namespace authority grant. Options: drop `-`-prefix inference entirely (explicit `parent` only), confine it to namespaces sharing an owner, or keep the inference for *lookup* while refusing to let an inferred ancestor contribute a governance policy. Needs a compatibility assessment — the `ai-memory-tests` → `ai-memory` → `ai` walk is documented behaviour.

Files: `src/mcp/tools/namespace.rs:136` (caller-supplied `parent`), `src/storage/mod.rs` (`set_namespace_standard`, `auto_detect_parent`, `build_namespace_chain`), `src/handlers/hook_subscribers.rs` (HTTP twin).

Related: #2537 / #2540, #1921 (positional/namespace-shaped visibility grants, the same failure mode), #1569 (allow-on-silence governance defaults this grafting subverts).

alphaonedev:
Closes #2542

Enterprise-federation cert-core (tenant-isolation / governance). Closes a tenant-isolation + approval-bypass primitive on the namespace-standard governance chain. Approach is the 5-agent-vote-ratified two-route fix; shipped in-tree patterns mirrored, nothing invented. Both backends (sqlite + postgres) enforce identically.

## The defect
`build_namespace_chain` walks `namespace_meta.parent_namespace`, and `resolve_governance_policy` layers `metadata.governance` (write/promote/delete/**approver**) along that chain. Two routes let that chain be constructed by someone other than the namespaces involved:
1. **Caller-supplied `parent`** — `set_standard { parent: victim-ns }` stored the victim as a governance ancestor with NO check on the parent (the #929 gate only checked the *bound* standard memory).
2. **`-`-prefix `auto_detect_parent`** — `acme-corp-attacker` silently adopts `acme-corp` as its governance ancestor by naming coincidence.

## Route 1 — declared-parent entitlement (the bind)
Extended the shipped pure gate `mcp::authorize_namespace_standard_bind` to also entitle the declared `parent`.

**Authorization rule (fail-closed):** for a non-daemon caller, the bind is refused unless the declared parent namespace is **UNOWNED** (no standard bound / severed) **or** its bound standard's `metadata.agent_id` equals the caller. A parent standard owned by a different principal → refuse loudly (`caller does not own this declared parent namespace standard (...)`).

Wired through **every** local `set_standard` funnel:
- **MCP** `handle_namespace_set_standard_inner` — resolves the parent's bound standard; the declared parent is authorized even on the first-write / absent-bound-memory arm (new `authorize_namespace_standard_parent`).
- **HTTP sqlite** — delegates to the MCP funnel, so parent authz flows through it (its own pre-checks pass `None`).
- **HTTP postgres** `set_namespace_standard_inner` — resolves the parent standard under the same bypass-visibility probe ctx (`for_admin(AI_HTTP_INTERNAL)`) the bound-memory gate uses, so a foreign `scope=private` parent standard is RESOLVED for the ownership check rather than folded into `NotFound → skip` (the #2709 precedent); enforced on both the resolved-bound and bound-absent arms.

This brings the local funnels to parity with the federation edge (`inbound_namespace_meta_authorized`), which already authorizes the declared parent (via peer scope, #2479/#2536). The forensic audit chain records the true `refuse` verdict.

## Route 2 — inferred ancestors vs governance
Kept the `-`-prefix inference for **lookup/display/resolution** (the documented `ai-memory-tests → ai-memory → ai` convenience is untouched in `build_namespace_chain`). Added a **governance view** of the chain — `build_namespace_governance_chain` (sqlite) and the postgres twins — that **excludes** `-`-inferred `parent_namespace` links from governance/approver/owner **layering**. Only an EXPLICITLY-declared parent chain contributes governance layers.

A link `child → parent` is treated as inferred iff `auto_detect_parent(child) == Some(parent)` (fail-closed: a coincident explicit parent is treated as inferred until provenance is persisted). Routed **every** chain-walking layerer through the governance chain: `resolve_governance_policy`, `resolve_require_approval_above_depth`, `resolve_skill_promotion_min_depth`, `namespace_owner`, plus both pg chain builders (`build_namespace_chain` + `build_namespace_chain_in_tx`). The `/`-derived G1 hierarchy and the global `*` substrate default are structural, never a graft vector, and unchanged. Postgres never `-`-auto-detects on write, so the pg filter is the fail-closed backstop for an inferred parent that reaches pg via federation / portability import (K3 parity).

## Deferred (out of scope — separate narrow vote)
On-disk **provenance PERSISTENCE** (recording the explicit-vs-inferred origin of a `parent_namespace` link) — a T4 on-disk sub-piece. Until it lands, an EXPLICIT parent that coincides with the `-`-auto-detect pattern is conservatively treated as inferred (fail-closed: grants no cross-namespace governance). Operators use a non-coincidental parent name or a `/`-hierarchy for explicit hierarchical governance. Only the in-memory authorization + governance-layering behavior ships here.

## R-203 evidence (both backends)
**sqlite** — `tests/ns_standard_parent_graft_2542.rs` (5 passed):
- `r1_unowned_parent_graft_allowed`, `r1_same_principal_parent_allowed`, `r1_cross_principal_parent_graft_refused_loudly`
- `r2_inferred_ancestor_does_not_contribute_approver_layer` (also asserts the inference survives for lookup), `r2_explicit_parent_does_contribute_approver_layer`

**postgres** — `tests/ns_standard_parent_graft_2542_pg.rs` (3 passed) — **run locally against the enterprise-fed tier** (pg18.6 + age + vector, mTLS verify-full; per-test schema isolation, tier left clean):
- `r1_cross_principal_parent_graft_refused_pg` (403 + no binding written), `r1_same_principal_and_unowned_parent_allowed_pg`, `r2_inferred_ancestor_excluded_explicit_included_pg`

Regression (green): governance_inheritance 8/8, governance_audit_verdict_2634 8/8, g_phase_e passthrough 8/8, delete_governance_strip_2503 10/10, recursive_learning_task2 16/16; live-tier governance_postgres_inheritance 10/10, ns_standard_set_ownership_2709_pg 17/17, cov_postgres_governance 3/3.

## CI-parity
- `cargo fmt --all --check` clean.
- clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` clean on **default / sal / sal-postgres / vectorlite**.
- `cargo check --examples` clean.
- Ceilings bumped in lockstep with dated comments: `src/store/postgres.rs` module-size 34_200 → 34_320; QUAL-7 `Result<(), String>` 40 → 42 (two sibling authz helpers mirror the shipped `authorize_namespace_standard_bind` whose String refusal already flows to the 403 body + audit record).
- No MCP tools / HTTP routes / CLI verbs / schema versions added → no SSOT pins to reconcile. CHANGELOG `[Unreleased]` updated.


alphaonedev:
## Adversarial-review rework pushed (rebased onto release tip `9339efd9`)

Per-finding response. The core change is that Route 2's governance exclusion is now **OWNERSHIP-based** (the reviewer's own "unowned or same-principal" legitimacy criterion, re-checked at resolution), which is provenance-independent and fixes Findings 1 + 2 uniformly.

### Finding 1 (BLOCKER) — FIXED. Confirmed + mechanism
**Confirmed the schema fact you asked about:** `storage::set_namespace_standard` PERSISTS the `-`-auto-detected parent into `namespace_meta.parent_namespace` indistinguishably from an explicit one, so a non-null `parent_namespace` is **not** reliably "explicit" for existing data — the stored-vs-null distinction is unavailable without the deferred provenance-persistence.

So instead of stored-vs-null, the governance chain now excludes a parent link by **ownership**, which is what "legitimate" means per your own key insight: a `parent_namespace` link layers governance **only when the parent is UNOWNED or owned by the SAME principal as the namespace being resolved** — the exact Route-1 bind rule, re-checked at resolution. Implemented in `build_namespace_governance_chain` (sqlite `ChainView::Governance`) via `namespace_standard_owner` + `parent_link_governance_entitled`, and the pg twin `pg_namespace_chain(.., governance = true)` via `pg_namespace_standard_owner_{pool,in_tx}`. A legitimately-declared same-owner parent — **including a flat `-` hierarchy `acme-corp` → `acme-corp-frontend`** — now KEEPS its Approve/approver/`require_approval_above_depth` layer (regression test `r2_explicit_same_owner_flat_hierarchy_inherits_approval_gate` asserts both the policy and the depth gate). A cross-tenant parent is dropped with a loud WARN (`ai_memory::governance::chain_graft`). "fail-closed" wording corrected to the truthful **de-restriction** everywhere (CHANGELOG + code + test comments).

### Finding 2 (TOCTOU) — CLOSED by the same mechanism
The resolution-time ownership re-check catches a parent that was unowned at bind time and later bound by another tenant. Test `r2_toctou_unowned_parent_bound_later_is_excluded`: attacker grafts `parent: victimns` while victimns has no standard (Route 1 allows), victim later binds a victim-owned standard → `resolve_governance_policy(attackerns)` returns `None`.

### Finding 3 (sqlite fails open) — FIXED
`resolve_namespace_standard_memory` now returns `Result<Option<Memory>, String>`; a DB read fault propagates and the funnel refuses the bind (fail-closed), matching the pg twin. `Ok(None)` is returned only for the genuinely-unowned states (no standard / severed pointer), parity with the pg `NotFound → None`.

### Finding 4 (backend-divergent trait) — FIXED
Reverted the pg `build_namespace_chain` filter: the **trait method is now the full LOOKUP chain on both backends**. Governance uses the distinct `pg_namespace_chain(.., governance = true)` (and `build_namespace_chain_in_tx(.., governance)`), so a future display consumer can never get an inference/ownership-dropped chain via the trait.

### Finding 5 (NIT doc) — FIXED
`authorize_namespace_standard_owner` doc corrected: only EMPTY / exact `system` / `SYSTEM_PRINCIPAL` pass — there is no `system:`-prefix match.

### Deferred provenance-persistence — no residual hole
NOT needed: the ownership re-check is provenance-independent. Its only remaining effect would be to also drop a SAME-OWNER `-`-inferred link from governance; keeping it is safe (same tenant; an ancestor can only ADD a restriction, never bypass one).

### Verification (live tier `AI_MEMORY_TEST_POSTGRES_URL`)
- sqlite `tests/ns_standard_parent_graft_2542.rs` — 9/9 pass: `r1_unowned/same/cross_principal_*`, `r2_inferred_cross_tenant_graft_excluded_from_governance`, `r2_same_owner_inferred_hierarchy_keeps_governance`, `r2_explicit_same_owner_flat_hierarchy_inherits_approval_gate` (**Finding 1**), `r2_explicit_non_coincident_same_owner_parent_inherits`, `r2_explicit_cross_tenant_parent_excluded_from_governance`, `r2_toctou_unowned_parent_bound_later_is_excluded` (**Finding 2**).
- pg `tests/ns_standard_parent_graft_2542_pg.rs` — 3/3 pass on the enterprise-fed tier: `r1_cross_principal_parent_graft_refused_pg`, `r1_same_principal_and_unowned_parent_allowed_pg`, `r2_cross_tenant_parent_excluded_same_owner_included_pg` (cross-tenant excluded + Finding-1 same-owner kept). pg Route 2 seeds THROUGH THE STORE so store-reads see the rows (post-#3055 the store normalizes its search_path; a raw side-pool write can land in a different schema).
- Regression green: `governance_inheritance` 8/8, `governance_audit_verdict_2634` 8/8, `g_phase_e…passthrough` 8/8, `delete_governance_strip_2503` 10/10, live-tier `cov_postgres_governance` 3/3.
  - **Heads-up (pre-existing, NOT this PR):** on the enterprise-fed LOCAL tier, `governance_postgres_inheritance` (9) and `ns_standard_set_ownership_2709_pg` (2) fail because they seed/read via a raw unqualified side-pool that lands in a different schema than the store post-#3055 (`$user` vs `public`, the exact CVE-2018-1058 split #3055 addresses). Verified identical failures with my `postgres.rs` reverted to base, so this is a #3055/local-tier infra interaction, not #2542; those suites use `/`-hierarchy inheritance which my ownership gate never touches, and they pass on CI's clean pg.
- fmt + clippy (`-D warnings -D clippy::all -D clippy::pedantic`) clean on default/sal/sal-postgres/vectorlite; hardcoded-literal + module-size + QUAL-7 ceilings green (bumped in lockstep).

alphaonedev:
## Federation regression FIXED (rebased onto release tip `592ea8ef`)

### Root cause
The CI failure (`control_parent_in_scope_applies_and_takes_effect_2479`) was a bug in HOW my ownership gate picked the comparison owner, not a missing federation exception. `seed_standard_memory` mints every standard in that suite with `x-agent-id: ai:victim`, so `alpha` (the federated in-scope namespace declaring `parent: victim`) and `victim` share an owner. But I compared the parent's owner against the **leaf** (`alpha/sub`, which has no standard → unowned) instead of the **declarer** of the link (`alpha`, owned by `ai:victim`). So the same-owner in-scope parent was wrongly dropped.

### How the federation-authorized parent entitlement is represented
Changed the check to **PER-HOP against the DECLARER**: for each `child → parent` `parent_namespace` hop, entitled iff the parent is UNOWNED **or** owned by the same principal as `child` (the namespace whose `namespace_meta` row declared the link). A `/`-child inherits its `/`-parent's entitled links structurally, so an unowned leaf still inherits.

This IS the reconciliation the review asked for:
- **(a) unowned** and **(b) same-principal** — as before, but now against the correct owner (the declarer).
- **(c) federation-peer-authorized in-scope parent** — the #2479 control mints `alpha` and `victim` same-owner at the receiver, so the link is **same-principal AT THE HOP** and kept by (b). It's exactly the Route-1 bind rule re-checked at the hop where the link was declared — which is precisely what `inbound_namespace_meta_authorized` gated at push time for the in-scope case. No provenance-persistence needed, no schema change.
- The **cross-tenant local graft stays refused**: at bind by Route 1 (forward), and at resolution the per-hop check still drops a declarer→parent hop with different owners (inference coincidence, TOCTOU-bound-later, pre-#2542 legacy). Confirmed by `r2_inferred_cross_tenant_graft_excluded_from_governance`, `r2_explicit_cross_tenant_parent_excluded_from_governance`, `r2_toctou_unowned_parent_bound_later_is_excluded` — all still `None`.

### New regression test (both backends)
`r2_federation_in_scope_same_owner_parent_applies_through_unowned_leaf` (sqlite) and `…_pg` (pg): `victim` (owner V, `write:any`) + `alpha` (owner V, `parent: victim`) + unowned leaf `alpha/sub` → `resolve_governance_policy("alpha/sub") == Any`. This pins the exact leaf-vs-declarer bug so it can't regress again.

### Verification (the previously-missed suites)
- `federation_ns_meta_scope_2479` — **11/11 GREEN** (was `control_parent_in_scope…` red).
- `federation_namespace_meta_descendant_2536` — **10/10 GREEN**.
- `tests/ns_standard_parent_graft_2542.rs` — **10/10** (9 prior + the new federation regression).
- pg twin `tests/ns_standard_parent_graft_2542_pg.rs` — **4/4** on the enterprise-fed live tier (`AI_MEMORY_TEST_POSTGRES_URL`, pg18.6+age+vector, mTLS verify-full).
- Governance regression: `governance_inheritance` 8/8, `delete_governance_strip_2503` 10/10, live-tier `cov_postgres_governance` 3/3.
- fmt + clippy (`-D warnings -D clippy::all -D clippy::pedantic`) clean on default/sal/sal-postgres/vectorlite; module-size + QUAL-7 + hardcoded-literal ceilings green.

Docs (CHANGELOG + code + both test module docs) updated to state the per-hop/declarer rule and the #2479 reconciliation accurately.

Issue #3033, opened 2026-08-16
asi-hard does NOT pin the OUTER federation transport gates — AI_MEMORY_FED_REQUIRE_SIG=0 (+ NONCE/PEER_ENROLLMENT/PUSH_NAMESPACE_SCOPE/POLICY_CURRENT/PERMISSIONS_MODE) boots clean under the hardened no-disable profile


alphaonedev:
A7-F6, needs T3 vote. `src/security_profile.rs::KNOBS` (17) pins the four INNER sig lanes (WRITE/SIGNAL/TRANSITION/CHECKPOINT_SIG) but NOT the outer /sync/push envelope gate AI_MEMORY_FED_REQUIRE_SIG (#29), nor NONCE/REQUIRE_PEER_ENROLLMENT/REQUIRE_PUSH_NAMESPACE_SCOPE/REQUIRE_POLICY_CURRENT/PERMISSIONS_MODE/GOVERNANCE_FAIL_OPEN_ON_ERROR. Empirical: SECURITY_PROFILE=asi-hard + FED_REQUIRE_SIG=0 → boot proceeds, [PASS] asi-hard pinned knobs, [FAIL] only under --posture. asi-hard advertises no-disable; an operator can disable per-message federation envelope sig verification under it. FIX: add the knobs to KNOBS — but a T3 posture change → 5-agent vote. Severity MED. Verified pm-v3.3 (fresh subprocess) + codegraph-cross-correlated. Hive-NHI feature campaign 2026-08-16.

alphaonedev:
Fixed via PR #3094 (merged 39d7891b). asi-hard now pins the four OUTER federation-transport gates (incl. AI_MEMORY_FED_REQUIRE_NONCE) fail-closed; boot refuses when an operator loosens any pinned knob. Verified: security_profile subprocess-dispatch + unit tests green. Verified per prime directive pm-v3.

alphaonedev:
**Independently corroborated** by an ox-alpha security-review swarm (Fable-verified against release/v1.0.0): `authorize_remote_transition` / `authorize_remote_checkpoint_resolution` (`src/federation/receive_auth.rs:66`, :141) both have permissive `else { Accept }` arms for BOTH unsigned and signed-but-unenrolled authority transitions when `require_sig=false`:
```rust
if signature.is_empty() { return if require_sig { RejectUnsigned } else { Accept }; }
[...]

Issue #2923, opened 2026-08-13
doctor --posture: 'asi-hard pinned knobs' row reports 17/17-at-floor PASS while SECURITY_PROFILE is standard (vacuous or mislabeled check)


alphaonedev:
Found by Fable 5's evidence-bundle audit of PR #2917 (2026-08-13), visible in the committed artifact docs/compliance/evidence/cert-54/posture-bare-env.out.

Under AI_MEMORY_NO_CONFIG=1 with no knobs set, the enterprise-federation posture output shows:

- `AI_MEMORY_SECURITY_PROFILE` -> `[FAIL] actual: (unset -> standard)`
- next row: `[PASS] asi-hard pinned knobs / required: all 17 at hard floor / actual: 17/17 at floor`

A check named for asi-hard reporting 17/17 at floor while the resolved profile is `standard` reads as either vacuous (nothing is pinned under standard, so "at floor" is trivially true) or mislabeled. Since this artifact is committed, publicly-auditable cert evidence, the row should either (a) report N/A / SKIPPED when the profile is not asi-hard, or (b) be renamed to state what it actually verifies under standard.

Evidence anchor: src/enterprise_federation_posture.rs (evaluate; the asi-hard pins row) + src/cli/doctor.rs:561 run_posture. Overall verdicts are unaffected (the SECURITY_PROFILE row itself fails the bare leg), so this is a truthfulness/legibility defect in one check row, not a gate bypass.

alphaonedev:
Fixed via PR #2927 (merged in release, head 1dcd2c97): under a standard profile the 'asi-hard pinned knobs' posture row now FAILs with an honest actual ('profile=standard — asi-hard pins not in force; floor not evaluated') instead of a vacuous 17/17-at-floor PASS; byte-identical under real asi-hard. Both postures pinned by subprocess-isolated tests (#2905 discipline). The corrected 18-row output flows into the cert-54 evidence re-capture in the cert-refresh commit. Retest evidence: full CI green (42 pass) on the merged head.

PR #3114, opened 2026-08-22
fix(storage): migration ladder fails OPEN — a lost core relation no longer stamps silently (#3113)

### LINKED ISSUE #3113: [data-integrity] migration ladder fails OPEN: existence-probe arms skip a lost core relation, tail stamps the tip anyway
## Confirmed against the current tree (release/v1.0.0 @ 14ce0bb5)

Several sqlite ladder arms in `src/storage/migrations.rs::migrate` are gated on an existence probe and SKIP when the relation is absent:

- v34 — `signed_events` `prev_hash`/`sequence` chain columns + backfill (`src/storage/migrations.rs:2243`)
- v66 — `governance_rules` severity-CHECK widening rebuild (`src/storage/migrations.rs:3151`)
- v73 — `signed_events.cause_hash` column (`src/storage/migrations.rs:3334`)
- v36 — `memory_links` relation-CHECK rebuild (`src/storage/migrations.rs:2327`)

e.g. v66:

```rust
let has_governance_rules = conn.prepare("SELECT 1 FROM governance_rules LIMIT 0").is_ok();
if has_governance_rules { conn.execute_batch(MIGRATION_V66_SQLITE)?; }
```

The in-code comments justify the probes ONLY via test fixtures, and that justification is real: a fixture may stamp a high `schema_version` over a database bootstrapped from the `SCHEMA` constant alone, and `SCHEMA` deliberately omits the later-added ladder-only relations. Skipping is correct there.

## The hole

Nothing distinguishes that benign case from a POPULATED database whose relation was LOST — corruption, a partial file-level restore, an operator `DROP`. After the skips, the tail of `migrate` unconditionally stamps:

```rust
conn.execute("DELETE FROM schema_version", [])?;
conn.execute("INSERT INTO schema_version (version) VALUES (?1)", params![CURRENT_SCHEMA_VERSION])?;
Ok(())
```

So such a database "upgrades successfully" to v89 with the signature-atomicity trigger restoration, the relation CHECK rebuild and the severity-CHECK widening never applied. Subsequent writes then bypass integrity constraints the version stamp asserts are in force — fail-OPEN where integrity gating must fail closed.

## Why it is production-reachable, not test-only

The relations involved are created UNCONDITIONALLY by their arms (`archived_memories` v4, `namespace_meta` v5, `signed_events` v26, `governance_rules` v30) and are NOT in the bootstrap `SCHEMA` constant. So every genuine ladder upgrade has them, and their absence under a high stamp is an unambiguous signal that the create arm never ran against that file. Partial file-level restore and catalogue corruption are exactly the scenarios the North Star names.

## Proposed fix

Report unconditionally, enforce opt-in:

1. An SSOT of ladder-created core relations + a probe (`storage::schema_integrity`).
2. A loud structured WARN at migrate time naming each missing relation, the version that introduced it, the integrity control not in force, and the live-corpus row count (the populated-vs-fixture discriminator).
3. A standing `core_relations` fact in `ai-memory doctor`, raising Storage to Critical.
4. `AI_MEMORY_MIGRATION_REQUIRE_CORE_TABLES=1` to escalate to a REFUSAL, sited inside the migrate transaction BEFORE the stamp so a refusal rolls back and leaves the database exactly as found (old version, rows intact, still readable, reversible).

Default-off enforcement is deliberate: defaulting to refuse would turn every existing high-stamp fixture and archive-less deployment into a hard boot failure — a fleet-wide availability regression for no data-integrity gain, since the check mutates nothing.

Found by an adversarial migrations audit (FINDING-2). The same audit's FINDING-4 (`encrypted_envelope` lost on archive→restore) is REFUTED — migration v68 mirrors the column onto `archived_memories`, both sqlite archive writers and `restore_archived` carry it, and `PostgresStore::migrate_v68` matches.

alphaonedev:
Closes #3113
Partially addresses #3159 (`agent_quotas` only — see *Scope*)

> **Status:** queue HEAD. Rebased onto `origin/release/v1.0.0` = `55c0486e` (#3115 merge). Head `537dfcff` (SSH-signed `%G?=G`). Knob count **22** is the SSOT (`security_profile::PINNED_KNOB_COUNT` / docs-vs-ssot `ASI_HARD_PINNED_KNOB_COUNT`). Fable-gate `COUNT(*)` short-circuit is in `refusal_required`. CI-parity green locally. Do **not** re-run CI.

### FINDING-4 `encrypted_envelope` lost on archive→restore — **REFUTED**

The claim was that no arm ever extends `archived_memories` with `encrypted_envelope`. The audit enumerated the v49/v85/v87 parity passes and stopped there — it missed **migration v68**, which exists for precisely this:

- `src/storage/migrations.rs:3201-3231` — v68 (#228/#1728) mirrors the column, and its comment states the exact harm the finding predicts as the thing it prevents.
- Archive writers carry it on **both** sides of the `INSERT ... SELECT`: `archive_memory_no_tx` (`src/storage/mod.rs:4411`, `:4419`), `archive_memory_insert_only` (`:4500`, `:4508`).
- Restore carries it back: `restore_archived` (`:13475`, `:13509`).
- Postgres matches: `PostgresStore::migrate_v68` (`src/store/postgres.rs:4232-4249`) adds the BYTEA to both tables; pg archive INSERTs carry it (`:5950`, `:6375`).
- `load_archived_as_memory` deliberately omits it, and says why (`:13862-13870`): it feeds only the governance pre-write hook, and selecting the envelope would fail-close every restore whose key rotated — while the restore `INSERT ... SELECT` moves the ciphertext verbatim.

No fix made; the round-trip is already lossless on both backends.

### FINDING-2 existence-probe arms fail OPEN — **CONFIRMED**, fixed here

Arms gated on an existence probe skip when the relation is absent — v34 `signed_events` chain columns (`:2243`), v66 `governance_rules` severity CHECK (`:3151`), v73 `signed_events.cause_hash` (`:3334`) — and the tail of `migrate` then stamps unconditionally.

The probes' stated justification (test fixtures) is genuine and is preserved. What was missing is the other reading of the same observation: nothing distinguished a fixture from a **populated database that lost the relation** (corruption, partial file-level restore, operator `DROP`). Such a database "upgraded successfully" to the tip with the signature-atomicity triggers, relation CHECK and widened severity CHECK never applied, and later writes bypassed constraints the stamp asserts are in force.

**Production-reachable.** The relations involved are created *unconditionally* by their arms and are *not* in bootstrap `SCHEMA`, so every genuine upgrade has them; absence under a high stamp means the create arm never ran against that file.

## The fix

New `src/storage/schema_integrity.rs` — an SSOT of ladder-created core relations plus the probe. A relation qualifies only if its arm creates it unconditionally **and** it is absent from bootstrap `SCHEMA` (a `SCHEMA` relation is re-created empty on every `db::open`, so its absence is unobservable and a row would be dead).

| Relation | Created at | Not in force while absent |
|---|---|---|
| `archived_memories` | v4 | archive/restore has no destination |
| `namespace_meta` | v5 | namespace standards + parent inheritance |
| `signed_events` | v26 | append-only audit chain; v34 + v73 columns |
| `agent_quotas` | v28 | K8 blast-radius caps; v50 per-namespace PK |
| `governance_rules` | v30 | rule storage; v66 widened severity CHECK |

- **Detection is unconditional** — a structured WARN naming each missing relation, its introducing version, the control not in force, and the corpus row count (recorded as `Some(n)`/`None`, never flattened to a sentinel).
- **Standing doctor signal** — a `core_relations` fact raising Storage to `Critical`, readable long after the migration that skipped.
- **Enforcement is opt-in** — `AI_MEMORY_MIGRATION_REQUIRE_CORE_TABLES=1`, and **pinned ON by `asi-hard`** (`KNOBS` 21 → 22; the first schema-integrity pin).

### Why the gate sits where it does, and why refusal needs demonstrable loss

The check runs **inside the migrate transaction, immediately before the stamp**, so a refusal rolls the ladder back and leaves the database *exactly as found* — old version, rows intact, still readable — reversible by unsetting the knob.

Refusal additionally requires a **positively observed populated corpus**. An empty database with a high stamp is the ordinary fixture / archive-less shape — no lost data, because no data — and since `asi-hard` pins enforcement ON, refusing it would brick every fresh hardened deployment, making the hardened posture *more* fragile than the standard one for no integrity gain. An unreadable corpus does not refuse either: the check refuses only on a fact it positively established, never on an absence of information.

**The gate cannot itself lose data.** No DDL, no DML — `sqlite_master` plus one `COUNT(*)`, and the count is short-circuited away on the healthy path so no migration pays for it inside `BEGIN EXCLUSIVE`. `the_check_issues_no_ddl_and_no_dml` round-trips the catalogue and row count across a probe. **No new migration was written** — `CURRENT_SCHEMA_VERSION` is untouched at 89.

One deliberate difference from pre-#3113: a `sqlite_master` probe **read error** propagates and fails closed rather than being coerced into "nothing missing" — reporting integrity as intact on a failed read is the same fail-open shape being removed (#2445 precedent). Documented at every site claiming byte-identical behaviour.

## Scope — what this PR does *not* do

**Partially addresses #3159.** Of the three further probe-skips reported, only one is this class:

- **`agent_quotas` (v50 skip) — folded in**, recorded at its **v28 CREATE site**, not the v50 skip site. Recording the skip site would have silently stopped checking every database stamped 28..49 — a one-directional, invisible under-report. Pinned by `agent_quotas_is_expected_from_v28_not_v50`.
- **v56 — not an integrity hole.** Its guard is `archived_memories`, covered here since v4, so a lost table already fires. The object it skips is a **non-UNIQUE** index: losing it degrades `list` performance and cannot lose or corrupt a row.
- **`agent_lineage` (v80) — cannot be covered by this mechanism.** It ships in bootstrap `SCHEMA` (`migrations.rs:397`), so it always exists by the time the ladder runs and a row would be permanently dead.

#3159 stays open for the residual.

**Explicit non-goal: #3172 (BLOCKING).** `agent_lineage` has a worse exposure this PR does not attempt and that no table-existence check can detect: bootstrap `SCHEMA` silently **re-creates it EMPTY** after a drop, the v80 probe then finds a table, the rebuild succeeds on zero rows, and the identity-lineage chain is gone with **no skip logged**. That is a SCHEMA-masks-data-loss class needing row-count/content attestation, not existence probing.

## Also fixed (self-review)

- The doctor Storage section **composes** its note instead of overwriting — a `dim_violations` note was being dropped when both branches fired.
- A failed core-relation probe raises `Warning` (unverified ≠ verified-good) without downgrading an already-raised `Critical`.
- **The `asi-hard` pinned-knobs doc table is now mechanically pinned to `KNOBS` by set equality.** It had silently fallen **two rows behind** before this branch (`FED_REQUIRE_TRANSITION_SIG`, `FED_REQUIRE_CHECKPOINT_SIG` — 19 rows vs 21 knobs) with nothing failing; a count-only check would not have caught it, since the count lived in prose that drifted too. `pinned_knobs_doc_table_matches_the_knobs_ssot_exactly` asserts both directions and is **mutation-proven**: deleting a row fails with `pinned by KNOBS but MISSING a table row: [...]`; a phantom row fails with `documented as pinned but ABSENT from KNOBS: [...]` — the more dangerous direction, since the docs would otherwise advertise a hardening guarantee the binary does not enforce.

- **The knob COUNT is now derived too — and the gate that should have caught its drift had a fail-open of its own.** The set-equality test above pins the doc table's MEMBERSHIP; the count itself was still a hand-maintained integer in six places, including a literal `== 22` inside the very test that claimed to police it. Three changes close it:
  - `security_profile::PINNED_KNOB_COUNT = KNOBS.len()` — one derived SSOT. `asi_hard_pins_documented_set` now asserts `pinned_knobs()` *and* the module doc table's row count against it, so no number can be hand-bumped.
  - A new `ASI_HARD_PINNED_KNOB_COUNT` rule in `scripts/check-docs-vs-ssot.sh` resolves the canonical from the `KnobSpec` entries themselves and fails on any prose site that disagrees. **Verified per-site, not assumed** — a rule whose regex matches nothing in a listed file is a no-op that still reports PASS, so all seven currently-matching lines were mutated `22 -> 21` one at a time and each was CAUGHT (`CLAUDE.md:541`, `docs/compliance/ENTERPRISE-FEDERATION-CERTIFICATION.md:333,343`, `src/security_profile.rs:434,1052`, `src/enterprise_federation_posture.rs:31,907`, `scripts/check-bootstrap-cert-gate.sh:66`).
  - **A FAIL-OPEN INSIDE THE DOCS GATE.** `check_narrative_count_rule`'s historical-line guard treated a leading `#` as a markdown HEADING for *every* file type. In a `.sh`/`.env` scan target — where `#` is the COMMENT marker — that skipped **every comment line**, so a rule listing such a file covered nothing in it and the gate still printed PASS: the #2444 "reports success while doing nothing" shape, arrived at by accident. The anchor is now scoped to `.md`/`.html`. No existing rule changes behaviour — every other scan set (`DOC_FILES` / `HOOK_DOC_FILES` / `HTML_DOC_FILES` / `PGVECTOR_DOC_FILES`) is markdown and html only. **This is not hypothetical, and it is not only mine:** applying PR #3169's proposed knob-count anchors with the pre-fix guard shows its rule silently skipping `docs/deploy/enterprise-federation.env:54` — *the exact line that PR exists to correct*. A new `--self-test` leg pins both directions (a stale count in a shell comment is REJECTED; the same count in a genuine markdown heading is still SPARED, so the guard was scoped rather than deleted), and it fails when the one-line scoping is reverted.
- **`docs/deploy/asi-hard.env` is now pinned knob-by-knob.** CLAUDE.md's env row #130 stated the template "names all N correctly and is pinned by `tests/deploy_templates.rs`". It was not: that file's only link to the SSOT was `!pinned_knobs().is_empty()`, true of any non-empty set, so a knob added to `KNOBS` could leave the operator-facing deploy enumeration silently short — a procurement document describing a weaker posture than the binary enforces. `asi_hard_env_names_every_pinned_knob` requires every pinned knob to be NAMED, matching **whole identifier tokens rather than substrings**: `AI_MEMORY_FED_REQUIRE_SIG` is a strict prefix of `AI_MEMORY_FED_REQUIRE_SIGNAL_SIG`, so a `contains` check would report the shorter knob as documented purely because the longer one is present. Mutation-proven in both shapes (dropping `MIGRATION_REQUIRE_CORE_TABLES` fails; removing the standalone `FED_REQUIRE_SIG` mentions while `FED_REQUIRE_SIGNAL_SIG` stays fails). The CLAUDE.md sentence is rewritten to describe what is now actually enforced.

## Rebase onto `55c0486e` (#3115) + remaining 22-count surfaces

Queue-head ping after #3115 merged. Linear rebase of the 12 commits onto `55c0486e` (the inherited merge `22a5b005` dropped). CHANGELOG `[Unreleased]` kept **both** sides — #3115's Linux-CI / hosted-runner entries plus this PR's `#3113` fail-open section.

Fable-gate item (quote): `refusal_required` short-circuits `missing.is_empty()` **before** `corpus_row_count` so a healthy migrate never pays `SELECT COUNT(*) FROM memories` inside `BEGIN EXCLUSIVE` (`src/storage/schema_integrity.rs`):

```rust
if missing.is_empty() {
    return false;
}
refusal_required_with(
    missing,
    crate::config::migration_require_core_tables(),
    corpus_row_count(conn),
)
```

The surfaces previously listed as "deliberately NOT fixed here" **landed** in this PR (operator: fix 100% of every verified finding):

- `PERFORMANCE.md` pinned-knobs table completed 15 → 22 and SET-equality pinned by `performance_md_pinned_knobs_table_matches_the_knobs_ssot_exactly`. Mutation-proven both directions: drop `AI_MEMORY_MIGRATION_REQUIRE_CORE_TABLES` → `MISSING a PERFORMANCE.md pinned-knobs row`; add `AI_MEMORY_THIS_KNOB_IS_NOT_PINNED` → `ABSENT from KNOBS`. Restore verified `git diff --exit-code -- PERFORMANCE.md`.
- `SECURITY.md:63` (count **and** enumeration), `SECURITY.md:169`, `README.md:43`, `docs/deploy/README.md` (17 → 22 bullets), `docs/enterprise-deployment.md:1939`, `docs/deploy/enterprise-federation.env:54` all read **22**.
- Leftover CURRENT-state `21` comments in `src/enterprise_federation_posture.rs` moved to 22; two #2923 historical `21/21` render examples rephrased as count-independent `N/N`.

`#3169` independently adds a duplicate knob-count rule that still says **21**. **This rule is canonical** (strict superset of #3169's scan set). #3169's duplicate collapses into this one at **#3169's rebase**; its `ENTERPRISE_FEDERATION_CHECK_COUNT` rule is #3169's and must survive. Historical CHANGELOG / cert-doc pre-#3033 notes / `infra/federation-lab` campaign logs are left intact.

QUAL-10 re-measured after the rebase: `src/storage/mod.rs` 29_201 / ceiling **29_260** (#3115); `src/store/postgres.rs` 34_692 / ceiling **34_720**; `src/storage/migrations.rs` 5_777 / ceiling 5_850. No bump.

Postgres: this binary is schema **v89**. Shared enterprise-fed PG is **v90**. No `AI_MEMORY_ALLOW_SCHEMA_AHEAD`; CI isolated PG is the authority. This PR does not touch `src/store/postgres.rs` / `postgres_schema.sql`.

## Tests + CI parity (re-run after rebase onto `55c0486e`, PATH-pinned rustc 1.96.0)

- `security_profile` **12/12** including `performance_md_pinned_knobs_table_matches_the_knobs_ssot_exactly` (mutation-proven both directions).
- `deploy_templates` **6/6**.
- `migration_core_relation_integrity_3113` **7/7**; `migration_core_relation_empty_corpus_3113` **1/1**; `migration_core_relation_refusal_3113` **1/1**.
- `qual_10_module_size_ceiling` **2/2** (no ceiling bump); `qual_6_7_legacy_error_type_ceiling` **2/2**; `doc_claims_integrity` **4/4**.
- Clippy at the exact CI flag set (`-D warnings -D clippy::all -D clippy::pedantic`) on all **five** invocations (`cargo clippy --all-targets`; `--all-targets --features sal`; `--features sal-postgres --tests`; `--features sal`; `--all-targets --features vectorlite`); `cargo check --all-targets` (default + `sal` + `sal-postgres`); `cargo check --examples`; `cargo fmt --all --check`; `check-docs-vs-ssot.sh` **and its `--self-test` (27 PASS, including the #3113 shell-comment fail-open leg)**; `check-hardcoded-literals.sh` (baseline untouched).

The two env-mutating tests each live **alone in their own binary** (`2905-posture-test-env-leak`). The per-relation contract is **table-driven over the SSOT**. `asi_hard_actually_enables_the_migration_core_relation_gate` proves the pin *enforces* end-to-end.

No new MCP tools, HTTP routes, CLI verbs, or schema versions.

Rules: ERRORS-01/02/06/09/19, CONST-01, API-01/29, OWNERSHIP-02/13, TOOL-03, TEST-01/02.


alphaonedev:
**Fable final-gate review — HOLD, two items (otherwise approved in substance).**

Substance ✅: the `encrypted_envelope` REFUTATION is correct (v68 at `migrations.rs:3201-3231` mirrors the column; archive writers carry it on both sides on both backends) — no fix was the right call. The fail-open ladder fix is the correct shape: a no-DDL/no-DML `sqlite_master`+`COUNT(*)` check immediately before the stamp (`migrations.rs:3862` → stamp at `:3872`), refusal rolls the ladder back leaving the DB untouched, failed catalogue read propagates (fail-closed), enforcement tests isolated per the #2905 lesson, clippy clean on all 4 legs. Filed #3113.

Before merge:
1. **Required CI red — hardcoded-literal ratchet**: `"governance_rules"`, `"namespace_meta"`, `"signed_events"` each +3 over baseline 0 (the new `CORE_TABLES` literals duplicate the migration arms). Hoist to single consts (reuse existing table-name consts if present); no `--update-baseline`.
2. **Pin the refuse mode in the certified posture**: warn-by-default is right for the fleet, but `AI_MEMORY_MIGRATION_REQUIRE_CORE_TABLES=1` must be pinned in the asi-hard KNOBS (`src/security_profile.rs`) so the certified config fails closed — same class as #3033. Plus a confirmation that an empty corpus at a tip stamp never refuses under REQUIRE=1 (fresh deployments must not brick).

Sent to the implementing Opus 5 instance.

alphaonedev:
## Final-gate follow-ups addressed (f9f50904)

### 1. Hardcoded-literal ratchet — PASS, baseline untouched

Grepped for an existing table-name const first, as directed. The same-valued consts in the tree are each bound to a different subsystem: `export_scope::OMITTED_CLASS_ARCHIVED_MEMORIES` is a portability omission *label*, `erasure::archive_sync::ARCHIVED_TABLE` selects an erasure-store table, `store::postgres::TABLE_ARCHIVED_MEMORIES` is postgres-local and private, and `signed_events::CHAIN_SIGNED_EVENTS` names a chain, not a relation. Reusing any of them would bind a schema invariant to an unrelated subsystem's naming.

So `storage::schema_integrity` now carries the storage-layer SSOT — `TABLE_ARCHIVED_MEMORIES` / `TABLE_NAMESPACE_META` / `TABLE_SIGNED_EVENTS` / `TABLE_GOVERNANCE_RULES` — referenced from the `CORE_TABLES` entries, the unit tests, and all three integration binaries. Each name is typed exactly once, with a doc comment recording why the existing consts were not reused.

`scripts/check-hardcoded-literals.sh` → **PASS (no duplicated string literal rose above baseline)**. No `--update-baseline`.

On the env-var name: `AI_MEMORY_MIGRATION_REQUIRE_CORE_TABLES` is defined once (`config::ENV_MIGRATION_REQUIRE_CORE_TABLES`) and referenced from the security-profile pin and the tests. The remaining occurrences you listed are doc comments and operator-facing message prose, not code paths — the gate does not flag them and it reads correctly in a WARN.

### 2. asi-hard pin — done, 21 → 22

`ENV_MIGRATION_REQUIRE_CORE_TABLES` is in the `KNOBS` SSOT with `hard_value: "1"` / `meets_floor: is_truthy` — the first **schema-integrity** pin, alongside the crypto/attestation/durability/network ones. Lockstep: `asi_hard_pins_documented_set` (count 22 **plus an explicit assertion for the new pin**, so a silent removal fails), `docs/deploy/asi-hard.env` (pinned by `tests/deploy_templates.rs`), CLAUDE.md row #130 enumeration + count, and the cert doc's `pinned_knobs().len()` clause (`22/22` on re-capture). CHANGELOG notes it.

### 3. Empty corpus under REQUIRE=1 — CONFIRMED, and you caught a real defect

You were right, and this was a genuine bug in the first cut, not just a missing test: the populated-corpus discriminator only *coloured the WARN*; it did not gate the refusal. An empty database with a high stamp and a missing relation **would have been refused** — and with the knob now pinned by `asi-hard`, that would have bricked every fresh hardened deployment. Strictly worse than the fail-open hole it replaces.

Fixed. `refusal_required_with(missing, enforced, corpus_rows)` requires **all three**: a missing relation, engaged enforcement, and a **positively observed non-empty corpus**. `corpus_rows == None` (unreadable) also does not refuse — the check refuses only on a fact it positively established, never on an absence of information. The predicate takes the flag and row count as parameters so the invariant is provable without touching process-global env.

**The tests you asked for, by name:**

| test | binary | asserts |
|---|---|---|
| `enforced_posture_never_bricks_an_empty_corpus` | `tests/migration_core_relation_empty_corpus_3113.rs` | REQUIRE=1, **two** relations dropped, stamp 80, **0 rows** → `db::open` **succeeds** and reaches the tip |
| `an_empty_corpus_never_refuses_even_when_enforced` | unit (`schema_integrity`) | `refusal_required_with(missing, true, Some(0)) == false` |
| `an_unreadable_corpus_never_refuses_even_when_enforced` | unit | `… , true, None) == false` |
| `a_populated_corpus_refuses_when_enforced` | unit | `… , true, Some(1)) == true` |
| `a_populated_corpus_reports_only_when_enforcement_is_off` | unit | default posture never refuses |

The empty-corpus integration test is alone in its own binary (env mutation + concurrent tests within a binary — the `2905-posture-test-env-leak` lesson), which is now three single-purpose binaries.

### Also fixed on self-review (not requested)

- The doctor Storage section **composed** its note instead of overwriting it — a `dim_violations` note was being dropped whenever both branches fired. Dropping an operator diagnostic to make room for mine is its own small data loss.
- A failed core-relation probe now raises `Warning` (unverified ≠ verified-good) without downgrading an already-raised `Critical`.

### Gates

`check-hardcoded-literals.sh` PASS · `check-docs-vs-ssot.sh` PASS · qual-10 PASS · `cargo fmt --check` clean · `cargo check --all-targets` clean · clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` clean on **all four** legs (default / sal / sal-postgres / vectorlite) · 15 unit + 8 integration + `deploy_templates` 5/5 + `security_profile` 9/9 + `cli::doctor` 75/75, all green.

One unrelated pre-existing finding while running the gates: `scripts/check-const-name-literals.sh` HARD-BLOCKs on `tests/signed_events_dlq_replay_1046.rs:109` (`payload_hash` embeds `0xad`), last touched by 247a438e. Untouched by this branch — flagging it rather than fixing it here.

alphaonedev:
**Fable gate — all three conditions met in `f9f5090`; APPROVED, merge on green required CI (after rebase onto `15fe44e4`, the new self-hosted base).**
1. Ratchet: table names routed through an SSOT (`TABLE_ARCHIVED_MEMORIES` / `TABLE_NAMESPACE_META` / `TABLE_SIGNED_EVENTS` / `TABLE_GOVERNANCE_RULES`) — CI's literal gate will confirm the counts no longer exceed baseline.
2. Certified posture fails closed: `ENV_MIGRATION_REQUIRE_CORE_TABLES` pinned in the asi-hard KNOBS (`security_profile.rs:376`, asserted at `:830`).
3. Fresh deployments can't brick: `an_empty_corpus_never_refuses_even_when_enforced` (unit) + `enforced_posture_never_bricks_an_empty_corpus` (integration) pin the populated-corpus predicate.
Substance unchanged from my earlier review (no-DDL/no-DML pre-stamp check, fail-closed on catalogue-read error, encrypted_envelope finding correctly refuted).

alphaonedev:
## Rebased onto the new base + toolchain note

Merged `origin/release/v1.0.0` (now `15fe44e4`, #3109 self-hosted 2x2 CI) into the branch as `22a5b005`. The merge was **clean** — no conflicts in `CHANGELOG.md` or `CLAUDE.md`. #3109's `### Changed` / `### Fixed` / `### Removed` entries and this PR's `### Fixed (migration ladder fail-open …)` section sit in different places under `[Unreleased]`, and #3109 touched no env-table row. Verified afterwards: no conflict markers anywhere, both sides' CHANGELOG content present.

All gates re-run against the merged tree, single push:

- `cargo check --all-targets` clean
- clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` clean on **all four** legs (default / sal / sal-postgres / vectorlite)
- 15 unit + 8 integration (3 binaries) + `deploy_templates` 5/5 + qual-10 2/2 + qual-6/7 2/2 — all green
- `check-docs-vs-ssot.sh` PASS · `check-hardcoded-literals.sh` PASS (baseline untouched) · `cargo fmt --check` clean

Also folded in `b35b50a8`: the knob **name** now reaches every runtime message through `config::ENV_MIGRATION_REQUIRE_CORE_TABLES` rather than re-typed prose — the migrate WARN carries it as a structured `enforce_env` field (queryable in the log pipeline, and it cannot go stale on a rename), and the refusal message + doctor note interpolate the const. Only doc comments still spell it literally, which is where a reader wants to see it.

### Environment note for the box (not a PR issue)

Mid-session the box's rustup default flipped from `1.96.0` to `stable`, and `stable` here is **1.95.0** — below the crate's MSRV, so cargo aborts with `rustc 1.95.0 is not supported`. It was intermittent: `RUSTUP_TOOLCHAIN=1.96.0` sometimes still resolved 1.95.0, which points at concurrent `rustup` mutation rather than a stale shell — plausibly the new self-hosted runners installing toolchains on the same machine. I did **not** change the box default (not mine to change); I pinned my own invocations to the absolute toolchain path (`~/.rustup/toolchains/1.96.0-.../bin/cargo`), which is immune to it. Worth knowing if another agent hits a spurious MSRV abort, and worth deciding whether the self-hosted runners should own their toolchain rather than the shared default.

alphaonedev:
**Queue head.** #3115 merged. Rebase onto current `origin/release/v1.0.0`, keep knob count **22** + `ASI_HARD_PINNED_KNOB_COUNT` SSOT, apply Fable-gate items, one `--force-with-lease` push. Do **not** write `ci-allow.txt`.

Issue #1, opened 2026-04-02
[CRITICAL] No authentication on HTTP API


alphaonedev:
All 20 HTTP endpoints are completely unauthenticated. Any network client can read/modify/delete all memories.

**Mitigation:** Server binds to localhost (127.0.0.1) by default. Documented as design choice for local-machine tool.

**Status:** Accepted risk for localhost-only service. Users deploying to network must add reverse proxy with auth.

Issue #2503, opened 2026-07-30
[SECURITY][data-integrity] a CORRECTLY-AUTHORIZED in-scope federated delete strips the governance policy of a namespace the peer is denied — via storage::delete's unqualified namespace_meta cleanup


alphaonedev:
## Summary

A federated `deletions[]` entry that the #2447/#2488 namespace gate **correctly ALLOWS** — an in-scope row, right peer, default config, no knobs — destroys the governance policy of a namespace that peer is explicitly **denied**. Through the global `*` standard, one in-scope delete can strip the policy of the entire namespace tree.

Probed end-to-end. Present at parent, so **not** introduced by #2497 — but it runs straight through the funnel that PR exists to confine, which makes it the sharpest counterexample to "namespace scope is the control on this lane."

## Mechanism

`storage::delete` (`src/storage/mod.rs:3484`) unconditionally executes

```sql
DELETE FROM namespace_meta WHERE standard_id = ?1     -- src/storage/mod.rs:23-24
```

with **no namespace predicate**. It exists as the #1642 fix — a delete must not leave a dangling `namespace_meta.standard_id` pointing at a reaped memory. Correct in intent.

But `set_namespace_standard` (`src/storage/mod.rs:16736`) **permits cross-namespace binding** — its own comment says *"allow cross-namespace — shared policy"*. So the standard a namespace points at need not live in that namespace. Deleting an in-scope row therefore clears the binding of every namespace that pointed at it, wherever they live.

And `resolve_governance_policy` (`:17021-17043`) **fails open**: no policy found ⇒ `GovernanceDecision::Allow`.

## Probe results

Peer scoped `public/*`, deleting a `public/ok` row — the gate correctly returns ALLOW:

```
X8a bind=secure/ops       BEFORE resolve_governance_policy(secure/ops)      = Some(policy)
                          AFTER  deleted=1 skipped=0 namespace_meta_cleared=0  ->  None
X8b bind=*  (global)      BEFORE resolve_governance_policy(secure/ops)      = Some(policy)
                          AFTER  deleted=1 skipped=0 namespace_meta_cleared=0  ->  None
X8c bind=secure           BEFORE resolve_governance_policy(secure/ops/deep) = Some(policy)
                          AFTER  deleted=1 skipped=0 namespace_meta_cleared=0  ->  None
```

[...]

Issue #1823, opened 2026-06-28
[TRACT-gap G6] Append-only broken at write AND erase — the spine break (in-place UPDATE + hard-DELETE)


alphaonedev:
## Gap G6 — TRACT §26 canonical-gaps catalog (UNTRACKED → tracked per v0.8.1 W7)

**Priority tier:** P1  ·  **Milestone:** v0.9  ·  **TRACT ref:** L1 append-only frozen-core / signed tombstone leaves

### Gap statement
The substrate mutates rows in place and hard-DELETEs on forget + federation delete, so the record is **not append-only**: prior states are destroyed, not superseded. This is the spine break that gates every "append-only / no silent delete / immutable record" claim. (Note: v0.8.1 W2/G30 builds the *minimal* signed FORGET tombstone primitive that G6.2 reuses; this issue is the full append-only refactor.)

### Anchor(s) (CodeGraph-verified on `release/v0.8.0`)
- `storage::update` + `ON CONFLICT DO UPDATE` is the default write path
- `EditSource::Human` in-place mutation — `src/models/memory.rs:817`
- `storage::forget` hard-DELETEs — `src/storage/mod.rs:2850`
- federation hard-deletes (CALL site) — `src/handlers/federation_receive.rs:1184`

### Fix shape (from the canonical doc)
- **G6.1** SUPERSEDE-not-UPDATE default (new row + RELATE edge, old row retained)
- **G6.2** signed FORGET tombstone leaf (the same primitive v0.8.1 W2 lands minimally)
- **G6.3** no hard-DELETE on federation / autonomy paths
- **G6.4** typed forgetting operators — a closed verb set **EXPIRE** (TTL-natural, tombstoned) / **EVICT** (capacity GC, recoverable from archive) / **REDACT** (content-scrub, row+lineage retained) / **DISTILL** (superseded-into-summary, RELATE-linked), each emitting a distinct signed leaf

### Provenance
Filed per the **v0.8.1 (Patch 1)** work order §W7 and the §24 prime directive (every gap is a tracked defect — no UNTRACKED gap may rot in a queue; the v0.9.0 epic starts on a clean tracker). Sources: `docs/design/TRACT-v0.8.0-DEVELOPMENT-GAPS-CANONICAL-opus.md`, `docs/reviews/corpus-completeness-21-agent-OPUS.md`, ROADMAP §26. Orchestration hub: #1821.

🤖 Filed by Claude Code (Opus 4.8, 1M context).

[...]

Issue #1628, opened 2026-06-11
HTTP PUT /memories/{id} on postgres silently ignores If-Match — Gap-1 optimistic concurrency inert on the postgres backend


alphaonedev:
## Root cause (one-liner)
`src/handlers/memories.rs:213-221` parses if_match_version; the postgres branch (:229-303) calls `app.store.update(&ctx,&id,patch)` and never references it (only the sqlite branch uses it, :407). `PostgresStore::update_with_expected_version` (postgres.rs:2580) — the method that honors it — has zero production callers. The comment at postgres.rs:9073-9076 documents an intended handler split that does not exist.

## Trigger
pg daemon, row at version=2; `PUT ... If-Match: 1` → sqlite: 409 VersionConflict; postgres: 200 + write applied (lost update).

## Violated requirement
Memory::version doc (src/models/memory.rs:572-579): "Two callers writing against the same expected_version race exactly one winner; the loser receives a typed CONFLICT envelope."

## Proposed fix (size: S)
pg branch: when if_match_version.is_some(), route to update_with_expected_version (also un-deadcodes that method). Cross-backend If-Match parity test.

Found by the Class-C deep-audit reviewer. Severity: HIGH.

alphaonedev:
Fixed via commits df9cf447 (pg PUT branch routes through update_with_expected_version when If-Match present; 409 CONFLICT envelope byte-equal to sqlite; the method gains the #1412 caller-owns gate) and 37401500 (gate hoisted into shared assert_caller_owns_for_mutation with consts per pm-v3.1). Retest: pg_fix3_parity_tests If-Match cases RAN against live pg (stale If-Match → 409; correct → 200 + version bump). Battery b4 green. Verified per pm-v3.3.

PR #3115, opened 2026-08-22
fix(parity): close four sqlite/postgres write-funnel divergences (archive reason, batch + delete atomicity, SAL caller-owns gate)

alphaonedev:
Closes four sqlite/postgres **write-funnel** divergences found by a parity audit of `store`/`update`/`upsert`/`delete`/`archive`. Three are data-integrity issues; one is an authorization gap. (`Closes` will not auto-fire on a `release/**` base.)

Every finding was **verified against the real code before fixing** — see the confirmed/refuted table.

## Findings

| # | Divergence | Verdict | Impact |
|---|---|---|---|
| 1 | `archive_reason` reason-less default: sqlite `'archive'` vs postgres `'manual'` | **CONFIRMED** | wrong-results (audit-trail / report drift) |
| 2 | `archive_by_ids`: sqlite per-id tx vs postgres one batch tx | **CONFIRMED** | data-loss (partial batch committed) |
| 3 | `delete`: sever + tombstone-leaf + delete as 3 autocommit statements | **CONFIRMED** | data-loss (stranded governance sever) |
| 4 | SAL caller-owns gate present only on postgres `update`/`delete` | **CONFIRMED** | cross-tenant write via direct SAL usage |
| 5 | `archive_restore` erasure-bundle path only on sqlite | **DEFERRED** (capability gap, security-neutral — as scoped) |

### 1 — one SSOT const for the reason-less archive default

sqlite stamped `'archive'` (**both** funnels: `archive_memory_no_tx` *and* `archive_memory_for_caller`); postgres `archive_by_ids` stamped `'manual'`. Identical operations produced different audit trails, and every reason-filtered query / `archive_stats` report disagreed across backends.

Corroboration that `"archive"` is canonical: the **HTTP archive handler has always defaulted to `"archive"`** and passes it explicitly to both backends — so postgres' own SAL default *contradicted the daemon's own contract* and was reachable only by a direct SAL caller. All four sites now read `models::field_names::ARCHIVE_REASON_DEFAULT`.

### 2 — `SqliteStore::archive_by_ids` is now ALL-OR-NOTHING

postgres wraps the batch in one `pool.begin()`. sqlite looped `db::archive_memory`, which opens its **own** `BEGIN IMMEDIATE` **per id** — so a mid-batch failure left a **partially archived batch committed**: the prefix of ids had their live rows deleted out of `memories` and moved to `archived_memories` while the remainder stayed live, and the caller got only an `Err` with no way to learn how far it got. Replay/DR tooling could not assume the same post-failure state machine on the two backends.

### 3 — `storage::delete` is now statement-atomic

The namespace-standard SEVER (#2503), the append-only TOMBSTONE leaf (#1823 G6), and the row DELETE ran as **three separate autocommit statements**. A failure between them stranded a **severed governance binding** — an unrecoverable policy downgrade — with the memory **still live**, i.e. caused by a delete that never happened. Body moved to a private `delete_inner` (the established `archive_memory` / `archive_memory_no_tx` split).

### 4 — sqlite `update`/`delete` gain the SAL caller-owns gate

`PostgresStore` enforces `assert_caller_owns_for_mutation` on both verbs (#1412/#1628); the sqlite adapter took `_ctx` and never inspected it. The check lived **only** in the sqlite HTTP handler, so any non-HTTP SAL caller (CLI, internal surfaces, federation, future paths) could rewrite or delete another tenant's row by id while postgres refused — exactly the fragility class #1412 closed on postgres. The two wire-pinned refusal strings are hoisted to `crate::store` (one declaration site, pm-v3.1) so both backends emit byte-identical errors.

**Behaviour change, called out deliberately:** a row with **no `metadata.agent_id` stamp** (legacy / pre-v0.6.3 / migrated) is now refused for *tenant* mutation on sqlite and must be cleaned up via the admin path — the same conservative fail-closed posture postgres has had since #1628.

**Blast radius audited:** every `MemoryStore::update`/`delete` caller already satisfies the identical postgres gate today (handlers are backend-agnostic; postgres daemons run them). The auto-tag worker deliberately runs as `CallerContext::for_agent(job.agent_id)` — the **original writer**, so it owns the row — and the hook-subscriber bind paths use `CallerContext::for_admin` (bypass).

## Transaction-awareness (why this can't regress)

Both atomicity fixes use `conn.is_autocommit()` — the existing `update_with_expected_version` precedent, adopted there because a nested `BEGIN` fails with *"cannot start a transaction within a transaction"*. This matters: **`delete` IS called from inside an open transaction** by `consolidate`'s legacy hard-DELETE arm. Opening our own tx only when none is active is structurally fail-safe — it cannot produce a nested BEGIN no matter which caller appears later — and when the caller owns the tx these writes join it, which is exactly the atomicity that caller already wanted.

## Tests — `tests/parity_write_funnels.rs` (7 sqlite + 1 live-pg)

- reason-less archive stamps the shared const — **sqlite and a live-pg twin**, since fix 1 changed the *postgres* side;
- a trigger-injected **mid-batch failure leaves the first id still live and unarchived** (pre-fix it was already committed);
- a successful batch commits every id;
- a trigger-injected DELETE failure **rolls back the namespace-standard sever** (pre-fix the sever committed on its own);
- `delete` inside a caller-held transaction does not nested-BEGIN and is undone by the caller's ROLLBACK (guard for the fix itself);
- bob cannot update or delete alice's row; alice can; an admin bypass can;
- an unstamped row refuses tenant mutation with the shared wire text and still yields to the admin path.

## Rust standard (`rust-1.98`)

`CONCURRENCY-04` (atomicity/ordering — the point of 2 and 3), `CONCURRENCY-24` (the sqlite gate is a plain `fn`: it has no `.await`, and `rusqlite::Connection` is `!Sync`, so an `async fn` holding `&Connection` yields a non-`Send` future `#[async_trait]` rejects), `ERRORS-19` (no silently dropped `Result`; the best-effort ROLLBACK uses an explicit `let _ =`), `ERRORS-01`/`ERRORS-09` (typed `PermissionDenied`/`NotFound`, no panics), `OWNERSHIP-05`.

## Verification

- `cargo check --features sal-postgres --all-targets` — clean
- clippy CI-exact `-D warnings -D clippy::all -D clippy::pedantic --all-targets` — clean on **all four legs**: default, `sal`, `sal-postgres`, `vectorlite`
- `cargo fmt --all` — clean
- `parity_write_funnels` — 7/7 pass; `qual_10_module_size_ceiling` and `qual_6_7_legacy_error_type_ceiling` pass (no ceiling bumps needed)


alphaonedev:
**Fable final-gate review — APPROVED in substance; merge conditional on green required CI *including the postgres leg*.**

Verified: all four divergences CONFIRMED against real code with quoted evidence; #5 correctly deferred as scoped. Fix 1 (`'archive'` canonical, corroborated by the HTTP handler's own default — the pg side was the outlier); fix 2/3 use the `is_autocommit()` precedent so no nested-BEGIN, and a trigger-injected DELETE failure test proves the namespace-standard sever now rolls back; fix 4 mirrors `assert_caller_owns_for_mutation` with the admin/`bypass_visibility` path pinned by test (line 334). The called-out behaviour change (unstamped legacy rows refused for *tenant* mutation on sqlite) is consistent — such rows are already invisible to tenant reads (absent scope ⇒ private, no owner), and it matches pg today. Commit `1b45c2d1` SSH-signed.

Gate conditions (per the #3112 lesson): the live-pg twin test is URL-gated and was not evidenced as run locally — so this merges only once the **postgres CI leg** (Postgres feature gate + coverage pg tests) is green. Also requested from the implementer: explicit confirmation that the auto-tag worker / hook-subscriber contexts pass the new sqlite gate (the parity argument — they already satisfy the identical pg gate — is sound, but I want it stated).

alphaonedev:
## Fable's two requests — evidenced

### (1) Live-pg twin test, EXECUTED against the local certified tier

`export AI_MEMORY_TEST_POSTGRES_URL="$(cat .local-runs/ENTERPRISE-FED-TEST-PG-URL.txt)"` (PG 18.6 + AGE 1.8.0 + pgvector, TLS `verify-full`), then
`build-gate.sh cargo test --features sal-postgres --test parity_write_funnels -- --nocapture`:

```
running 8 tests
test pg::pg_archive_by_ids_reason_less_default_matches_the_sqlite_funnel ... ok
test delete_still_works_when_the_caller_already_holds_a_transaction ... ok
test delete_rolls_the_namespace_standard_sever_back_with_the_row ... ok
test archive_by_ids_commits_the_whole_batch_on_success ... ok
test sqlite_sal_mutation_of_an_unstamped_row_is_refused_for_tenants ... ok
test archive_by_ids_is_all_or_nothing_on_a_mid_batch_failure ... ok
test archive_by_ids_reason_less_default_is_the_shared_ssot_const ... ok
test sqlite_sal_update_and_delete_enforce_the_caller_owns_gate ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.02s
```

The `pg::` twin ran (it is **not** in the skip path — a skipped run prints `skip: AI_MEMORY_TEST_POSTGRES_URL not set` and no `pg::` line). So fix #1's *postgres* side — the backend that actually changed — is proven on a real server.

### (2) How each caller passes the new sqlite caller-owns gate

I went further than the two you named and audited **every** `MemoryStore::update`/`delete` call site. Result: **no production caller can be refused by the new gate.**

**auto-tag worker** — `src/background/auto_tag_worker.rs:328`, inside `async fn apply_tags_postgres`:

```rust
let ctx = crate::store::CallerContext::for_agent(job.agent_id.clone());
```

Owner context, and deliberately so. The module doc (lines 49-58) states it outright:

> The read+apply runs under `CallerContext::for_agent(job.agent_id)` — the ORIGINAL WRITER's own identity (threaded through `AutoTagJob` from the enqueue site), not an admin/privacy-bypass context. […] so `for_admin`'s `bypass_visibility` was unwarranted privilege

So the worker mutates the row **the same agent authored** → `existing_owner == ctx.effective_principal()` → gate passes. (It is also inside the `StorageBackend::Postgres` branch at line 352, so it never reaches `SqliteStore` anyway.)

**hook_subscribers** — the `app.store.update(&ctx, &standard_id, patch)` at line 618 uses the tenant caller's own ctx (line 354):

```rust
let ctx = if let Some(h) = headers {
    crate::handlers::parity::http_caller_ctx(h, None)
} else {
    crate::store::CallerContext::for_agent(&caller)
};
```

Not a bypass — but it is **unreachable from sqlite**: that update sits inside the `#[cfg(feature = "sal")] if matches!(app.storage_backend, StorageBackend::Postgres) {` branch opened at **line 343**, so it only ever dispatches to `PostgresStore::update`, which has enforced the identical gate since #1412. It is additionally preceded by `crate::mcp::authorize_namespace_standard_bind` (line 513), the #929/#2542 ownership gate.

**Full sweep — nearest enclosing postgres guard for every other site:**

| call site | postgres-gated |
|---|---|
| `handlers/memories.rs:387`, `:392` (update) | yes — guard at 269 |
| `handlers/memories.rs:956` (delete) | yes — guard at 874 |
| `handlers/memories.rs:1420` (update) | yes — guard at 1312 |
| `handlers/subscriptions.rs:674` (delete) | yes — guard at 619 |
| `handlers/hook_subscribers.rs:618` (update) | yes — guard at 343 |
| `background/auto_tag_worker.rs:379` (update) | yes — guard at 352 |
| `autonomy.rs:1060` (delete) | **no guard — backend-agnostic** |

The single backend-agnostic path is `autonomy::reverse_rollback_entry_store`, and both of its real callers pass an **admin** context:

```rust
// src/cli/curator.rs:600 and :1247
let ctx = CallerContext::for_admin(crate::identity::sentinels::AI_CURATOR);
```

`CallerContext::for_admin` sets `bypass_visibility: true` (`src/store/mod.rs:564`), and the gate's first statement is `if ctx.bypass_visibility { return Ok(()); }` — so it short-circuits.

**Conclusion:** today `SqliteStore::update`/`delete` have **no production caller the gate can refuse**. Finding #4 is therefore pure defense-in-depth: it closes the *future/direct-SAL-caller* hole (the #1412 fragility class) with no behavioural change to any shipping path. The one behaviour change that *could* bite an operator remains the documented one — a legacy row with **no `metadata.agent_id` stamp** is refused for tenant mutation and must be cleaned up via the admin path, matching postgres since #1628.

alphaonedev:
**Fable gate — HOLD (real regression signal, not a fixture collision).** Required `Per-Module Coverage Thresholds` on `1b45c2d`: `handler_postgres_branches_fake_pg` → `pg_update_memory_happy` (:374) and `pg_delete_memory_after_create` (:500) FAIL with `caller "anonymous:req-…" does not own memory (owner: "anonymous:req-…")` → 403 where 200 was expected. Requests without `X-Agent-Id` get a per-request anonymous identity, and the gate now refuses anonymous→anonymous mutation. These tests pass on base, so the PR changed pg/handler behaviour on the **default no-identity posture** — the dev daily-driver. Sent back to the implementer: bisect the PR's own diff against these tests and restore the pre-existing contract exactly; if tightening the default posture is believed correct, that is a separate #2046-class decision, not a parity PR. Substance approval stands; merge blocked until this is resolved + the live-pg twin is evidenced.

alphaonedev:
## fake-pg red — root cause, and the posture question I'm escalating

Rebased onto `15fe44e4`; new head **`5bd91ae5`** (both commits `%G?` = `G`, base is an ancestor, CHANGELOG keeps both sides).

### Reproduced locally first

`build-gate.sh cargo test --features sal-postgres --test handler_postgres_branches_fake_pg pg_update_memory_happy pg_delete_memory_after_create` → reproduced (`EXIT=1`), same panic.

### Root cause: the fixture, and a harness fidelity gap it was hiding

`handler_postgres_branches_fake_pg` deliberately wires **`storage_backend = Postgres` with an `SqliteStore`** as the `dyn MemoryStore` (its own module doc says so), to drive the postgres handler branches. The pg handler branch does **not** call `handlers::parity::require_caller_owns_memory` — it delegates to `app.store.update`/`.delete` and relies on the **adapter's** gate.

- Pre-change `SqliteStore` had no gate → 200.
- A **real `PostgresStore` would have returned 403** for the identical requests: `assert_caller_owns_for_mutation` has enforced this since **#1412/#1628** and has **no anonymous carve-out**.

So the two tests asserted behaviour the backend they stand in for does not have. Adding the sqlite gate made the harness faithful and surfaced it.

Why the two principals differ: both requests are sent with **no `X-Agent-Id`**, and `identity::resolve_http_agent_id` synthesizes a **fresh `anonymous:req-<uuid>` per request** (and logs a WARN). The POST stamps `metadata.agent_id` with one principal; the later PUT/DELETE arrives as another. A header-less caller has no stable identity *by design*.

### Did I tighten the default no-identity posture? No — evidence

| path | anonymous→anonymous mutate | changed by this PR? |
|---|---|---|
| real **postgres** | already **403** (`assert_caller_owns_for_mutation`, #1412) | **no** |
| real **sqlite HTTP** | already **403** (`require_caller_owns_memory`, `memories.rs:569` — owner non-empty ≠ caller) | **no** |
| **fake-pg harness** | 200 (adapter had no gate) | yes — now matches real pg |

The contract the codebase actually encodes is *"supply a stable `X-Agent-Id` and own the row"* — the real sqlite HTTP smoke test passes `Some("ai:smoke-agent")` for **both** the create and the update (`route_put`, `tests/integration.rs`). The anonymous→anonymous flow is not a tested contract on either real backend.

### Fix — the file's own idiom, assertions unchanged

The suite already carries `post_json_as` / `get_uri_as` / `delete_uri_as` / `put_json_as`, added because "handlers … now require the `X-Agent-Id` header for authentication". Both tests now create **and** mutate under one stable identity. **The assertions still expect 200**, so the happy-path branches they exist to cover are still driven end to end — I did not loosen them. `put_json_as` gains a doc comment describing the per-request-anonymous trap.

`cargo test --features sal-postgres --test handler_postgres_branches_fake_pg` → **90 passed; 0 failed** (was 88/2).

---

## ESCALATION — one genuine posture decision for you, not shipped silently

There **is** one place where my sqlite gate is stricter than sqlite's own pre-existing contract, and per your rule I'm surfacing it rather than deciding it:

- Pre-existing sqlite HTTP gate `require_caller_owns_memory` has a documented **legacy-unowned carve-out**: `owner.is_empty()` → **allowed**.
- My new sqlite SAL gate mirrors **postgres** (#1628): an **unstamped** row (no `metadata.agent_id`) → **refused** for tenant mutation.

So on the unstamped arm, sqlite-SAL now matches postgres rather than sqlite-HTTP. It is **behaviourally inert today** — as documented in the PR body and re-verified this pass, `SqliteStore::update`/`delete` have no production caller that can reach the gate (every HTTP/worker site sits inside a `StorageBackend::Postgres` branch; the one backend-agnostic path, `autonomy::reverse_rollback_entry_store`, is driven only by `CallerContext::for_admin`, which sets `bypass_visibility`). My test `sqlite_sal_mutation_of_an_unstamped_row_is_refused_for_tenants` pins it deliberately.

**Options:**
- **(A) keep as-is** — full postgres parity, fail-closed, which is what parity finding #4 asked for; or
- **(B) mirror sqlite-HTTP's `owner.is_empty()` carve-out** — preserves sqlite's own prior contract exactly, at the cost of leaving that one arm non-parity.

I've kept **(A)** because it is the requested parity and is fail-closed, and because it changes no shipping behaviour — but it's your call and I'll switch to (B) on your word. Flagging it as the #2046 secure-by-default class you asked me to escalate.

### Local verification on the rebased head
- `cargo fmt --all` clean
- clippy CI-exact `-D warnings -D clippy::all -D clippy::pedantic --all-targets` → green on **`sal`** and **`sal-postgres`** (the only legs that compile the two changed test files; both are `#![cfg(feature = "sal")]`, so `default`/`vectorlite` are unaffected and were green on the identical source last pass)
- `parity_write_funnels` + `handler_postgres_branches_fake_pg` → all pass
- live-pg tier (`AI_MEMORY_TEST_POSTGRES_URL`, PG 18.6 + AGE 1.8.0 + pgvector, TLS verify-full): `parity_write_funnels` **8 passed / 0 failed**, including the `pg::` twin

alphaonedev:
**Fable gate — escalation decided: Option B (mirror sqlite's own `owner.is_empty()` carve-out at the SAL gate).**

Rationale: a parity PR restores each backend's pre-existing contract; refusing unstamped rows on sqlite is a posture tightening (and turns today-writable legacy rows into inaccessible ones — a data-loss mode under the prime directive). Independent 3-reviewer panel: 3-0 for B. Cross-backend unification (stamp legacy rows via migration, then refuse everywhere) is now **#3124** — cite it in the gate comment.

Also accepted: the anonymous→anonymous regression was a fake-pg fixture artefact (already refused on both real backends); no default-posture change shipped. HOLD lifted pending the B switch + green required legs; final diff review then.

alphaonedev:
**Queue head.** #3116 merged. Rebase onto current `origin/release/v1.0.0`, apply remaining Fable-gate items, one `--force-with-lease` push. `ci-allow.txt` is this branch.

alphaonedev:
## Queue-head rebase — landed

Rebased onto `origin/release/v1.0.0` = **`060c668e`** (Merge #3116). New head **`39c0b8d3`**. All five commits `%G?` = `G`. CHANGELOG keeps both sides. `ci-allow.txt` untouched.

### Remaining Fable-gate items (already in the replayed commits; not reversed)

- **Option B** (`7e495768`): sqlite SAL caller-owns gate delegates to `visibility::caller_owns_for_mutation` — unstamped rows stay mutable (sqlite's own HTTP/MCP carve-out). `#3124` cited in the gate docblock, CHANGELOG, and the carve-out test. Unification is a FOLLOW-UP.
- **fake-pg `X-Agent-Id` pin** (`a5fec1a4`): `pg_update_memory_happy` / `pg_delete_memory_after_create` create-and-mutate under one stable identity.
- **Gate-doc scope** (`921547d2`): SAL gate covers `MemoryStore` trait callers only.

### QUAL-10 re-measure after rebase

| module | measured | ceiling | action |
|---|---|---|---|
| `src/store/postgres.rs` | **34_692** | 34_720 (#3112) | under; no bump |
| `src/storage/mod.rs` | **29_196** | was 29_160 | **bump 29_160 → 29_260** (+64 headroom) in `39c0b8d3` |

### CI-parity (PATH-pinned 1.96.0, one `build-gate.sh` admission)

- `cargo fmt --all -- --check` green
- `cargo check --all-targets` default / `sal` / `sal-postgres` + `--examples` green
- clippy `-D warnings -D clippy::all -D clippy::pedantic --all-targets` × 5: default, `sal`, `sal-postgres`, `vectorlite`, `--examples` green
- `qual_10_module_size_ceiling` 2 passed (after the bump)
- `qual_6_7_legacy_error_type_ceiling` 2 passed
- `parity_write_funnels` (sqlite, `--features sal`) **8 passed**: `delete_still_works_when_the_caller_already_holds_a_transaction`, `delete_rolls_the_namespace_standard_sever_back_with_the_row`, `unstamped_row_is_allowed_through_sqlite_sal_gate`, `archive_by_ids_reason_less_default_is_the_shared_ssot_const`, `sqlite_sal_inbox_recipient_may_delete_but_not_update`, `archive_by_ids_is_all_or_nothing_on_a_mid_batch_failure`, `sqlite_sal_update_and_delete_enforce_the_caller_owns_gate`, `archive_by_ids_commits_the_whole_batch_on_success`
- `handler_postgres_branches_fake_pg` **90 passed; 0 failed** (includes `pg_update_memory_happy`, `pg_delete_memory_after_create`)
- `scripts/check-hardcoded-literals.sh` PASS
- `scripts/check-docs-vs-ssot.sh` PASS (schema=89)

### Live-pg twin — SchemaAheadOfBinary (hatch NOT set)

Shared enterprise-fed PG is **schema v90** (PR #3216). This binary is still `CURRENT_SCHEMA_VERSION = 89`. `pg::pg_archive_by_ids_reason_less_default_matches_the_sqlite_funnel` failed closed:

```
connect postgres: SchemaAheadOfBinary { detail: "… is on schema v90, but ai-memory 1.0.0 understands up to v89. Refusing to operate it …" }
```

**`AI_MEMORY_ALLOW_SCHEMA_AHEAD` was NOT set** (fail-closed; a hatch here would let an older binary write a newer schema). CI isolated PG is the authority for the postgres leg. sqlite + fake-pg coverage of the four funnels is green locally.

alphaonedev:
**MERGED** via `fable-merge.sh`. Head `39c0b8d3` 34/34 required-green. Option B (#3124 carve-out) kept. Next queue head: **#3114**.

Issue #2887, opened 2026-08-11
[data-integrity] autonomy/curator guarded-restore paths have a narrow non-atomic probe→restore window (needs a restore-safe atomic primitive)


alphaonedev:
## Root cause
Surfaced during #2878 (MCP/import lost-update fix). The autonomy `reverse_rollback_entry_store` (`src/autonomy.rs:1013`) and curator `rollback_consolidation` (`src/curator/compaction.rs:599`) are guarded RESTORE paths: probe `find_by_title_namespace` → refuse if a *different* id owns the key → else `store.store` restore-by-id. Between the probe and the restore there is a narrow non-atomic window.

## Why the #2771 primitive does NOT fit (deliberately not force-fit)
`db::insert_no_overwrite` / `store_with_embedding_no_overwrite` refuse ANY `(title,namespace)` collision — including the LEGITIMATE same-id idempotent restore these paths perform. Applying it would regress restore availability. So the #2878 fix correctly did NOT extend the create no-overwrite primitive here.

## Scope / severity (honest)
Narrow: single curator daemon / serial `curator --rollback`; low concurrency; already documented (guardrail G4 'EXACT PARITY with the rusqlite twin'). It is NOT the create error/no-overwrite lost-update class (#2771/#2874/#2878).

## Proposed fix
A NEW restore-safe atomic primitive: an INSERT-or-update-ONLY-IF-same-id CAS (e.g. `ON CONFLICT (title,namespace) DO UPDATE ... WHERE memories.id = excluded.id`, else refuse) so a same-id restore is atomic while a different-id collision is still refused — a T1 SAL trait-shape change requiring its own 5-agent vote. Regression test: concurrent restore vs foreign-id write.

## Related
#2878 (MCP/import lost-update, this is the flagged residual); #2771/#2874. Filed per prime directive.

alphaonedev:
[...]

Issue #1774, opened 2026-06-23
Consolidation: un-embedded rows merge on Jaccard-only (cosine safety-gate bypassed) → false-positive merge can destroy a distinct memory's links


alphaonedev:
**Auto-filed by the 2026-06-22 destructive-ops audit** (Agent 2).

## Defect
The consolidation clusterer's pair gate (`src/curator/cluster.rs:72-85` `pair_merges`) is a dual AND-gate: Jaccard pre-filter AND cosine gate — BUT only "when an embedding is available". For a row with **no stored embedding** (keyword-tier / never-embedded / oversize-skip), `cos = None` ⇒ `pair_merges` returns true on the **Jaccard signal alone** (≥ `JACCARD_THRESHOLD` 0.55). On a keyword-tier or un-embedded corpus the cosine safety-gate is therefore **fully bypassed**, and merges happen on lexical overlap alone — a materially higher false-positive-merge surface exactly where the safety gate is absent. Combined with the cascade-link-loss issue (sibling), a wrong Jaccard-only merge **silently destroys the merged-away memory's links**.

## Proposed fix
- Under `AI_MEMORY_COMPACTION_ENABLED=true`, require a stored embedding for both sides before merging (skip un-embedded pairs), OR raise the Jaccard-only threshold materially above the cosine-backed one, OR emit a one-shot operator WARN documenting that un-embedded corpora consolidate on lexical overlap alone. `src/curator/cluster.rs`.

Severity: MEDIUM (opt-in #1749 path; default-OFF, but autonomy Pass-1 shares the cluster contract). Evidence: audit mem 88e8050d.

alphaonedev:
Fixed via e7538167 on release/v0.8.0 — 8/8 CI green (CI 27m35s incl Postgres-gate, Coverage 30m25s). 5-agent vote (4d3ea1c5 → A).

Consolidation now requires a stored embedding on BOTH sides — `pair_merges` (src/curator/cluster.rs) `None => false`, and the default-ON autonomy Pass-1 mirror (src/autonomy.rs find_consolidation_clusters) changed to match. Pre-fix, un-embedded pairs (keyword-tier/never-embedded/oversize) merged on the Jaccard lexical pre-filter ALONE, bypassing the cosine safety gate; since consolidation is destructive (merge+delete sources), a false-positive lexical merge silently destroyed a distinct memory + its links.

[...]

Issue #816, opened 2026-05-16
FEAT: curator daemon — wire auto_persona post-reflect hook with daemon keypair (deferred from #811)


alphaonedev:
## Symptom

The curator daemon (`curator --daemon --interval-secs 300`) does not auto-generate personas from new reflections. Operators must call `memory_persona_generate` explicitly for every entity that needs a persona row. This is the deferred-scope follow-up flagged in #811 and commit f928980.

## Evidence

Cold-boot smoke against the post-#814 binary on 2026-05-16 (curator PID 31961, ai-memory 0.7.0):

A fresh entity `e8667693-2228-4d10-9780-051f8f43ca4f` (namespace `scratch/curator-gap-smoke-2026-05-16`) was created with one observation and one reflection that explicitly `reflects_on` the entity. The reflection was deliberately left in its un-curated state (no manual `mentioned_entity_id` patch) so the curator pass must do the entity-tagging AND persona-gen itself.

After one full curator cycle (interval=300s) elapsed past the reflection's creation:

```
sqlite> SELECT id, memory_kind, persona_version, source FROM memories
        WHERE entity_id='e8667693-...' OR mentioned_entity_id='e8667693-...';
(no rows)
```

No persona row was created.

The most recent curator self-report (`_curator/reports`):

```json
{
  \"auto_tagged\": 8,
  \"clusters_formed\": 0,
  \"contradictions_found\": 2,
  \"cycle_duration_ms\": 1297845,
  \"cycle_ts\": \"2026-05-16T21:43:40.659891+00:00\",
  \"errors_total\": 17,
  \"memories_consolidated\": 0,
  \"memories_forgotten\": 0,
  \"priority_adjustments\": 0,
  \"rollback_entries_written\": 0
}
```

No `personas_generated` counter exists. The curator's report schema confirms it doesn't track persona work — because it doesn't do any.

## Root cause

`src/curator/mod.rs::run_once` (lines 135-260) calls auto_tag → detect_contradiction → `autonomy::run_autonomy_passes`. None of these:
- Invoke `memory_persona_generate` for entities mentioned in newly-curated reflections.
- Install a `post_reflect` callback on the storage::reflect hook surface.

[...]

PR #3112, opened 2026-08-22
fix(security): enforce the SAL #910 scope=private gate on the pg reciprocal-provenance search (#3110)

alphaonedev:
Fixes the cross-agent private-row leak in `PostgresStore::search_with_source_uri` reported as #3110. (`Closes` will not auto-fire on a `release/**` base — please close #3110 manually on merge.)

## Confirmed against the real code

Verified before fixing (R-405), not taken on the report's word:

| | sqlite twin | postgres (pre-fix) |
|---|---|---|
| `search_with_source_uri` | fail-closed since **#1720 A3** — owner-keyed `visibility_clause(11,15,"m")`, caller bound at `?15`, documented *"None = fail-closed (no private rows)"* | **NO scope/caller predicate at all** |
| trait `search` | gated | gated since **#942** (`$7` caller ladder + in-process `is_visible_to_caller`) |

So the postgres reciprocal-provenance surface returned full `Memory` payloads with no visibility filter, while both of its siblings were gated. **CONFIRMED.**

### Reachability — stated plainly

The method is `pub` but has **no production caller today**: the postgres HTTP search path dispatches through the SAL trait `search` (which *is* gated) and drops `source_uri` entirely, so the only current callers are tests. This closes a **latent** leak and restores the fail-closed parity contract before a caller gets wired — it is **not** a fix for an actively exploited route. Flagging that explicitly so the severity is not overstated.

## The fix — mirror the pg `search` gate exactly

- Thread `ctx: &CallerContext` through the signature and every call site.
- Push the owner-keyed predicate into SQL as `$7` (`LIMIT` renumbered `$7` to `$8`):

```sql
AND ( $7::text IS NULL
      OR COALESCE(m.metadata->>'scope','private') <> 'private'
      OR m.metadata->>'agent_id'        = $7
      OR m.metadata->>'target_agent_id' = $7 )
```

Absent `scope` defaults to `private` (fail-closed); the owner is exempt; `target_agent_id` is the **#3070** inbox carve-out so a row addressed *to* the caller stays visible.

- Derive the bind with the canonical `caller_opt` ladder (`ctx.bypass_visibility` gives `NULL`), keeping admin / migrate / federation-catchup / GC as the one sanctioned bypass.
- Re-apply `is_visible_to_caller` **in-process** after the fetch — byte-identical to what pg `search` already does — so the SQL predicate can never drift from the canonical `crate::visibility` semantics.

## Regression test

[...]

PR #3116, opened 2026-08-22
fix(curator): autonomy/curator correctness — namespace cfg, op-cap, silent-error surfacing

### LINKED ISSUE #2941: agents bind-key: intermittent silent no-op (register row created but metadata.agent_pubkey unset → signed store 403s)

alphaonedev:
## Summary

Fixes nine defects found by a rust-1.98-grounded code review of `src/curator/*` and `src/autonomy.rs` (ox-alpha `review-curator-autonomy`, 13 reported items). Every item was **re-verified against the code at `release/v1.0.0` before being fixed** — none were taken on the report's word.

Every fix is the same class of problem: the curator daemon either **ignored operator configuration**, **exceeded its documented budget**, or **self-reported work it did not do**. All three make the fleet less manageable and the cycle report less truthful, which is exactly what the North Star's "manageable at scale / degrade-never-corrupt" constraint rules out.

Closes #2941

## Verification verdict per reported item

| # | Item | Verdict |
|---|---|---|
| 1 | `persona_sweep` ignores include/exclude namespace config | **CONFIRMED** — fixed |
| 2 | Autonomy passes run uncapped vs `max_ops_per_cycle` | **CONFIRMED** — fixed |
| 3 | `adjacent_memory` `Err` silently dropped | **CONFIRMED** — fixed |
| 4 | `persist_auto_tags`/`persist_contradiction` silent no-op counted as success | **CONFIRMED** — fixed |
| 5 | Embedding-read failures indistinguishable from "no embedding" | **CONFIRMED** — fixed (both clusterers) |
| 6 | Dry-run inflates `rollback_entries_written` | **CONFIRMED** — fixed |
| 7 | Destructive action proceeds after rollback-log write fails | **CONFIRMED** — fixed |
| 8 | `block_on` in `pub run_once` panics from an async runtime | **CONFIRMED — and reachable in production**, see below |
| 9 | Conserve edge signed with an arbitrary key | **CONFIRMED** — fixed |
| 10 | `usize as i64` LIMIT cast (PERF-07) | **CONFIRMED** — fixed |
| 11 | Collision probe reads only 50 rows | Confirmed real, **out of this batch's scope**; additionally mitigated by the load-bearing `insert_restore_same_id` CAS (#2887) |
| 12 | `large_enum_variant` suppressed instead of boxed | Confirmed real, **out of this batch's scope** (touches every `RollbackEntry` match site + the serialized rollback log) |
| 13 | Stale `#[allow(unused_assignments)]` on the Pass-3 loop | **CONFIRMED as a stale suppression** — removed; the build is clean under `-D warnings`, so it was hiding nothing |

### HIGH — `persona_sweep` ignored the operator's namespace configuration
The sweep's scan filtered on the reserved `_`-prefix only, so `PersonaGenerator` ran — and wrote **signed** persona rows — in namespaces the operator had excluded, or outside a non-empty `include_namespaces`, while `run_size_gc_pass` **in the same file** filtered correctly.

Both lists are now pushed into the scan SQL as **bound parameters** (so out-of-scope rows cannot consume the `LIMIT` window and starve in-scope entities in a busy corpus) and re-checked per row as a fail-closed backstop. `needs_curation`, the size-GC pass, and the persona sweep now all route through **one** `namespace_in_scope` predicate, so scope can never be honoured by one pass and ignored by another.

### HIGH — autonomy passes ran outside the documented per-cycle op cap
`max_ops_per_cycle` is documented as a *"Hard cap on LLM-invoking operations per cycle"*, but `run_autonomy_passes` had no budget check anywhere: Pass 1 issued one `summarize_memories` call per cluster over a candidate batch capped at `max_ops_per_cycle * 4`, so a single cycle could make several times the authorised number of LLM invocations.

`run_once` now threads its **remaining** budget in. Over-budget clusters are deferred to the next cycle and counted in `operations_skipped_cap` — deferred, never dropped — and the ops spent fold back into the cycle counters so the persona sweep downstream sees a true remaining budget. Passes 2 and 3 invoke **no** LLM and stay bounded by the candidate batch, so they are deliberately left uncharged; charging them would silently cut priority-feedback throughput without any contract requiring it.

### MED — a destructive action proceeded after its rollback-log write failed
The corpus was already mutated, the only reversibility record had failed to persist, and the cycle carried straight on issuing further irreversible actions with one line in `report.errors` as the sole trace. The passes now **halt** for the rest of the cycle and set a new `rollback_log_degraded` flag that surfaces in the self-report JSON. Fail-closed: fewer actions, never un-reversible ones.

### MED — silent no-op counted as success in the persist helpers
A row whose `metadata` column holds non-object JSON made both helpers write the metadata back **unchanged**, return `Ok(())`, and let `run_once` increment `auto_tagged` / `contradictions_found`. A curator that claims work it did not do is worse than one that refuses: they now return an error naming the row and the offending JSON shape (ERRORS-19).

### MED — `adjacent_memory`'s `Err` half discarded
`if let Ok(Some(..))` dropped the error with no report entry, so a transient DB failure silently skipped contradiction detection while the cycle report looked clean. Now recorded in `report.errors` (ERRORS-19).

### MED — `block_on` reachable from a public API panicked (reachable in production)
The SAL consolidation pass builds its own current-thread runtime; reached from a thread with an ambient tokio runtime, that `block_on` panics with *"Cannot start a runtime from within a runtime"* — and only a doc comment stood in the way of a `pub fn`.

**This was not theoretical:** `src/cli/curator.rs` called `curator::run_once` directly from an `async fn`, so `ai-memory curator --once` with `compaction.enabled` would panic. (Both daemon entry points in `daemon_runtime.rs` already used `spawn_blocking`, which is why it had not been hit.) Two fixes: the pass now detects the in-runtime case and degrades to a **reported skip** (ERRORS-08 / CONCURRENCY-22 — the corpus is untouched either way, but a reported skip is recoverable and a panic is not), and the CLI now runs the sweep under `spawn_blocking` so it actually works.

### MED — conserved `contradicts` edge signed with an arbitrary key
`forget_if_superseded` loaded *"the lexicographically-first key under the active key dir"* ambiently, so on a multi-key host the edge could be attested to an identity that is **not** the one the daemon runs as — muddying exactly the #816 attestation trail. The authenticated `active_keypair` is now plumbed from `run_once` through `run_autonomy_passes` into the pass, the ambient loader is deleted, and the CLI passes its own identity through (it had been passing `None`).

### LOW
- Embedding-read failures (`.ok().flatten()`) in **both** clusterers made a failing store indistinguishable from an un-embedded corpus — consolidation silently did nothing, cycle after cycle, with nothing in the logs. Both now log before degrading; the degrade itself is unchanged and still blocks the merge per #1774.
- Dry runs inflated `rollback_entries_written` (documented as *rows persisted*); they now report `rollback_entries_simulated`.
- `limit as i64` → `i64::try_from(..).unwrap_or(i64::MAX)` (PERF-07).
- Stale `#[allow(unused_assignments)]` removed.

### HIGH — the op-cap fix could starve Pass-1 consolidation forever

Charging the autonomy passes against `max_ops_per_cycle` (item 2 above) shares **one**
cycle budget between the auto-tag / contradiction loop and the passes, and both are fed by
the **same** `needs_curation` predicate. The loop runs first and is entitled to spend the
whole cap, so whenever the eligible backlog reached the cap (default 100), Pass-1
consolidation was handed a budget of exactly **zero**.

That is not a transient stall. It is permanent whenever those backlog rows keep yielding
empty auto-tags: an empty tag list persists nothing, so the row stays eligible forever and
re-fills the cap every cycle. Consolidation would then never run again on a busy corpus,
silently — the cycle report looked healthy throughout.

**Falsified before/after.** With the loop bounded by the pre-fix `cfg.max_ops_per_cycle`,
`run_once_reserves_autonomy_budget_when_autotag_backlog_fills_cap` fails with
`autonomy_ops_budget: 0`, `operations_attempted: 4` (the whole cap eaten by the loop) and
the autonomy pass reporting `operations_skipped_cap: 1` — a cluster formed with no budget
to act on it. Bounding the loop by its own share turns the test green. (Note
`clusters_formed` is **not** budget-sensitive — it reads 1 even while starved — so the
budget assertion, not the cluster count, is the load-bearing witness.)

The reserve is `max_ops_per_cycle / AUTONOMY_OP_RESERVE_DIVISOR`, floored at one op and
carved **out** of the cap (the loop's own budget shrinks by exactly that amount), never
added on top of it — so the documented hard cap still holds, which the test asserts
directly. A cap of 0 or 1 has nothing to split and reserves nothing.

Deferred auto-tag work is counted in `operations_skipped_cap` (postponed, never dropped).
New `CuratorReport.autonomy_ops_budget` publishes what Pass 1 actually received: a zero
there against a non-zero `memories_eligible` is the starvation signal. It prints in the
`curator --once` report and in the daemon's per-cycle log line alongside `deferred`,
neither of which was visible before.

### Behaviour change — `curator --once` now ARMS the auto-persona sweep

Threading the daemon's signing identity into the one-shot path passes `Some(keypair)`
where the pre-change code passed `None`, and the `run_once` auto-persona sweep is gated on
exactly that. A `curator --once` invocation can therefore mint signed `__persona_*` rows
where it previously wrote none. The identity is ambient (lexicographically-first key in the
resolved key dir), so it is now logged at INFO on load. Documented in the CHANGELOG.

## #2941 re-register wipe (13 preserve-list sites)

**Verdict on the filed title: NOT-A-DEFECT. The real defect is a different, worse one.**

The issue was filed as "`agents bind-key` intermittent silent no-op". Re-verified at
`release/v1.0.0`: `bind_agent_pubkey` is ONE atomic `UPDATE` that bails loud when
`affected == 0` (`src/storage/mod.rs`), so the bind itself cannot silently no-op.

The real cause of the observed end-state is a **re-register wipe**, and it is not
intermittent — it is deterministic on every re-register:

- `storage::register_agent` is documented "Register or refresh" and builds **fresh**
  metadata with no `agent_pubkey` (`src/storage/mod.rs`), then ends in `insert()`.
- `insert()`'s `ON CONFLICT … DO UPDATE` overwrites `metadata` wholesale, preserving
  only `agent_id, derived_from, consolidated_from_agents`.
- The postgres twin is identical (`src/store/postgres.rs`).

So **any** idempotent re-register — MCP `memory_agent`, `POST /agents`, the SAL trait,
the CLI — erased the bound key on **both** backends.

### Why this is a data-integrity issue, not an availability one

| Posture | Result after a re-register |
|---|---|
| strict (`AI_MEMORY_REQUIRE_AGENT_ATTESTATION=1`) | every later signed write 403s — loud |
| **DEFAULT permissive** | a genuinely signed write is persisted as `attest_level=claimed` instead of `agent_attested` — **no error, no WARN, no counter** |

The default posture is the dangerous one: the downgrade is durable and, after the fact,
indistinguishable from an agent that never signed at all. That is exactly the
"degrade, never corrupt" line — the substrate kept serving while silently
recording weaker provenance than it actually had.

### Fix — two layers

1. **`register_agent` carries the bound-key PAIR through its existing pre-read**, on
   both backends. sqlite previously read only `$.registered_at`; it now reads the whole
   metadata object. Absent keys stay absent, so a never-bound or explicitly **revoked**
   agent does not re-acquire a key.
2. **The upsert preserve-list is now a reserved-key SET**,
   `ai_memory::RESERVED_UPSERT_METADATA_KEYS`, covering
   `agent_pubkey` + `pubkey_bound_at`. **This is the load-bearing half**: it is evaluated
   INSIDE the conflicting statement, so a `bind-key` landing between any caller's
   pre-read and its upsert cannot be clobbered either. Layer 1 alone has a TOCTOU window;
   Layer 2 does not.

The same reserved set also now guards two sibling funnels that had the identical hole:
`update_with_expected_version` (a full-object metadata patch that OMITS the pair) and
`insert_if_newer` (a newer-wins federation merge unbinding a locally-bound key).

**Rotation and revocation are unaffected.** `bind_agent_pubkey` / `revoke_agent_pubkey`
write through their own explicit `UPDATE` (sqlite `json_set`, postgres `jsonb`
set-and-strip), never this upsert funnel, so "existing wins" here never freezes a key.

### Anti-drift gate

The preserve set is open-coded in **thirteen** SQL literals (3 sqlite, 10 postgres).
Restructuring all thirteen to splice the shared text at compile time would touch far more
of the hottest write path than this fix warrants, so
`tests/reserved_upsert_metadata_keys_2941.rs` scans both source files and asserts every
preserve-site renders exactly `reserved_upsert_metadata_keys_sql!()`, and that this list
names exactly `RESERVED_UPSERT_METADATA_KEYS`. Drift between any two sites — or between
the SQL and the typed set — is not mergeable.

**Decision flagged for review:** this is a source-scanning gate rather than a literal
single-const splice. It is uniformly binding across all 13 sites, but it is a test-time
guarantee, not a compile-time one. Raised deliberately rather than silently.

## Rules applied

ERRORS-08 (no panic reachable from a safe public API on a caller-context violation), ERRORS-19 (never silently drop a `Result`) ×4, CONCURRENCY-22 (no blocking call inside an async task), PERF-07 (`TryFrom`, never `as`, for lossy conversions), API-09 (a named predicate instead of three copies of a bool condition), TEST-02/TEST-03.

## Tests (14 new regressions)

- `persona_sweep_honours_exclude_namespaces` / `persona_sweep_honours_include_namespaces` — both halves, each with a **positive control** so a filter that merely disabled the sweep could not pass.
- `namespace_in_scope_predicate` — including exclude-wins-over-include (fail closed).
- `persist_helpers_refuse_non_object_metadata` — plus an object-metadata control that must still succeed.
- `run_autonomy_passes_charges_pass1_against_llm_op_budget` — two mergeable clusters, budget 1: asserts exactly one LLM call, one consolidation, one counted skip.
- `run_autonomy_passes_zero_llm_budget_consolidates_nothing` — asserts zero LLM calls and that **both source rows survive** (a skipped consolidation is a no-op, never a partial merge).
- `dry_run_reports_simulated_not_written_rollback_entries` / `live_run_reports_written_not_simulated_rollback_entries`.
- `consolidation_pass_degrades_instead_of_panicking_inside_a_runtime` — a `#[tokio::test]` that pre-fix panicked and took the test thread with it.

Pass-1 starvation (this fold):
- `run_once_reserves_autonomy_budget_when_autotag_backlog_fills_cap` — 5 eligible rows against a cap of 4; asserts Pass 1 gets a non-zero budget AND that `operations_attempted <= max_ops_per_cycle` (the reserve is carved out, never added on top) AND that the deferred auto-tag work is **counted**, not dropped.
- `autonomy_op_reserve_is_a_share_of_the_cap_never_an_addition` — property-style sweep over caps `0,1,2,3,4,7,99,100,10_000`: the reserve never exceeds the cap, and any splittable cap reserves at least one op.

#2941 (`tests/reserved_upsert_metadata_keys_2941.rs`):
- `reregister_preserves_bound_pubkey_and_signed_store_stays_agent_attested` — register → bind → **re-register** → the key and its bind stamp survive, `registered_at` preservation is unaffected, and a signed write still lands `agent_attested` rather than silently downgrading to `claimed`.
- `pg_reregister_preserves_bound_pubkey` — the live-Postgres twin, gated on `AI_MEMORY_TEST_POSTGRES_URL`.
- `reserved_upsert_metadata_keys_sql_matches_the_typed_set` + `both_backends_preserve_the_same_reserved_key_set` — the anti-drift gate over all 13 preserve-sites.

## Compatibility

`run_autonomy_passes` gains two parameters and `AutonomyPassReport` gains four fields (all `#[serde(default)]`, so previously-written self-report rows still deserialize). Both are pre-1.0 internal-daemon surface; no MCP tool, HTTP route, CLI verb, or schema version changed.

## Verification

- `cargo check --all-targets` — default, `sal`, `sal-postgres`: clean
- `cargo check --examples`: clean
- **clippy with the exact CI flag set** (`-D warnings -D clippy::all -D clippy::pedantic`, `--all-targets`) on **all four legs** — default, `sal`, `sal-postgres`, `vectorlite`: clean
- `cargo fmt --all`: applied
- `qual_10_module_size_ceiling` + `qual_6_7_legacy_error_type_ceiling`: pass (no ceiling bump needed)
- `append_only_spine_guard_g7` (source guard over `forget_if_superseded`) + `curator_conserves_contradiction_g7`: pass
- Unit: `curator::` (136) + `autonomy::` (37) + `cli::curator::` (24) — all pass
- Integration: `curator`, `persona`, `persona_signing_pipeline_813`, `issue_809_nhi_self_persona_any_agent`, `confidence_source_curator_derived_1242`, `rollback_evidence_1946` — all pass
- `scripts/check-docs-vs-ssot.sh`: PASS


alphaonedev:
**Fable final-gate review — APPROVED, merge on green required CI.**

Verified against the diff: all 10 findings CONFIRMED (none refuted — correct, I re-read the key sites). Highlights:
- **#8 was a real production panic, not a doc contract** — `ai-memory curator --once` with compaction enabled called `run_once` from an async fn (`src/cli/curator.rs:297`). Fixed both ways: `Handle::try_current()` guard that degrades to a *reported* skip (ERRORS-08 / CONCURRENCY-22), and `spawn_blocking` at the CLI so the pass runs. Good catch.
- **#2 op-cap**: budget charged against Pass 1 (the sole LLM-invoking pass, matching the config field's documented contract), with `operations_skipped_cap` counted + surfaced before `break` — nothing silently zeroed.
- **#7 halt-over-reverse**: accepted — strictly fail-closed (halting never takes an un-reversible action), surfaced via `rollback_log_degraded` and an explicit "last action is irreversible" message. Reversing from a failure state would re-enter the restore path; halt is the safer choice.
- **#1 namespace cfg**: persona sweep now honours include/exclude, with positive controls so a filter that merely disables the sweep can't pass.
- API break (2 params, 4 report fields, `serde(default)`) is pre-GA internal and backward-deserializable.

Judgement calls endorsed. 10 regression tests with positive controls; clippy exact CI flags clean on all 4 legs; unit 136/38/24 + the curator/persona/rollback integration suites green locally. Commits `056fee46`/`850dab84` SSH-signed. Out-of-scope items #11/#12 noted for a follow-up (both real; #12 touches every `RollbackEntry` site + the serialized log — own PR).

alphaonedev:
**Queue head.** #3112 merged → base `faac3e3a`. Strict protection: rebase onto `origin/release/v1.0.0`, apply any remaining Fable-gate items, CI-parity, one `--force-with-lease` push. `ci-allow.txt` is this branch. Re-review the #2941 preserve-list commit at the gate (`wip(handoff)` locally at `a671e7e7`; remote still `057265ae`).

alphaonedev:
**Grok T7 progress.** Rebased onto `faac3e3a`. Head **`11486e47`** (4 commits, all `%G?=G`). Local CI-parity: 5 clippy + fmt + checks + QUAL-10 (storage 29150/29160, postgres 34683/34720) + 157 curator lib tests + #2941 sqlite tests + load-bearing falsification (revert both layers → 2 tests red, restored).

Live-PG `pg_reregister_preserves_bound_pubkey` hit `#2445` `SchemaAheadOfBinary` (shared `ai_memory_test` is **v90** from PR #3216's pg lane; this binary is v89). Did not set `AI_MEMORY_ALLOW_SCHEMA_AHEAD`. CI's isolated PG at v89 is the authoritative twin.

`ci-allow.txt` restored to this branch after a sibling worktree overwrote it. Waiting 34/34.

alphaonedev:
**MERGED** via `fable-merge.sh`. Head `11486e47` 34/34 required-green. #2941 preserve-list falsification proved load-bearing. Next queue head: **#3115**.

Issue #2991, opened 2026-08-16
R40 m-of-n signed-approval gate has NO production trigger — route_escalation_to_approval_gate has zero src callers, so mandatory signed quorum never fires


alphaonedev:
## Summary
Hive-NHI campaign (A4), codegraph-verified. `route_escalation_to_approval_gate` (`src/approvals.rs:636`) — the only fn that stamps `requires_signed_approval` on a pending payload — has **zero callers in src/** (codegraph blast radius: 1 caller, tests only: `tests/r40_airgapped_approval.rs:208`). So nothing in production routes a `Decision::Escalate` into a signed-approval-required pending; `AI_MEMORY_APPROVER_PUBKEYS`/`AI_MEMORY_APPROVAL_THRESHOLD` (#1957 R40) can only fire when a caller VOLUNTARILY presents an `approvals` array — never mandatory. Compounding: the gate exists on one surface only (grep: `verify_quorum_from_env`/`pending_requires_signed_approval` → src/approvals.rs + src/mcp/tools/pending.rs), and the CLI approve path executes unsigned regardless (sibling issue).

## Proposed fix
Wire the escalate→pending-requires-signed-approval router at the governance decision point; route all approve funnels (CLI/HTTP/MCP) through one shared chokepoint that honors the gate. Medium. Verified pm-v3.3 (call-graph + surface greps) + codegraph.

alphaonedev:
Fixed via PR #3106 (merged to release/v1.0.0). All SIX approve-and-execute funnels (MCP handle_pending_approve + HTTP approval_decide/approve_pending sqlite+pg + CLI pending approve) now route through ONE evaluate_signed_approval_gate above their finalizers; requirement = server-side OR(stored escalation flag, live namespace policy); missing/forged/unenrolled → fail-closed refuse; quorum-not-met → pending; record_quorum_event chains on every surface. L1-6 Decision::Escalate PRODUCER wired via route_or_block_escalated_write (single-use CID-bound exemption, keyless fail-closed). Posture check #20 (producer armable). Adversarial-reviewed: missed 6th (CLI) funnel + check-20 test caught & fixed; live-pg twins verified. GA Wave-2 (#3089).

Issue #2954, opened 2026-08-15
Audit: federation newer-wins merge (insert_if_newer / PostgresStore::apply_remote_memory) overwrites content in place without a memory_revisions leaf when armed


alphaonedev:
## Context

Surfaced while closing #2948 (the `db::insert` create-funnel `ON CONFLICT DO UPDATE SET content` append-only bypass). Widening the G6 static guard (`tests/append_only_spine_guard_g6.rs`) to a P4 shape catching `ON CONFLICT(title, namespace) DO UPDATE SET content = ...` revealed a SECOND, distinct content-overwrite class that also bypasses the signed `memory_revisions` ledger when the append-only spine (#1823/#2947, `AI_MEMORY_APPEND_ONLY`) is armed.

#2948 was deliberately scoped to the UNCONDITIONAL create-funnel overwrite (`content = excluded.content`). This issue tracks the CONDITIONAL federation newer-wins LWW merge, which is a separate funnel class.

## The finding (verified in code)

Two funnels resolve a `(title, namespace)` collision with a CONDITIONAL newer-wins overwrite:

- sqlite `db::insert_if_newer` (`src/storage/mod.rs`) — `content = CASE WHEN excluded.updated_at > memories.updated_at OR (excluded.updated_at = memories.updated_at AND excluded.id > memories.id) THEN excluded.content ELSE memories.content END`.
- postgres `PostgresStore::apply_remote_memory` (`src/store/postgres.rs`) — the same newer-wins CASE arm.

When the inbound row WINS the LWW tiebreak, these rewrite an existing local row's durable authored `content` IN PLACE. Neither emits a `memory_revisions` Supersede leaf, even when `append_only_enabled()`. Evidence: `tests/append_only_spine_flagon_g6.rs::same_id_supersede_converges_identically_flag_on_vs_off` asserts the total leaf count stays at 1 (from the earlier `db::update`) after an `insert_if_newer` call that overwrites `local-v2` -> `remote-v2` — i.e. the federation merge produced ZERO leaves for a real content overwrite.

The same-id federation full-row overwrite path (`overwrite_full_row_by_id`) IS sanctioned and DOES emit a Supersede leaf; it is only the `(title, namespace)` newer-wins merge arm that is unledgered.

## Why it matters (prime directive)

[...]

Issue #2758, opened 2026-08-06
[governance][integrity] PreRecall / PreSearch / PreTranscriptStore hooks are enum+table-present but production-UNFIRED (same class as #2637)


alphaonedev:
## Class
Same advertised-vs-real governance-gate-integrity class as #2637 (PreCompaction/PreArchive). Surfaced by the #2637 scout while proving those two were test-only.

**Enterprise-ready standard:** a hook the substrate advertises must actually fire, or it must not be advertised — a gate that never fires is a claim of enforcement that does not exist.

## The gap
`PreRecall`, `PreSearch`, and `PreTranscriptStore` appear in the `HookEvent` enum (+ `is_pre_event`/`event_class` tables and serialization tests) but have **zero production fire sites** — no `.fire`/`consult_pre_event_gate` at any recall / search / transcript-store path. As shipped they are inert: they inflate `HOOK_EVENTS_COUNT`, are selectable in `[hooks].required_events`, and advertise an enforcement point that does nothing.

## Proposed fix (mirror #2637's adjudicated pattern)
Per-event 5-agent vote (`4d3ea1c5`) on **wire-in vs remove**, exactly as #2637 (W-C):
- **Wire in** where a real production destructive/sensitive path exists and enforcement is coherent (consult via the `crate::mcp::consult_pre_event_gate` process-global gate, #1885/#1924/#2356 precedent; add to `ELIGIBLE_REQUIRED_EVENTS`; boot-install the gate on the serving process).
- **Remove** (variant + tables + tests + docs, decrement `HOOK_EVENTS_COUNT`, fail `[hooks].required_events` parse loudly) where the op is read-only / non-sensitive or the only fire site is a self-deadlock hazard — the #2637 `PreArchive` disposition.
- Note: recall/search are READ paths (recall is pure since #1869/#1953) — a pre-READ governance gate may be the wrong abstraction; the vote must decide honestly rather than wire a gate onto a read for symmetry.

## Acceptance
- Each of the 3 events is either FIRED at a named production site with an R-203 test proving Deny changes behavior, OR fully removed with `HOOK_EVENTS_COUNT` truthful.
- `check-docs-vs-ssot.sh` green on the new count.

Refs #2637. auto-filed-by-agent (Fable 5 orchestrator, v1.0.0 GA cert loop).

alphaonedev:
[...]

PR #3118, opened 2026-08-22
docs(claims): make every published claim truthful — claims-audit batch 1

alphaonedev:
Re-verified the pre-flagged CLAIMS-AUDIT overclaims against the code at `release/v1.0.0` and corrected every published statement that overstated what v1.0.0 actually does.

**Docs, doc-comments, and two machine-facing MCP advertisement strings only — zero behaviour change.** Claims that turned out to be accurate were left alone.

## REFUTED — verified accurate, left untouched

| Claim | Verification |
|---|---|
| README 94 routes / 80 unique paths | `EXPECTED_PRODUCTION_ROUTES_COUNT = 94`, `EXPECTED_PRODUCTION_UNIQUE_PATHS_COUNT = 80` (`src/lib.rs`), pinned by `tests/route_count_invariant.rs` |
| README 92 CLI subcommands sal / 90 default | `EXPECTED_CLI_SUBCOMMANDS_{SAL,DEFAULT}`, pinned by `tests/cli_subcommand_count_invariant.rs` |
| README 103 MCP entries full / 8 core | derived from `Profile::full().expected_tool_count()`; no restated literal |
| ROADMAP schema v89 + 103-tool anchors | `CURRENT_SCHEMA_VERSION = 89` both backends; lockstep tests pin both directions |
| ROADMAP `#1952` decorrelation `off → advisory` default | `ReflectDecorrelationMode` derives `Default` with `Advisory` annotated `#[default]` |
| ROADMAP 22-event count | live `HookEvent` enum has exactly 22 variants (the *count* is right — see below for the firing caveat) |
| CHANGELOG `#2542` namespace-standard entries (3) | fail-closed authz + sqlite/postgres parity verified in both twins |
| Release-notes 59/21 pg-parity disclosure, MCP-stdio SQLite-only | byte-exact against `postgres_gate.rs` + the CI-pinned inventory |
| `docs/CLI_REFERENCE.md` link relations | already lists all 9 correctly |

### 1. Hook pipeline: 11 of 22 advertised events never fire

The docs presented all 22 lifecycle events as live. Verified at `release/v1.0.0`:

- **Wired (11)** — all 10 decision-class `pre_*` events, plus `post_signal_ack`. **Hook-based enforcement is real**: a `Deny` from any `pre_*` hook genuinely refuses the operation.
- **Advertised but inert (11)** — `post_store`, `post_recall`, `post_search`, `post_delete`, `post_promote`, `post_link`, `post_consolidate`, `post_governance_decision`, `post_reflect`, `on_index_eviction`, `on_compaction_rollback`. These parse from `hooks.toml`, classify, and render in `ai-memory doctor --hooks` — but **no production code path fires them**.

This is the same `#2444` advertised-but-inert class that `#2637`/`#2758` addressed for other events, against this project's own stated standard: *"a hook the substrate advertises must actually fire, or it must not be advertised."*

Two source claims were flatly false and are corrected at the source:

- the `#2758` rationale asserted the retained `post_recall`/`post_search` siblings *"fire on real production read paths"* — neither has a fire site; `RecallResult`/`SearchResult` are constructed nowhere outside `events.rs`;
- `src/curator/compaction.rs` asserted its Stage-6 rollback *"fires the notify-only `OnCompactionRollback`"* — the only `HookEvent::OnCompactionRollback` reference in that file is inside its `#[cfg(test)]` module.

`on_index_eviction` is the subtlest: producer (`hnsw` eviction send) and consumer (`spawn_eviction_observer` → `fire_on_index_eviction`) **both exist**, but nothing calls `set_eviction_sink` outside `#[cfg(test)]`, so the channel is never connected and the sink stays `None`. Reading either half alone suggests it fires.

Corrected in `docs/hook-pipeline.md` (per-event firing table + status note), `docs/ADMIN_GUIDE.md`, `docs/production-deployment.md`, `README.md`, `ROADMAP.md`, `src/hooks/events.rs`, `src/curator/compaction.rs`. **Wiring or removing these 11 is a code change and is left for a follow-up** — see the behaviour-gap list below.

### 2. `memory_link` advertised 5 of its 9 relations

`LinkTool::docs()` and the always-served `memory-workflow` MCP prompt listed `related_to | supersedes | contradicts | derived_from | reflects_on`, omitting `derives_from`, `decomposes_into`, `depends_on`, `advances` — all four accepted by the handler and enforced by `MemoryLinkRelation` (`COUNT == 9`), `VALID_RELATIONS`, and the SQL CHECK constraints. The strings **understated** the tool's real surface to every MCP client.

Both now list all 9 and flag the `derived_from` (N→1 consolidation-merge) vs `derives_from` (1→N atomisation-split) opposite-direction footgun. The byte-pinned `tests/snapshots/tool_definitions_pre_d1_6.json` is amended in lockstep, with the rationale recorded in the pinning test's own module doc so the next reader sees why the pin moved. Same fix applied to the TypeScript SDK `Relation` union (4 of 9) and the README CLI `link` row (4 of 9).

### 3. `ARCHITECTURAL_LIMITS.md` promised two things Postgres does not do

- §5: *"Postgres adapter removes the daemon-as-bottleneck entirely; agents connect directly to Postgres over its wire protocol."* Every request still terminates at the axum daemon behind `postgres_route_gate` (59 of 80 paths served, 21 fully `501`), with a scratch SQLite still opened alongside.
- §7: *"Postgres adapter exposes CDC natively."* There is no logical-replication, `wal2json`, publication, or change-stream code anywhere in `src/` — zero hits.

### 4. `docs/README.md` — unqualified "first-class backend"

Advertised `postgres+AGE first-class backend` with no parity caveat while the root README already carried the honest Backend parity section. Now carries the 59/21 split and the SQLite-only stdio MCP limitation.

### 5. `SECURITY.md` M8 overstated `asi-hard`

Claimed the profile pins the fail-closed gates *"so a compromised host cannot silently weaken them."* But the profile is selected by a **host-controlled environment variable** (`AI_MEMORY_SECURITY_PROFILE`) and enforced **in-process**; unset, it resolves to `standard` with no pins in force, **silently**. A host that can set the process environment simply does not select `asi-hard`. Rescoped to what it genuinely protects against — operator error and configuration drift, once `asi-hard` *is* selected — and states that defending the selection needs an out-of-process control ai-memory does not ship.

### 6. Certification-doc evidence staleness

- The committed four-leg posture proof in `docs/compliance/evidence/cert-54/` is an **18-check** capture (`grep -c '[PASS]'` = 18) while `ENTERPRISE_FEDERATION_CHECK_COUNT` is now **20** (`#2954` added check 19, `#2991` added check 20). The doc disclosed only the 19-check step. The note now covers both and states explicitly that the post-`#2954`/`#2991` tallies are **derived, not measured**, with the re-capture called out as outstanding.
- §5.4(3)'s heading claimed *"executed … evidence (at certified versions)"* while **every cited run ID executed the superseded 18.4 / 1.7.0 / 0.8.5 triple**; no `actions/runs/<id>` citation for the current 18.6 / 1.8.0 / 0.8.6 triple exists in the document. Heading plus a new evidence-status note now say plainly: executed at the prior triple, CI-asserted at the current one.

### 7. `docs/postgres-age-guide.md` asserted feature parity it then disproved itself

Its lead called postgres+AGE a *"first-class backend"* with no caveat, and its schema-parity section stated *"every feature that works on sqlite works on postgres"* — contradicted by the **same document's own "21 fully-501 paths" inventory further down** (no `skills` table, since `migrate_v82` is a version-stamp no-op; no `governance_rules` table). A matching `CURRENT_SCHEMA_VERSION` does not imply a matching set of tables when ladder arms are no-ops. Both places now carry the 59/21 split and the SQLite-only stdio MCP limitation up front. Two stale *"schema parity at v78"* statements — in a doc that states `CURRENT_SCHEMA_VERSION = 89` two lines away — corrected to v89.

### 8. `src/lib.rs` stale CLI-subcommand prose

The block comment still described the v0.7.0-era *"82 variants / 80 default"*; one const's doc said *"86 variants … leaving 84"* against a value of 90; both bump trails were missing the `#2676` `Features` entry (89→90 default, 91→92 sal). The consts themselves (90 / 92) were correct and machine-pinned — only the surrounding prose was wrong.

## Behaviour gaps found — NOT fixed here, flagged for follow-up

1. **`ai-memory forget --tier` parses leniently — silent over-deletion (data-loss class).** `src/cli/forget.rs` does `args.tier.as_deref().and_then(Tier::from_str)`; `Tier::from_str` returns `None` on an unknown value, and `db::forget` treats a `None` tier as *unconstrained* in the DELETE. So `--tier Long` (capitalised) **silently widens the erase to every tier** and still prints success. Fix is to fail closed with `ok_or_else`, mirroring `src/cli/store.rs`. Highest priority of the three — it destroys data on a typo.
2. **`ai-memory backup restore` publishes in place with no guard.** `run_restore` renames the live DB aside, moves the `-wal`/`-shm` sidecars, then publishes via in-place `std::fs::copy` with no daemon-liveness check, no confirmation, no atomic rename, and no rollback if the copy fails.
3. **The 11 unwired hook events** (§1 above) need a disposition: wire them, or remove them as `pre_archive` / `pre_recall` / `pre_search` were removed.
4. **Cert evidence re-capture** — re-run the four-leg posture proof at 20 checks and trigger one `cert-postgres-age.yml` run at the release tip, then bind §2/§3/§8 to the new artifacts (this PR makes the doc honest about the gap; it cannot close it).
5. **Minor:** postgres's namespace-owner filter checks `o != "system"` while sqlite's also treats `sentinels::SYSTEM_PRINCIPAL` as unowned — harmless iff `SYSTEM_PRINCIPAL == "system"`, worth a one-line confirmation.

## Gates

`check-docs-vs-ssot`, `check-doc-surface-completeness`, `check-doc-symbol-anchors`, `check-benchmark-claims`, `check-capacity-claims`, `check-sdk-route-paths`, `check-hardcoded-literals`, `check-cert-expiry` — all green (and all green at baseline before the edits, so no pre-existing red is masked). `cargo check` green.

**rust-1.98 rules applied:** TOOL-03 (docs must not rot — every corrected item's doc states its real behaviour), TOOL-05 (intra-doc links in the rewritten variant docs), ERRORS-09 (an advertised-but-inert gate is an illegal representable state; named as such rather than left implied), ERRORS-14 (concise, non-overstated claim text), TEST-01 (the snapshot-pin amendment is recorded in the pinning test's own module doc).


alphaonedev:
**Fable gate — APPROVED (final review).** Independently verified on the branch: `git log -S CERT-REMOVAL-PROOF-MUTATION base..head` = 0 commits, marker absent from the tree, `src/federation/receive_auth.rs` absent from the diff, all 3 commits %G?=G (the #3118 near-miss is structurally closed by #3123 + #3119). `src/` delta = comments + exactly 4 non-comment lines: the two `memory_link` description strings — checked against `src/validate.rs::VALID_RELATIONS` (exactly 9: related_to, supersedes, contradicts, derived_from, reflects_on, derives_from, decomposes_into, depends_on, advances) and `validate.rs:796` (unknown relation → error), so the new claim is true. The 11-of-22 inert hook lifecycle events finding stays documented-as-truth for GA; wiring them is a post-GA issue. Merge when required legs are green (snapshot + token-budget gates cover the description change).

alphaonedev:
Fable gate — **APPROVED with doc FOLLOW-UP** (scout-verified @ 93d09145; ox 'fully wired overclaims' REFUTED — the `http_pre_event_gate` wrapper covers delete/promote/consolidate/links/route_1111/create/bulk; 92/2 command counts and 9 relations verified). Fold at rebase: `src/hooks/events.rs:72,75,94` still say 21/21/27 lifecycle events while the PR's own header says 22 (enum has exactly 22 variants) — better derive from a count-pinned invariant; `src/curator/compaction.rs:543` new comment 'the WARN below is the only signal' is false (`report.errors` at :557-560 also carries it); `src/hnsw.rs:539-547` producer doc still claims daemon_runtime populates the sink — only `vectorlite.rs:324/:654` forward one and `spawn_eviction_observer` is test-only (the PR's 'never connected' claim is right; this is the stale other half). Not touched by #3169 either.

alphaonedev:
**Rebased onto `85b4d0e4` (Merge #3133) and pushed `--force-with-lease`.**

Head **`bef6bfae` G**. Five commits, all `%G?=G`. Contains base. `scripts/check-docs-vs-ssot.sh` PASS (hooks=22, asi_hard_knobs=22, schema=89).

Fable FOLLOW-UP folded at rebase:
- `src/hooks/events.rs` 22 lifecycle events (was 21/27)
- compaction signals comment + `report.errors`
- hnsw eviction sink producer is vectorlite not daemon_runtime
- #2373 closed — pending postgres CHECK-side claim dropped (`f88fb8e4` → `bef6bfae`)

`ci-allow.txt` = `docs/claims-audit-truthfulness` + `release/v1.0.0`. 34-watcher armed on `bef6bfae`. Do not rerun the old `dc5fd630` SHA.

alphaonedev:
**Allowlist race:** `ci-allowlist-enforcer` restores `ci-allow.txt` from `ci-allow.ssot` every loop. After #3133 merge I updated only `ci-allow.txt`; SSOT still had `fix/security-fail-open-gates`, so the first wave on `bef6bfae` was cancelled.

Both files now list `docs/claims-audit-truthfulness` + `release/v1.0.0`. Empty retrigger commit **`1b6ab760` G** (FF; do **not** rerun `bef6bfae`). 34-watcher re-armed; cancelled-only is treated as pending, not fail.

PR #3122, opened 2026-08-22
fix(gate): segment-aware const-name-literal matcher — unblock base tree (#3121)

### LINKED ISSUE #3121: check-const-name-literals.sh: false positives at base (hex-byte substring matched inside identifiers; size suffixes) — HARD-BLOCKs every local gate sweep
Two Opus 5 instances independently hit this on 2026-08-22 while running the local gate sweep on clean worktrees of `release/v1.0.0` @ 15fe44e4. The script is **not** wired into CI (no workflow references it), so it is not a merge blocker, but it HARD-BLOCKs on the unmodified base:

```
tests/conformance_corpus.rs:676: name `zero32` embeds its value `32` — rename semantically
tests/signed_events_dlq_replay_1046.rs:109: name `payload_hash` embeds its value `0xad` — rename semantically
const-name-literals: HARD-BLOCK — value-encoding identifier name(s) above
```

Both are false positives:
- `payload_hash = vec![0xde, 0xad, ...]` — the matcher finds the hex byte `ad` as a substring of `paylo**ad**`. Hex-byte values must be matched only against whole identifier tokens / `_`-delimited segments, and two-hex-digit values are too short to match inside words at all.
- `zero32` — `32` is the array length (a 32-byte zero buffer), a legitimate size suffix (`u8`, `zero32`, `buf64`, `key256`). Treat `[a-z]+(8|16|32|64|128|256|512)` size suffixes as semantic, not value-encoding.

Fix the matcher (segment-aware + hex min-length + size-suffix allowlist), add both cases to `--self-test`, and verify the script is clean at base with **no** `--update-baseline`. Keep the script's true-positive power (its own self-test injection must still be caught).

alphaonedev:
## The bug, precisely

`scripts/check-const-name-literals.sh` HARD-BLOCKed on the **unmodified** `release/v1.0.0` tree (@ `15fe44e4`) with two false positives. The matcher was an arbitrary-**substring** test:

```python
flat_name = name.replace("_", "")   # payload_hash -> "payloadhash"
...
if digits in flat_name:             # "ad" in "payloadhash"  -> TRUE
```

* `tests/signed_events_dlq_replay_1046.rs:109` — `let payload_hash = vec![0xde, 0xad, 0xbe, 0xef, 0x10, 0x46];`. The hex byte `0xad` yields digits `"ad"`, which is a substring of `payloadhash` (the "paylo-**ad**" tail). Nothing about the name encodes a value.
* `tests/conformance_corpus.rs:676` — `let zero32 = [0u8; 32];`. `32` is the array LENGTH (a 32-byte zero buffer) — a semantic width suffix, not a tuned value.

## The fix (script + `--self-test` only)

New `encodes_value(name, digits, is_hex)` predicate replaces the substring test:

1. **Segment-aware** — a value counts as embedded only when it equals a whole `_`-delimited name segment (or the whole identifier), or is the digit run of an otherwise-alphabetic segment (`chunk500`, `500ms`). Never an arbitrary substring.
2. **Hex** digit runs only ever match a **whole segment** (`MAGIC_DEADBEEF`), and two-hex-digit byte values (`0xde`, `0xad`) are skipped outright — far too short to be a value-encoding name.
3. **Size-suffix allowlist** — `[A-Za-z]+(8|16|32|64|128|256|512)` on an alphabetic stem (`zero32`, `buf64`, `key256`) is a width TERM, exempt. A bare digit segment is NOT exempt, so `const BATCH_64 = 64` still blocks.

The new predicate is **strictly narrower** than the old one (every new hit was also an old hit), so no new hits are possible and the baseline — already empty — cannot grow. **No `--update-baseline` was run**, and neither identifier was renamed: they are correct as written; the script was wrong.

## `--self-test` (requirement 3)

Each case is probed in isolation so one shape cannot mask another:

* **(a) true positives, must HARD-BLOCK** — `SELFTEST_TIMEOUT_8000_MS = 8_000` (whole digit segment), `SELFTEST_CHUNK500 = 500` (stem+digits, non-size), `SELFTEST_MAGIC_DEADBEEF = 0xdead_beef` (whole hex segment).
* **(b) #3121 known-good, must NOT be flagged** — `payload_hash = vec![0xde, 0xad, ...]`, `zero32 = [0u8; 32]`, `buf64 = [0u8; 64]`, `key256 = [0u8; 256]`.

Both directions were **mutation-proven load-bearing**: reverting the matcher to the old substring test fails all four negative cases; neutering the matcher (`return False`) fails all three positive cases.

## Verification

```
$ bash scripts/check-const-name-literals.sh          # BEFORE
tests/conformance_corpus.rs:676: name `zero32` embeds its value `32` — rename semantically...
tests/signed_events_dlq_replay_1046.rs:109: name `payload_hash` embeds its value `0xad` — ...
const-name-literals: HARD-BLOCK — value-encoding identifier name(s) above    (exit 1)

$ bash scripts/check-const-name-literals.sh          # AFTER
const-name-literals: clean                                                   (exit 0)

$ bash scripts/check-const-name-literals.sh --self-test
self-test OK: gate is load-bearing and free of the #3121 false positives      (exit 0)
```

Also verified: documented true positives `const BATCH_64: usize = 64;` and `let chunk_500 = 500;` still HARD-BLOCK; semantic names (`gc_chunk_rows = 500`, `ttl_ms = 8_000`) stay clean; `bash -n` clean; the probe file is removed on every path. `check-hardcoded-literals.sh`, `check-vendor-literals.sh` and `check-docs-vs-ssot.sh` all still PASS. `shellcheck` is not installed on this host and CI runs actionlint with `-shellcheck=`, so it could not be run.

No Rust changes. CHANGELOG `[Unreleased]` -> Fixed entry added (base entries kept).

Closes #3121


alphaonedev:
**Fable gate — APPROVED (final review).** Matcher is segment-aware: a value hits only as a whole `_` segment / whole identifier, or as the digit run of an alphabetic segment excluding width suffixes (`zero32`/`buf64`/`key256`); hex only as a whole segment and only at ≥3 hex digits. Traced: `SELFTEST_TIMEOUT_8000_MS`, `SELFTEST_CHUNK500`, `SELFTEST_MAGIC_DEADBEEF` still HARD-BLOCK; `payload_hash`/`zero32` no longer flagged; baseline untouched; self-test covers both directions. Merge when required legs are green.

alphaonedev:
Fable gate — **APPROVED with FOLLOW-UP** (scout-verified @ 93d09145; ox BLOCK REFUTED — base's matcher had no `.lower()` so the claimed lost true-positives never existed; GitHub's CONFLICTING flag is stale cache, merge-tree is clean). Follow-ups (fold if cheap): (1) the `size_suffix` arm exempts ANY fused `<alpha><width>` segment (`POOL_SIZE256 = 256` now passes silently, baseline burned EMPTY) — put `zero32` in `scripts/qc-allowlists/const-name-literals-allow.txt` and drop the regex; add a positive probe keeping `SELFTEST_POOL_SIZE256 = 256` BLOCKED. (2) hex arm: `MAGIC_DEAD_BEEF = 0xdeadbeef` no longer flagged — accept a segment-aligned span match for hex runs ≥3. (3) self-test: add a `let` positive probe (all positives are `pub const`). (4) the gate is wired into NO workflow — add run + `--self-test` to c8-precheck mirroring `check-hardcoded-literals.sh` + the required-contexts mirror entry. (5) document multi-run segments.

alphaonedev:
**Rebased onto `16dcb8fb` (Merge #3208) + Fable FOLLOW-UP folded.**

Head **`50e40870` G**. Contains base. `scripts/check-const-name-literals.sh --self-test` OK; scan clean; `check-required-contexts.sh` OK.

1. `size_suffix` regex dropped; `zero32`/`buf64`/`key256` exact allowlist. `SELFTEST_POOL_SIZE256 = 256` BLOCKED.
2. Hex consecutive-segment span: `MAGIC_DEAD_BEEF = 0xdeadbeef` flagged.
3. `let selftest_chunk_500 = 500` positive probe.
4. `c8-precheck.yml` job `Const-name-literal identifier gate (#3121)` (run + `--self-test`); **not-required at introduction** (rule f ledger). Mirror-first; do not add to live required set in this PR.
5. Header documents multi-run hex segments.

`ci-allow.txt` **and** `ci-allow.ssot` = `fix/3121-const-name-literals-false-positives` + `release/v1.0.0`.

Issue #2473, opened 2026-07-29
CI job name silently truncated by YAML comment parsing: L3-boundary gate loses '#1853)' — declared name ≠ emitted context


alphaonedev:
## Summary

`.github/workflows/c8-precheck.yml:75` declares:

```yaml
    name: L3-boundary perma-ban gate (§25.3 S5 / RQ-10 #1853)
```

In YAML, ` #` (whitespace then hash) in an **unquoted** scalar starts a comment. So the job name parses as 44 characters and `#1853)` is silently discarded. GitHub emits the check-run as:

```
L3-boundary perma-ban gate (§25.3 S5 / RQ-10
```

Verified: `gh api repos/alphaonedev/ai-memory-mcp/commits/e1f16679/check-runs --jq '...|{name,len:(.name|length)}'` → `{"len":44,"name":"L3-boundary perma-ban gate (§25.3 S5 / RQ-10"}`.

## Why this matters beyond cosmetics

**It is a live footgun on branch protection.** That truncated name is now one of the 22 required status checks on `release/v1.0.0`. Quoting the name to "fix" it **renames the context**, and the required check would then never report — a permanently deadlocked release branch, which is exactly the failure mode #2443 documents.

So this cannot be fixed in isolation. Any fix must update the workflow and the branch-protection required set **together**.

## Not a general problem — check the boundary

Sibling job names survive because their `#` is preceded by `(` rather than whitespace, which YAML does not treat as a comment start:
- `Vendor-monoculture + SECS_PER_* lint-gate (#1174 PR10)` — intact
- `Test-env $HOME-lock gate (#2146)` — intact
- `Test-reachable stdin-read gate (#1989)` — intact

`L3-boundary` is the only one with ` #`. Worth a mechanical check across all workflows rather than trusting that observation.

## Proposed fix

1. Quote the name in `c8-precheck.yml:75`.
2. In the **same** change, update the required-context list on `release/v1.0.0` (and `main`/`develop` if applicable) to the new string.
3. Add a gate asserting every `jobs.<id>.name` round-trips through YAML unchanged — i.e. no unquoted ` #`. This is the same 'declared ≠ actual' class as #2443 and should be caught mechanically, not by eye.

🤖 auto-filed-by-agent

alphaonedev:
## Confirmed live, and the deadlock coupling is now ARMED — sequencing is critical

Verified directly against the live API and the workflow file at `4f28f1d8`:

```
[...]

PR #3051, opened 2026-08-18
fix(cli): close the CLI-surface security parity cluster (#2990 #2992 #3036 #2988)

alphaonedev:
Wave-C1 of the v1.0.0 hive-campaign: CLI-surface **security** parity. Four CLI paths diverged from the safe MCP/HTTP behaviour; each fixed by applying the guard the reference surface already enforces (precedent-copy — the fail-closed posture is already decided, so no 5-agent vote).

Closes #2990 #2992 #3036 #2988 (manual close — Closes-N does not fire on the non-default release branch).

- **#2990 (HIGH)** CLI recall visibility was FAIL-OPEN (returned other agents' scope=private rows). Now post-filters with `visibility::is_visible_to_caller`, mirroring the MCP `apply_visibility_filter`.
- **#2992 (HIGH)** CLI `pending approve` executed escalation-gated pendings on a bare `--agent-id`. Now enforces the same m-of-n Ed25519 quorum the MCP `approvals` param requires (new `--approval PUBKEY_B64:SIG_B64` flag); fails closed.
- **#3036** CLI `link` never signed edges (unattestable under the certified posture). Now signs via `create_link_signed` with the active keypair, mirroring MCP `memory_link`.
- **#2988** CLI recall_observations bound a namespace into the `agent_id` column (#1705 replay guard inert). Now binds `vis_caller` on `recall.rs` + `shell.rs`.

**Parity tests:** `recall_hides_cross_agent_private_row_parity_2990`, `recall_ledger_binds_caller_identity_not_as_agent_2988`, `cli_approve_refuses_escalated_pending_without_signature_2992` (+`parse_cli_approvals_splits_on_first_colon_2992`), `cli_link_signs_edge_when_keypair_present_parity_3036`.

Locally verified: all `cli::` lib tests **1059/0**. fmt + clippy (all 4 feature legs) clean; ceilings unchanged; no SSOT-pin changes.

## AI involvement
Implemented by an Opus hard-coder subagent; reviewed, independently re-tested locally (all 5 parity tests + 1059 cli:: tests), and signed by the orchestrator (AlphaOne). Each fix copies an existing MCP/HTTP precedent; no crossroads vote was required. Base: release/v1.0.0 @ 0d126021.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

## Rebase onto `release/v1.0.0` @ `15fe44e4` (2026-08-22)

This PR sat unmerged while the base advanced from `0d126021`. **Two of its four
[...]

PR #3123, opened 2026-08-22
fix(security): #3119 trap-restore the cert removal-proof harness + repo-wide mutation-marker gate

### LINKED ISSUE #3119: check-cert-removal-proof.sh: mutation is not restored on interrupt — add an EXIT/INT/TERM trap (near-miss: disabled receive_auth control reached a PR branch)
## What happened (2026-08-22, PR #3118 — never merged, base never affected)

An Opus 5 instance ran `scripts/check-cert-removal-proof.sh` while preparing a docs-only PR, killed it at a 2-minute timeout, and the in-place mutation was left in the tree. A later `git add -A` swept it into a pushed commit:

- `src/federation/receive_auth.rs` — `inbound_write_namespace_authorized` short-circuited to `return true; // CERT-REMOVAL-PROOF-MUTATION` → cross-tenant inbound federated-write authorization bypass.
- On the remote PR branch ~25 min. Caught by local clippy (dead-code/unused-variable errors after the `return true`). Branch history rewritten so no commit ever contains it; verified by Fable: `git log -S"CERT-REMOVAL-PROOF-MUTATION" base..head` = 0, marker absent from tree, file absent from the PR diff.

## Root cause (structural, not operator)

The harness restores each mutated file only on its happy/handled paths (`git checkout -- "$tf"` after the broken run and on the two explicit error returns). There is **no `trap`** — SIGINT/SIGTERM/timeout-kill, `set -e` exits, or a crash inside the cargo run leave the control DISABLED in the working tree with nothing in `git status` a human would read as "security control off".

## Required fix

1. `trap` on EXIT/INT/TERM/HUP that restores every file the harness has mutated (track them in an array; `git checkout -- "${MUTATED[@]}"`), then re-raises.
2. At start: refuse to run if the tree already contains `CERT-REMOVAL-PROOF-MUTATION` (a prior aborted run) unless `--force-restore` is passed, which restores and exits.
3. At end (and in the trap): assert `git grep -q CERT-REMOVAL-PROOF-MUTATION -- src` is **empty**; non-zero exit + loud message if not.
4. A repo-wide gate (pre-commit hook and/or CI lint step) that fails if the marker string appears anywhere under `src/` — defence in depth so the mutation can never be committed regardless of how it got there.
5. Test: run the harness under `timeout 5` and assert the tree is clean afterwards.

## Also record
Agent SOP: never `git add -A` after running any gate/harness script — several gates in this repo mutate the tree by design. Stage explicitly.

alphaonedev:
Closes #3119

## The defect

`scripts/check-cert-removal-proof.sh` is a mutation-testing harness: for each cited security control it rewrites the production source to an always-allow disposition, runs the control's lane test, and asserts the test goes RED. Restoration happened only on the happy and handled paths — **there was no `trap`** — so a SIGINT, a `timeout` kill, or a crash inside the cargo run left the control **disabled** in the working tree, with nothing in `git status` that reads as "a security control is off".

On 2026-08-22 an aborted run left `inbound_write_namespace_authorized` short-circuited to always-allow — a cross-tenant inbound federated-write authorization **bypass** — and a later `git add -A` swept it onto a pushed PR branch for ~25 minutes (#3118; never merged, branch history rewritten, no commit ever contained it).

## The fix — four layers between a mutation and a commit

**1. Trap (req 1).** Each target is registered in a `MUTATED` array **before** it is rewritten, so a crash *between* "mutate" and "run" is still covered. An `EXIT`/`INT`/`TERM`/`HUP` trap restores all of them with `git checkout HEAD -- "${MUTATED[@]}"` — **index and worktree**, which unwinds the staged `git add -A` variant that an index-only `git checkout --` would have written straight back into the tree. Restoration is idempotent (the array is cleared as it is consumed; a second call is a no-op).

Exit-status propagation:

- `on_exit` captures `rc=$?` as its very first statement, clears its own handlers, restores, asserts, and `exit "$rc"` — a PASS still exits 0, a CERT-RED still exits 1, and only a *success that failed to restore* is upgraded (to 9). A failure is never downgraded.
- `on_signal` deliberately does **not** `exit`. It clears the handler and **re-raises** with `kill -s "$sig" $$`, so the shell dies *of the original signal* and every parent — `timeout`, a shell job, a CI runner, a `wait` in a driver — observes `WIFSIGNALED` / 128+N exactly as it would with no trap installed. `exit 130` would merely *look* like a signal death and would lie to anything that distinguishes "cancelled" from "failed".
- The guard test now runs **backgrounded under `wait`**. Bash defers a trap until the current *foreground* command returns, so the entire length of a `cargo test` run was previously an un-interruptible window — precisely the window #3118 died in. `wait` is interruptible; the trap also reaps the child.

**2. Start guard (req 2).** A marker already present in `src/` (an aborted earlier run) makes the harness refuse to start — loudly, exit 4 — rather than stack a second mutation on a disabled control. `--force-restore` is the deliberate recovery path: it **backs the mutated files up** under `.local-runs/cert-54-evidence/force-restore-<ts>/` first (reversibility — the only durable record of what the aborted run left behind), then resets them to `HEAD` and exits 0. The guard also covers `--list` and `--self-test`: nothing about this harness should proceed on a tree whose controls are off.

**3. End assertion (req 3).** End-of-run *and* the trap both assert the marker is absent from `src/`, printing `SECURITY: mutation still present in <files>` and exiting non-zero otherwise. The scan fails **closed**: a `git grep` that errors is reported as "could not scan — refusing to report clean", never as clean.

**4. Repo-wide gate (req 4).** New `scripts/check-mutation-marker.sh` fails if the literal marker appears anywhere under `src/`, however it got there — a crash the trap could not catch, a `git add -A`, a rebase that resurrects a dropped hunk. It cannot trip itself: it scans `src/` only, and the marker string lives exclusively outside `src/` (this gate, the harness, archived evidence logs).

## Workflow wiring

Added as steps to the **existing** `l3-boundary-gate` job in `.github/workflows/c8-precheck.yml` — the job that already runs `check-l3-boundary.sh` next to `check-hardcoded-literals.sh`, and structurally the same shape (a perma-banned string in `src/`):

```yaml
      - name: Run cert removal-proof mutation-marker gate (#3119)
        run: bash scripts/check-mutation-marker.sh
      - name: Self-test the mutation-marker gate (regression evidence)
        run: bash scripts/check-mutation-marker.sh --self-test
      - name: Prove the harness restores an interrupted run (#3119)
        run: bash scripts/check-cert-removal-proof.sh --self-test
```

It rides in that job rather than a new one **because that job's `name:` is a declared required context** (`scripts/qc-allowlists/required-contexts-release.txt`), so the gate is merge-blocking from the first commit; a fresh job would have been an unrequired context a merge could ignore. The job name is **not** changed — it is the exact string branch protection requires, and #2473 is the record of what renaming it costs.

## Tests (req 5)

`--self-test` gains two interrupt-safety legs that need **no cargo**. They drive the **real** harness against a **real** control in the **real** `src/` tree, with the cargo step stubbed via `CERT_PROOF_STUB_RUN` to a sentinel-touch + long sleep — so the harness is *guaranteed* to be sitting on a live, disabled security control when the signal lands. The sentinel is asserted: if the harness never reached the guard step, the leg **fails** rather than falsely passing. The stub can never manufacture a green verdict (it never returns 0, so a stubbed row always reports CERT-RED).

```
  [ok] return-shape: first-statement injected, body preserved
  [ok] body-shape: whole body replaced, braces balanced
  [ok] subst-shape: unique OLD replaced, marker embedded
  [ok] subst-shape: non-unique OLD rejected (exit 3)
  [ok] timeout-2: SIGTERM mid-mutation → src/ clean, marker absent, rc=124
  [ok] sigint-reraise: died OF SIGINT (rc=130), src/ clean, marker absent

self-test: PASS — three mutation shapes sound; interrupted runs leave a clean tree
```

Leg 5 asserts `wait` reports **130** — that is the assertion that pins the *re-raise*, since a trap ending in `exit 130` is indistinguishable to `$?` but not to a parent inspecting `WIFSIGNALED`. (`set -m` is required there and is commented as such: with job control off, a bash script's asynchronous children start with SIGINT *ignored*, and a signal ignored on entry cannot be trapped.)

**Negative control** — with the four `trap` lines commented out in a scratch copy, both legs go RED and reproduce the exact #3118 artefact:

```
  [FAIL] timeout-2: src/ left DIRTY after the interrupt:
            M src/federation/receive_auth.rs
  [FAIL] timeout-2: SECURITY — mutation marker still present in src/ after the interrupt
self-test: FAIL
```

Also verified by hand: planting the marker → gate exit 1 and harness start-guard exit 4; `git add`-ing the mutation (the literal #3118 shape) → `--force-restore` backs it up, resets index *and* worktree, exits 0.

Hardening found on the way: a `subst` payload whose replacement omits the marker would be invisible to the trap, the start guard, the end assertion **and** the repo-wide gate — the exact blind spot #3118 lived in. It is now rejected outright (exit 3).

## SOP

`docs/AI_DEVELOPER_WORKFLOW.md` §5.6 (new) and `CLAUDE.md` §Build & Test Commands: gate scripts mutate the tree by design; `git add -A` / `git add .` / `git commit -a` are banned after running one; stage explicitly and read the staged diff.

## CI parity run locally

`actionlint` (1.7.7) clean on `c8-precheck.yml`; `check-required-contexts.sh`, `check-docs-vs-ssot.sh`, `check-ci-job-claims.sh`, `check-doc-symbol-anchors.sh`, `check-doc-surface-completeness.sh`, `check-capacity-claims.sh`, `check-benchmark-claims.sh`, `check-l3-boundary.sh`, `check-hardcoded-literals.sh`, `check-vendor-literals.sh`, `check-mutation-marker.sh` all PASS; the full `l3-boundary-gate` job simulated end-to-end locally, green, `src/` clean afterwards. No Rust changed, so no cargo gate is in scope. CHANGELOG `[Unreleased]` → `### Security`.

### DIFF
--- a/.github/workflows/c8-precheck.yml
+++ b/.github/workflows/c8-precheck.yml
@@ -129,6 +129,15 @@ jobs:
     # gate HARD-BLOCKS `rqgm|epoch_manifest|red.?queen` in src/. The
     # ruled public identifiers (SignableEpochManifest, epoch.manifest_applied,
     # EpochAdvance, EPOCH_APPLIED) are gate-clean by construction.
+    #
+    # This job now carries a SECOND perma-banned-string-in-src/ gate (#3119,
+    # below). It rides HERE, as extra steps, rather than in a job of its own
+    # because this job's `name:` is a DECLARED REQUIRED CONTEXT (see
+    # scripts/qc-allowlists/required-contexts-release.txt) — so the new gate is
+    # merge-BLOCKING from the first commit, whereas a fresh job would be an
+    # unrequired context that a merge could ignore. The name is deliberately
+    # NOT changed: it is the exact string branch protection requires, and #2473
+    # is the record of what renaming it costs.
     steps:
       - uses: actions/checkout@v4
       - name: Run L3-boundary perma-ban gate
@@ -137,6 +146,30 @@ jobs:
         # Proves the gate is load-bearing: passes a clean tree, fails on
         # a planted `epoch_manifest` violation in a tmpdir, cleans up.
         run: bash scripts/check-l3-boundary.sh --self-test
+      - name: Run cert removal-proof mutation-marker gate (#3119)
+        # DEFENCE IN DEPTH for the #3118 near-miss.
+        # scripts/check-cert-removal-proof.sh rewrites production security
+        # controls in place to prove they are load-bearing; an aborted run left
+        # `inbound_write_namespace_authorized` short-circuited to always-allow
+        # (a cross-tenant inbound federated-write authorization BYPASS) in the
+        # working tree, and a `git add -A` pushed it to a PR branch. #3119
+        # fixed the harness (EXIT/INT/TERM/HUP trap + start guard + end
+        # assertion); THIS is the independent second line — the marker can
+        # never reach a merged commit under src/, however it got there.
+        run: bash scripts/check-mutation-marker.sh
+      - name: Self-test the mutation-marker gate (regression evidence)
+        # Plants the exact #3118 artefact in a throwaway src/ tree under
+        # .local-runs/ (must FAIL), against a clean near-miss control that
+        # names the harness without carrying the marker (must PASS), and an
+        # empty scan set that must fail CLOSED.
+        run: bash scripts/check-mutation-marker.sh --self-test
+      - name: Prove the harness restores an interrupted run (#3119)
+        # The other half of #3119: the harness's own --self-test drives the
+        # REAL harness against a REAL control under `timeout 2` and under a
+        # re-raised SIGINT, with the cargo step stubbed (no toolchain), and
+        # asserts `git status --porcelain src/` is empty and the marker absent
+        # afterwards. Removing the trap turns these two legs RED.
+        run: bash scripts/check-cert-removal-proof.sh --self-test
 
   hardcoded-literal-gate:
     name: Hardcoded-literal duplication ratchet (pm-v3.1)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,58 @@ write is permission-evaluated as; see the operator note below.
   Both now bind `resolve_read_visibility_caller()`, the same value the MCP/HTTP
   writer stamps. `--as-agent` remains the namespace-scope visibility knob.
 
+### Security
+
+- **The cert removal-proof harness can no longer leave a DISABLED security control
+  in the working tree** (#3119, closing the #3118 near-miss).
+  `scripts/check-cert-removal-proof.sh` is a mutation-testing harness: for each cited
+  control it rewrites the production source to an always-allow disposition, runs the
+  control's lane test, and asserts the test goes RED. Restoration previously happened
+  only on the happy and handled paths — there was **no `trap`** — so a SIGINT, a
+  `timeout` kill or a crash inside the cargo run left the control off, with nothing in
+  `git status` that reads as "a security control is disabled". On 2026-08-22 an aborted
+  run left `inbound_write_namespace_authorized` short-circuited to always-allow (a
+  cross-tenant inbound federated-write authorization **bypass**) and a `git add -A`
+  swept it onto a pushed PR branch for ~25 minutes (#3118 — never merged, no commit
+  ever contained it, base never affected). Four layers now stand between a mutation
+  and a commit:
+  - **Trap.** Every target is registered in a `MUTATED` array *before* it is rewritten
+    (so a crash between "mutate" and "run" is still covered), and an `EXIT`/`INT`/
+    `TERM`/`HUP` trap restores all of them with `git checkout HEAD --` — index *and*
+    worktree, which is what unwinds the staged `git add -A` variant that an index-only
+    `git checkout --` would have written straight back. Restoration is idempotent. The
+    `EXIT` trap carries the run's real status through unchanged (upgrading only a
+    success that failed to restore); the signal traps clear their handler and
+    **re-raise** (`kill -s "$sig" $$`) so the process dies of the original signal and
+    every parent observes `WIFSIGNALED` / 128+N exactly as with no trap installed.
+    The guard test now runs backgrounded under `wait` rather than in the foreground,
+    because bash defers a trap until the foreground command returns — the whole length
+    of a `cargo test` run was previously an un-interruptible window.
+  - **Start guard.** A marker already present in `src/` (an aborted earlier run) makes
+    the harness refuse to start, loudly and non-zero, instead of stacking a second
+    mutation on a disabled control. `--force-restore` is the deliberate recovery path:
+    it backs the mutated files up under `.local-runs/cert-54-evidence/` first, then
+    resets them to `HEAD` and exits 0.
+  - **End assertion.** The end of the run and the trap both assert the marker is absent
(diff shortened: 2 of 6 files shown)


alphaonedev:
**Fable gate — APPROVED (final review).** Verified in the diff: targets registered in `MUTATED` before rewrite; restore = `git checkout HEAD --` (index AND worktree — the #3118 shape was a staged mutation) guarded by the pre-mutation `git diff --quiet HEAD` check so developer edits are never clobbered; `on_exit` preserves rc (upgrades only success-without-restore → 9); `on_signal` clears traps and re-raises with `kill -s $sig $$` so WIFSIGNALED/128+N are truthful; guard run backgrounded under `wait` so the trap is not deferred behind a foreground cargo; start guard exit 4 + `--force-restore` with backup; scanning fails closed; `check-mutation-marker.sh` scans `src/` only and is wired as steps inside the already-required `l3-boundary-gate` job (gate → gate self-test → harness self-test ordering correct). Negative control reproducing the #3118 artefact accepted as evidence. Merge when required legs are green.

alphaonedev:
Fable gate — **FIX-BEFORE-MERGE** (scout-verified @ 93d09145; the ox 'script/trap/CI wiring/CHANGELOG don't exist' BLOCK is REFUTED — all are in the diff; design approved). (1) `scripts/check-mutation-marker.sh:56-62` `run_gate`: header promises exit 2 fail-CLOSED on an unscannable tree, but only the missing-`src/` arm does; `grep` rc=2 (unreadable file) / rc=127 fall through to PASS because the grep is an `if` condition. Fix: capture rc; 0→report+return 1; 1→return 0; else return 2; self-test leg with a `chmod 000` file asserting exit 2. (2) `check-cert-removal-proof.sh` `marked_src_files` uses `git grep` (tracked-only) — an untracked mutated file is invisible to preflight, end-assert and `--force-restore`; use `git grep --untracked`, keep the rc>1→2 arm. (3) `run_one()` step 3→4: assert `git diff --quiet HEAD -- "$tf"` after restore before the GREEN leg. (4) `:656-663` on revert failure: restore + dump + exit non-zero, do NOT continue to the next MAP row. (5) self-test fixed path → `mktemp -d` + RETURN trap (sibling `check-l3-boundary.sh:52-55`). (6) assert the MARKER literal still appears in the harness in `run_self_test`. (7) trap QUIT PIPE too; widen the marker scan to `src tests conformance examples sdk`. Governance nit (follow-up, not blocking): the gate rides as steps on `l3-boundary-gate` so it has no own context/ledger entry.

alphaonedev:
**Queue head after #3051.** Rebased onto `4ddf8bfc` (clean). Applied Fable FIX-BEFORE-MERGE at `e6fc7ed0`+this commit: (1) `run_gate` captures grep rc (0→found, 1→clean, else 2) + chmod-000 self-test; (2) `marked_src_files` uses `git grep --untracked` over src/tests/conformance/examples/sdk; (3) post-restore `git diff --quiet HEAD -- "$tf"`; (4) revert rc=3 dumps + `exit 3` (does not continue MAP); (5) marker self-test `mktemp -d` under `.local-runs` + RETURN trap; (6) both self-tests assert the MARKER literal still lives in the harness; (7) trap QUIT/PIPE. `--self-test` PASS on both scripts. Did **not** run a live removal-proof mutation of production. One `--force-with-lease` push.

Issue #3125, opened 2026-08-22
Design decision: should permissions enforce-mode REFUSE writes to namespaces with no governance configured (allow-on-silence → fail-closed)? Opt-in knob AI_MEMORY_PERMISSIONS_REQUIRE_GOVERNED_NAMESPACE shipped default-off


alphaonedev:
## Context (from the security fail-open remediation, 2026-08-22)

Finding FO-3 (ox-alpha security sweep, Opus 5 D verified): under `permissions.mode = enforce`, a write into a namespace with **no** governance policy configured is ALLOWED (allow-on-silence) on both backends. The fix-spec asked for a blanket refusal, on the premise that the compiled default mode is `Advisory`.

Facts established while implementing:
1. The premise is false: `AppConfig::effective_permissions_mode` (`src/config.rs:~7657`) resolves an unconfigured `[permissions]` block to **`Enforce`** (`governance::resolve_v07_default_mode(None)` → `default_v07_secure_mode()`). So `enforce` + zero policy IS the out-of-the-box posture — a blanket refusal would refuse every write on every fresh install.
2. Allow-on-silence is a **cutline-protected contract**: `tests/ship_gate_governance_inheritance.rs` ("failures here are release blockers") pins that an ungoverned subtree still allows while a sibling subtree is governed. A narrowing ("refuse only where governance is configured somewhere in the substrate") passed every sqlite suite and was caught RED only by the **pg lane** (S60 no-policy-leak assertion).

What shipped instead (PR from `fix/security-fail-open-gates`): the strict posture as an **opt-in** knob `AI_MEMORY_PERMISSIONS_REQUIRE_GOVERNED_NAMESPACE=1`, default off, byte-identical behaviour by default, with `ungoverned_namespace_allows_by_default_under_enforce` pinning no regression.

## Decision this issue owns (same class as #3111 capability Deny→Allow joiner)

Whether v1.x flips the default to fail-closed (refuse ungoverned namespaces under `enforce`), and if so with what migration/ramp:
- fresh installs must not brick (needs a bootstrap path: either an implicit root policy on first boot, or the knob pinned only in `asi-hard` with `doctor` reporting ungoverned namespaces first);
- `ship_gate_governance_inheritance` S60 contract must be consciously re-cut, not silently broken;
- both backends in lockstep with pg-lane evidence.

[...]

Issue #3111, opened 2026-08-22
[security][design-decision] capability-token grant joiner flips an EXPLICIT operator Deny→Allow — should a rule-derived Deny be terminal?


alphaonedev:
## Mechanism (VERIFIED in code, release/v1.0.0)
`apply_capability_grant` (`src/governance/capability.rs:838`) treats `Deny` AND `Ask` as flippable:
```rust
// Only Deny / Ask(pending) are flippable.
if !matches!(base, Decision::Deny(_) | Decision::Ask(_)) { return (base, GrantOutcome::NoOp); }
match verify(tok, cfg, req) {
    Ok(..) => (Decision::Allow, GrantOutcome::Granted{..}),   // Deny -> Allow
    Err(reject) => (base, GrantOutcome::Rejected(reject)),
}
```
The doc comment confirms this is intentional: *"a token that verifies and covers the request flips a Deny/Ask base to Allow"*. Wired into `enforce_governance` via `apply_at_gate` (`src/storage/mod.rs:~18849`, #1827 G10.1). `verify()` checks version/issuer/HMAC-chain/caveats/issuer-ceiling — it does **not** consult the permission rules that produced the base `Deny`.

## The question (NOT an unambiguous bug — a design decision)
Because token caveats and operator deny-rules are independent universes, a holder of a valid delegated token whose caveats cover the tuple can flip an **explicit operator Deny** (e.g. `[permissions] enforce` + "no writes to secrets/*") to Allow. This is intentional per the docs, but it appears to conflict with the standing directive quoted in `wire_check.rs` module docs — that permission rules "can NEVER be bypassed by AI/AI Agents — 100% of the time."

**Decision needed:** should the flip be restricted to `Ask/Pending` bases only, making a rule-derived `Deny(refusal)` **terminal** (never token-flippable)? Or is delegation-over-deny the intended trust model (a capability holder is, by construction, operator-authorized)?

## Suggested resolution
Owner ruling / T3 vote. If Deny must be terminal: change the guard to `if !matches!(base, Decision::Ask(_))` so only pending/approve bases flip, and add a test that an explicit deny rule survives a covering token.

---
_Surfaced by an ox-alpha security-review swarm; mechanism **independently verified** against release/v1.0.0 by Fable. Filed as a design-decision (not a confirmed defect) per truthful-labeling — the behavior is intentional-per-docs; the question is whether it should be._

alphaonedev:
[...]

PR #3133, opened 2026-08-22
fix(security): close fail-open governance + federation-trust gates

alphaonedev:
## What

Five security gates that **failed open** now fail closed. Every finding was
re-verified against `release/v1.0.0` source before any edit (R-405). One
candidate was **REFUTED**; one fix-spec premise turned out to be **factually
wrong**, and implementing that spec verbatim would have been a release-blocking
product break — that correction is the most important thing in this PR.

Related but **NOT changed here** (by instruction, and confirmed untouched by
this diff):
- **#3111** — the capability-token grant joiner flipping an explicit operator
  `Deny` → `Allow`. A design decision awaiting a ruling.
- **#3033** — the unsigned/unenrolled `Accept` arms in
  `receive_auth::authorize_remote_transition` /
  `authorize_remote_checkpoint_resolution` under `AI_MEMORY_FED_REQUIRE_SIG=0`.

## Findings: CONFIRMED vs REFUTED

| # | Finding | Verdict | Fixed |
|---|---|---|---|
| FO-1 | capability joiner `Deny`→`Allow` | CONFIRMED (mechanism) | **No — #3111 design decision** |
| FO-2 | `parse_presented_token` fails OPEN (invalid token → `None`, request proceeds) | **CONFIRMED** | Yes |
| FO-3 | `enforce_governance` allow-on-silence when no policy resolves | **CONFIRMED mechanism** — but the spec's premise is wrong and the default is a cutline-protected contract | Ships as an **opt-in posture**; default flip deferred to **#3125** |
| FO-4 | agent-action wire-points are no-ops when the `OnceLock` hook is unset | **CONFIRMED (shape)** | Partially — see below |
| FT-1 | unsigned / signed-but-unenrolled transition `Accept` arms | CONFIRMED | **No — #3033** |
| FT-2 | `peer_attestation::namespace_allowed` "unknown/absent peer" arm | **REFUTED** | n/a |
| FT-3 | mTLS↔`x-peer-id` binding skipped for unbound certs under `enforce` | **CONFIRMED** | Yes |
| FT-4 | FED-RQ-03 stale-policy gate maps a governance-read `Err` → ACCEPT | **CONFIRMED** | Yes |

## CORRECTION to the FO-3 fix-spec — please read before reviewing

The spec's plan was *"under `enforce`, absence of policy REFUSES"*. Implementing
it verbatim would have been a release-blocking product break, for two reasons
the spec did not account for.

**1. The spec's stated premise is factually wrong.** It asserted *"the compiled
default is Advisory (`config.rs:5398` `fn default() -> Self { Self::Advisory }`)"*
and concluded a blanket refusal would be inert out of the box.
`AppConfig::effective_permissions_mode` (`src/config.rs:7657`) deliberately
bypasses that serde default: an absent or `mode`-less `[permissions]` block
resolves through `governance::resolve_v07_default_mode(None)` →
`default_v07_secure_mode()` → **`PermissionsMode::Enforce`** (pinned by
`governance/mod.rs` `default_v07_secure_mode_is_enforce` and
`resolve_v07_default_mode_unconfigured_yields_enforce_with_warning`). So
`enforce` + zero configured policy IS the shipped out-of-the-box posture, and a
blanket refusal would refuse **every write on every fresh install**.

**2. Allow-on-silence is a CUTLINE-PROTECTED ship-gate contract, not an
oversight.** `tests/ship_gate_governance_inheritance.rs` ("CUTLINE-PROTECTED …
Failures here are release blockers") asserts an ungoverned subtree still Allows
*while a sibling subtree IS governed* — "ungoverned subtrees remain opt-in
(compatibility preserved)". `tests/governance_postgres_inheritance.rs` S60
asserts the same across siblings, as the proof that a parent's policy does not
LEAK into an unrelated namespace. My first narrowing ("refuse only on a
substrate that has governance configured somewhere") passed every sqlite suite I
ran and was then **caught red by the pg lane** on exactly that S60 assertion —
which is the honest reason the pg-lane evidence requirement exists.

Refusing writes into tenant B's ungoverned subtree because tenant A configured
governance is arguably worse than the hole being closed, so flipping this
default is a product-semantics decision (owner ruling / T3 vote), not
remediation.

**What landed instead:** the strict posture ships as an **opt-in** knob,
`AI_MEMORY_PERMISSIONS_REQUIRE_GOVERNED_NAMESPACE=1`. Engaged, it refuses under
`enforce` with a typed `GovernanceRefusal` naming every remedy, on both backends
from one SSOT. **Unset — the default — every code path is byte-identical to
before**, pinned by `ungoverned_namespace_allows_by_default_under_enforce`,
which reproduces the exact governed-sibling shape the ship gate protects. The
certified enterprise-federation deployment is told to engage it
(`docs/compliance`). **The default flip is deferred to #3125**, filed alongside
#3111 and cited in the knob's doc comment, the CHANGELOG entry and below.

### FT-2 — why REFUTED

`namespace_allowed`'s fall-through arms are **default-DENY**, not fail-open:
`sync_trust_peer_bypass()` is `false` unless the operator sets the documented
legacy `AI_MEMORY_FED_SYNC_TRUST_PEER=1`. Independently, no load-bearing caller
reaches a permissive arm by accident: `/sync/since` resolves the scope itself
with an explicit `None => false` (`federation_sync_since.rs:319-330`), the
push/delete lanes gate through the `has_allowlist()` /
`peer_declares_namespace_scope` wrappers (#2447/#2491), and the third call site
(`federation_receive.rs:1300`) is WARN-only observability (#2479). Changing it
would remove a documented operator opt-out without closing anything.

### FO-4 — why only partially fixed

The unset-hook `Ok(())` leg is real, but it is the **documented CLI one-shot
exemption** (`wire_check.rs` layering rationale 3: *"CLI one-shot binaries MUST
NOT install the hook — the operator's direct ops stay unimpeded"*). Rather than
delete an operator-facing design, this PR adds `wire_check::check_governed`,
which refuses on an unset hook, and migrates only the sinks that are
**structurally daemon/MCP-only** (verified by call-graph, not assumption):

- `mcp/tools/skill_export.rs` ×2 — reachable only via MCP dispatch
  (`mcp/registry.rs` + the `mcp/mod.rs` re-export); `run_mcp_server` installs
  the hook before dispatching.
- `hooks/executor.rs` ×2 — the only production `ExecutorRegistry::from_hooks`
  sites are in `src/mcp/mod.rs`.
- `federation/sync.rs` ×1 (`post_once`) — every `broadcast_*_quorum` /
  `bulk_catchup_push` caller lives in `src/handlers/` or
  `mcp/tools/store/transport.rs`; the `ai-memory sync` daemon does NOT reach it
  (its cycle issues its own signed GET against `/sync/since`).

`llm.rs::check_outbound` deliberately stays on `check` because it IS
CLI-reachable (`cli/curator.rs`, `cli/commands/atomise.rs`,
`cli/commands/expand.rs`). The four raw-`rusqlite::Connection` CLI write sites
catalogued in `storage/connection.rs:314-321` are untouched: a
`--bypass-governance` operator flag is a new CLI surface with its own SSOT
reconciliation and belongs in its own PR.

## Behaviour changes (call-outs)

All are in the CHANGELOG, and the posture-relevant ones are also in
`docs/compliance/ENTERPRISE-FEDERATION-CERTIFICATION.md`.

1. **NEW opt-in `AI_MEMORY_PERMISSIONS_REQUIRE_GOVERNED_NAMESPACE=1`** — under
   `enforce`, a namespace with no resolvable policy is REFUSED. **Default OFF =
   zero behaviour change** (see the correction above); **default flip deferred
   to #3125**. `advisory` / `off` unaffected in every case.
2. **A presented-but-unparseable capability credential ⇒ 403
   `GOVERNANCE_REFUSED`** (was: silently dropped; the request proceeded
   anonymously on the bare ACL, byte-indistinguishable from presenting
   nothing). OMITTING the credential is unchanged, and
   `[capabilities].enabled = false` still returns `Ok(None)` without parsing
   (zero new audit bytes). The forensic `capability-reject` edge-parse row's
   decision changes `warn` → `deny`.
3. **`AI_MEMORY_FED_CERT_PEER_BINDING=enforce` refuses an unbound cert and a
   bound cert with no `X-Peer-Id`** → `401 peer_id_cert_unbound`. `warn` (the
   DEFAULT and the documented rollout posture) and `off` are unchanged.
   Documented honestly in the fn docs: because `ClientCertPeerId` is `None`
   both for an unmapped fingerprint AND for a connection presenting no client
   cert, `enforce` also refuses a non-mTLS request arriving over the
   peer-binding acceptor.
4. **FED-RQ-03 refuses `503 policy_read_unavailable` (retryable) when the local
   governance policy read keeps failing**, instead of accepting. Three
   attempts, linear 10/20 ms backoff, and the retry budget applies only when
   the gate is enabled so the disabled path costs exactly one read as before.
   The happy path is byte-identical.

`wire_check::check_governed` is a structural guarantee, not a posture flip:
`serve` and `run_mcp_server` both install the hook during bootstrap before
dispatching anything, so no production behaviour changes today.

## SSOT / gates touched

- `CLAUDE.md` env-table **row 168** for the new
  `AI_MEMORY_PERMISSIONS_REQUIRE_GOVERNED_NAMESPACE` const — mechanically
  required by `scripts/check-docs-vs-ssot.sh`'s `ENV_VAR_CENSUS` rule, which
  hard-fails when production code declares an `AI_MEMORY_*` const the table
  never mentions. Only that one table row was added.
- `tests/qual_10_module_size_ceiling.rs`: `src/storage/mod.rs` 29_160 → 29_260,
  bumped in lockstep with a dated comment (measured 29_170 post-fmt).
  `src/store/postgres.rs` stays at its existing ceiling — the final shape
  removed more lines there than it added.
- `scripts/check-hardcoded-literals.sh` green with **no baseline update** —
  `"ai_memory::governance"` was hoisted to
  `governance::GOVERNANCE_GATE_TRACE_TARGET` and every pre-existing site now
  references the const (the literal's site count goes to zero, not up).

## asi-hard: decided NO, and why

Checked against the #3114 precedent for a default-OFF opt-in knob
(`AI_MEMORY_MIGRATION_REQUIRE_CORE_TABLES`), which registered a CLAUDE.md
env-table row + a `doctor` fact and was **not** added to the `asi-hard` pin set.
Note `docs/deploy/asi-hard.env` does not enumerate knobs at all — it sets
`AI_MEMORY_SECURITY_PROFILE=asi-hard` and the profile pins the set in
`security_profile::KNOBS`, count-asserted by `asi_hard_pins_documented_set`
(21).

**This knob is deliberately NOT pinned there**, for a reason that is a property
of the table itself: every existing entry already defaults fail-closed, which is
what makes pinning *"a NO-OP for a compliant deployment — it only removes the
ability to DISABLE them"* (the table's own #3033 comment). This knob defaults
**OFF**, so pinning it to `1` would be the first entry that is **not** a no-op:
it would change behaviour for every existing `asi-hard` deployment, refuse
writes into their ungoverned subtrees, and thereby **pre-empt #3125 for that
slice of the fleet** — deciding by side effect the exact question that was
deferred. The knob-count stays 21 and `asi-hard.env` is untouched.

Revisit as part of **#3125**: if the default flips ON, the `asi-hard` pin
becomes a no-op and should follow in lockstep (21 → 22). Recorded that way in
the knob's doc comment and in `docs/compliance`.

**Doctor surface** (the #3114-style registration): `ai-memory doctor` now emits
a `require_governed_namespace` fact in the governance section, deliberately
adjacent to the pre-existing `namespaces_without_policy` — together those two
facts ARE the blast radius. When the posture is engaged under `mode = enforce`
and that count is non-zero, the section raises to `Warn` with a note naming the
count and both remedies.

## Follow-up recommended (not folded in)

1. **#3125 — flip the `AI_MEMORY_PERMISSIONS_REQUIRE_GOVERNED_NAMESPACE`
   default to ON.** The actual closure of FO-3. Needs an owner ruling / T3
   vote: it breaks a cutline-protected ship gate and changes multi-tenant
   semantics.
2. **Pin the knob in `security_profile::KNOBS` (21 → 22)** — correct only
   *after* #3125 lands ON, at which point the pin becomes a no-op. Touches the
   **certified** doc (`ENTERPRISE-FEDERATION-CERTIFICATION.md`), so it is a
   cert-scope change. See the asi-hard section above.

Also observed, not changed: the `MemoryStore::enforce_governance_action` trait
DEFAULT (`store/mod.rs:3323`) returns `Allow`. No production backend uses it
(sqlite and postgres both implement), but it is a fail-open default for any
future adapter.

## Rule IDs

ERRORS-01 (recoverable failure ⇒ `Result`, not accept), ERRORS-02, ERRORS-06,
ERRORS-09 (make the illegal state — "credential presented but unverifiable" —
representable rather than aliased to Allow), ERRORS-10/12 (`CapReject` is
already `Error + Send + Sync + 'static`), ERRORS-15 (`map_err` preserves the
cause), ERRORS-19 (no silently discarded `Result`), ERRORS-14 (lowercase, no
trailing punctuation), API-07, PERF-03, CONCURRENCY-16 (the `OnceLock` usage
itself is sound), TEST-02/TEST-03.

## Tests

Every fix has a regression test that reproduces the fail-open pre-fix and
asserts the refusal post-fix, plus a control proving the documented opt-out
still works.

- `tests/permissions_mode_gate.rs` —
  `enforce_mode_ungoverned_namespace_is_deny_under_strict_posture` (the
  reproduction, knob engaged), **`ungoverned_namespace_allows_by_default_under_enforce`**
  (the no-regression guard: reproduces the governed-sibling shape the
  cutline-protected ship gate pins, and asserts Allow),
  `advisory_mode_no_policy_still_allows_under_strict_posture`. The pre-existing
  `ship_gate_governance_inheritance` and pg `S60` suites run **UNMODIFIED** as
  the compatibility proof.
- `src/governance/capability.rs` —
  `presented_but_unparseable_token_is_refused_not_downgraded`;
  `parse_presented_token_is_feature_gated_and_fail_closed` (renamed from
  `…_and_lenient`).
- `tests/governance_wire_check_fail_closed.rs` (**new crate**, never installs a
  hook so the unset-hook leg is observable) — reproduces `check`'s fail-open leg
  AND pins `check_governed`'s refusal on every `AgentAction` variant.
- `tests/mtls_cert_peer_binding_2045.rs` —
  `legacy_cert_without_binding_is_refused_under_enforce` (**flipped** from
  `…_never_bricks_under_enforce`),
  `bound_cert_without_asserted_peer_id_is_refused_under_enforce`, plus two
  `warn`-posture controls.
- `tests/fed_rq_03_policy_refuse_stale.rs` —
  `policy_read_fault_refuses_503_and_does_not_apply` (the fault is induced by
  renaming the column the count query reads, and asserted non-vacuous),
  `policy_read_fault_still_accepts_when_gate_disabled`.
- `tests/g10_capability_tokens.rs` — edge-parse assertions updated to the
  fail-closed shape; the feature-OFF leg is pinned unchanged.

---


alphaonedev:
**Fable gate — APPROVED (final security review).** Read every `src/` hunk. (1) `parse_presented_token` → `Result<Option<_>, CapReject>`: presented-but-unparseable credential refused (403 `GOVERNANCE_REFUSED`, audit `deny`) across 8 HTTP / 6 MCP / 3 CLI edges; absent credential and `[capabilities].enabled=false` unchanged (`Ok(None)`). (2) `wire_check::check_governed` fails closed on an uninstalled hook for the 5 daemon/MCP-only sinks; `check` keeps the documented CLI exemption (`llm.rs`) — structural, no production change. (3) FO-3 ships as opt-in `AI_MEMORY_PERMISSIONS_REQUIRE_GOVERNED_NAMESPACE`, default byte-identical on BOTH backends (sqlite `ungoverned_namespace_decision` / pg twin share `governance::ungoverned_namespace_refusal`), S60 cutline contract untouched, default-flip → #3125, not pinned in asi-hard (correct: every existing pin is a no-op for compliant deployments; this one would not be). (4) cert↔x-peer-id: unbound cert / missing header now 401 under `enforce` only; default `Warn` verified in `tls.rs:487`. (5) FED-RQ-03: bounded retry (3×, 10/20ms) then 503 `policy_read_unavailable` (retryable) when `require_policy_current_enabled()`; `=0` opt-out still accepts. Refutation of the `namespace_allowed` finding accepted (default-DENY). Literal hoist to `GOVERNANCE_GATE_TRACE_TARGET` keeps the ratchet baseline unchanged. pg-lane evidence (34/0 with `--include-ignored`) + sqlite 62/0 accepted. Merge when required legs are green; queued after the code PRs ahead of it.

alphaonedev:
Fable gate — **FIX-BEFORE-MERGE** (scout-verified R-405 @ 93d09145; ox batch-2 review, base-tree false positives discarded). (1) **HIGH, functional regression:** `src/mcp/tools/skill_export.rs:165` and `:246` switched to `wire_check::check_governed`, which refuses when `GOVERNANCE_PRE_ACTION` is unset; the hook installer has exactly two callers (`daemon_runtime.rs:5684` inside `bootstrap_serve`, `src/mcp/mod.rs:3975`) but `ai-memory skill export` is a CLI one-shot (`src/cli/commands/skill.rs:499` → `handle_skill_export`) → hard-refuses with `HOOK_NOT_INSTALLED_REASON`, contradicting the PR's own contract at `wire_check.rs:134`. Revert both sites to `wire_check::check` (llm.rs CLI-exemption rationale) or install the hook on the CLI skill path; add an end-to-end CLI regression test. (2) `src/governance/refusal.rs:169-174` `require_governed_namespace()` accepts only `1|true` while the repo grammar is `1|true|yes|on` (`governance/audit.rs:1443 env_flag_enabled`) — `=yes` silently stays FAIL-OPEN; reuse `env_flag_enabled`. (3) `src/handlers/mod.rs:204-207` `capability_from_headers`: a PRESENTED non-UTF-8 header collapses to `None` = absent = bare ACL — the exact fail-open the PR says it closed; return the 403. (4) `federation_receive.rs` `policy_read_unavailable_response` echoes the raw rusqlite error to a federation PEER — opaque token on the wire, raw under the trace target. (5) `src/cli/doctor.rs:1192-1196` the new warning is overwritten by the unconditional pending-age `note = Some(…)`; use `notes: Vec`. Documented scope note (no change required): `federation_receive.rs:143` no-extension arm still proceeds under `enforce` — state the single-mTLS-listener precondition in the cert doc. UNVERIFIED to check before merge: `doctor_governance_coverage` counts `namespace_meta` rows only — namespaces existing only as memory rows would be refused but not counted in the blast-radius warning.

alphaonedev:
**Grok gate — Fable HIGH FIX-BEFORE-MERGE applied on rebase onto `ae3e0aec`.** Head `732c995a` G.

Rebase conflicts: CLAUDE.md KEEP BOTH (`AI_MEMORY_MIGRATION_REQUIRE_CORE_TABLES` row 168 + `AI_MEMORY_PERMISSIONS_REQUIRE_GOVERNED_NAMESPACE` row 169); `federation/sync.rs` keeps HEAD `Result<RequestBuilder, String>` return + `check_governed` (daemon NetworkRequest); QUAL-10 `storage/mod.rs` MEASURED 29_314, ceiling 29_260 → 29_400.

Fable HIGH (2026-08-22T15:11:56Z), all five:

1. **HIGH:** `src/mcp/tools/skill_export.rs` SKILL.md + resource writes reverted `check_governed` → `check` (CLI `ai-memory skill export` never installs `GOVERNANCE_PRE_ACTION`). Pinned by `tests/governance_wire_check_fail_closed.rs::skill_export_uses_check_not_check_governed`.
2. `require_governed_namespace()` now uses `env_flag_enabled` (`1|true|yes|on`). Pinned by `require_governed_namespace_uses_shared_truthy_grammar`.
3. `capability_from_headers`: PRESENTED non-UTF-8 header is 403 `GOVERNANCE_REFUSED`, not `None`/bare ACL. Pinned by `presented_non_utf8_capability_header_is_403`.
4. `policy_read_unavailable_response` no longer echoes raw rusqlite/IO `detail` to a federation peer (opaque tag + retry note on the wire; raw under attestation trace). Pinned by `fed_rq_03` assertion `body.detail` is absent.
5. `section_governance` composes notes via `append_note` so the strict-admission blast-radius warning is not overwritten by pending-age.

Cert doc: single-mTLS-listener precondition for the no-extension `enforce` arm recorded.

Local CI-parity: fmt + 5 clippy legs + examples + targeted sqlite tests + QUAL-10/6-7 + docs-vs-ssot/hardcoded-literals/symbol-anchors/surface-completeness GREEN. pg-lane on scratch DB `ai_memory_3133_scratch` (not shared v90): `permissions_mode_gate` **8/8 RAN** including the new grammar test (`--include-ignored`). Full `store::postgres` suite was a script-parse mistake and was aborted — not a product failure.

Waiting 34/34 on `732c995a`.

alphaonedev:
**macos-fed sqlite Check** failed `tests/wire_check_sole_path_pin.rs` (executor + federation/sync still pinned to `wire_check::check(...)` after those sinks moved to `check_governed`). Lockstep on `ee43b64c` G. Skill export / llm stay on `check`. 9/9 local. Waiting 34/34 on `ee43b64c` (do not rerun `732c995a`).

alphaonedev:
**Lint / Postgres feature gate / vectorlite** failed `clippy::doc_markdown` on `tests/wire_check_sole_path_pin.rs` module docs (`ProcessSpawn` / `NetworkRequest` needed backticks). Local default clippy does not compile tests. Fixed `6a394835` G after `cargo clippy --all-targets --features sal -- -D warnings -D clippy::all -D clippy::pedantic` green. Waiting 34/34 on `6a394835`. Do not rerun `ee43b64c`/`732c995a`.

alphaonedev:
**Per-Module Coverage** failed `cov_federation_sync_quorum_arms` (20 tests): `check_governed` refused loopback fanout because the coverage binary never installs `GOVERNANCE_PRE_ACTION` (not `bootstrap_serve`). DLQ reason was `HOOK_NOT_INSTALLED` instead of `http 500`. Fix `f9ecc746` G: Allow-all hook once per process. 25/25 local `--features sal`. Waiting 34/34 on `f9ecc746`. Do not rerun `6a394835`.

alphaonedev:
**Per-Module Coverage** failed next on `cov_ga2_pg_adapter2` (4 quorum-ack tests) — same `check_governed` unset-hook refuse. Remaining fanout helpers now install Allow-all `GOVERNANCE_PRE_ACTION` (`878e4040` G). 6/6 `broadcast_*` local `--features sal-postgres`. Waiting 34/34 on `878e4040`. Do not rerun `f9ecc746`.

alphaonedev:
**Lint fmt** failed wrapping on the Allow-all hook-install lines. rustfmt `7cb67e1b` G. Waiting 34/34 on `7cb67e1b`. Do not rerun `878e4040`.

alphaonedev:
**Per-Module Coverage** failed next on `fed_outbound_checkpoint_route_2391` (W=2 got 202 not 200 — `check_governed` refused the mock peer). Remaining coverage fanout crates now install Allow-all hook (`61f91bfd` G). 4/4 local `--features sal`. Waiting 34/34 on `61f91bfd`. Do not rerun `7cb67e1b`.

alphaonedev:
**Coverage follow-up on `61f91bfd` → `ace0b328` G (FF, no force).**

`61f91bfd` was 33/34: only Per-Module Coverage Thresholds RED (job 97184135810). llvm-cov `--tests` failed closed on first binary:

`tests/federation_stable_peer_id_2442.rs::replay::decommissioned_peer_dlq_row_is_never_delivered_to_a_survivor_2442`
left=`[]` right=`["for-C"]` — `replay_once` hit `check_governed` with `GOVERNANCE_PRE_ACTION` unset (no `serve` bootstrap under coverage), so the positive-control delivery never landed.

Fix: first-writer-wins Allow-all hook in `replay::config_for` (same rustfmt as sibling fanout suites). Production `check_governed` unchanged.

Local: `cargo test --features sal,sal-postgres --test federation_stable_peer_id_2442 -- --test-threads=1` = **10 passed / 1 ignored** (pg `#[ignore]` without URL; sqlite replay arms ran).

Fable HIGH 1–5 still ESTABLISHED at `ace0b328` (quoted):
1. `skill_export.rs:171/:254` `check` not `check_governed`
2. `refusal.rs:170-171` `env_flag_enabled`
3. `handlers/mod.rs:196-241` non-UTF-8 header → 403
4. `federation_receive.rs:259-276` no rusqlite `detail` on the wire
5. `doctor.rs:813` `append_note`

Waiting 34/34 on `ace0b328`. Do not `gh run rerun` `61f91bfd`.

alphaonedev:
**Coverage on `ace0b328` 33/34 RED** — Per-Module Coverage Thresholds job 97188657068. Next binary after #2442 DLQ replay: `tests/g3_hooks_stderr_drain.rs` (0/3). `DaemonExecutor`/`ExecExecutor` now `check_governed` ProcessSpawn; coverage `--tests` has no `serve` bootstrap → HOOK_NOT_INSTALLED (timeout test reported 500ms Timeout on the refused spawn).

**`22930753` G** (FF, no force): first-writer-wins Allow-all in `cfg_for`/`write_script` of `g3_hooks_stderr_drain`, `hooks_executor_test`, `hooks_pre_recall`, `hooks_timeout_budget`, `hooks_hot_reload`. Production `check_governed` unchanged. Local `g3_hooks_stderr_drain` 3/3.

Waiting 34/34 on `22930753`. Do not rerun `ace0b328`.

alphaonedev:
**`22930753` required red:** vectorlite feature gate + Postgres feature gate (`clippy --all-targets --features vectorlite` and `clippy --features sal-postgres --tests`). `clippy::items_after_statements`: `write_script` ran `GOVERNANCE_PRE_ACTION.set` then `use std::io::Write`.

**`2d3fc61d` G** (FF): uses first, then the set. Local both clippy legs green. Production unchanged.

Waiting 34/34 on `2d3fc61d`. Do not rerun `22930753`.

alphaonedev:
**`2d3fc61d` 33/34 RED** — Coverage job 97193544512. Next binary: `tests/integration.rs` (219 passed, 1 failed). `test_quorum_partial_failure_with_timeout` agent-register **202** (under-replicated) because `check_governed` refused outbound POSTs with HOOK_NOT_INSTALLED.

**`97e1871f` G** (FF): Allow-all in `federation_cfg_for_test`. Local 1/1 with `AI_MEMORY_REQUIRE_AGENT_ATTESTATION=0` (suite opt-out already used by `cmd()` children; CI's other 219 OneshotDaemon 201s already run under that posture). Production `check_governed` unchanged.

Waiting 34/34 on `97e1871f`. Do not rerun `2d3fc61d`.

alphaonedev:
**`97e1871f` required red:** Check (ubuntu-latest,sqlite) job 97197961274 — `tests/integration.rs::test_curator_autonomy_end_to_end_cycle` (215 passed, 1 failed). Curator `--json` stdout was not JSON (`expected value` line 1 column 1). `--json` is `CuratorArgs`, not a global; a leading `--json` is not applied.

Coverage / Lint / vectorlite / Postgres feature gate were green or still in progress. Do not rerun `97e1871f`.

**`e47e2cfb` G** (FF): `--json` after `curator`; dump stdout/stderr on parse failure. Local 1/1. Waiting 34/34 on this SHA.

alphaonedev:
**`e47e2cfb` still Check (ubuntu-latest,sqlite) RED** (job 97200308495). Same test: stdout WAS a valid CuratorReport JSON (dry_run=true, errors=[no LLM client]) prefixed so parse failed at column 1; stderr empty. Cause: GitHub-hosted Check inherits `RUST_LOG=info`; curator INFO cycle line lands on stdout ahead of `--json`.

Lint / vectorlite green; Coverage still in progress. Do not rerun `e47e2cfb`.

**`13be0c85` G** (FF): `cmd()` pins `RUST_LOG=error`; e2e test parses from the first `{`. Local 1/1 under inherited `RUST_LOG=info`. Waiting 34/34 on this SHA.

alphaonedev:
**`13be0c85` required red:** vectorlite + Postgres feature gate — `clippy::map_unwrap_or` at `tests/integration.rs:10905` (`find('{').map(...).unwrap_or(...)`). Ubuntu sqlite / Coverage were still in progress. Do not rerun `13be0c85`.

**`44b22fb7` G** (FF): `map_or`. Local clippy `--test integration --features vectorlite -D pedantic` green. Waiting 34/34 on this SHA.

alphaonedev:
**Fable gate — APPROVED (final code + security review) at `44b22fb7`.** Shadow-QC lane, 2026-08-23.

All five FIX-BEFORE-MERGE items ESTABLISHED on this head: (1) `skill_export.rs:171/:254` back on `wire_check::check` + regression pin `tests/governance_wire_check_fail_closed.rs:114-135`; (2) `refusal.rs:171` → `env_flag_enabled` (repo truthy grammar); (3) `handlers/mod.rs:206-213` presented non-UTF-8 capability header → 403, not `None`; (4) `federation_receive.rs:262-265` opaque `POLICY_READ_UNAVAILABLE_TAG` on the wire, detail only under the trace target; (5) `doctor.rs:1844-1915` `notes: Vec` joined — the blast-radius warning is no longer overwritten.

**Allow-all `GOVERNANCE_PRE_ACTION` installs in the test fan-out (commits `f9ecc746`…`97e1871f`) — reviewed and ACCEPTED, with the reasoning recorded so it is not mistaken for a gate-weakening later:** the three `check_governed` sinks on this head are `federation/sync.rs:123` (`build_governed_peer_post` ← `post_once` / `bulk_catchup_push` ← `handlers/bulk.rs:785` / daemon only) and `hooks/executor.rs:671/:1046` (`HookChain::fire` ← `mcp/mod.rs:1675`, `hooks/recall.rs:167` only). No `src/cli/**`, curator, or background path reaches any of them. Both installers are UNCONDITIONAL (`daemon_runtime.rs:5718` in `bootstrap_serve`; `mcp/mod.rs` → `install_governance_pre_action_hook`), so every production process that can reach a sink has the hook before the sink is reachable; the failing tests were harnesses that bypass bootstrap, and `tests/governance_wire_check_fail_closed.rs:520-549` still proves HOOK-absent ⇒ refuse in its own process. First-writer-wins `OnceLock` means the Allow-all can never displace a real installer.

Follow-ups (not blocking; file post-merge if not already tracked): (a) the skill-export regression pin is structural (`check` vs `check_governed`), not an end-to-end `ai-memory skill export` subprocess test — an e2e pin would also catch a future hook-installer refactor; (b) `doctor_governance_coverage` (`doctor.rs:1208`) still counts `namespace_meta` rows only — the UNVERIFIED note from the prior gate stands as a documented limitation of the blast-radius warning, not a fail-open.

Merge when 34/34 on `44b22fb7` via `fable-merge.sh`.

alphaonedev:
**`959883cb` G** (FF on `44b22fb7`): Per-Module Coverage Thresholds RED on `44b22fb7` — `handlers/mod.rs` measured **79.63% < floor 90%** (GLOBAL 92.75% PASS; llvm-cov tests all passed). Floor not lowered.

Cause: fail-closed `capability_from_headers` grew ~54 production lines; only the non-UTF-8 403 arm had a unit test.

Added `capability_from_headers_tests` arms (6/6 local `--lib` + clippy lib `-D warnings -D clippy::all -D clippy::pedantic` green):
- omitted header → `Ok(None)` (bare ACL; R9 additive-only)
- blank UTF-8 → `Ok(None)`
- `[capabilities].enabled=false` + garbage UTF-8 → `Ok(None)` (feature-off short-circuit)
- enabled + unparseable UTF-8 → 403 via `map_err`
- enabled + wrong-version envelope (`cap9:`) → 403
- existing non-UTF-8 403 unchanged

Do **not** rerun `44b22fb7`. 34-watcher armed on `959883cb` against live protection API. Fable APPROVED at `44b22fb7`; this commit is test-only coverage of those same arms.

alphaonedev:
**Fable gate — APPROVED extended to `959883cb`.** Delta since `44b22fb7` is one commit, `src/handlers/mod.rs` `#[cfg(test)]` only (coverage-floor tests for `capability_from_headers` arms: absent→`Ok(None)`, disabled→`Ok(None)`, non-UTF-8→403). No production hunk. Merge when 34/34 on `959883cb`.

Issue #3127, opened 2026-08-22
[data-integrity][pg-parity] postgres HTTP/MCP search path DROPS the source_uri filter (dispatches through trait search) — returns WRONG results, not fewer; pg search_with_source_uri has no production caller


alphaonedev:
Surfaced by PR #3112 (#3110 gate review, 2026-08-22). Quoting the PR's reachability statement, verified by Fable against the handlers:

> The pg `search_with_source_uri` method is `pub` but has **no production caller today**: the postgres HTTP search path dispatches through the SAL trait `search` (which is gated) and **drops `source_uri` entirely**, so the only current callers are tests.

Consequences:
1. On the postgres backend a `?source_uri=<uri>` search (q + source_uri compose path, `src/handlers/memories_query.rs` / `src/mcp/tools/search.rs`) silently ignores the filter → the caller receives rows from every source, i.e. **WRONG results** (prime directive: worst case must be fewer results, never wrong ones). Sqlite honours it (`crate::storage::search_with_source_uri`, fail-closed visibility at ?15).
2. The reciprocal-provenance compose feature (#1580) is therefore sqlite-only in practice — a parity gap that the claims audit should check against the docs (`docs/` advertise `source_uri` filtering without a backend caveat?).

Fix (pg-parity lane): wire the postgres branch of both callers to `PostgresStore::search_with_source_uri(ctx, …)` — now visibility-gated by #3112 — with a `handler_postgres_branches_fake_pg` + live-tier test asserting the filter is applied on pg (rows from another source_uri excluded), and a parity test twin in `tests/parity_*`. Until wired, either document the pg caveat or make the pg path **refuse** `source_uri` (400/unsupported) rather than silently drop it — silent drop is the one unacceptable option.

Depends on #3112 merging first (signature with `ctx`).

alphaonedev:
**Closed by merge #3252** (`1080ea68` G, parent `df7d36d5`).

Evidence: pg keyword search SSOT `search_with_source_uri` honours Filter.since/until/source_uri. Tests RAN: `pg_search_filter_ssot_3185` 3 passed on :5445 TLS. CI 34/34 at `df7d36d5`. Grok gate: https://github.com/alphaonedev/ai-memory-mcp/pull/3252#issuecomment-5418600343

Issue #3110, opened 2026-08-22
[security][data-integrity] postgres search_with_source_uri has NO caller visibility gate — cross-agent private-row leak (sqlite fails closed, pg fails OPEN)


alphaonedev:
## Summary
`PostgresStore::search_with_source_uri` (`src/store/postgres.rs:6569`) returns `scope=private` memories to **any** caller, regardless of ownership — while the sqlite SSOT twin (`src/storage/mod.rs:6637`) correctly fails closed. A cross-backend privacy divergence: the same query leaks other agents' private rows on postgres.

## Verified evidence (release/v1.0.0)
**sqlite twin** (`storage/mod.rs:6637`) — has the visibility gate:
- takes `caller: Option<&str>` (param ?15), embeds `visibility_clause(11, 15, "m")` in its WHERE, documented **"None = fail-closed (no private rows)"** (#1720 A3).

**postgres** (`store/postgres.rs:6569`) — WHERE clause is:
```sql
WHERE m.tsv @@ plainto_tsquery('english', $1)
  AND ($2 IS NULL OR m.namespace = $2)
  AND ($3 IS NULL OR m.tier = $3)
  AND (m.expires_at IS NULL OR m.expires_at > NOW())
  AND ($4/$5 created_at bounds)
  AND ($6 IS NULL OR m.source_uri = $6)
```
No `scope`/`metadata->>'agent_id'`/caller predicate, and no SAL `is_visible_to_caller` post-filter (result is `rows.iter().map(Self::row_to_memory_scan)...collect()`). Binds are only query/namespace/tier/since/until/source_uri/limit — **caller is never passed**.

Contrast: the pg `search` (`postgres.rs:18995`) and `list` (`:18745`) both carry the `COALESCE(metadata->>'scope','private') <> 'private' OR metadata->>'agent_id' = $caller` gate; `search_with_source_uri` was missed.

## Impact
Postgres backend + a `source_uri`-scoped query returns other tenants' `scope=private` rows to an unauthorized caller. Same class as the earlier IDOR/private-leak fixes; violates the North-Star (never leak private data) and fails-open where it must fail-closed.

## Fix
Add the caller visibility predicate to the pg WHERE (mirror `search`/`list` at :18995/:18745) and thread the caller through the bind list; or apply the SAL `is_visible_to_caller` post-filter. Add a cross-backend parity test asserting a private row is withheld from a non-owner on both backends.

---
[...]

PR #3134, opened 2026-08-22
fix(parity): pg/sqlite divergences batch 2 — signal prune, action edge atomicity, AGE projection fail-open, store-backend gaps

### LINKED ISSUE #2373: Postgres check-side db_id parity for the #2370 per-database rollback anchor (verify-audit-trail + any future pg open-time check)
## Context (deferred from #2370 / PR fix/2370-per-db-head-anchor)

#2370 scoped the #1946 off-table head-anchor log PER DATABASE via a signed v3 `db_id` (genesis-derived, `db_id_from_genesis_parts`). That PR shipped:

- sqlite OPEN-time check + `verify_audit_trail` db_id filtering (full matrix in `tests/rollback_evidence_1946.rs`);
- v3 db_id EMISSION on BOTH backends (`try_emit_audit_head_witness` sqlite + `pg_emit_audit_head_witness_in_tx` postgres, incl. the C2b NULL-`sequence` legacy-chain fallback to v2).

**Explicitly deferred (this issue):** the postgres CHECK side. `PostgresStore::verify_audit_trail` passes `db_id = None` into `compute_rollback_verdict_for_report`, i.e. the CONSERVATIVE count-every-anchor posture — byte-identical to the pre-#2370 verdict (over-detect toward Evidence, never suppression). A second healthy pg-backed database sharing one witness mount can therefore still surface a FALSE `RollbackCheck::Evidence` in its `verify-audit-trail` REPORT (not a boot refusal — pg has no sqlite-style open-time refuse path today).

## Proposed fix

- Thread an async genesis fetch (`GENESIS_ROW_SQL`, already shared with the emission twin) into the pg `verify_audit_trail` flow BEFORE the sync shared verdict fn, passing `db_id.as_deref()` — mirroring `src/signed_events.rs::verify_audit_trail`'s sqlite wiring (~15 LOC in `src/store/postgres.rs` around the `compute_rollback_verdict_for_report(head_sequence, None)` call).
- Extend the `#[ignore]`+`AI_MEMORY_TEST_POSTGRES_URL` suite with a two-sibling-DB no-cross-Evidence report assertion (K3 parity with `two_dbs_sharing_one_anchor_log_do_not_cross_contaminate`).
- If/when a pg open-time rollback check is added, it must consume the same db_id filter from day one.

## Safety posture until fixed

Fail-closed direction only: the deferred side can over-report Evidence on pg, never suppress genuine evidence. Disclosed in the #2370 PR body and in `src/store/postgres.rs` at the `compute_rollback_verdict_for_report` call site.

alphaonedev:
## What

Eight cross-backend PARITY defects in the **#3110 class** — a gate/check/cap the sqlite SSOT holds that the postgres adapter is missing or implements divergently. Every item is a **leak, wrong-results, or unbounded/unreconciled-state** defect, not a capability gap. Findings came pre-verified from the ox-alpha tier; each was **re-verified against current `release/v1.0.0` code** before any fix (verdicts below).

| # | Divergence | Class |
|---|---|---|
| 1 | `signal_inbox` re-served ACKED signals on pg (no `acknowledged_at IS NULL`, while `signal_ack` stamps-and-KEEPS) | wrong results / work re-dispatched forever |
| 2 | expired signals NEVER reaped on pg (`expires_at` persisted, no reaper anywhere) | unbounded accumulation; documented TTL contract false |
| 3 | `action_add_edge` cycle probe + INSERT were two independent pool statements | **corrupted ordering DAG** under concurrency → wedged frontier |
| 4 | AGE projection/unprojection failures SWALLOWED on the sync paths | permanent relational↔graph divergence, no self-heal, no signal |
| 5 | `search_with_source_uri` silently dropped `tags_any`/`agent_id` **before** `LIMIT` | wrong page, not merely a wider one |
| 6 | `kg_query` UNCAPPED on both pg branches (sqlite clamps 200/1000) | unbounded result set + divergent results |
| 7 | `POST /api/v1/notify` stored a DIFFERENT priority per backend | divergent durable row; pg wrote out-of-domain values |
| 8 | pg `verify_audit_trail` passed `db_id = None` (#2373) | backend-dependent tamper verdict; **false rollback conviction** |

Plus PERF-07/ERRORS-07 hygiene on two `usize as i64` limit casts and one provably-guarded `unwrap()`.

## Re-verification (R-405)

**CONFIRMED (9 of 10 findings, all fixed here):** pg `search_with_source_uri` tags/agent drop; AGE fail-open on hard-delete + link-insert; `prefix_upper_bound` unwrap; both `limit as i64` casts; pg signal prune gap; `action_add_edge` pg atomicity; pg `signal_inbox` ack gate; notify priority narrowing; pg `verify_audit_trail` db_id.
**Also confirmed and fixed:** pg `kg_query` row cap (checkpoints-kg F-1).
**REFUTED / not carried:** `chrono_ms as i64` timestamp narrowing (no such symbol in tree); `find_paths` visibility error-path split (unproven by the verifier, left unfixed rather than fixed on speculation).
**Confirmed but DISCLOSED rather than implemented** (capability gaps, per the #3064 minimum): the skills plane (hard 501 on pg), persona mint/read, and the erasure cold tier.

## Design notes worth a reviewer's attention

- **#3008 lock granularity.** Endpoint-row `FOR UPDATE` is provably insufficient: two *disjoint* new edges can co-close a cycle, so a writer locking only its own endpoints cannot see the other. One tx-scoped global advisory key reproduces exactly what the sqlite connection mutex already does, on a low-volume coordination surface. `_xact_` scope means it cannot leak on an error path; single-key means it is trivially deadlock-free (CONCURRENCY-04).
- **AGE fail-open closed with NO schema change.** Reuses `kg_projection_outbox`. The unprojection marker rides a reserved `relation` value that is unforgeable because `memory_links.relation` is a closed typed set *plus* a SQL CHECK — pinned by `kg_unproject_marker_is_not_a_valid_link_relation`, which fails the build if a future variant ever collides. Without that property the marker would be a caller-triggerable node-deletion primitive; with it, the reconciler can only ever detach a node whose relational row it re-confirms absent *inside the same tx*. The #1783 vote's best-effort posture is preserved (the delete is never undone); what changed is that the residual is now durable and reconcilable instead of a log line.
- **The drainer now runs in sync mode too.** Recording drift is pointless if nothing drains it. On a healthy sync deployment the queue is empty and each tick is one indexed no-op count.
- **#2373 is a de-escalation of a FALSE POSITIVE, not of detection.** A foreign anchor carries a different `db_id` by construction, so scoping suppresses no evidence about *this* store. Removal proof below shows the pre-fix pg store convicting itself off a sibling's anchor.
- **`normalize_priority` fixes a latent inversion on ALL surfaces.** The old inline expression `i32::try_from(raw).unwrap_or(i32::MAX).clamp(1,10)` saturated a value below `i32::MIN` *upward*, so `priority = i64::MIN` was stored as the MAXIMUM priority 10. The SSOT clamps in `i64` space first, making the narrowing total and the saturation directional. sqlite behaviour is otherwise unchanged.

## Test evidence — live enterprise-fed pg18.6 + AGE tier

`AI_MEMORY_TEST_POSTGRES_URL` = the enterprise-fed tier; every pg test run with `--include-ignored`.

```
tests/pg_parity_batch_2.rs                        11 passed; 0 failed
  pg_action_add_edge_concurrent_opposing_arcs_cannot_both_land_3008 ... ok
  pg_action_add_edge_refuses_self_and_cycle_3008 ................... ok
  pg_gc_reaps_expired_signals_3011 ................................. ok
  pg_kg_query_is_row_capped ........................................ ok
  pg_list_by_namespace_prefix_saturates_a_huge_limit ............... ok
  pg_search_with_source_uri_honours_tags_any_and_agent_id .......... ok
  pg_signal_inbox_excludes_acked_3011 .............................. ok
  sqlite_action_add_edge_refuses_self_and_cycle_3008 ............... ok
  sqlite_gc_reaps_expired_signals_3011 ............................. ok
  sqlite_kg_query_is_row_capped .................................... ok
  sqlite_signal_inbox_excludes_acked_3011 .......................... ok

tests/rollback_evidence_1946.rs                   13 passed; 0 failed
  postgres_emission::pg_check_side_scopes_rollback_verdict_by_db_id_2373 ... ok
tests/store_parity_gaps.rs                        47 passed; 0 failed
tests/cov_ga2_postgres.rs                         19 passed; 0 failed
tests/cov_postgres_kg.rs                          10 passed; 0 failed
cargo test --features sal --lib -- notify priority   56 passed; 0 failed
tests/qual_10_module_size_ceiling.rs               2 passed
tests/qual_6_7_legacy_error_type_ceiling.rs        2 passed
```

**Removal proof** (fixes reverted, tests re-run against the live tier — both fail, so the tests are load-bearing):

```
pg_signal_inbox_excludes_acked_3011 ... FAILED
pg_check_side_scopes_rollback_verdict_by_db_id_2373 ... FAILED
  assertion `left == right` failed
   left: Evidence { anchored_head: 5, db_head: 3 }   <- pg falsely convicted by a FOREIGN sibling's anchor
  right: NotApplicable
```

*(One incidental note: `store_parity_gaps.rs` uses fixed-id fixtures — `pg-g1`, `pg-g3-*`, `pg-1024-version` — that are not re-run-idempotent against a persistent shared DB; a second run trips `expected_version=1 but stored version=2`. Pre-existing, unrelated to this PR, and I reset those rows on the shared tier after observing it. Worth a follow-up.)*

## CI-parity checklist

- `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` — **green on all four legs**: default, `sal`, `sal-postgres`, `vectorlite`
- `cargo check --examples` — green
- `cargo fmt --all` — applied
- `scripts/check-hardcoded-literals.sh` — PASS (no baseline update; four repeated SQL literals hoisted to named consts, a net burn-down)
- `scripts/check-vendor-literals.sh` — PASS; `scripts/check-docs-vs-ssot.sh` — PASS (no SSOT pin moves: no new MCP tools, routes, CLI verbs, or schema versions)
- `qual_10` module-size ceiling bumped 34_620 → 35_060 in lockstep with a dated rationale; `qual_6_7` unchanged and green

## Merge-order note

This PR touches `PostgresStore::search_with_source_uri` in the **same lines** as **#3112** (which adds the `&CallerContext` visibility gate to that method). The two changes are orthogonal and additive — #3112 adds the caller/scope predicate, this PR adds the `tags_any`/`agent_id` predicates — but they will conflict textually. The combined shape is `$7 = tags`, `$8 = agent_id`, `$9 = caller`, `LIMIT $10`; whichever merges second should rebase to that. This PR does not overlap **#3115** (archive lanes).

**#3127** (pg HTTP/MCP search dropping `?source_uri=`) is deliberately **not** folded in: its wiring depends on #3112's new signature, and stacking a third change on the same method would compound the merge risk above. Recommend a follow-up PR once #3112 lands.

Closes #2373 (the postgres check-side db_id parity follow-up — the only issue this PR fully resolves).

Related, NOT closed here: **#3110** is the visibility half of the same method and is owned by **#3112**; this PR only adds the orthogonal `tags_any`/`agent_id` half. **#3011** (coordination-plane retention) is partly advanced — the `signals` lane now reaps on postgres, matching sqlite — but the actions/checkpoints/routine_runs lanes remain deferred there, so it stays open. **#3064** (postgres-unreachable MCP tools) gains the exact disclosure it asked for but no implementation, so it stays open. **#3008** is already closed; the postgres atomicity hole was a residual its sqlite fix did not cover.


alphaonedev:
**Fable gate — APPROVE-CONDITIONAL (final review in progress).** Accepted on the diff: tx-scoped `pg_advisory_xact_lock` on `action_add_edge` (correctness over throughput, documented; endpoint-row locks provably insufficient); `__ai_memory_unproject__` outbox marker — verified on the live tier that `kg_projection_outbox` has no `relation` CHECK/FK so the enqueue works, and `memory_links`' closed typed set makes the marker unforgeable (unit test pins `from_str` = None); `models::normalize_priority` as the cross-surface SSOT (fixes the i64::MIN saturation inversion; in-band values pass through); `kg_query_row_cap` mirrors sqlite 200/1000; signal inbox `acknowledged_at IS NULL` + expired-signal reaper; #2373 `db_id` scoping with removal proof. **Condition:** the marker → drainer → `unproject_memory_from_age` reconciliation path has no live-tier test (only the `from_str` unit check). That path IS the fail-open closure; it needs pg-lane evidence: enqueue marker (via fault injection or a minimal test seam) → drainer tick → `projected_at` set + AGE vertex gone, plus the dedupe negative. Approved for merge once that test is in and green on `:5445`; queued after the code PRs ahead.

alphaonedev:
## Review condition addressed — marker lane now has live-tier evidence

Head is now `9d09ee5f47d73e85d57c315a4f60632c8d11b00c` (single push, SSH-signed `%G? == G`). No CI re-run triggered from my side.

### Writing the test found a real defect in the marker arm — fixed in the same commit

The drainer's marker arm called `unproject_memory_from_age`, which is the wrapper that **deliberately swallows an AGE runtime failure and returns `Ok(())`** for the hot delete path. Inside the drainer that meant a failed detach would stamp the marker `projected_at` **while the ghost node stayed attached** — the reconciler reporting work it did not do, and silently defeating the retry/quarantine ladder for exactly this lane. It now calls the propagating `unproject_memory_from_age_inner`, so an AGE failure routes into the drainer's own attempt-bump/quarantine path, symmetric with how a failed `project_link_into_age` is already handled. The `Err` arm rolls the per-row tx back, so an aborted tx is not a problem.

### On the seam

I took the `#[doc(hidden)] pub` option you named — `record_unreconciled_unprojections_for_test`, a thin forwarder following the crate's existing convention (`config::override_active_permissions_mode_for_test`). I checked first that neither the *Surface stability* job nor `check-doc-surface-completeness.sh` enumerates `pub fn`s on `PostgresStore` (they pin named symbols and CLI/route/skill-tool doc coverage), so this adds no gate obligation.

I did look for a fault knob to drive the real failure instead. There is no `AI_MEMORY_TEST_*` AGE-fault hook, and the only available injection is revoking the role's AGE access — which on a **shared** tier would break every concurrent test. So the tests reproduce the exact post-failure **state** (relational row deleted by raw SQL, ghost vertex still attached) and then drive the real recording fn and the real drainer arm. Everything except the induced AGE outage is production code.

### Three live-tier tests (AGE-gated on `store.kg_backend()`; skip cleanly on a CTE backend)

```
test pg_age_unprojection_marker_records_dedupes_and_reconciles ... ok
test pg_age_unprojection_marker_never_detaches_a_recreated_id .. ok
test pg_age_quarantined_marker_does_not_block_the_drain ........ ok

tests/pg_parity_batch_2.rs        14 passed; 0 failed
tests/rollback_evidence_1946.rs   13 passed; 0 failed
```

Mapped to your numbered asks: (1) memory + link so an AGE vertex exists, asserted `= 1`; (2) recording driven through the seam after a raw-SQL relational delete, with the ghost vertex re-asserted as surviving; (3) marker row present with `projected_at IS NULL`; (4) `drain_kg_projection_outbox(256)`; (5) `projected_at` set **and** cypher vertex count `= 0`; (6) second enqueue deduped to exactly one pending row.

Two additions beyond the list, both cheap and both about not turning a repair tool into a destructive one:

- **`..._never_detaches_a_recreated_id`** — if the id came back before the drain, the stale marker is dropped and the **live** projection survives (`count == 1`). Without the liveness guard the reconciler would delete a legitimate vertex.
- **Stamped, not deleted** — the consumed marker keeps a `projected_at` row so the operator retains a record of what was repaired.

**Quarantine** (your "if cheap"): seeded a marker at `MAX_AGE_PROJECTION_ATTEMPTS` rather than burning 100 live attempts. It asserts the poison row stays **pending** — never silently stamped reconciled, so the unrepaired drift stays visible in the pending-depth gauge — **and** that a healthy marker enqueued behind it still reconciles in the same pass (no head-of-line block).

### Removal proof

Disabling the drainer's marker arm fails the reconcile test on the discriminating assertion:

```
pg_age_unprojection_marker_records_dedupes_and_reconciles ... FAILED
  assertion `left == right` failed: the ghost :Memory vertex MUST be gone ...
   left: 1
  right: 0
```

### Honest scope

The marker arm's **failure** path (AGE down during reconciliation → attempt bump → quarantine) is covered by the shared drainer machinery and by the exclusion test, **not** by an induced live AGE outage. Calling that out rather than implying full coverage.

### Gates re-run after the change

clippy exact CI flags green on all four legs (default / `sal` / `sal-postgres` / `vectorlite`); `cargo check --examples`; `cargo fmt --all`; `check-hardcoded-literals.sh`, `check-vendor-literals.sh`, `check-docs-vs-ssot.sh` all PASS; `qual_10` + `qual_6_7` green with no further ceiling bump needed (the tests live in `tests/`, and the `src/store/postgres.rs` delta is ~20 lines inside the 63 lines of headroom from the earlier bump). Scope held to the new tests plus the seam and the one-line detach-call correction — no other changes.

alphaonedev:
**Fable gate — APPROVED (final review) at `9d09ee5f`.** Condition met and it surfaced a real defect: the drainer's marker arm used the swallowing `unproject_memory_from_age` (hot-path wrapper) so a failed AGE detach would have stamped `projected_at` with the ghost vertex still attached — now the propagating `_inner`, symmetric with `project_link_into_age`. Live-AGE tests accepted: record→dedupe→reconcile (ghost count 1→0, marker STAMPED not deleted), never-detach-a-recreated-id (repair tool can't become destructive), quarantined marker does not head-of-line-block; removal proof on the discriminating assertion. Test seam `#[doc(hidden)] pub …_for_test` follows the crate convention and adds no surface-gate obligation (verified by owner). Honest scope note on the induced-AGE-outage path accepted. Merge when required legs are green; queued after #3133.

alphaonedev:
Fable gate — **APPROVED with FOLLOW-UP** (scout-verified @ 93d09145; both ox 'high' findings REFUTED: `signals.expires_at` is BIGINT (`postgres_schema.sql:1334`) so the i64 bind is correct; sqlite is also `tags_first`; zero `#[ignore]` in the test file). Fold into this PR at rebase if cheap, else they become the follow-up issue: (1) `tests/pg_parity_batch_2.rs:626-650` PERF-07 saturation test is VACUOUS — one seeded row passes on the unfixed `LIMIT 1` collapse; seed ≥3 rows, assert all returned at `limit=usize::MAX`, add a `limit=2` leg. (2) `src/store/postgres.rs:8716`/`:8811` both call `kg_query_row_cap(None)` — the `limit` param is dead in production and the new unit test exercises unreachable arms; sqlite honours the caller's limit (`storage/mod.rs:10897-10898`); also the doc claims 'identical to what sqlite returns' but the tie-break differs (`storage/mod.rs:11002-11004`) → thread the limit, soften the doc. (3) `live_pg()` `:178-187` turns a DEAD postgres in the pg CI tier (URL set) into a silent GREEN — distinguish env-unset⇒skip from connect-failed⇒panic. (4) `Filter.tags_any` (`store/mod.rs:742`) advertises OR-across-tags that neither backend delivers (both `.first()`) — rename or implement.

alphaonedev:
**Rebased onto \`76417e85\` G (Merge #3136).** Head \`a6483ea5\` G.

Conflict in \`search_with_source_uri\`: kept BOTH #3110 visibility (\`$7\` caller_opt + \`is_visible_to_caller\`) AND batch-2 \`tags_any\`/\`agent_id\` (\`$8\`/\`$9\`, \`LIMIT \$10\`). QUAL-10 postgres.rs measured 35_166 / ceiling 35_200.

Fable FOLLOW-UPs (vacuous PERF-07 test, \`kg_query_row_cap(None)\`, \`live_pg()\` soft-skip, \`tags_any\`≡first) **not folded** — not cheap during the 5-hunk postgres.rs conflict. Remain follow-ups.

Waiting 34/34 on \`a6483ea5\` then \`fable-merge.sh 3134\`.

alphaonedev:
**Postgres feature gate fix pushed** \`10da1324\` G. Did **not** rerun the old head.

E0061: #3110 added \`CallerContext\` as arg 1 of \`search_with_source_uri\`; batch-2 tests still used the 3-arg shape. Four call sites now pass \`&ctx\` + \`Some(uri.as_str())\`.

Waiting 34/34 on \`10da1324\`.

PR #3129, opened 2026-08-22
ci(throughput): re-home no-pg/no-cargo gates to GitHub-hosted; keep tier + warm-cache jobs self-hosted

alphaonedev:
Closes #3128

## The problem

#3109 (merged as `15fe44e4`) re-homed CI onto the self-hosted enterprise-fed fleet **wholesale** — **39 job definitions / 43 expanded** onto **six** runner instances (four `linux-fed` on one Linux host, two `macos-fed` on one macOS host) shared across every open pull request, with no priority ordering.

`Classify changes` was among them. It is the **root of `ci.yml`'s `needs:` graph** — `lint`, `actionlint`, `msrv`, `check` (×4), `postgres-feature`, `mobile-cross-compile` (×2), `sal-feature`, `vectorlite-feature` and `dockerfile-validate` all wait on it — so while it sat `queued`, the four REQUIRED `Check (…)` legs could not start at all. With 7 PRs open, whole runs sat queued 40+ minutes. Head-of-line blocking on the one job that must never wait.

The slots were largely spent on jobs the fleet does nothing for: all 22 `c8-precheck.yml` gates are `bash scripts/check-*.sh` over the checked-out tree, and `actionlint`, `Dockerfile build (no push)`, `Build-script custom-build ledger gate (#2635)` and `Surface stability (load-bearing symbols)` are the same shape.

## The rule

Stated **once**, at the head of `.github/workflows/ci.yml`; every re-homed job carries a `# RUNNER PLACEMENT` comment citing it.

| Class | Test | Home |
|---|---|---|
| `pg-tier` | reads the node-local tier-URL file, `CREATE DATABASE`s an ephemeral db on `:5445`, or points `AI_MEMORY_TEST_POSTGRES_URL` / `AI_MEMORY_TEST_AGE_URL` at it | self-hosted |
| `warm-cache` | `cargo build` / `test` / `clippy` over a non-trivial slice of the crate, where warm-vs-cold is minutes | self-hosted |
| `script-only` | everything else — `bash scripts/check-*.sh`, git/grep/python3, container-action steps, dependency-graph *resolution* that never compiles, Docker builds whose layer cache is `type=gha` | `ubuntu-latest` |

Plus one absolute: **a job that other jobs `needs:` must never be self-hosted unless it is itself `pg-tier` or `warm-cache`.** A queued decider stalls its whole graph.

## Placement — before / after

[...]

PR #3137, opened 2026-08-22
fix(ci): bench --report-only — CLI smoke must not be a perf gate on non-reference hardware

alphaonedev:
## Defect

`cmd_bench` (`src/daemon_runtime.rs`) `anyhow::bail!`s whenever ANY operation's measured p95 exceeds its `target_p95_ms` by more than 10% (`budget_failed`). **Two** integration tests ran `bench` and asserted `status.success()`, so both were **undeclared performance gates living inside the required `Check` matrix**.

The one that fired: `tests/integration.rs::test_cli_smoke_canonical_paths` step 19. Measured by Fable on CI run [32556214239](https://github.com/alphaonedev/ai-memory-mcp/actions/runs/32556214239), job **`Check (linux-fed,sqlite)`** on runner **f2-linux-fed-2** — it ran while the rest of the suite had the box at **load 36** and panicked at `tests/integration.rs:12447` ("bench failed"). On GitHub-hosted VMs the same step ran isolated and passed.

`PERFORMANCE.md` pins the budgets to reference hardware (`ubuntu-latest`, hardware multiplier 1.0), and the performance gate is **one dedicated job** — `.github/workflows/bench.yml`, which builds `--release` and runs the workload as that job's only work. A debug binary on a contended shared runner measures nothing about the performance contract.

## Change

1. **`BenchArgs` gains `--report-only`.** Every operation is still measured and its `pass`/`fail` status still printed, and regression rows still render. `"report_only": true` is added to the `--json` envelope and a one-line notice goes to **stderr** in table mode (mirroring the existing `--history` notice, so the rendered table on stdout stays byte-identical). Only the process verdict changes. The doc comment states why.
2. **The verdict moves into the pure `bench::verdict`** (`src/bench.rs`) so the fatal-vs-advisory decision is unit-testable without spawning the binary. Fatal semantics are **byte-identical when the flag is absent** — the same three message shapes (budget / regression / both). `--report-only` downgrades BOTH `budget_failed` and `regression_failed` to advisory.
3. **Step 19** passes `--report-only`, keeps the existing JSON assertions, and adds: `report_only == true`; `results` is a **non-empty array**; each result has `operation`, `measured_p95_ms`, `target_p95_ms`, and `status` in {`pass`, `fail`}.
4. **`test_cli_bench_emits_json_with_seven_results_and_passes_budget`** (the latent flake flagged in review) likewise passes `--report-only`. **Every shape assertion is kept** — exactly 8 results, all eight documented per-result fields, `status` in {`pass`, `fail`} — plus the new `report_only == true`. Renamed **`test_cli_bench_emits_json_with_eight_results_report_only`**: the old name claimed *seven* results while asserting *eight*, and claimed a budget pass it no longer makes. Its doc comment now states where the budget verdict lives and why it is not asserted there.
5. **Unit tests** (`src/bench.rs` `mod tests`) pin both directions — a synthetic `Status::Fail` row is `Err` containing `"exceeded its p95 budget"` without the flag and `Ok` with it; the same for a regressed baseline row; plus a passing-run control.
6. **`MACOS_BUDGET_MULT` doc comment** cited the renamed test *and justified itself by that test asserting exit 0* — a rationale that no longer holds. Rewritten to state what still makes the multiplier load-bearing (it decides the `Status::Pass`/`Fail` every result reports, in every caller). **The constant's value and behavior are unchanged.**
7. **Docs/SSOT/CHANGELOG**: `docs/CLI_REFERENCE.md` bench section gains the flag row, an example, and a paragraph on why; `CHANGELOG.md` `[Unreleased]` -> **Fixed**, covering both tests (base entries untouched).

`BenchArgs` carries `#[allow(clippy::struct_excessive_bools)]` with a comment: the fourth bool is a 1:1 CLI flag, and collapsing them into an enum would change the published CLI surface.

## Audit of every `bench` invocation in `tests/`

| Site | Asserted exit 0? | `--baseline`? | Disposition |
|---|---|---|---|
| `tests/integration.rs` step 19 (`test_cli_smoke_canonical_paths`) | yes | no | **fixed** — `--report-only` + shape assertions |
| `tests/integration.rs:8020` `..._seven_results_and_passes_budget` | yes | no | **fixed** — `--report-only`, all shape assertions kept, renamed `..._eight_results_report_only` |
| `tests/bench_verified_dispatch_1961.rs:17` (2 tests) | **no** — already documents the exact failure mode | no | already immune, no change |
| `tests/cov_ga2_misc.rs:465` `daemon_cmd_bench_scale_path_runs` | **no** — accepts `Ok` or a `bench:` error | no | already immune, no change |
| `tests/cli_integration.rs:106`, `tests/integration.rs:12229` | `bench --help` only | n/a | no change |
| `tests/integration.rs` `FAILURE_CASES` | asserts **failure** on invalid args | n/a | no change |

No `bench` invocation in `tests/` asserts a budget verdict any more.

## Rebase onto 48e52192 (post-#3129)

#3129 moved `bench.yml` **back to GitHub-hosted `ubuntu-latest`** and codified
the runner-placement rule at the head of `ci.yml`. That **resolves** the
doc-vs-reality drift an earlier revision of this branch had to hedge around:
the enforcing gate now demonstrably runs on the hardware the budgets are
calibrated to, so every comment in this PR states that directly instead of
hedging. The required `Check (…)` legs remain `[self-hosted, linux-fed]`
(warm-cache), so the defect this PR fixes is unchanged.

Rebase was clean — no CHANGELOG conflict; base entries intact, this PR's entry
is a pure addition.

## Verification (local, CI-parity)

- `cargo check --all-targets` — pass
- `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` on **default**, **sal**, **sal-postgres**, **vectorlite** — all four pass
- `cargo fmt --all`
- `cargo test --test integration` -> `test_cli_smoke_canonical_paths` + `test_cli_bench_emits_json_with_eight_results_report_only` — both pass
- `cargo test --lib bench::tests` — 28 pass
- `cargo test --test qual_10_module_size_ceiling --test qual_6_7_legacy_error_type_ceiling` — 4 pass, no ceiling bump needed
- `scripts/check-docs-vs-ssot.sh` — PASS (no pin change required)
- `scripts/check-vendor-literals.sh` — PASS

## North Star

No data path touched. `bench` seeds a disposable `:memory:` DB and never reads or writes the operator's DB. The change is exit-code-only and **fails open in the direction of reporting, never of hiding**: the `fail` status of every over-budget operation is still measured, still printed, and still machine-readable in the JSON envelope. `--report-only` is opt-in and never the default, so the enforcing `bench.yml` gate is unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **The required `Check` legs are no longer undeclared performance gates:
+  `ai-memory bench` gains `--report-only`.** `cmd_bench` bails non-zero
+  whenever any operation's measured p95 exceeds its target by more than 10%, so
+  the TWO integration tests that ran `bench` and asserted `status.success()`
+  were in effect asserting the performance contract from inside the required
+  `Check` matrix. On GitHub-hosted VMs they ran isolated and passed; on the
+  self-hosted `Check (linux-fed,sqlite)` leg
+  `tests/integration.rs::test_cli_smoke_canonical_paths` step 19 ran while the
+  rest of the suite had the box at load 36 and went red (CI run 32556214239).
+  `PERFORMANCE.md` §"Measurement methodology" calibrates the budgets to
+  GitHub-hosted `ubuntu-latest` (hardware multiplier 1.0) and exactly one gate
+  measures them there — `.github/workflows/bench.yml`, which runs on that
+  runner class and builds `--release`. A debug binary on a contended
+  self-hosted box compared against an `ubuntu-latest`-calibrated target is a
+  measurement read off the wrong machine. The new `--report-only` flag
+  still measures every operation, still prints each `pass`/`fail` status and
+  every regression row, and adds `"report_only": true` to the `--json` envelope
+  plus a one-line stderr notice in table mode — it only stops the
+  budget/regression verdict from failing the process. Both tests now pass
+  `--report-only` and assert envelope SHAPE instead: step 19 gains non-empty
+  `results` plus per-row `operation`, `measured_p95_ms`, `target_p95_ms` and
+  `status` in `{pass, fail}`; `test_cli_bench_emits_json_with_seven_results_and_passes_budget`
+  keeps all eight-operation and per-field assertions, gains `report_only ==
+  true`, and is renamed
+  `test_cli_bench_emits_json_with_eight_results_report_only` (the old name
+  claimed seven results while asserting eight, and claimed a budget pass it no
+  longer makes). Fatal semantics are unchanged when the flag is absent.
 - **Record-stop enforcement no longer silently degrades at the SAL layer when
   the sqlite DB path resolves through a symlink** (e.g. the macOS
   `/var -> /private/var` temp dir). The `SqliteStore` write-funnel gate keyed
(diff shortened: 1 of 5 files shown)


alphaonedev:
**Fable gate — APPROVED (final) at `06cd5830`.** `bench::verdict` carries the three fatal arms verbatim (byte-identical without the flag) with an early `Ok` under `--report-only`; `report_only: true` in the JSON envelope + stderr notice in table mode; smoke step 19 and the (renamed, eight-result) budget test now assert envelope shape only, with the budget verdict left to `bench.yml` on reference hardware — the only place p95 budgets are valid. Queued immediately after #3129 because it de-flakes a REQUIRED leg for every PR; expect a rebase ping if the base moves first.

Issue #2571, opened 2026-07-31
[portability] neither export mode carries archived_memories or namespace_meta — now DECLARED, still not round-trippable


alphaonedev:
Found while closing #2490. PR #2568 added `archived_memories` + `namespace_meta` to the `excludes` scope marker so the artifact stops asserting a scope it does not have — a **truthfulness** fix. The **capability** gap is this issue.

## Verified by execution (release/v1.0.0 @ `b2a9f5bd`, unchanged by #2568)

A source database holding 2 live memories, 2 archived rows and 1 `namespace_meta` binding, exported and imported into a fresh destination:

| table | source | destination after round trip |
|---|---|---|
| `memories` | 2 | 2 |
| `memory_links` | 1 | 1 |
| `archived_memories` | 2 | **0** |
| `namespace_meta` | 1 | **0** |

`storage::export_all` reads only the live `memories` table; `namespace_meta` is read by neither exporter. `ExportEnvelope` (`src/portability/emit.rs`) has no field for either class.

## Why each matters

- **`archived_memories`** is where #1725 in-place-edit snapshots, `gc` evictions and operator archives live. `ai-memory undo-edit` and `memory_archive_restore` both read it. A portability artifact that drops it silently discards the only copy of every pre-edit revision and every evicted memory the operator chose to keep. Note the v85/#2035 and v87/#2333 work went to some trouble to make archive→restore lossless *within* a database — that fidelity does not survive an export.
- **`namespace_meta`** carries `standard_id` (which memory holds the governance policy) and `parent_namespace` (the explicit inheritance link `build_namespace_chain` walks). A restored corpus therefore comes back **ungoverned** — allow-on-silence (#1569) — even though the standard memories themselves round-trip. Post-#2503 that is sharper still: a namespace whose binding is absent is treated differently from one whose binding is severed.

## Scope note against the spec

`docs/spec/PORTABILITY-V2.md` §V2-2 does not enumerate either class, so this is **not** a spec violation — it is a spec gap. Either the spec adds them to the L1 data classes (and `ExportEnvelope` grows two arrays, which is a `deny_unknown_fields` wire change and therefore needs to land before any v1.0.0 binary is in the field), or the spec states plainly that archive + governance bindings are `backup`-only and the docs stop implying `export --full` is a complete portable copy.

That fork is a T1/T4 decision and wants its own 5-agent vote.

## Related

- #2490 / PR #2568 — the marker now declares the omission
[...]

Issue #2209, opened 2026-07-18
[BLOCKING][#2006/PR#2205] v2 import pre-commit re-verify covers signed_events only — tampered memory_revisions/agent_lineage commit unverified; non-empty-dest merge can commit forked chains


alphaonedev:
**Fable pre-merge audit finding on PR #2205 (`src/portability/import.rs`), severity: BLOCKING (fail-closed guarantee does not hold for two signed spine classes).**

## Defect

The fail-closed spine gate in `import_full_envelope` is exactly: `verify_audit_trail` → `chain_intact && sequence_gaps.is_empty() && !TruncationCheck::Detected`. That predicate covers the **`signed_events` chain only**. Two other §V2-2 signed spine classes are staged RAW and **never re-verified before commit**:

1. **`memory_revisions`** — rows cross with their raw `prev_hash`/`sequence` and are `INSERT OR IGNORE`d with **no chain walk** (there is no revisions-chain verifier in the codebase at all — only `read_revision_chain_head`; `canonical_revision_chain_bytes` exists but nothing replays it). Proof: take a valid bundle, mutate `env.memory_revisions[0].namespace` (or its `prev_hash`) → import **succeeds and commits** the tampered row. On an unsigned daemon (`signature=None`, the default posture) the tamper is then **undetectable at rest forever**. The PR's own claim — "a malformed / tampered / truncated bundle is REJECTED" — is false for this class.
2. **`agent_lineage`** — PK is `(agent_id, epoch)` with `INSERT OR IGNORE`. `verify_audit_trail` *does* compute a lineage verdict (`AuditTrailReport.lineage: LineageCheck`, src/signed_events.rs:1698, via `identity::lineage` walk), but the import gate **ignores `audit.lineage`** — a bundle whose lineage records are Forged/tampered commits. Worse, on a dest that already holds lineage for agent X: overlapping epochs keep dest rows, non-overlapping bundle epochs append rows whose `prev_record_hash` points at the *bundle's* predecessor → a **forked/broken key-succession chain is committed**, poisoning that agent's future identity verification.

Note `memory_revisions.sequence` has **no UNIQUE constraint** (migrations/sqlite/0056_v72), so a non-empty-dest import commits duplicate/conflicting revision sequences with broken links — manufactured integrity-fault state in the destination's tamper-evidence lane (the dual-chain audit-head witness binds the revisions head).

## Fix

[...]

Issue #1956, opened 2026-07-09
[v1.0][P1][R56] Crypto-erase (per-record key destruction) + tombstones on every delete path + erasure attestation


alphaonedev:
Gate-2. Extends G30 (v71 tombstones, single-node, forget-path only) to ALL delete paths + per-record envelope keys destroyed on erasure + a signed erasure-attestation enumerating ungoverned-copy classes. Unlocks the banned provable-erasure claim only when shipped. Cross-link #1832.

Filed under the v1.0.0 development epic #1940 (operator grant memory f9a0f397). Gate-item carrier from the 2026-07-09 kickoff audit (run wf_3c42dc07-761). P0/P1 substrate items are T1+T4 5-agent-vote crossroads; Rust skills (/rust-skills + /rust-microsoft) + the no-hardcoded-literals gates are mandatory on all src/ work.

alphaonedev:
Landed on main via `b7c4f8e0`. **Key finding + design:** the #228 at-rest scheme used per-AGENT keys (no per-record key to destroy), so I implemented true envelope encryption as envelope version `0x03` — a random per-record DEK encrypts content, wrapped under the per-agent KEK, the wrapped DEK embedded in the existing `encrypted_envelope` BLOB (NO schema change, decryption stays centralized in open_content). **Crypto-erase** overwrites the envelope with a `0x04` destroy marker → the random DEK is unrecoverable → ciphertext unrecoverable even by a KEK holder. **Mandatory tombstones + attestation** wired on forget/forget_for_caller/gc/size_gc (sqlite; postgres forget at parity) via `substrate.crypto_erase` signed events {id, erasure-kind, actor, timestamp}, chained + verify_audit_trail-surfaced. Archive-moves deliberately excluded (recoverable, tombstoning would block restore). **Honest boundary** (docs/security/crypto-erase.md): ATTESTABLE = key destruction + signed event; ESTIMABLE = depends on encryption-enabled-at-write + master-key custody; legacy 0x02/plaintext rows = delete+tombstone only, not crypto-erasable. Schema stays v80. Gates: default+sal+sal-postgres, all clippy combos, crypto_erase 7 + **#228 encryption_at_rest regression 11/11** + erasure_fanout 7 + 5 encryption unit, docs-vs-ssot. Residue: postgres gc/size_gc eviction-path attestation + archive-reaper tombstones (backend-agnostic primitive already callable; needs a live PG to verify) — tracked follow-up.

PR #3210, opened 2026-08-22
fix(federation,portability): close four fail-OPEN defects on the catch-up egress and bundle-import funnels (#3148 #3149 #3150 #3151)

### LINKED ISSUE #3148: [security] federation bulk_catchup_push bypasses the governance NetworkRequest egress gate (wire_check) that post_once honours, and omits CREDENTIAL_HEADER/CHAIN_HEADER (FED-P3a/P4d)
Scout-verified at cda57210. `src/federation/sync.rs:95` is the ONLY `wire_check::check` call in the file (inside `post_once`; #3133 renames it to `check_governed` but adds no gate elsewhere). `bulk_catchup_push` (`:2360`) builds its own `client.post(&url)…send()` with no governance gate and without the credential/chain headers used only at `:162`/`:175`. An operator refuse-rule for a peer host blocks per-row fanout, but the S40 catch-up still ships full memory batches there with no `governance.check` signed_event (audit asymmetry) — ungoverned egress of memory content.

Repro: refuse-rule a configured peer's host; force a catch-up leg via bulk_create with one missed row → the peer receives the `X-Catchup: bulk` POST.
Fix: one shared outbound request builder that gates (`check_governed`) + attaches FED-P3a/P4d headers for BOTH lanes; test that a refused host blocks the bulk lane and emits the signed_event. GA: BLOCKING.

### LINKED ISSUE #3149: [data-integrity] portability import counts INSERT OR IGNORE'd forget_tombstones / model_attestations as staged — ImportReport asserts rows that did not land
Scout-verified at cda57210. `src/portability/import.rs:449-452` and `:1224-1227`: `conn.execute("INSERT OR IGNORE INTO forget_tombstones …")?; report.forget_tombstones += 1;` (same for `model_attestations`) — the affected-row count is discarded. When the destination already holds a different-bytes row at the PK (two nodes forgot the same id with different forgotten_at/signature; a divergent TOFU attestation), the bundle row is suppressed while the committed report claims full staging — falsifying the struct's own "a report can only exist for a bundle that fully landed" contract. Contrast `:1002-1035` where the signed lanes probe on n==0 and refuse (#2209).

Repro: dst tombstone (m1,T1,S1), bundle tombstone (m1,T2,S2) → report.forget_tombstones == 1, SELECT shows T1/S1.
Fix: bind `execute`'s n; add the #2209 identical-or-refuse probe; tests for both lanes. GA: BLOCKING (false integrity report).

### LINKED ISSUE #3150: [security][data-integrity] archived-memories import lane bypasses redact_inbound_before_attestation / apply_import_attestation / validate_memory that the live lane enforces — a crafted bundle lands forged attestation + oversized/secret content
Scout-verified at cda57210. Live lane: `src/portability/import.rs:551` redact, `:564` apply_import_attestation (HIGH-1 'NEVER trust wire attestation'), `:581` validate_memory (HIGH-2). Archived lane (`:806-901`): DTO → liveness probe → seal_content → raw INSERT, justified only by an in-code comment 'No re-validation otherwise (an archived row was already vetted when originally stored)' — trusting the PRODUCER, while the module itself declares bundles 'UNAUTHENTICATED input … earns no implicit trust (fail-closed, #2211)'.

Repro: spec_version 2, empty signed_events, archived_memories[0] with forged `attest_level=agent_attested` + write_signature + PEM-bearing content on a refuse-mode destination → Ok(report), row present untouched; the same row via env.memories is refused/downgraded.
Fix: run the three gates on the archived lane (PERF-15, ERRORS-01) + tests mirroring the live-lane ones. GA: BLOCKING.

### LINKED ISSUE #3151: portability import: divergent same-id rows on the #2571 lanes (archived_memories, archived_memory_links, namespace_meta) are silently dropped — no identity probe, no warning, no *_skipped_divergent counter
Scout-verified at cda57210. `src/portability/import.rs:901`, `:967`, `:931` use INSERT OR IGNORE / ON CONFLICT DO NOTHING with no byte-identity probe and no warning — unlike the signed lanes (`:1002-1035`, #2209 'Refuse instead') and the live lane (warns on divergent idempotent re-import, `:500-505`). A restore/merge where the destination holds an older snapshot keeps the stale bytes and drops the bundle's newer snapshot with zero signal. (Counters here ARE honest — they use n.) namespace_meta's DO-NOTHING policy is deliberate; only the unaccounted drop needs a signal.

Fix: identical-or-refuse probe, or per-row WARN + `*_skipped_divergent` counter in ImportReport. GA: DEFERRABLE.

alphaonedev:
## What

Four fail-OPEN defects on two funnels that both handle memory CONTENT from a
source this node does not control: the federation S40 catch-up egress, and the
Portability-v2 bundle importer.

| Issue | Class | Defect | Disposition after |
|---|---|---|---|
| #3148 | security, GA-BLOCKING | `bulk_catchup_push` shipped full memory batches past the `NetworkRequest` governance egress gate and without the FED-P3a/P4d identity headers | one shared gated request builder for BOTH lanes |
| #3149 | data-integrity, GA-BLOCKING | `forget_tombstones` / `model_attestations` import counted `INSERT OR IGNORE`'d rows as staged | count only what landed; identical-or-refuse |
| #3150 | security + data-integrity, GA-BLOCKING | the `archived_memories[]` import lane ran NONE of the three gates the live lane runs | same three gates, same order |
| #3151 | data-integrity, deferrable | divergent rows on the three #2571 lanes were dropped with zero signal | probe + keep destination + WARN + counter |

## #3148 — ungoverned federation egress

Verified at `cda57210`: `src/federation/sync.rs:95` was the ONLY
`wire_check::check` call in the file, inside `post_once`; the credential
(`x-memory-cred`) and chain (`x-memory-cred-chain`) headers were attached only
there too. `bulk_catchup_push` (`:2360`) open-coded `client.post(&url)…send()`
and honoured neither. An operator `refuse` rule for a peer host blocked the
per-row fanout while the catch-up lane still delivered FULL memory batches to
that host — and emitted no `governance.check` `signed_events` row, so the
bypass was not even auditable.

Both lanes now build their request through one `build_governed_peer_post`. The
only way to obtain a peer `RequestBuilder` in this module is to pass the gate
first, so a future lane cannot forget it (ERRORS-09 — make the ungated state
unrepresentable). Lane-specific headers (`Idempotency-Key`, `X-Catchup`) are
attached by the caller on the returned builder; the wire is otherwise
byte-identical to the per-row lane. The DLQ replay path was already gated —
it calls `post_once` — so `bulk_catchup_push` was the last ungated content
egress in `src/federation`.

**Tests** (each its own binary; `GOVERNANCE_PRE_ACTION` and the outbound
credential holder are process-global one-shots):
`federation_bulk_catchup_governed_egress_3148` installs a REAL `refuse` rule
through the real rule engine and asserts the catch-up returns the
operator-authored refusal, ZERO bytes reach the peer, and a `governance.check`
`signed_events` row is emitted; `federation_bulk_catchup_identity_headers_3148`
captures the batch POST at a live axum peer and asserts the credential + chain
headers ride alongside signature / nonce / api-key / peer-id / `X-Catchup`.

## #3149 — a report that asserted rows which never landed

`ImportReport`'s own doc contract is "a report can only exist for a bundle that
fully landed". Two lanes broke it by discarding `execute`'s affected-row count
(pre-fix `src/portability/import.rs:452` and `:1227`, both `report.X += 1;`
after an `INSERT OR IGNORE`). A different-bytes row already at the key — two
nodes forgot the same id with a different `forgotten_at`/erasure signature; a
divergent TOFU model pin — was suppressed while the committed report claimed
full staging.

Both lanes now bind `n`, count only what landed, and apply the #2209
identical-or-refuse discipline the signed lanes at `:1002`-`:1035` already use.
The attestation probe matches on EVERY column, because that table has two
suppressing constraints (the `id` PK and
`UNIQUE (provider, model_ref, model_family, agent_id)`).

## #3150 — the archived import lane skipped every gate

Live lane gates, pre-fix line numbers: `:551` redact, `:564`
`apply_import_attestation`, `:581` `validate_memory`. The archived lane
(`:812`-`:901`) went DTO -> liveness probe -> `seal_content` -> raw INSERT,
justified by an in-code comment at `:807` — "No re-validation otherwise (an
archived row was already vetted when originally stored" — i.e. it trusted the
bundle PRODUCER, while this module declares bundles unauthenticated input.

It now runs restamp -> redact -> attestation re-derivation -> `validate_memory`.
The restamp matters past the archive: `restore_archived_for_caller` gates
ownership on `metadata.agent_id`, so pre-fix the bundle decided who could
promote its archived rows back to LIVE. `snapshot_dest_enrolled_keys` covers
archived authors too — resolving them lazily inside the transaction would have
reopened the in-bundle self-enrollment hole that snapshot exists to close.

## #3151 — silent drops, and a deliberate departure from the issue text

The issue offered "identity probe + refuse, OR per-row WARN + a
`*_skipped_divergent` counter". **This PR takes the second option on all three
#2571 lanes** (the first draft of this branch refused; that was changed
deliberately):

- Refusing the bundle for a divergent ARCHIVED snapshot while the LIVE lane
  merely warns and keeps the durable row (`:500`-`:505`) inverts the severity —
  an archived snapshot is strictly less critical than the live row it
  supersedes.
- #1725 writes an `in_place_edit` archived snapshot on every edit, so two
  independently-running federated nodes hold DIFFERENT snapshots under the same
  id as an ordinary steady state. Refusing would pin a repeat restore
  permanently red on a merge condition — the objection-O9 reasoning already
  recorded in `cli::io`'s import exit code — and an always-on alarm gets
  `|| true`-ed away.
- Nothing is lost: bundle intact, destination intact, report honest. The
  fail-closed contract requires not ASSERTING a row that never landed; it does
  not require destroying the rest of a legitimate restore.

The write-once SIGNED lanes (#3149) keep their refusal — a conflicting erasure
receipt or TOFU pin IS an integrity signal, not a merge outcome.

The probe is also structurally unable to fail an import. `row_to_memory` is
fail-closed on an at-rest envelope it cannot open, and a crypto-erased row
(#1956 destroys the wrapped DEK on forget) is undecryptable by design — a
propagating probe would have let one forgotten id abort an otherwise valid
restore. An unreadable survivor is treated as divergent, which is also the
covenant-correct outcome: the erased row stays and the bundle's pre-erasure
copy is not resurrected
(`unreadable_destination_archived_row_does_not_abort_the_import_3151`). And
since `INSERT OR IGNORE` swallows CHECK / NOT NULL failures as well as PK
conflicts, `n == 0` does not always mean "already present": the lanes now
distinguish "a divergent twin is there" from "nothing landed and nothing is
there", and the #3149 refusal messages are reworded for the same reason.

Archived divergence is measured through the existing
`storage::imported_row_diverges` SSOT (title + content, never the restamped
metadata) plus `archived_at`/`archive_reason`, and compared DECRYPTED via the
new `portability::read::read_archived_memory`, which shares ONE row mapper with
the export path's `read_all_archived_memories`. The archived insert re-seals
with a fresh per-record DEK, so a raw ciphertext compare would have flagged
every byte-identical re-import as divergent.

## Scope notes

- **No Postgres lane.** `import_full_envelope` takes a `rusqlite::Connection`
  and has exactly one caller (`src/cli/io.rs:436`); `src/portability/` contains
  no Postgres reference at all, and `export --full` / `import` are documented
  SQLite-only verbs. `federation::sync` is backend-agnostic (HTTP + the
  governance wire-point). So there is no pg twin to keep in parity here.
- **One truthfulness fix carried along.** The live lane's comment claimed that
  under the default `refuse` screen mode `validate_memory` "refuses
  secret-bearing rows outright". It does not: the redact step above it masks
  the credential under every non-`off` mode, so the caller screen never fires
  on this funnel. Masking is the correct behaviour (capture-first); only the
  comment was wrong, and it is corrected rather than copied onto the new lane.
- `docs/spec/PORTABILITY-V2.md` still described all three #2571 lanes as raw
  byte-preserved inserts; updated in lockstep.
- No MCP tools / HTTP routes / CLI verbs / schema versions added — every SSOT
  pin is unchanged and `scripts/check-docs-vs-ssot.sh` +
  `scripts/check-hardcoded-literals.sh` both pass. No qual_10 / qual_6_7
  ceiling moved (neither touched file is in the qual_10 table, and the one new
  `String`-error signature is `Result<RequestBuilder, String>`, which QUAL-7
  does not count).

Closes #3148
Closes #3149
Closes #3150
Closes #3151


alphaonedev:
**Rebased onto \`19f6d1ee\` G (Merge #3052).** Head now \`5236801a\` G. \`git merge-tree\` was CHANGELOG-only and rebased clean (4/4). All four commits \`%G?=G\`. No Fable FIX-BEFORE-MERGE on this PR. Waiting 34/34 on \`5236801a\` then \`fable-merge.sh 3210\`.

alphaonedev:
**fmt fix pushed.** Lint failed on Check formatting only (`src/portability/import.rs` `report.warnings.iter().any` wrapping). Head now `f3ea8f60` G. Did **not** `gh run rerun` the old head (concurrency group would cancel the new run). Waiting 34/34 on `f3ea8f60`.

Issue #3164, opened 2026-08-22
[availability] production-reachable panics (scout-verified): db_op/db_read_op re-panic JoinError (transport.rs:86, read_pool.rs:259); forensic audit writer thread spawn .expect kills boot (audit.rs:164); deferred_audit supervisor exhausts retries into a silent dead drainer (2597/2618); checkpoint.rs:296 one config flag from a remote panic


alphaonedev:
Scout-verified at cda57210 (7 of the audit's 13 PROD-REACHABLE claims confirmed, 6 refuted; 5 more found by the scout's own sweep — two audit lanes were empty).
- `src/handlers/transport.rs:86` `.expect("PERF-1: db_op spawn_blocking worker panicked or runtime shut down")` and `src/handlers/read_pool.rs:259` same shape: a panic in the blocking closure or a spawn_blocking cancelled during graceful shutdown re-panics the request task; on the write side this is the trigger for the open-transaction hazard (separate issue). Fix: return `Result<T, JoinError>` → logged 500.
- `src/governance/audit.rs:164` `.expect("spawning the forensic audit writer thread")` on the main thread at boot (`main.rs:162 init_forensic_audit`): clone(2) EAGAIN (pids cgroup cap / RLIMIT_NPROC on a dense host) kills the process exit 101. Fix: `map_err` → anyhow (ERRORS-01/02).
- `src/governance/deferred_audit.rs:2597` and `:2618` `panic!("deferred_audit supervisor: … exhausted max_restarts")`: deliberate fail-loud, but the panic kills only the supervisor task — the daemon keeps serving with the audit drainer permanently dead and no typed terminal state, no metric, no resumption (non-fleet-manageable). Fix: keep fail-loud semantics but return `JoinHandle<Result<(), DrainError>>`, expose a terminal-state metric + doctor fact + resumption.
[...]

Issue #3140, opened 2026-08-22
macOS Check leg: unit test hang burns the full 80-min job cap (#1492 watchdog is Linux-only) — candidates curator::run_once_with_llm_dry_run_skips_writes / llm::wiremock_tests::test_generate_returns_error_on_500


alphaonedev:
Observed 2026-08-22 on PR #3137 run 32565464216, job `Check (macos-fed,sqlite)` (97013361525, f1-macos-fed): the lib unittest binary printed its last completion at 10:03:53 (`storage::tests::visibility_private_owner_keyed_matrix_1720`) and then nothing for 72 minutes until the 80-min `timeout-minutes` cancelled the job. The same leg took 19 min on two earlier runs today (32556214239, 32552244493), so this is a nondeterministic hang, not contention (f1 was at load 1.3 afterwards; `spindump` fired).

Diffing completed unit tests against the green run 32556214239 leaves exactly two tests that completed there but never completed here:
- `curator::tests::run_once_with_llm_dry_run_skips_writes`
- `llm::wiremock_tests::test_generate_returns_error_on_500`

Both involve an HTTP mock (wiremock) / LLM client path. Hypothesis: an HTTP client call with no request timeout waiting on a mock server whose response never arrives (or a port-bind race on macOS). Not related to #3137 (bench-only change).

Two fixes, both needed:
1. **Test-level:** give the LLM client path used by these tests a bounded timeout (and/or `tokio::time::timeout` in the tests) so a stuck mock fails in seconds with a message, never hangs. Reproduce by looping the two tests on f1 (`cargo test --lib curator::tests::run_once_with_llm_dry_run_skips_writes` ×50).
2. **CI-level:** the #1492 hung-test watchdog is gated to Linux because macOS has no GNU `timeout`. `coreutils` (`gtimeout`) is now installed on f1 (both runners); extend the watchdog to use `gtimeout` when present so a macOS hang is cut at the same per-invocation cap as Linux instead of the 80-min job cap. Fold into the CI follow-up PR with #3139.

alphaonedev:
[...]

Issue #1580, opened 2026-06-10
v0.8.0 perf: WAL read-pool for HTTP sqlite (4.1x proved) — fold into #1488 Pillar 4


alphaonedev:
From the v0.7.0 final-gate performance audit (#1579, Tier C1).

**Finding (measured, P2):** the single `Arc<Mutex<Connection>>` behind all HTTP traffic flat-lines throughput from c=1→32 (reads ~55 rps @10k rows, ~3–5 rps @100k regardless of concurrency); in mixed 50/50 load at c=16/100k, writes are head-of-line blocked to p50 2.08s = **1300×** their standalone 1.6ms. A WAL parallel-reader prototype (8 reader processes on the identical worst-case recall) proved **4.1×**; the postgres sqlx pool on the same workload scales 12–16×.

**Work:** read-only connection pool for recall/search/list/get + dedicated writer connection; move the DLQ replay and replication workers off the shared handle (also resolves Reviewer-5 F5.11 — `SqliteDlqSink::take_pending` holding the shared mutex across SELECT); read-your-writes semantics within a request; the `Db` tuple type is threaded through ~88 routes.

**LOE:** ~300–600 LOC. **AI-NHI:** ~6–10h incl. gates + load test. Aligns with #1488 Pillar 4 admission control — implement together.

Evidence: `.local-runs/perf-audit/p2/` (concurrency matrices, read-pool prototype).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

alphaonedev:
GA-ledger re-validation (2026-06-11, v0.7.0 ship gate): confirmed genuinely deferred — WAL read-pool is a pure perf enhancement folding into #1488 Pillar 4; no v0.7.0 correctness dependency surfaced in the audit wave. Label intact; not a mislabeled GA gap. Re-verified during the 100%-coverage reconcile pass per the RESUME-FABLE5-GA bucket-B contract.

alphaonedev:
## WAL read-pool for HTTP sqlite — SHIPPED (4.1× recall path now concurrent)

Landed on `release/v0.8.0` (`db2387c1`..`436dedb8`, 8 commits). All 8 CI workflows green on `436dedb8`; full `cargo test --features sal --tests` suite green (394 suites, 0 failures).

### Design — Option B (per 5-agent adversarial vote `4d3ea1c5`)
[...]

Issue #982, opened 2026-05-21
[v0.7.0 #972 D1.1] MCP registry: foundation — schemars dep + McpTool trait + PoC on memory_capabilities


alphaonedev:
**Parent**: #972 (Tier-D1 split, sub-issue 1 of 8). Per operator directive 2026-05-21 ("Take all 9 in v0.7.0") this MCP tool registry tightening lands in-cycle. Filed by Claude (Opus 4.7 1M context).

## Scope (1-2 dev-days)

Introduce the foundation abstraction for per-tool schema binding. Subsequent sub-issues build on this.

### Tasks

1. Add `schemars = "0.8"` (or current stable) to `Cargo.toml`. Add to `[dependencies]` (not just `[dev-dependencies]`).
2. Define `pub trait McpTool` in `src/mcp/registry.rs`:
   ```rust
   pub trait McpTool {
       fn name() -> &'static str;
       fn description() -> &'static str;
       fn docs() -> &'static str;
       fn input_schema() -> serde_json::Value;
       fn family() -> &'static str;
   }
   ```
3. PoC the trait against `memory_capabilities` (always-on bootstrap, lowest-risk first migration). Keep the existing `json!({...})` block on the wire — the new derive should produce a byte-identical (or equivalent) schema; verified by snapshot test.
4. No behavior change at this stage. The trait exists alongside the existing `tool_definitions()` macro.

### Files

- `Cargo.toml`
- `src/mcp/registry.rs` (define trait + PoC integration)
- `src/mcp/tools/capabilities.rs` (PoC implementation)
- `tests/` (snapshot test)

### Acceptance

- `cargo build` clean with schemars added
- `tools/list` byte-shape for `memory_capabilities` unchanged
- Snapshot test pins the shape

🤖 Filed by Claude (Opus 4.7 1M context). Per prime directive pm-v3 — fix→test→4 gates→commit→close protocol applies.

alphaonedev:
**Audit-trail note (cross-ref):** commit `844a48328 perf(#982)` forward-referenced `#982` for recall handler lock-acquisition order inversion BEFORE this issue existed. That earlier work is tracked retroactively at #993. This issue (D1.1 of the #972 split) holds the actual #982 slot and is unrelated.

alphaonedev:
[...]

PR #3211, opened 2026-08-22
fix(storage,availability): RAII WriteTxn + typed db_op panics (#3163 #3164)

### LINKED ISSUE #3163: [data-integrity] a panic unwind inside a write path leaves the shared sqlite writer connection in an OPEN write transaction forever (tokio::sync::Mutex does not poison; explicit BEGIN IMMEDIATE/COMMIT, no RAII guard)
Scout-verified at cda57210. `Cargo.toml:538-546` keeps `panic = "unwind"`; there is no `CatchPanicLayer` and no `panic::set_hook` in `src/`. A panic in an axum handler kills only that connection task. BUT the writer `Db` is a `tokio::sync::Mutex` (does NOT poison) and the codebase drives transactions via `execute_batch("BEGIN IMMEDIATE")` … `COMMIT` (`src/storage/connection.rs:75` and callers) rather than RAII `rusqlite::Transaction`. Consequence: an unwind between BEGIN and COMMIT (any panic in the closure under `db_op` — e.g. the `.expect` at `src/handlers/transport.rs:86` re-panicking a JoinError, or any downstream unwrap) leaves the single writer connection inside an open write transaction; every subsequent writer takes the mutex, issues `BEGIN IMMEDIATE` inside an already-open txn ("cannot start a transaction within a transaction") or silently nests, and nothing ever rolls back — partial writes become visible or all writes fail until restart.

Fix (structural, fail-closed): a drop-guard around every explicit BEGIN (issue `ROLLBACK` if not committed when dropped — including on unwind), or migrate the hot paths to `rusqlite::Transaction`; add a test that injects a panic between BEGIN and COMMIT under `db_op` and asserts the next write succeeds with the prior partial write rolled back; a `catch_unwind`/poison-detect in `db_op` that marks the conn dirty and re-opens it. GA: BLOCKING (prime directive: never leave mixed/corrupt state).

alphaonedev:
## What

Two GA-blocking defects in the same failure family: a panic unwind (or a failed
`COMMIT`) could leave the daemon's **single shared sqlite writer connection inside an
open write transaction**, and the helper that dispatches every sqlite operation
**re-panicked the request task** instead of reporting the failure.

## #3163 — RAII `WriteTxn`: an unwind or a failed COMMIT can no longer strand the writer

`Cargo.toml` keeps `panic = "unwind"` and there is no `CatchPanicLayer`, so a panic in a
write path unwound straight past the hand-written `match result { Err(_) => ROLLBACK }`
arm at all **34** manual `BEGIN … COMMIT` sites. The writer is a SINGLE
`rusqlite::Connection` behind a `tokio::sync::Mutex`, which — unlike `std::sync::Mutex` —
does **not** poison, so the unwind released the mutex with the connection still inside an
open transaction and the next caller inherited it.

The nested-`BEGIN` outcome is loud and recoverable. The dangerous one is **silent**: the
three `owns_tx = … && conn.is_autocommit()` funnels —

- `src/storage/mod.rs:1882` `insert_inner`
- `src/storage/mod.rs:3556` `update_with_expected_version`
- `src/storage/mod.rs:14259` `insert_if_newer`

— see `is_autocommit() == false`, conclude *"the caller owns the transaction"*, skip their
own BEGIN/COMMIT, and let their writes **JOIN the orphaned transaction**, whose durability
is then decided by whoever eventually errors or restarts. That is mixed state and
unintentional data loss, which the prime directive ranks above every other concern.

**Fix — structural, not by discipline.** `storage::connection::WriteTxn` issues `ROLLBACK`
from `Drop`, which runs on an unwind exactly as it runs on an early `?` return. The worst
case after a panic is a **rolled-back transaction on a usable connection** — degrade,
never corrupt.

`commit()` disarms the guard **only on success**, which closes the issue's scout addendum:

> the house pattern `conn.execute_batch(SQL_COMMIT)?` propagates a COMMIT failure WITHOUT
> attempting ROLLBACK … a failed COMMIT leaves the shared writer mid-transaction → next
> caller sees `is_autocommit()==false`, skips BEGIN, joins the stranded txn, never commits.

The same destructor covers the panic path, the failed-COMMIT path, and the non-panic `Err`
path. `ROLLBACK`-from-`Drop` is infallible by construction (no-op when already in
autocommit, ERROR log otherwise) because a panicking destructor during an unwind aborts
the process — rust-1.98 **OWNERSHIP-25**.

Converted at all 34 sites — `storage/mod.rs` (23), `storage/reflect.rs`,
`storage/migrations.rs` (`BEGIN EXCLUSIVE`), `storage/connection.rs`, `store/sqlite.rs`
(the SAL store's *separate* shared writer, a second non-poisoning mutex with the identical
hazard), `quotas.rs` (2), `atomisation/mod.rs`, `federation/push_dlq.rs`,
`mcp/tools/store/synthesis.rs`, `cli/io.rs` (2, via `begin_deferred`).
`mcp/tools/skill_register.rs` and `skill_retire.rs` already used RAII
`rusqlite::Transaction` and are unchanged. Zero raw `execute_batch(…BEGIN…)` remain in
`src/` outside `connection.rs` itself and two deliberately-unguarded transport tests.

**Defense in depth at the mutex boundary.** `db_op` runs
`storage::connection::ensure_autocommit` on **both** sides of the closure and contains the
unwind. A connection found inside a transaction on acquisition, or left inside one on
release, is rolled back; if it cannot be rolled back the request is **refused** rather than
run inside an unknown transaction, and the next acquisition retries the sweep — so a
transient failure self-heals with no poison flag, no reopen, and no extra state on the `Db`
tuple. A closure that returns `Ok` while leaving a transaction open now **fails closed**
(`OrphanedTransaction`) instead of reporting a write the sweep just erased: a wrong result
is never an acceptable degradation, fewer results are. `db_read_op`'s writer-connection
FALLBACK arm sweeps too, because a read issued inside a foreign transaction would observe
that transaction's uncommitted rows.

**Postgres lane audited, no change required.** Every transaction goes through
`pool.begin()` → `sqlx::Transaction`, whose `Drop` calls `start_rollback` (sqlx-core
0.8.6); the only `BEGIN` literals in `store/postgres.rs` are PL/pgSQL
`DO $$ BEGIN … END $$` migration blocks, which are not transaction control. Postgres also
takes a connection from a **pool** rather than sharing one behind a mutex, so an unwind
cannot hand a dirty connection to the next caller.

## #3164 — production-reachable panics

1. **`db_op` / `db_read_op` return `Result` instead of re-panicking the request task.**
   Both ended in `.expect("… spawn_blocking worker panicked or runtime shut down")`
   (`handlers/transport.rs:86`, `handlers/read_pool.rs:259` at `cda57210`), so a panic in
   any handler closure — or a `spawn_blocking` cancelled during graceful shutdown —
   propagated into the connection task instead of producing a 500. Panics are now
   contained, logged with their payload, and surfaced as a typed `DbOpError`; `flatten_db_op`
   folds dispatch failures into the closure's own error type so handlers keep exactly one
   error path. **The PERF-1 (#982) and #1580 fast paths are unchanged** — same
   `spawn_blocking` hop, same pooled read-only connection, and `catch_unwind` costs nothing
   when nothing panics. Degradation is chosen per call site: `/health`'s sqlite probe
   reports a dispatch failure *as* a liveness failure; the corpus-gauge cold prime and
   recall's observation-ledger write log and continue (best-effort telemetry); substrate
   read/write handlers 500.
2. **The forensic-audit writer thread spawn no longer aborts boot.**
   `governance/audit.rs:164` `.expect()`ed `thread::Builder::spawn` on the **main thread**
   during `init_forensic_audit`, so a `clone(2)` refusal (`EAGAIN` from a pids-cgroup cap or
   `RLIMIT_NPROC` on a dense host — a resource condition, not a programmer bug) killed the
   process with exit 101 before it served a request. Now `Result` (ERRORS-01/02); every
   caller degrades; a failure does not poison the `OnceLock`, so a later call retries.
3. **A dead deferred-audit drainer is visible while the daemon still serves.** The
   supervisor `panic!`ed on `max_restarts` exhaustion (`deferred_audit.rs:2597`/`:2618`) —
   deliberately fail-loud, but a panic in a `tokio::spawn`ed task kills only *that* task:
   the daemon kept serving with the audit drainer permanently dead and nothing observed it
   until shutdown finally awaited the `JoinHandle`. Now
   `JoinHandle<Result<(), DrainError>>` with a typed `DrainTerminalState`
   (`sink_unresolved` / `sink_panicked`) published first-writer-wins on the shared
   `DeferredAuditMetrics` **and** on a new gauge
   `ai_memory_deferred_audit_drainer_terminal_state` (0 running/graceful, 1 unresolved,
   2 panicked). Fail-loud semantics unchanged — the supervisor still refuses to advance past
   an unresolved occurrence and `close_and_flush` still refuses to report a clean drain, now
   with a machine-readable cause.
4. **`CheckpointResolutionAuthz::Accept` carries the verifying key (ERRORS-09).** The
   verdict was returned bare from three branches, only one of which had authenticated
   anything: the two permissive (`require_sig == false`) branches returned `Accept` with
   `enrolled_key == None`. `mcp/tools/checkpoint.rs:296` turned that into
   `enrolled.expect("Accept implies an enrolled verifying key")`, sound **only** because
   that call site hardcodes `require_sig = true` two lines up — while its HTTP sibling
   (`handlers/federation_receive.rs:3541`) already passes the **runtime** flag. Wiring the
   same flag into the MCP tool would have made a remote peer's unsigned resolution panic an
   MCP tool. The unsound pairing is now unrepresentable: `Accept(VerifyingKey)` for the
   authenticated verdict, `AcceptUnverified` for the permissive rollout window. Behaviour at
   both call sites is byte-identical.

`src/llm.rs:172/176/190` is **not** touched here — it is routed to the #3140 bridge PR.

## Not in scope: a `doctor` fact for (3)

The drainer's terminal state is process-local to the daemon. `doctor`'s local lane sees
only the DB; its `--remote` lane carries no credential (`DoctorArgs` has no api-key field)
and `/metrics` sits behind `handlers::api_key_auth`; and persisting the state to the DB
would write through the very sink that just failed. The Prometheus gauge is the correct
fleet surface, documented in `docs/API_REFERENCE.md` with a page-on-any-non-zero note.
#3164 marks this bullet DEFERRABLE.

## Tests

- `write_txn_rolls_back_on_panic_unwind_and_connection_stays_usable` — injects a **real
  panic between BEGIN and COMMIT** and asserts the partial write is invisible, the
  connection is back in autocommit, and the **next write on the same connection succeeds**.
- `write_txn_rolls_back_when_commit_itself_fails` — forces a **genuine COMMIT failure** via
  a DEFERRED foreign-key violation (SQLite reports it at COMMIT time and deliberately leaves
  the transaction OPEN) and asserts the same three properties.
- `db_op_panic_between_begin_and_commit_leaves_a_usable_writer_3163` — the same regression
  end-to-end on the real shared `Db`.
- `db_op_sweeps_an_unguarded_open_transaction_3163`,
  `db_op_fails_closed_when_a_closure_returns_with_an_open_transaction_3163`,
  `db_op_happy_path_returns_the_closure_value`, `db_read_op_contains_a_closure_panic_3164`.
- `writer_is_fallible_and_idempotent`, `flush_blocking_never_panics_without_an_initialised_sink`.
- `supervisor_exhaustion_returns_typed_terminal_state_sink_{panicked,unresolved}_3164`.
- `checkpoint_accept_carries_the_verifying_key_and_permissive_cannot_produce_it_3164`.
- Plus the guard's commit / explicit-rollback / drop-without-commit and the two
  `ensure_autocommit` sweep cases.

Local (all under `( umask 022; … )`, toolchain 1.96.0): `storage::connection` 19/19 ·
`transport_helpers_tests` 16/16 · `handlers::read_pool` 5/5 · `governance::deferred_audit`
85/85 · `governance::audit` 46/46 · `federation::receive_auth` 27/27 ·
`governance_deferred_log_audit` 17/17 · `f53_claimed_vs_attested_e2e` 4/4 ·
`signed_events_chain_v34` 10/10 · `deferred_audit_soak` 1/1 · `qual_10` 2/2 · `qual_6_7` 2/2
(neither ceiling needed a bump) · archive owner-gate + `http_source_uri_query` +
`claim_bitemporal_recall_1834` + `http_if_match_concurrency` all green.

CI parity run locally and green: `fmt --check`, **all five** clippy invocations
(`-D warnings -D clippy::all -D clippy::pedantic` on default / `--all-targets --features sal`
/ `--features sal-postgres --tests` / `--features sal` / `--all-targets --features vectorlite`),
`check --examples`, `check-hardcoded-literals.sh` (no ratchet bump — named consts),
`check-docs-vs-ssot.sh`.

## SSOT

No new MCP tool, HTTP route, CLI verb, or schema version, so no tool-count badge,
`EXPECTED_PRODUCTION_ROUTES_COUNT`, param-names census, token-budget, or schema-version pin
moves. One new Prometheus gauge, documented in `docs/API_REFERENCE.md`. CHANGELOG
`[Unreleased]` updated with both sections.

Closes #3163
Addresses #3164 (resumption residual — stays OPEN)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -407,6 +407,57 @@ backend and missing or divergently implemented on its twin). Pinned by
   from a POOL rather than sharing one behind a mutex, so an unwind cannot hand a
   dirty connection to the next caller.
 
+### Fixed (availability: production-reachable panics — JoinError re-panic, boot-time thread spawn, silent dead audit drainer, checkpoint authz; #3164)
+
+- **`db_op` / `db_read_op` return `Result` instead of re-panicking the request
+  task.** Both helpers ended in `.expect("… spawn_blocking worker panicked or
+  runtime shut down")`, so a panic in any handler closure — or a
+  `spawn_blocking` cancelled during graceful shutdown — propagated into the
+  connection task instead of producing a 500. Panics are now contained, logged
+  with their payload, and surfaced as a typed `DbOpError`. The PERF-1 (#982)
+  and #1580 dispatch fast paths are unchanged: the same `spawn_blocking` hop,
+  the same pooled read-only connection, and `catch_unwind` costs nothing when
+  nothing panics.
+- **The forensic-audit writer thread spawn no longer aborts boot.**
+  `governance::audit::writer()` `.expect()`ed the `thread::Builder::spawn`
+  result on the MAIN thread during `init_forensic_audit`, so a `clone(2)`
+  refusal (`EAGAIN` from a pids-cgroup cap or `RLIMIT_NPROC` on a dense host) —
+  a resource condition, not a programmer bug — killed the process with exit
+  101 before it served a request. It now returns `Result` (ERRORS-01/02) and
+  every caller degrades: the boot path applies its existing "continuing
+  unsigned" policy, `flush_blocking` logs and returns, and emission surfaces a
+  typed error. A failure does not poison the `OnceLock`, so a later call
+  retries the spawn.
+- **A dead deferred-audit drainer is now visible while the daemon is still
+  running.** The supervisor `panic!`ed when a sink exhausted `max_restarts`.
+  That was deliberately fail-loud, but a panic in a `tokio::spawn`ed task kills
+  only THAT task: the daemon kept serving with the audit drainer permanently
+  dead, and nothing observed it until shutdown finally awaited the
+  `JoinHandle`. `spawn_supervised_drainer` now returns
+  `JoinHandle<Result<(), DrainError>>` with a typed `DrainTerminalState`
+  (`sink_unresolved` / `sink_panicked`) published on the shared
+  `DeferredAuditMetrics` and on a new Prometheus gauge
+  `ai_memory_deferred_audit_drainer_terminal_state` (0 = running/graceful,
+  1 = unresolved, 2 = panicked) that a fleet can alert on. Fail-loud semantics
+  are unchanged — the supervisor still refuses to advance past an unresolved
+  occurrence, and `close_and_flush` still refuses to report a clean drain,
+  now with a machine-readable cause (`DrainFlushError`) instead of a panic
+  payload.
+- **`CheckpointResolutionAuthz::Accept` now CARRIES the verifying key
+  (ERRORS-09).** `authorize_remote_checkpoint_resolution` returned a bare
+  `Accept` from three branches, only one of which had authenticated anything:
+  the two permissive (`require_sig == false`) branches returned `Accept` with
+  `enrolled_key == None`. `mcp::tools::checkpoint` turned that into
+  `enrolled.expect("Accept implies an enrolled verifying key")`, sound ONLY
+  because that one call site hardcodes `require_sig = true` — while its HTTP
+  sibling in `handlers::federation_receive` already passes the RUNTIME flag.
+  Wiring the same flag into the MCP tool would have made a remote peer's
+  unsigned resolution panic an MCP tool. The unsound pairing is now
+  unrepresentable: `Accept(VerifyingKey)` for the authenticated verdict, a
+  distinct `AcceptUnverified` for the permissive rollout window. Behaviour at
+  both call sites is byte-identical; the split exists so no caller can mistake
+  a permissive accept for an authenticated one.
+
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -351,6 +351,62 @@ backend and missing or divergently implemented on its twin). Pinned by
   - This file: the #3129 "unchanged on the fleet" list and the nine-job
     `rust-cache` enumeration are marked SUPERSEDED by #3141 — the current total
     is **5 self-hosted legs**.
+### Security (data integrity: RAII write transactions — an unwind can no longer strand the shared sqlite writer mid-transaction; #3163)
+
+- **Every explicit `BEGIN IMMEDIATE` in the substrate is now closed by a
+  destructor, not by discipline.** `Cargo.toml` keeps `panic = "unwind"` and
+  there is no `CatchPanicLayer`, so a panic inside a write path unwound past
+  the hand-written `match result { Err(_) => ROLLBACK }` arm at all **34**
+  manual `BEGIN … COMMIT` sites (`rg -n 'WriteTxn::begin' src/` outside
+  `#[cfg(test)]`). Because the daemon's writer is a SINGLE
+  `rusqlite::Connection` behind a `tokio::sync::Mutex` — which, unlike
+  `std::sync::Mutex`, does **not** poison — the unwind released the mutex with
+  the connection still inside an open write transaction and the next caller
+  inherited it. The nested-`BEGIN` outcome is loud and recoverable; the
+  dangerous one is silent: the three `owns_tx = … && conn.is_autocommit()`
+  write funnels (`insert_inner`, `update_with_expected_version`,
+  `insert_if_newer`) see `is_autocommit() == false`, conclude "the caller owns
+  the transaction", skip their own BEGIN/COMMIT, and let their writes JOIN the
+  orphaned transaction — whose durability is then decided by whoever eventually
+  errors or restarts. That is mixed state and unintentional data loss, which
+  the project ranks above every other concern. New
+  `storage::connection::WriteTxn` issues `ROLLBACK` from `Drop`, which runs on
+  an unwind exactly as it runs on an early `?` return, so the worst case after
+  a panic is a ROLLED-BACK transaction on a usable connection. Converted at all
+  34 sites across `storage/mod.rs` (23), `storage/reflect.rs`,
+  `storage/migrations.rs` (`BEGIN EXCLUSIVE`), `storage/connection.rs`,
+  `store/sqlite.rs` (the SAL store's separate shared writer), `quotas.rs` (2),
+  `atomisation/mod.rs`, `federation/push_dlq.rs`,
+  `mcp/tools/store/synthesis.rs`, and `cli/io.rs` (2 — the `mine` importer's
+  chunked `BEGIN`/`COMMIT` loop, via `WriteTxn::begin_deferred`).
+  `mcp/tools/skill_register.rs` and `mcp/tools/skill_retire.rs` already used
+  RAII `rusqlite::Transaction` and are unchanged.
+- **A failed `COMMIT` no longer strands the connection either.** The house
+  pattern propagated a COMMIT failure with `?` and never rolled back, so the
+  writer was left mid-transaction for the next caller to silently join.
+  `WriteTxn::commit` stays ARMED on failure: the drop that follows the `?`
+  rolls the transaction back and the commit error is still returned. Pinned by
+  a deterministic regression test that forces a real COMMIT failure through a
+  DEFERRED foreign-key violation (SQLite reports it at COMMIT time and
+  deliberately leaves the transaction OPEN).
+- **Defense in depth at the mutex boundary.** `handlers::transport::db_op` now
+  runs a `storage::connection::ensure_autocommit` sweep on BOTH sides of the
+  closure and contains the unwind with `catch_unwind`. A connection that is
+  inside a transaction on acquisition, or left inside one on release, is rolled
+  back; if it cannot be rolled back the request is REFUSED rather than writing
+  into an unknown transaction, and the next acquisition retries the sweep — so
+  a transient failure self-heals with no poison flag, no reopen, and no extra
+  state on the `Db` tuple. A closure that returns `Ok` while leaving a
+  transaction open now fails CLOSED (`DbOpError::OrphanedTransaction`) instead
+  of reporting a write that the sweep just erased.
+- **Postgres lane audited, no change required.** Every Postgres transaction goes
+  through `pool.begin()` → `sqlx::Transaction`, whose `Drop` calls
+  `start_rollback` (sqlx-core 0.8.6); the only `BEGIN` literals in
+  `store/postgres.rs` are inside PL/pgSQL `DO $$ BEGIN … END $$` migration
+  blocks, which are not transaction control. Postgres also takes a connection
+  from a POOL rather than sharing one behind a mutex, so an unwind cannot hand a
+  dirty connection to the next caller.
+
 
 ### Changed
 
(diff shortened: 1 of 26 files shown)


alphaonedev:
**Fable gate — FIX-BEFORE-MERGE** at `7cb9a593` (branch is BEHIND `ae3e0aec`; GitHub shows CONFLICTING). Shadow-QC lane 2026-08-23. Read every `src/` hunk; R-405'd the PR body against the head and against the release tip.

**Accepted (ESTABLISHED):** `WriteTxn` (`storage/connection.rs:94-247`) — commit-only-disarms, `finish()` idempotent + `is_autocommit()` short-circuit, ROLLBACK failure logged never panicked (OWNERSHIP-25); the three `owns_tx` funnels (`storage/mod.rs:1882/3556/14259`) carry `Option<WriteTxn>` so a caller-owned tx is still joined, not nested. `db_op` (`transport.rs:89-126`): sweep-before, `catch_unwind`, sweep-after, `Ok`-with-open-tx ⇒ `OrphanedTransaction` fail-closed; panic detail reaches the log only — `handler_error_500` body is the fixed `INTERNAL_SERVER_ERROR` slug (`errors.rs:196-203`, pinned by `handler_error_500_is_sanitized`). `db_read_op` fallback arm sweeps the writer. `audit.rs:writer()` fallible, `OnceLock` not poisoned on spawn failure, `run_writer` exits on disconnect so a lost race leaks nothing. `CheckpointResolutionAuthz::Accept(VerifyingKey)` / `AcceptUnverified` — MCP `checkpoint.rs:300-309` no `expect`, HTTP arm byte-identical. Supervisor returns `Result<(), DrainError>` with first-writer-wins terminal state + gauge; `close_and_flush` maps `Drain(..)` to an error. Postgres lane: `sqlx::Transaction` Drop-rollback, pooled — no change needed, agreed. Residual raw `SQL_ROLLBACK`s at `synthesis.rs:329/:433` and `storage/mod.rs:10679` are idempotent with a live guard — acceptable.

**FIX-BEFORE-MERGE**
1. **HIGH (ESTABLISHED) — two addendum-named sites are unguarded on the release tip and this branch does not touch them.** The #3163 addendum names `src/storage/mod.rs` `delete` (owns_tx wrapper) and the SAL `archive_by_ids` wrapper "new in #3115". #3115 merged (`55c0486e`) AFTER this branch's merge-base `93d09145`; at `origin/release/v1.0.0` they are raw: `storage/mod.rs` `pub fn delete` (`owns_tx` → `execute_batch(SQL_BEGIN_IMMEDIATE)` … `SQL_COMMIT`/`let _ = … SQL_ROLLBACK`) and `store/sqlite.rs` `archive_by_ids` (same owns_tx shape around `archive_memory_no_tx`). On this head `delete` is still the pre-#3115 single-statement body and `archive_by_ids` still loops `db::archive_memory` — so after rebase the count of raw sites outside `connection.rs` is 3, not 0 (tip has 36 vs 33 at merge-base; the third is `cli/backup.rs` `"BEGIN EXCLUSIVE; ROLLBACK;"`, a self-terminating probe batch — accepted as-is). Convert both with the `Option<WriteTxn>` owns_tx pattern already used at `storage/mod.rs:1882`. Without this, a failed COMMIT in `delete` strands the writer exactly as the addendum describes — and `delete` is the erasure path.
2. **MED (ESTABLISHED) — `WriteTxn` needs `#[must_use = "dropping the guard rolls the transaction back immediately"]`.** A future `WriteTxn::begin(conn)?;` statement compiles silently, does BEGIN+ROLLBACK, and the following writes run in autocommit — a silent atomicity loss, the class this PR exists to remove. One attribute on the struct (`connection.rs:140`); `Result` alone does not catch it once `?` is applied.
3. **MED (ESTABLISHED) — add the structural pin the PR body asserts.** "Zero raw `execute_batch(…BEGIN…)` remain in `src/` outside `connection.rs`" was true at this head and is already false at the tip — the evidence that a test-time grep pin (same style as the repo's other source pins) is needed, allow-listing `storage/connection.rs` and `cli/backup.rs`'s probe. Cheap; prevents #3115-style regrowth.
4. **Process (#3164 closure):** bullet 3 of #3164 asks for terminal state + metric + doctor fact + **resumption**. Terminal state + gauge landed; the doctor-fact argument (process-local, no `--remote` credential, persisting through the failed sink) is sound and should be recorded on the issue; **resumption is neither implemented nor argued.** Under the 100%-no-deferral directive, do NOT close #3164 on this merge: either land a resumption path (e.g. SIGHUP re-arm of the supervisor alongside the existing `[llm]` reload) or record a refutation with reasoning on the issue and leave it open for the residual bullet. Ledger row stays not-DONE until then.

**Follow-ups (non-blocking):** (a) `read_pool.rs:131` `try_lock()` skips a POISONED slot forever after one contained panic (`Err(TryLockError::Poisoned)` ≠ `Ok`); recover with `into_inner` like the fallback at `:137` so pool capacity does not decay panic-by-panic. (b) `DbOpError` is `pub` without `#[non_exhaustive]` (API-07). (c) When a closure returns a domain `Err` AND leaves a tx open, `db_op` reports `OrphanedTransaction` and drops the domain error — fine for fail-closed, but log the inner error in that arm so the root cause is not lost. (d) `db_op` sweep + `catch_unwind` on every call — the PR's "costs nothing when nothing panics" is PLAUSIBLE, not measured; a PERF-1 A/B is owed post-merge (see `ab-benchmark-parent-binary-validity`).

**Do not merge until** items 1–3 are on the rebased head and item 4 is reflected in how #3164 is closed/left open; then re-request a gate on the new SHA.

alphaonedev:
**Rebased onto `714f8450` (Merge #3122) + Fable FIX-BEFORE-MERGE items 1–3 applied.**

Head **`42f8694d` G**. Contains base.

1. HIGH: `pub fn delete` and `SqliteStore::archive_by_ids` now use `Option<WriteTxn>` owns_tx (same pattern as `storage/mod.rs` insert funnel). Pin test `no_raw_sql_begin_immediate_outside_allowlisted_sites_3163` green.
2. MED: `#[must_use = \"dropping the guard rolls the transaction back immediately\"]` on `WriteTxn`.
3. MED: structural pin allow-lists `connection.rs`, `cli/backup.rs` (probe), `handlers/transport.rs` (db_op sweep tests that must open a raw BEGIN).

**#3164 stays OPEN** (item 4): terminal state + gauge landed; resumption of the dead audit drainer is neither implemented nor refuted. PR must not `Closes #3164`.

alphaonedev:
**Fable QC post-merge audit (merge `78f7c1c3`, verified at tip):** items 1–3 of the FIX-BEFORE-MERGE gate are ESTABLISHED — `storage/mod.rs` `delete` and `SqliteStore::archive_by_ids` route through `Option<WriteTxn>` (`sqlite.rs:1220/:1986` shape), `#[must_use]` on the guard (`connection.rs:142`), structural pin present (`connection.rs:1214`), #3164 correctly left OPEN for the drainer-resumption residual. The remaining `BEGIN IMMEDIATE` strings in `src/` are error messages / the `WriteTxn` internals — no raw sites.

Two notes:
1. **Protocol (§13):** this merged on `42f8694d` without a Fable APPROVED on that head — the gate was FIX-BEFORE-MERGE, and the rule is a later APPROVED gate on the head you merge. This time the application was complete, so no code action; but on the next FIX-gated PR, post the re-push and wait for (or request) the re-gate — the #3051/#2988 incident is exactly what this step exists to catch. Same applies to #3122 (merged at `714f8450` on rebased head `50e40870` without a re-gate; its follow-up fold looks applied per your comment — I am treating my two earlier gates plus your fold note as the verdict of record there).
2. **Matcher gap (follow-up, non-blocking):** the pin matches only `execute_batch` + `SQL_BEGIN_{IMMEDIATE,DEFERRED,EXCLUSIVE}` const names — a future raw site written as a string literal (`execute("BEGIN IMMEDIATE")`, `execute_batch("BEGIN …")`) evades it. Extend the matcher to the literal strings (minus comments) in the next touch of `connection.rs`.

Issue #2679, opened 2026-08-02
GA BLOCKER: AI_MEMORY_STORE_URL=postgres:// is SILENTLY IGNORED — daemon boots on SQLite and writes to the wrong store


alphaonedev:
Split out of #2676. **This is not the packaging issue. It is strictly worse, and it is a data-integrity defect.**

## Reproduction — the PUBLISHED v0.10.0 release binary, run just now

```
$ AI_MEMORY_NO_CONFIG=1 AI_MEMORY_DB=$PWD/probe.db \
    AI_MEMORY_STORE_URL='postgres://u:p@127.0.0.1:5432/db' \
    ai-memory serve --port 19312

INFO ai_memory::daemon_runtime: database: /…/probe.db
```

The daemon **booted**, created a real SQLite database (`probe.db` + a 1.1 MB WAL), and served traffic. **Zero warnings mention `store_url` or postgres.** It was not quiet in general — it emitted WARNs about the missing API key, `FED_REQUIRE_WRITE_SIG`, `FED_REQUIRE_SIGNAL_SIG`, and the permissions default. The silence is specific to the ignored PostgreSQL URL.

Contrast the argv channel, which fails **loud**:

```
$ ai-memory serve --store-url postgres://…
error: unexpected argument '--store-url' found
```

## Root cause

`resolve_store_url` (`src/daemon_runtime.rs:2573`) is called **only** from `build_store_handle` (`:4116`, `:4144`), which is `#[cfg(feature = "sal")]`. In a default build that call site does not exist, so the env var is never read by anything. There is no else-branch and no diagnostic.

So **#1927's non-argv credential channels are dead in every shipped artifact** — the channels added specifically so the PostgreSQL password would not ride on world-readable `/proc/<pid>/cmdline`.

## Why this outranks "cannot connect"

Cannot-connect is **loud and safe**: the operator sees an error and fixes it. This is **silent and wrong**: the operator gets a healthy-looking daemon writing agent memory to a local SQLite file while believing it is in PostgreSQL. Nothing in the logs, `/health`, or the banner contradicts that belief.

And the operator is *following instructions*: `docs/postgres-age-guide.md:243` and `docs/production-deployment.md:100` both actively instruct exporting this variable. The postgres guide is the document a competent operator reads, and it is correct about everything except that the shipped binary ignores it.

[...]

PR #3136, opened 2026-08-22
fix(#3130,#3131): tier filters fail closed; restore publishes safely (staged, verified, atomic, reversible)

alphaonedev:
Closes #3130
Closes #3131

Two data-integrity defects from the v1.0.0 claims audit, one commit each.

## #3130 — an unrecognised tier WIDENED the operation (data loss)

`Tier::from_str` answers `None` for anything outside `short`/`mid`/`long`, and every
filter surface reads `None` as **"no tier constraint"**. Six sites chained the two
(`.and_then(Tier::from_str)`), so a typo did not narrow the request — it widened it.

| Site | Pre-fix behaviour on `--tier Long` |
|---|---|
| `src/cli/forget.rs:102` | matched **every** tier, erased the whole scope, printed `forgot N memories` |
| `src/mcp/tools/forget.rs:134` | same, via `memory_forget` |
| `src/cli/search.rs:64` | returned **unfiltered** results |
| `src/cli/crud.rs:127` (`list`) | returned **unfiltered** rows |
| `src/mcp/tools/search.rs:101` | returned **unfiltered** results |
| `src/mcp/tools/list.rs:93` | returned **unfiltered** rows |
| `src/cli/update.rs:95` | silently NO-OPed the retier, reported success |
| `src/mcp/tools/update.rs:182` | same, via `memory_update` |
| `src/handlers/subscriptions.rs:138` (HTTP notify, pg SAL branch) | wrote at the backend default tier while reporting the caller's requested one |

New SSOT in `src/models/memory.rs`: `Tier::VALUES_HINT` plus fail-closed
`Tier::parse_strict` / `Tier::parse_optional`. `parse_optional` preserves the
distinction the old shape collapsed — **ABSENT is genuinely unconstrained,
PRESENT-but-unparseable is a refusal**. Every site parses *before* the store is
resolved or a single id is collected, so a refusal deletes nothing and returns
nothing, with a non-zero exit and `invalid tier: <value> (use short, mid, long)`.

The two already-correct sites (`src/cli/store.rs:163`, `src/cli/io.rs:933`) now route
through the same helper, retiring their hand-copied literals.

**Four in-tree tests PINNED the defect and are rewritten to assert the fail-closed
contract**: `chunkc_forget_invalid_tier_string_silently_dropped`,
`mcp::tools::forget::tests::forget_with_invalid_tier_string_treated_as_none`,
`handle_list_invalid_tier_treated_as_none`, and the "silently falls through" arm of
[...]

PR #3212, opened 2026-08-22
fix(llm): bound the sync↔async bridge — a wedged runtime driver can no longer park a caller forever (#3140)

### LINKED ISSUE #3140: macOS Check leg: unit test hang burns the full 80-min job cap (#1492 watchdog is Linux-only) — candidates curator::run_once_with_llm_dry_run_skips_writes / llm::wiremock_tests::test_generate_returns_error_on_500

alphaonedev:
Closes #3140

## The defect

On PR #3137, run `32565464216`, job `Check (macos-fed,sqlite)` (`97013361525`, runner `f1-macos-fed`) printed its last unit-test completion at 10:03:53 and then went silent for **72 minutes**, until the 80-minute `timeout-minutes` cancelled the job. The same leg took 19 minutes on two other runs the same day (`32556214239`, `32552244493`). Diffing completed unit tests against the green run left exactly two that never finished, both on the LLM path:

- `curator::tests::run_once_with_llm_dry_run_skips_writes`
- `llm::wiremock_tests::test_generate_returns_error_on_500`

**The HTTP client was never the problem.** `src/llm.rs` builds every `reqwest::Client` with `.timeout(GENERATE_TIMEOUT)` (30 s) and `.connect_timeout(CONNECT_TIMEOUT)` (5 s), and overrides per-request where a tighter bound applies.

The unbounded chokepoint both hung tests traverse is the **sync↔async bridge**, `block_on_local` at `src/llm.rs:118` on base `93d09145`:

```rust
pub(crate) fn block_on_local<F, Fut, T>(make_fut: F) -> T { ... }
//   MultiThread arm:  tokio::task::block_in_place(|| handle.block_on(make_fut()))   // :133
//   current-thread:   std::thread::scope(..).join().expect("...panicked")           // :176-:190
//   no runtime:       Builder::new_current_thread()...build().expect("ephemeral runtime builds")
```

`block_in_place(|| handle.block_on(..))` parks a blocking-pool thread on the **ambient runtime's** driver. If that driver never advances the future — a wedged mock server, a time driver nothing polls, a runtime whose only worker is the thread just parked — the inner `reqwest` timeout never gets a chance to fire and the caller waits forever. The function's return type was `T`, so there was no channel through which a bound *could* be reported even if one existed.

## The fix

`block_on_local` is **retired**. It is replaced by:

```rust
pub(crate) fn block_on_local_bounded<F, Fut, T>(budget: Duration, make_fut: F) -> Result<T>
```

- **`tokio::time::timeout(budget, ..)` on all three runtime-flavor arms**, so expiry is enforced by whichever runtime is actually driving the future, not by a caller who may never be scheduled again.
- **An OS-thread wall-clock alarm on the multi-thread arm only** (`spawn_wallclock_alarm`). That is the one arm whose timer belongs to a runtime this thread does not drive, so a starved time driver would defeat a tokio timer too. The alarm sleeps on a plain OS thread and signals through a `tokio::sync::oneshot`, whose `send` wakes the parked `block_on` directly via its `Waker` — no timer wheel, no IO driver, no worker; nothing the wedged runtime can starve. Arms 2 and 3 each **own** the runtime driving their future, so their timer cannot be starved and they carry no alarm. The alarm runs at `budget + BRIDGE_ALARM_GRACE` (5 s) so on a healthy runtime the tokio timeout always wins and produces the canonical error; a `WallclockAlarmGuard` flips a completion flag on drop so the thread exits within one `BRIDGE_ALARM_POLL` (50 ms) of the call finishing. If the alarm thread cannot be spawned, the future is simply never ready — the caller degrades to the tokio bound rather than being handed a spurious expiry.
- **Fail-closed, never fabricated.** Every failure mode returns `Err`; none panics and none hangs. Losing the `select!` race drops the request future, so `reqwest` cancels the in-flight socket — the bridge does not merely stop waiting while the work continues.
- **Two panic sites closed by deletion.** `Runtime::build().expect("ephemeral runtime builds")` (×2) and `.join().expect("...bridge thread panicked")` are now `?`/`map_err`. A failed runtime build is a real fleet-density condition (fd/thread/memory exhaustion) that used to kill a curator daemon thread or a `spawn_blocking` worker mid-request; a bridge-thread panic used to be re-raised into an unrelated caller's stack.

### Budgets are derived, never guessed

Every budget strictly **exceeds** the largest inner `reqwest` timeout of the call it wraps, at `BRIDGE_BUDGET_FACTOR = 2`, so the bridge can only fire when the inner timeout has *already* failed to release the caller. It never truncates a request the HTTP client would have completed.

| Bridge budget | Value | Inner ceiling of the wrapped call |
|---|---|---|
| `BRIDGE_HEALTH_BUDGET` | 10 s | `HEALTH_TIMEOUT` 5 s (`is_available_async`, `new_with_url_async`) |
| `BRIDGE_GENERATE_BUDGET` | 60 s | `GENERATE_TIMEOUT` 30 s (chat + single embed) |
| `BRIDGE_PULL_BUDGET` | 240 s | `PULL_TIMEOUT` 120 s + a 10 s `/api/tags` list |
| `bridge_embed_batch_budget(n)` | 60 s × n | worst case one `GENERATE_TIMEOUT`-bounded request per text |
| `HOOK_BRIDGE_BUDGET` (mcp) | 1 min | `hooks::timeouts` per-class deadline, max 5 s |

`embed_text_with_budget(_, _, Some(b))` is safe against an operator-widened `AI_MEMORY_RECALL_EMBED_BUDGET_MS`: the underlying `embed_text_async` request carries `.timeout(GENERATE_TIMEOUT)` (`src/llm.rs:2432`), so the inner ceiling is `min(b, 30 s)` regardless of how large `b` is set. **No documented default is tightened by this PR.**

### Security: the MCP enforcement gates now refuse when unevaluable

`pre_signal_send_decision` and `consult_pre_event_gate` (`src/mcp/mod.rs`) drive the async `HookChain` synchronously through this same bridge — they are **deny-capable PRE gates**. Pre-#3140 an `Err` there was unreachable: the bridge panicked (killing the stdio worker / the request task mid-write) or hung. They now **fail closed** with `HOOK_BRIDGE_UNAVAILABLE_CODE = 503`, matching what `hooks::chain` already reports for an executor-level failure under `fail_mode=closed`. An enforcement gate that could not be evaluated is not an allow. The refusal string reuses the existing `"refused by hooks.enforce gate (code {code}): {reason}"` shape, so nothing downstream needs to learn a new format.

## Addendum riders (from the #3140 root-cause comment)

Same class, swept to zero across `src/` and `tests/`:

- **Every `reqwest::Client::new()` bounded.** `grep -rn 'reqwest::Client::new()' src/ tests/` now returns **only doc comments** — 26 call sites replaced, via a shared `tests/common::bounded_test_client()` (30 s) plus module-local builders in `src/llm.rs` (`HEALTH_TIMEOUT`/`CONNECT_TIMEOUT`) and `src/federation/receive.rs` (10 s/5 s, whose tests exist precisely to prove we survive a hostile peer). `tests/common::pg_test_client` gains the same bound, covering six Postgres suites at one site.
- **`wait_for_http_ready` bounded.** Its loop checks `elapsed >= timeout` only at the **top** of each iteration, so a single probe that connects and never answers defeated the whole 5-minute `DAEMON_READY_TIMEOUT`. The probe is now capped at 2 s by a `OnceLock`-cached client (a fresh client per probe would discard the connection pool and leak a resolver thread pool each iteration). Two suites that hand-rolled `for _ in 0..50 { reqwest::get(..) }` (`s75_capabilities_db_schema_version`, `wt_1_a_schema_migration`) now call it, and their `let _ = handle.await` shutdown joins are bounded too.
- **`-m` on every `curl` argv in `tests/`.** All 13 `Command::new("curl")` sites now carry `--max-time`: 12 added here, 1 (`b3_precompute_doesnt_block_serve`) already had the long form. Two named constants per file — `CURL_PROBE_MAX_SECS` (2 s, inside a bounded retry loop) and `CURL_REQUEST_MAX_SECS` (30 s, a real API call under `cargo-llvm-cov`, which is 3-5× slower).
- **Deadline+kill reaps replace bare `child.wait()` / `Command::output()` / `wait_with_output()`** in `governance/deferred_audit.rs` (3 sites), `tests/reembed_cli_1598.rs` and `tests/transcript_extractor.rs`. The two `output()`-shaped reaps redirect stdout/stderr to **files** rather than pipes, which is what makes `try_wait` a truthful liveness signal — reading two pipes to EOF with no deadline is itself the hang. Capture files land under `.local-runs/` (never `/tmp`, per the project rule).
- **Refusal probes are no longer vacuous.** `cli_write_verb_pg_refuse_ceiling_2572`, `backup_fail_closed_2444` and `store_url_fail_closed_2679` proved a `postgres://` store URL is REFUSED while pointing at `127.0.0.1:5432`. On any host actually running Postgres the probe could pass for the wrong reason — and a regression that moved the refusal *after* connect could have WRITTEN to that live database. They now point at `127.0.0.1:1`, where nothing listens. The #2679 probe's accepted reason text also drops the bare `Postgres` substring, which a mere connection error would have matched.
- **Wall-clock assertions that measured the runner, not the product, are now shape guards.** `record_stop_r45_1955::effect_latency_under_100ms` → `effect_refuses_next_write_synchronously`, ceiling 100 ms → 2 s. Stated plainly because the old assertion implied otherwise: the product's ≤100 ms record-stop figure is a **reference-hardware** claim that needs a bench producer to be enforced, and no such bench exists — the 100 ms assertion in a CI unit test never established it either. The load-bearing property (`Err(StoreError::Stopped { .. })` on the very next write, synchronously) is unchanged and now the thing the message describes. Also: the fold-before-gc race tests move from a 1.33× to a 5× expiry margin (900 ms TTL vs a monotonic sleep against a wall-clock `chrono` comparison); `b3_precompute_doesnt_block_serve` uses one 60 s budget in both execution modes instead of 10 s uninstrumented; the abandoned-download watchdog's detached sleeper drops 30 s → 5 s (still unconditionally over the assertion ceiling, so the property is intact).
- **`tests/transcript_extractor.rs` stops using a per-PID target dir.** Cargo already takes an exclusive lock on a target dir, so the PID suffix never prevented anything — it only bought a cold full rebuild and a multi-gigabyte tree per test-runner process.

Item 2 of the issue (the macOS hung-test watchdog via `gtimeout`) is **not** in this PR; it is #3141.

## Tests

New `bridge_budget_tests_3140` module in `src/llm.rs` — 6 tests:

- one per runtime-flavor arm (no ambient runtime / multi-thread / current-thread): an over-budget call returns `Err` whose message names the budget, within 5 s of a 50 ms budget against a 30 s sleep;
- `starved_time_driver_still_releases_the_caller` — the #3140 wedge reproduced **deterministically**: a 1-worker multi-thread runtime whose only worker is held by a task that never yields, so the runtime's time driver can never advance and the bridge's `tokio::time::timeout` is structurally unable to fire. Whichever bound wins, the caller is released with an error naming the budget;
- `wallclock_alarm_fires_after_its_budget` — the alarm does not fire EARLY (that would truncate a healthy call) and fires promptly after. SHA `8aa7ddd6` CI (`Check (ubuntu-latest, sqlite)` + SAL-only) failed this test at 97-99 ms vs a 100 ms budget because `Instant` was captured **after** `spawn_wallclock_alarm`; the worker starts its deadline inside the OS thread. `0637b94c` moves `t0` before spawn. Product budgets, `BRIDGE_ALARM_GRACE`, and documented defaults are unchanged. Local re-run: 6/6 + 20/20;
- `under_budget_call_returns_its_value` — the happy path is untouched.

Deliberately not tested, and said so in the source: the `Runtime::build()` failure path. Tokio exposes no injection point, and forcing it by exhausting the process fd/thread limit would corrupt every other test in the binary. The fix there is nonetheless load-bearing and type-enforced — the function returns `anyhow::Result<T>` and contains no `expect` on the builder.

Test-level backstops on the two tests that actually hung: `wiremock_tests::bridge_call` (a `tokio::time::timeout` + `spawn_blocking` wrapper, applied at 11 sites) and `curator::tests::run_under_hung_test_guard` (a detached worker thread + `recv_timeout`; detaching on expiry is what lets the assertion fire at all). Both 1 minute — far above any legitimate mock round-trip, far below the job cap.

`qual_10` ceiling for `src/llm.rs`: **6_360 → 6_780**, bumped in lockstep with a dated comment. Measured 6_714 (+66 headroom). Growth is bound-and-proof, not feature surface.

## AI involvement

- Agent: Claude Opus 5 (implementation) + Grok (successor: amend, umask-022 deferred_audit re-run, PR, wallclock-alarm test-clock fix `0637b94c`)
- Authority class: Standard
- Human approver(s) for Sensitive items: n/a
- Memory entries created/updated: `3c3d3bbc-c4a1-47aa-bec3-e570535b4416` (operator successor directive)

## CI parity run (local, `rust-toolchain.toml` pin 1.96.0)

All five clippy invocations used the exact CI flag set `-D warnings -D clippy::all -D clippy::pedantic`. Toolchain: cargo 1.96.0 / rustc 1.96.0, PATH-pinned (no rustup shim). Head: `7f1cfc5b`. Base: `origin/release/v1.0.0` = `93d09145`.

| Leg | Result |
|---|---|
| `cargo fmt --all -- --check` | PASS |
| `cargo check --all-targets` | PASS |
| `cargo check --examples` | PASS |
| `cargo check --all-targets --features sal` | PASS |
| `cargo check --all-targets --features sal-postgres` | PASS |
| `cargo clippy -- -D warnings -D clippy::all -D clippy::pedantic` | PASS |
| `cargo clippy --all-targets --features sal -- -D warnings -D clippy::all -D clippy::pedantic` | PASS |
| `cargo clippy --features sal-postgres --tests -- -D warnings -D clippy::all -D clippy::pedantic` | PASS |
| `cargo clippy --features sal -- -D warnings -D clippy::all -D clippy::pedantic` | PASS |
| `cargo clippy --all-targets --features vectorlite -- -D warnings -D clippy::all -D clippy::pedantic` | PASS |
| `scripts/check-hardcoded-literals.sh` | PASS (no ratchet bump; named consts) |
| `scripts/check-docs-vs-ssot.sh` | PASS |
| `cargo test --test qual_10_module_size_ceiling` | PASS (`src/llm.rs` ceiling 6_360 → 6_780, measured 6_714) |
| `cargo test --test qual_6_7_legacy_error_type_ceiling` | PASS |
| `cargo test --lib bridge_budget_tests_3140` | **6 passed, 0 failed** (5.08 s) |
| `curator::tests::run_once_with_llm_dry_run_skips_writes` ×20 | **PASS 20/20** |
| `llm::wiremock_tests::test_generate_returns_error_on_500` ×20 | **PASS 20/20** |
| both hung-tests together in ONE binary invocation ×20 (CI shape) | **PASS 20/20** |
| `cargo test --lib governance::deferred_audit` under `( umask 022; … )` | **84 passed; 0 failed** (4.93 s) |

`bridge_budget_tests_3140` names: `under_budget_call_returns_its_value`, `budget_expiry_is_an_error_with_no_ambient_runtime`, `budget_expiry_is_an_error_under_a_multi_thread_runtime`, `budget_expiry_is_an_error_under_a_current_thread_runtime`, `wallclock_alarm_fires_after_its_budget`, `starved_time_driver_still_releases_the_caller` (deterministic #3140 wedge reproduction).

Postgres: no change under `src/store/postgres.rs` / `postgres_schema.sql` / pg handlers. The pg files in this diff are *test* files only (`reqwest::Client::new()` → `common::bounded_test_client()`). `cargo clippy --features sal-postgres --tests` and `cargo check --all-targets --features sal-postgres` both PASS. Live `AI_MEMORY_TEST_POSTGRES_URL` not run (successor rule 7 does not trigger).

## Rust standard

`rust-1.98`: **ERRORS-01/06/19** (every bridge failure is a `Result`, no `unwrap`/`expect` left on the production path, the two `expect` panic sites deleted); **ERRORS-08** (the remaining panics are test-only contract violations, per **ERRORS-24**); **CONCURRENCY-08** (the alarm's completion flag is `Release`/`Acquire`); **CONCURRENCY-23** (dropping the future on expiry is the cancellation, and `reqwest` aborts on drop); **OWNERSHIP-25** (`WallclockAlarmGuard::drop` only stores a flag — it cannot panic); **PERF-03** (`saturating_add`/`saturating_mul` on every budget arithmetic); **PERF-07** (`u32::try_from` in `bridge_embed_batch_budget`, saturating to `u32::MAX` rather than an `as` truncation); **TEST-02** (every new assertion carries a context message).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -831,6 +831,76 @@ backend and missing or divergently implemented on its twin). Pinned by
     ship with `--self-test` legs proving they reject stale counts, spare the
     `PE-1 knobs` prose and the cert doc's historical evidence notes, and fail
     closed on an unresolvable SSOT only when a doc actually narrates the count.
+- **The sync&#8596;async LLM bridge is bounded: a wedged runtime driver can no
+  longer park a caller forever (#3140).** On PR #3137 the macOS `Check
+  (macos-fed,sqlite)` leg printed its last unit-test completion and then went
+  silent for **72 minutes**, until the 80-minute job cap cancelled it; the same
+  leg took 19 minutes twice that day. Diffing completed tests against the green
+  run left exactly two that never finished, both on the LLM path
+  (`llm::wiremock_tests::test_generate_returns_error_on_500`,
+  `curator::tests::run_once_with_llm_dry_run_skips_writes`). The HTTP client was
+  never the problem — every `reqwest` call already carries a `timeout` +
+  `connect_timeout`. The unbounded chokepoint was the **bridge**: the
+  multi-thread arm's `block_in_place(|| handle.block_on(..))` parks a
+  blocking-pool thread on the ambient runtime's driver, so when that driver
+  never advances the future the inner request timeout never gets a chance to
+  fire. The unbounded `block_on_local` has been **retired** and replaced by
+  `block_on_local_bounded(budget, ..)`, which applies
+  `tokio::time::timeout(budget, ..)` on all three runtime-flavor arms — expiry
+  is now enforced by the runtime that is actually driving the future, not by a
+  caller who may never be scheduled again. Every budget is derived as **twice**
+  the largest inner `reqwest` timeout of the call it wraps (`BRIDGE_HEALTH_BUDGET`,
+  `BRIDGE_GENERATE_BUDGET`, `BRIDGE_PULL_BUDGET`, and an input-scaled
+  `bridge_embed_batch_budget` for batched embeds), so the bridge can only fire
+  when the inner timeout has *already* failed to release the caller — it never
+  truncates a request the HTTP client would have completed. **Fail-closed:**
+  expiry returns a named error, never a partial or fabricated value, and
+  dropping the future also cancels the in-flight request. Two panic sites went
+  with it: a failed `Runtime::build()` (fd/thread/memory exhaustion — a real
+  fleet-density condition) used to `panic!` on a curator daemon thread or a
+  `spawn_blocking` worker, and a panicking bridge thread was re-raised into an
+  unrelated caller's stack; both are now `Err`. The two MCP hook-chain
+  enforcement gates (`pre_signal_send`, the pre-event `hooks.enforce` gate) run
+  through the same bounded bridge and **refuse** when the gate cannot be
+  evaluated (`503`) — an unevaluated deny-capable gate is not an allow.
+- **Unbounded waits removed across the test suites (#3140).** The same class,
+  swept: EVERY `reqwest::Client::new()` (no request timeout) in `src/` tests and
+  `tests/` replaced with a bounded client — the LLM tests, the federation
+  hostile-peer tests, the capabilities suites, and all ten Postgres-backed
+  suites, via a shared `common::bounded_test_client()`; `common::wait_for_http_ready`'s probe bounded (its
+  `elapsed >= timeout` check runs only at loop top, so ONE hung probe defeated
+  the whole 5-minute budget); `-m` riders on every `curl` argv in `tests/`;
+  deadline+kill reaps replacing bare `child.wait()` / `Command::output()` /
+  `wait_with_output()` in the deferred-audit, reembed-CLI and
+  transcript-extractor suites; bounded joins on the in-process daemon tasks in
+  the schema-version suites; and outer guards on the two tests that actually
+  hung. The transcript-extractor build also stops using a **per-PID** target
+  dir — cargo already locks a target dir, so the PID suffix bought nothing but a
+  cold rebuild and a multi-gigabyte tree per runner process.
+- **Fail-closed refusal probes are no longer vacuous (#3140).** Three tests
+  (`cli_write_verb_pg_refuse_ceiling_2572`, `backup_fail_closed_2444`,
+  `store_url_fail_closed_2679`) proved a `postgres://` store URL is REFUSED
+  while pointing at `127.0.0.1:5432`. On any host actually running Postgres the
+  probe could pass for the wrong reason — and a regression that removed the
+  guard could have WRITTEN to that live database. They now point at
+  `127.0.0.1:1`, where nothing listens, and the #2679 probe's accepted reason
+  text no longer admits a bare `Postgres` substring that a mere connection
+  error would also match.
+- **Wall-clock assertions that measured the CI runner, not the product, are
+  now shape guards (#3140).** The `record_stop` effect-latency assertion moved
+  from `<= 100 ms` to a `<= 2 s` SHAPE guard: it fails if the stop plane ever
+  becomes eventually-consistent (a background reconcile, a network round-trip)
+  and cannot fire on a loaded shared runner. Stated plainly, because the old
+  assertion implied otherwise: the product's `<= 100 ms` record-stop figure is
+  a REFERENCE-HARDWARE claim, it needs a bench producer to be enforced, and no
+  such bench exists — the 100 ms assertion in a CI unit test never established
+  it either. The load-bearing `Err(StoreError::Stopped { .. })` property is
+  unchanged. Elsewhere: the
+  fold-before-gc race tests moved from a 1.33x to a 5x expiry margin; the
+  `/health`-not-blocked-by-precompute test uses one 60 s budget in both
+  execution modes instead of 10 s uninstrumented; and the abandoned-download
+  watchdog test's detached sleeper drops from 30 s to 5 s (still
+  unconditionally over the assertion's ceiling, so the property is intact).
 
 - **The required `Check` legs are no longer undeclared performance gates:
   `ai-memory bench` gains `--report-only`.** `cmd_bench` bails non-zero
--- a/docs/compliance/ENTERPRISE-FEDERATION-CERTIFICATION.md
+++ b/docs/compliance/ENTERPRISE-FEDERATION-CERTIFICATION.md
@@ -900,6 +900,23 @@ this clause for the changed surface — the posture is strictly STRONGER,
 never weaker, than at `e22bc93c` (a silent-downgrade path in the
 declarative operator flow was closed).
 
+**Re-cert trigger — FIRED, pending re-affirmation (PR #3212, #3140
+bounded test HTTP client).** The §7-watched path
+`src/federation/receive.rs` changed. The delta is **test-only**: the
+`issue_1928_tests` hostile-peer probes replaced unbounded
+`reqwest::Client::new()` with a 10 s / 5 s `timeout` /
+`connect_timeout` builder (`bounded_test_client`). Production receive
+code, the `/sync/*` wire, the schema, and every `AI_MEMORY_FED_*`
+identifier are **byte-identical**. It REMOVES **no** certified control
+(the removal-proof set is unchanged). The change exists so a HOSTILE
+peer that accepts the TCP connection and then stalls cannot park the
+test forever — the posture those tests exist to prove we survive.
+**This PR does NOT re-mint the certification:** re-running
+§5.4(2)–(5) at the new SHA and re-affirming this document against it
+is the operator/reviewer gate. Until that re-affirmation lands, treat
+the certification as EXPIRED per this clause for the changed path —
+the shipped federation posture is UNCHANGED from `e22bc93c`.
+
 **Named signer.** The determination at `580d8427` is a **GitHub
 squash-merge** of PR #2910, committed by `GitHub` on behalf of the
 operator account (`alphaonedev`). GitHub signature verification is
--- a/src/curator/mod.rs
+++ b/src/curator/mod.rs
@@ -1738,6 +1738,48 @@ mod tests {
         let _ = stream.shutdown(std::net::Shutdown::Write);
     }
 
+    /// v1.0.0 #3140 — outer wall-clock guard for a curator test that drives
+    /// the sync↔async LLM bridge.
+    ///
+    /// `run_once_with_llm_dry_run_skips_writes` is a plain `#[test]`, so every
+    /// `auto_tag` / `detect_contradiction` call reaches the no-runtime arm of
+    /// the sync↔async bridge. That bridge is now bounded structurally
+    /// ([`crate::llm::block_on_local_bounded`]); this is the test-level
+    /// backstop so a future wedge fails in a minute with a message instead of
+    /// burning the whole CI job cap the way #3140 did (72 min on macOS).
+    const HUNG_TEST_GUARD: std::time::Duration = std::time::Duration::from_mins(1);
+
+    /// v1.0.0 #3140 — run `body` on its own thread and give up waiting after
+    /// [`HUNG_TEST_GUARD`].
+    ///
+    /// The thread is deliberately detached (not scoped) on expiry: abandoning
+    /// a wedged worker is what lets the assertion fire at all. A panic inside
+    /// `body` drops the sender, so it is re-raised immediately via `join`
+    /// rather than being misreported as a hang — and re-raised with
+    /// `resume_unwind`, which carries the ORIGINAL payload through, so a failed
+    /// assertion inside `body` is reported by the harness verbatim rather than
+    /// being flattened into this wrapper's message.
+    fn run_under_hung_test_guard<F>(what: &str, body: F)
+    where
+        F: FnOnce() + Send + 'static,
+    {
+        let (tx, rx) = std::sync::mpsc::sync_channel::<()>(1);
+        let worker = std::thread::spawn(move || {
+            body();
+            let _ = tx.send(());
+        });
+        match rx.recv_timeout(HUNG_TEST_GUARD) {
+            Ok(()) | Err(std::sync::mpsc::RecvTimeoutError::Disconnected) => {
+                if let Err(payload) = worker.join() {
+                    std::panic::resume_unwind(payload);
+                }
+            }
+            Err(std::sync::mpsc::RecvTimeoutError::Timeout) => {
+                panic!("{what} still running after {HUNG_TEST_GUARD:?} — hung (#3140)")
+            }
+        }
+    }
+
     /// Build an `OllamaClient` pointed at a running fake server.
     fn ollama_for(server: &FakeOllama) -> crate::llm::OllamaClient {
         crate::llm::OllamaClient::new_with_url(&server.url, "fake-model")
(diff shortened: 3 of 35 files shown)


alphaonedev:
**Fable gate — APPROVED with FOLLOW-UP** at `0637b94c` (shadow-QC lane, 2026-08-23). Read every `src/` hunk; `git merge-tree` against `ae3e0aec` is CLEAN (GitHub's CONFLICTING is the stale-cache shape seen on #3136/#3122) — the gate carries to the rebased head provided the `src/` delta is unchanged.

**Core (ESTABLISHED at `0637b94c`):**
1. `src/llm.rs` `block_on_local_bounded`: `tokio::time::timeout` on all three arms; multi-thread arm adds `spawn_wallclock_alarm` (OS thread → `oneshot`, wakes the parked `Handle::block_on` via its `Waker`, independent of the time driver). Losing the `select!` drops the request future → reqwest cancels the socket; no fabricated value on any path. Both `Runtime::build().expect(..)` sites and the `.join().expect(..)` are now `Err` (ERRORS-06/19). `WallclockAlarmGuard::drop` is a flag store only (OWNERSHIP-25). Alarm-thread spawn failure degrades to the tokio bound, never a spurious expiry. Budgets: `BRIDGE_*` = 2× the inner reqwest ceiling (`:608-640`); `embed_text_with_budget(Some(b))` inner is `min(b, GENERATE_TIMEOUT)` (`:2432` `.timeout(GENERATE_TIMEOUT)` confirmed) so 60 s > inner for any `b`. No documented default tightened.
2. `src/mcp/mod.rs:1697-1714`, `:1854-1869`: both deny-capable gates (`pre_signal_send`, pre-event `hooks.enforce`) FAIL CLOSED on bridge `Err` with `HOOK_BRIDGE_UNAVAILABLE_CODE=503`; the refusal string keeps the existing `refused by hooks.enforce gate (code N)` shape. `HOOK_BRIDGE_BUDGET` 60 s vs chain class deadline max 5 s (`hooks/timeouts.rs:114`, per-chain at `chain.rs:388`) — >10× headroom, cannot truncate a legitimate chain. Exec hooks are `kill_on_drop(true)` (`hooks/executor.rs:710`) so a dropped chain future does not leave a hook process doing work after the caller refused.
3. Tests are real, not fast-path: `starved_time_driver_still_releases_the_caller` pins a 1-worker multi-thread runtime whose only worker never yields, so the tokio timer is structurally unable to fire and only the alarm can release — deterministic reproduction of the #3140 wedge; the three per-arm expiry tests use a 50 ms budget against a 30 s sleep; `wallclock_alarm_fires_after_its_budget` asserts not-early (the `0637b94c` t0-before-spawn fix is correct — the deadline is taken inside the worker).
4. Issue #3140 100%: item 1 (bounded LLM path + test guards) is this PR; item 2 (macOS `gtimeout` watchdog) is already on base via merged #3141 (`ci.yml:1076-1084` on `ae3e0aec`). `Closes #3140` is therefore complete, not partial.
5. R-405 on the sweep claims: `reqwest::Client::new()` in `src/`+`tests/` → doc/comment matches only (confirmed); all 13 `curl` argv carry `-m`/`--max-time`; `wait_for_http_ready` probe bounded (`tests/common/mod.rs` `READY_PROBE_TIMEOUT` 2 s, `OnceLock` client); refusal probes moved to `127.0.0.1:1`; `qual_10` 6_360→6_780 with dated comment; cert doc §7 re-cert note is accurate (test-only delta under `src/federation/receive.rs`).

**Follow-ups (apply at rebase — cheap, and (a) is a claims-truth item):**
(a) MED, claims-truth: the CHANGELOG says "Unbounded waits removed across the test suites" and the PR says the class was "swept to zero", but `tests/integration.rs:13127, :13142, :13207, :13372, :13445` still use `reqwest::get(..)` — the default client has NO request timeout, the identical class — four of them in the hand-rolled `for _ in 0..50 { reqwest::get(..) }` readiness loop this PR replaced in two other suites. Either route them through `common::wait_for_http_ready` / `common::bounded_test_client()` or narrow the CHANGELOG/PR claim to `reqwest::Client::new()`. ESTABLISHED.
(b) LOW, pre-existing (CONCURRENCY-20): both gate closures hold the std `Mutex<ExecutorRegistry>` guard across `chain.fire(..).await`. `consult_pre_event_gate` is also reached from HTTP handlers on the multi-thread runtime, so concurrent writes serialize on the OS-level lock while one evaluation is in flight; the lock wait happens inside the bridged future before its first poll, so the tokio bound cannot fire during it — bounded only transitively by the holder's own 60 s + 5 s. Not introduced here; worth an issue (`fable-qc`) rather than a rider.
(c) LOW: `is_available()` maps a bridge `Err` (including an unbuildable ephemeral runtime) to `false`; correct degradation, but the `Err` text is lost — a `tracing::warn!` there would keep the fleet-density failure visible.

Merge when 34/34 on the rebased head with (a) applied (or the claim narrowed).

alphaonedev:
**Rebased onto `78f7c1c3` (Merge #3211) + Fable FOLLOW-UP (a)+(c).**

Head **`957bb959` G**. Contains base.

(a) `tests/integration.rs` readiness loops and the one-shot health GET now use `wait_for_http_ready` / `bounded_test_client` — no remaining default-timeout `reqwest::get` at the cited sites.
(c) `OllamaClient::is_available` warns on bridge `Err` before degrading to false.
(b) Mutex-across-await in gate closures: filing as fable-qc (pre-existing, not a rider).

`ci-allow.txt` **and** `ci-allow.ssot` = `fix/3140-bounded-llm-bridge` + `release/v1.0.0`.

alphaonedev:
**`957bb959` vectorlite + Postgres feature gate RED** — `clippy::needless_borrows_for_generic_args` at `tests/integration.rs` `.get(&format!(...))`.

**`41c52a6e` G** (FF): pass owned `format!(...)` to `.get`. Do **not** rerun `957bb959`.

Issue #2331, opened 2026-07-23
sqlite memory_update tier->long keeps stale short/mid expires_at — tier-blind GC reaps the explicitly-promoted permanent row (missing #1626 coupling)


alphaonedev:
## Root cause
`db::update_with_expected_version` (src/storage/mod.rs:2531-2549, UPDATE at 2679-2691) applies tier monotonicity but never clears `expires_at` when the effective tier becomes Long. The postgres twin implements the #1626 tier→long ⇒ expires_at=NULL CASE in BOTH update funnels (src/store/postgres.rs:5053-5064, 15721-15735); sqlite never got it. The GC predicate is tier-blind (`expires_at IS NOT NULL AND expires_at < ?1`, src/storage/mod.rs:11441-11443), so a long-tier row with a leftover short/mid TTL is archived — or hard-deleted + crypto-erased under archive_on_gc=false. `update_with_archive_on_supersede` shares the gap (new_tier can become Long while new_expires inherits the old TTL).

## Evidence
- src/storage/mod.rs:2545-2549 (expires_at: None ⇒ existing preserved verbatim; no tier CASE at 2679-2691)
- Contrast the sites that DO enforce the invariant: insert ON CONFLICT arm (mod.rs:1229-1230), fold auto-promote (2226-2236), MCP promote (src/mcp/tools/promote.rs:259-262), HTTP promote (handlers/memories.rs:1448-1450)
- Exposed callers with no compensation: src/mcp/tools/update.rs:487-503, handlers/memories.rs:505-521, store/sqlite.rs:251-271
- Regression pin exists only for postgres: tests/pg_fix3_parity_tests.rs

## Repro
1. `memory_store {tier:"mid"}` → row gets created_at+7d expiry (v54 backfill / effective_expires_at)
2. `memory_update {id, tier:"long"}` → tier='long', expires_at STILL created_at+7d
3. 7 days later GC archives/deletes the documented-permanent row. Identical call on postgres keeps it forever — cross-backend divergence on a data-loss path.

## Proposed fix
~10 LOC in src/storage/mod.rs: force the merged expires_at to None when `effective_tier == Tier::Long` in `update_with_expected_version`, and force `new_expires = None` when `new_tier == Tier::Long` in `update_with_archive_on_supersede`. Do NOT touch postgres.rs (pg already fixed, #1626). Regression tests: update short→long leaves expires_at NULL; gc() does not reap it (~40 LOC).

[...]

Issue #1945, opened 2026-07-09
[v1.0][P0][R4] Told/intervention epistemic kinds + channel-derived write defaults + kind-provenance in recall


alphaonedev:
Gate-1. Adjudication-MANDATED new issue (companion to #1862, refusal slice only). Untyped writes currently default to Observation (highest epistemic class); default to channel-derived; add told/instruction + intervention kinds; surface declared-vs-channel-verified provenance in recall. T4 vote.

Filed under the v1.0.0 development epic #1940 (operator grant memory f9a0f397). Gate-item carrier from the 2026-07-09 kickoff audit (run wf_3c42dc07-761). P0/P1 substrate items are T1+T4 5-agent-vote crossroads; Rust skills (/rust-skills + /rust-microsoft) + the no-hardcoded-literals gates are mandatory on all src/ work.

alphaonedev:
## DECISION — 2×5 vote wf_e3296610-469 (10/10 on resume) · cite `4d3ea1c5` + grant `f9a0f397`

The decisive move: **decouple R4 into three speeds** — the cheap-safe-additive halves ship now; the dangerous signed-genesis flip is phased.

**(a) Vocab — SHIP NOW (additive, zero DB migration):** add three named consts `KIND_TOLD="told"`, `KIND_INSTRUCTION="instruction"`, `KIND_INTERVENTION="intervention"` (appended after `Step`, append-only indices). *told* = received/hearsay (epistemically below Observation); *instruction* = a received imperative (fixes the CLAUDE.md L1 rule that mis-stamps operator directives as `decision`); *intervention* = enacted `do(X)` ground-truth (do-calculus alignment, distinct from Event/Decision). `memory_kind` has **no CHECK constraint** on either backend and no `COUNT` const → pure Rust enum change. Hoist the 11 currently-literal arms to consts in the same edit (honors the no-literals gate).

[...]

Issue #1629, opened 2026-06-11
postgres upsert ON CONFLICT arms drift from sqlite — 6 per-column deltas (citations wiped, confidence_source provenance flipped, priority/confidence downgradable, source/expires_at never updated, persona stickiness missing)


alphaonedev:
## Root cause (one-liner)
All four pg local-write upserts (store :8120-8164, store_batch :8389ff, store_with_embedding :8801ff, capture_turn_idempotent :8607ff) share ON CONFLICT arms that drift from sqlite insert() (storage/mod.rs:663-733) on six columns:

| column | sqlite | pg | consequence |
|---|---|---|---|
| citations | preserve unless non-empty (:712-714) | `= EXCLUDED` (:8150) | re-store w/o citations WIPES stored citations (contradicts the comment at :8147-8149) |
| confidence_source | replace only when != 'caller_provided' (:722-724) | unconditional (:8153) | re-store flips calibrated/auto_derived back — the #1608 provenance-flip family (likely the open federated-anchor flip remainder; same arm at :8845, :8632) |
| priority/confidence | MAX(...) (:666-667) | `= EXCLUDED` (:8128-8129) | re-store can DOWNGRADE both |
| source | `= excluded.source` (:668) | ABSENT | provenance never updates |
| expires_at | long→NULL + COALESCE (:674-675) | ABSENT | tier-raise-to-long keeps old expiry (expiring "long" row, second route to the promote bug); fresh expiry never extends |
| memory_kind stickiness | reflection AND persona (:696-698) | reflection only (:8145-8146) | persona row downgradable by same-title collision |

## Trigger (sharpest)
pg daemon: memory_store title T with citations=[...]; memory_store T again without citations → GET returns citations: [] (sqlite: preserved).

## Proposed fix (size: M)
Port the sqlite CASE arms verbatim to all four pg upserts; add a cross-backend upsert-parity test (store → re-store → assert all 26 fields, both backends).

Found by the Class-C deep-audit reviewer. Severity: MED-HIGH — every upsert collision on postgres.

alphaonedev:
Class-A reviewer adds a 7th delta to this table: entity_id COALESCE order is SWAPPED between backends — pg lets the new value win while sqlite is sticky (postgres.rs:8127-8158 vs storage/mod.rs:668-707). Fold into the same fix + parity test.

alphaonedev:
[...]

PR #3216, opened 2026-08-22
fix(integrity): sqlite/pg data-integrity cluster — v90 archive cid, merge field-set, long-tier permanence, zeroed-stamp guard (#2399 #2395 #2394 #2385 #2564)

### LINKED ISSUE #2385: [3x7-round23][N4] archive→restore re-mints the BLAKE3 cid (archived_memories has no cid cols) — genesis identity drifts, link source_cid/target_cid dangle
**Severity / category**: HIGH / data-integ (hard)

**Root cause**
`archived_memories` has no `cid`/`cid_genesis` columns, so archive→restore re-mints a fresh BLAKE3 cid instead of carrying the original — genesis identity drifts silently and `memory_links.source_cid`/`target_cid` mirrors dangle.

**Evidence**
- `src/storage/mod.rs:12323-12351` (restore re-mint)
- `archived_memories` schema: no cid columns on either backend

**Reproduction / impact**
Archive a memory that participates in lineage edges, then restore it: the restored row mints a new cid (restore-time fields differ from genesis), lineage traversals resolving by cid dangle, and the v74 genesis-identity contract is silently violated. Data-integrity class: identity of the durable row changes without any write intent.

**Proposed fix**
New migration (v88-class, both backends — the v49/#1025 + v85/#2035 + v87/#2333 archive-column-parity precedent): add `cid` + `cid_genesis` to `archived_memories`, carry them through the archive INSERT...SELECT and both `restore_archived*` paths, and stamp them in the federation funnel where applicable. `src/storage/mod.rs` + `src/store/postgres.rs` + `migrations/{sqlite,postgres}/`. Classified "hard"; expect migration + ~60 LOC.

---
Auto-filed by AI NHI per prime directive pm-v2/pm-v3 (3x7 round-2/3 net-new triage). Source: .local-runs/roadmap-review-3x7-round23-netnew.md

### LINKED ISSUE #2394: [3x7-round23][N13] upsert keeps sticky memory_kind but adopts incoming kind_provenance — provenance labels the rejected kind
**Severity / category**: MED / data-integ (coder)

**Root cause**
Upsert-merge keeps the EXISTING row's `memory_kind` (sticky) but adopts the INCOMING write's `kind_provenance` — the stored provenance then labels a kind that was rejected.

**Evidence**
- `src/storage/mod.rs:1256-1258` (sticky kind) vs `:1312` (incoming provenance adopted)

**Reproduction / impact**
Row stored as `Decision` (kind_provenance=declared); an upsert arrives claiming `Observation` via `llm`: the row stays `Decision` but now says kind_provenance=`llm`. Downstream consumers (decorrelation, typed-cognition analytics) read a false provenance for the surviving kind. Mixed/incompatible state on durable metadata.

**Proposed fix**
Make kind_provenance sticky together with memory_kind (adopt incoming provenance ONLY when the incoming kind is adopted). `src/storage/mod.rs` upsert-merge (+ the crdt/fed twin if it shares the skew); ~10 LOC + regression test.

---
Auto-filed by AI NHI per prime directive pm-v2/pm-v3 (3x7 round-2/3 net-new triage). Source: .local-runs/roadmap-review-3x7-round23-netnew.md

### LINKED ISSUE #2395: [3x7-round23][N14] confidence merges MAX but confidence_source/signals/decayed_at merge by a different rule — value and label from different operands
**Severity / category**: MED / data-integ (coder)

**Root cause**
Upsert/CRDT merge takes `confidence` by MAX but merges `confidence_source`, `confidence_signals`, and `confidence_decayed_at` by a DIFFERENT rule — the surviving value and its calibration label/signals can come from different operands.

**Evidence**
- `src/storage/mod.rs:1222` vs `:1282-1286`
- `src/.../crdt_merge.rs:533` vs `:602-607`

**Reproduction / impact**
Merge row A (confidence 0.9, source=X) with incoming B (0.4, source=Y): result can be confidence 0.9 labeled source=Y with B's signals — a self-inconsistent calibration record. Corrupts Form-5 confidence semantics durably on both the local merge and the CRDT lane.

**Proposed fix**
Select the confidence WINNER first, then carry that operand's source/signals/decayed_at as a unit, at both merge sites. `src/storage/mod.rs` + `crdt_merge.rs`; ~15 LOC + a merge test asserting field-set coherence.

---
Auto-filed by AI NHI per prime directive pm-v2/pm-v3 (3x7 round-2/3 net-new triage). Source: .local-runs/roadmap-review-3x7-round23-netnew.md

### LINKED ISSUE #2399: [3x7-round23][N19] fresh long-tier insert with caller expires_at is GC-reapable (HTTP POST + memory_update) while upsert pins NULL
**Severity / category**: MED / data-integ (coder)

**Root cause**
A FRESH long-tier insert that carries a caller-supplied `expires_at` keeps that expiry (HTTP POST + memory_update lanes), making the "permanent" row GC-reapable — while the upsert lane pins `expires_at` to NULL for long tier. Inconsistent long-tier permanence contract by lane.

**Evidence**
- `src/models/memory.rs:1118-1127`
- `src/.../parity.rs:230-241`

Cross-ref: sibling of OPEN #2331 (memory_update tier→long keeps stale expiry); this finding is the fresh-insert + caller-expiry lane.

**Reproduction / impact**
POST /api/v1/memories with tier=long + expires_at=+1h: row is created long-tier but reaped at the hour — silent loss of an explicitly-permanent memory. Same logical write via upsert survives forever. Data-loss class.

**Proposed fix**
Normalize at the write funnel: long-tier writes force `expires_at=NULL` (matching the upsert rule) on the insert + update lanes. `src/models/memory.rs` / validate funnel; ~10 LOC + lane-parity regression test.

---
Auto-filed by AI NHI per prime directive pm-v2/pm-v3 (3x7 round-2/3 net-new triage). Source: .local-runs/roadmap-review-3x7-round23-netnew.md

### LINKED ISSUE #2564: [#2445 residual] zeroing `schema_version` is the strictly better attack, and it is undefended — full v1 ladder replay with the safety snapshot suppressed
#2445 refuses when the stamp is too HIGH. It does nothing when the stamp is too LOW, and low is worse.

`SELECT COALESCE(MAX(version), 0)` means `DELETE FROM schema_version` (or an inserted `0`) reads as version 0. `src/storage/migrations.rs` then gates its pre-migration safety snapshot on `if version > 0`. Consequence: the entire v1 -> tip ladder replays over a POPULATED database, under `BEGIN EXCLUSIVE`, **with the safety snapshot skipped**, re-running the in-code DATA arms (v54 expiry backfill, v86/v87 rendering normalizations). Most structural arms are probe-guarded and therefore inert, but the snapshot suppression is not.

An adversary choosing between `999` (loud refusal) and `0` (silent full-ladder replay with the backup disabled) picks 0 every time.

A negative stamp is a related hole: `cli::boot::read_schema_version` does `u32::try_from(v).ok()`, so a negative stamp yields None -> `schema_supported=false` with NO warn emitted, and `observed > CURRENT` is likewise false, so the guard waves it into a full ladder replay.

**Proposed fix:** (a) make the pre-migration snapshot unconditional whenever any user table has rows — replace the `version > 0` predicate with a row-count probe; (b) make the guard a RANGE check that also refuses `observed < 0`; (c) pairs naturally with the structural corroboration in the sibling residual (stamp<tip but tip-marker present -> refuse rather than replay).

Surfaced by the 5-agent adversarial vote on #2445 (decision memory 4c4789ac), voter-5 A3.

alphaonedev:
Five sqlite/postgres data-integrity defects from the 3x7 round-2/3 net-new triage, plus the residual from the #2445 adversarial vote. Every one is a case where the durable tier ends up holding something that is *wrong* rather than merely *missing* — a self-inconsistent calibration record, a provenance attesting a rejected kind, a re-addressed row, a permanent memory with a live expiry, or a version stamp that lies about the schema it describes.

Base `release/v1.0.0`. Schema advances **v89 → v90** on both backends (#2385).

Closes #2399, closes #2395, closes #2394, closes #2385, closes #2564

---

## Per-issue

| # | Defect | Fix | Backends |
|---|---|---|---|
| **#2399** | A FRESH `long`-tier insert carrying a caller `expires_at` kept it. Every other lane pins NULL for long (the insert `ON CONFLICT` arms, both update funnels per #2331/FBL-01, the supersede funnel), and the GC reap predicate is TIER-BLIND (`expires_at IS NOT NULL AND expires_at < now`) — so a documented-permanent row was archived, or hard-deleted **and crypto-erased** under `archive_on_gc=false`, at the caller's deadline, while the identical logical write through the upsert lane survived forever. | The long gate runs FIRST in `Memory::effective_expires_at` (the shared store funnel every sqlite insert path and ~10 postgres write sites bind through) and in `handlers::parity::resolve_create_expires_at` (so the create response echoed to the caller is truthful, not just the row). Narrow: `short`/`mid` keep their caller expiry verbatim. | both (SSOT) |
| **#2395** | `confidence` merged by `MAX`/`GREATEST` while `confidence_source` / `confidence_signals` / `confidence_decayed_at` merged by a **different** rule. Stored `(0.9, auto_derived, S1)` + incoming `(0.4, calibrated, S2)` durably produced `0.9` labelled `calibrated` carrying `S2` — and on the federation lane every replica CONVERGES on that inconsistent tuple, i.e. permanently. | The field-set is ATOMIC: one selector carries the winning operand's whole tuple. Local funnels key on the `MAX` winner, with the pre-existing #1629 rule preserved verbatim as the exact-tie branch. Federation funnels key on the lexicographic `(confidence, updated_at, id)` order — whose first component IS the `MAX` — so the join stays total, commutative and idempotent. | 2 sqlite + 6 postgres funnels |
| **#2394** | `memory_kind` is sticky (a stored `reflection`/`persona` is never downgraded) but `kind_provenance` merged by a bare `COALESCE`, so a row that KEPT its stored kind was relabelled with the provenance of the kind the merge had just thrown away. | The `kind_provenance` CASE now mirrors that funnel's `memory_kind` CASE exactly, and is NULL-safe by construction. The #1945/#2393 COALESCE rule is preserved verbatim for the kind-unchanged case. | 2 sqlite + 7 postgres funnels |
| **#2385** | `archived_memories` never gained the v74/#1825 `cid`/`cid_genesis` pair, so every archive `INSERT…SELECT` dropped the BLAKE3 address and both restore paths RE-MINTED it from six reconstructed inputs. Any drift in any one — a rewritten `metadata.agent_id`, or a decrypt failure whose `unwrap_or` hashes the CIPHERTEXT placeholder — silently re-addressed the durable row and dangled every `memory_links.source_cid`/`target_cid` mirror. | **Schema v90**, additive on both backends. All 8 sqlite + 7 postgres archive funnels CARRY the pair; both sqlite restore paths and pg `archive_restore` bind it ATOMICALLY from one operand (`CASE WHEN cid IS NOT NULL THEN cid ELSE <re-mint> END`, same predicate for `cid_genesis` — mixing a carried address with a re-derived pre-image would produce a row whose own verify disagrees with itself). | both, schema v90 |
| **#2564** | #2445 refuses a stamp that is too HIGH. Too LOW did nothing: `COALESCE(MAX(version), 0)` means `DELETE FROM schema_version` reads as 0 → the ENTIRE v1→tip ladder replays over a POPULATED database under `BEGIN EXCLUSIVE` **with the pre-migration safety snapshot suppressed** (it was gated on `version > 0`). Offered `999` (loud refusal) or `0` (silent replay, backup disabled), an adversary picks 0 every time. | `SchemaStamp` gains `Zeroed`; every WRITE funnel reads it through `operable_version()`, which has **no path** from `Zeroed` to an `i64` (ERRORS-09 — the illegal state is unrepresentable) instead of `version()`, which coerced it to a permissive `0`. New typed refusal `StoreError::SchemaStampInvalid` / `SCHEMA_STAMP_INVALID`. `evaluate` is now a RANGE check that refuses `observed < 0`, and the `AI_MEMORY_ALLOW_SCHEMA_AHEAD` hatch can never authorise a negative stamp. Separately, the snapshot gate is no longer `version > 0` alone — it also fires on a durable-row probe. | both |

## #2385 — what the migration deliberately does NOT do

It **backfills nothing**. A pre-v90 archived row's genesis address cannot be proven from the archive alone, so those rows keep NULL and keep the legacy re-mint fallback. Inventing an address we cannot prove would be exactly the corruption this rung exists to stop — degrade, never corrupt.

It is additive and probe-guarded (SQLite has no `ADD COLUMN IF NOT EXISTS`; postgres uses it), idempotent, and performs **no full-table rebuild**, so every trigger survives — the v63/v65 lesson, asserted by `v90_migration_is_idempotent_and_preserves_every_trigger_2385`.

The Portability-v2 archive importer (`src/portability/import.rs`) deliberately does NOT bind the columns: a bundle's content address is remote-asserted, and the live-import funnel already recomputes it locally rather than trusting the bundle. Admitting a caller-supplied cid on an archived row would launder it onto a LIVE row via archive→restore.

## #2564 — the discriminator is structural, and the refusal is diagnosable

**No false positives on either legitimate stamp-0 database.** A fresh install has the tables but no rows (bootstrap DDL runs before the first stamp is written), and a legacy pre-v2 database has rows but cannot have `memories.confidence` — acquiring it IS the v2 step. The refusal fires only when a database claims version 0 while carrying BOTH durable rows AND a column no version-0 schema ever had.

**The refusal does not cost the operator their data.** `ai-memory backup` re-opens through the unmigrated funnel on this refusal exactly as it already did for #2445, so the refusal message's "snapshot it before doing anything else" is something the binary actually honours. `ai-memory doctor` reports `schema_stamp=invalid`. `ai-memory boot` reports a distinct `WarnSchemaStampInvalid` status whose copy is backup-first / restore-the-stamp — not the schema-ahead "consider upgrading" remedy (#2445 still owns ABOVE-MAX as `WarnSchemaUnsupported`). On the HTTP surface the postgres gate maps `SCHEMA_STAMP_INVALID` to **503**, not 500, so an orchestrator parks the node rather than crash-looping it into the replay.

### A bug this PR introduced and then closed, called out because it would have been catastrophic

The first cut of the postgres probe was one statement:

```sql
SELECT to_regclass('memories') IS NOT NULL
   AND EXISTS(SELECT 1 FROM memories LIMIT 1) AND ...
```

**PostgreSQL resolves relation references at PARSE time, so the `to_regclass` guard does not short-circuit.** Verified against the live test server, not assumed:

```
$ psql "$URL" -tAc "SELECT to_regclass('nope') IS NOT NULL AND EXISTS(SELECT 1 FROM nope LIMIT 1)"
ERROR:  relation "nope" does not exist
```

`connect_*` runs this gate **pre-bootstrap**, where "no `memories` table" is the normal state of every fresh install — so that probe would have failed EVERY fresh postgres deployment, turning an integrity guard into a total availability outage. The probe is now two catalogue-safe round trips (`schema_stamp_zero_is_impossible`): a `to_regclass` + `pg_attribute` step that can answer against a database with no application tables at all, short-circuited in Rust before the row test is ever prepared. `pg_attribute` rather than `information_schema.columns` because the latter matches a bare `table_name` in ANY visible schema, while `attrelid = to_regclass('memories')` resolves through `search_path` to exactly the relation this adapter reads and writes. Covered by `live_schema_stamp_zero_corroboration_probe_is_valid_sql_2564`, which asserts both that a populated database corroborates AND that the catalogue-only step **answers `false` rather than erroring** for a missing relation.

The SQLite twin now matches that contract: table-presence is asked of `sqlite_master` first (the legitimate no-corroboration case = Fresh); the row/column test then propagates with `?`. A probe that cannot answer returns Err, never Fresh (`corroboration_probe_error_is_not_fresh_2564`). `database_holds_durable_rows` uses the same split so a snapshot cannot be skipped because a row probe failed.

## Deliberate residuals (not regressions, not silently left)

1. **The pg funnel at `store_batch`'s `memory_kind` CASE carries no `persona` arm** (unlike every other pg funnel and both sqlite funnels). The #2394 `kind_provenance` CASE there mirrors *that funnel's* actual kind CASE rather than the canonical one — mirroring a kind rule the funnel does not implement would reintroduce the exact skew #2394 closes. The sqlite-parity gap itself is pre-existing and deliberately NOT widened here.
2. **`confidence` still merges by `MAX`/`GREATEST` independently of the content that won the LWW tiebreak.** That is the pre-existing lattice-join design and #2395 does not challenge it; this PR makes the *calibration tuple* coherent with the surviving *number*, which is what the issue is about.
3. **`scripts/check-docs-vs-ssot.sh` has two structural blind spots** that let ~30 stale `schema 89` claims ship under a green gate: `check_schema_version_rule()` walks `DOC_FILES` only and never `HTML_DOC_FILES`, and its phrasing regexes miss `schema v89`, `Current version: **89**`, `schema=v89`, and a bolded `advanced to schema **89**`. Every stale claim found is fixed in this PR, but **the gate itself is not widened here** — widening it is a separate change with its own blast radius, and doing it in the same PR would conflate "the docs are true again" with "the gate can now prove it". Recommend a follow-up issue.
4. **A fully-archived corpus (zero `memories` rows, all durable text in `archived_memories`) with a zeroed stamp is ALLOWED to replay, not refused.** Naming `archived_memories` in the corroboration probe would make the whole statement fail to PREPARE on a pre-v4 database — which has `memories` but not the archive table — turning the guard OFF for exactly the oldest databases it protects. The bounded harm is a no-op: the only ladder arms touching that tier (v86/v87) are instant-preserving, idempotent and fail-safe. The **snapshot** is not residual — `database_holds_durable_rows` probes BOTH tiers in separate statements for the same prepare-time reason, so an archive-only database is still backed up before a replay (`an_archived_only_database_still_gets_a_pre_migration_snapshot_2564`).
5. **`docs/audience/decision-maker.html` was NOT bumped to v90.** It claimed "Migration round-trips (now through schema v89) are tested against the operator's own DB on every commit" — an *evidence* claim, not a version claim, and the release notes' own §"Migration evidence, at its true bound" says the real-corpus dogfood attested **v78 → v86** only. Bumping the number would have manufactured evidence that does not exist, so the sentence was rewritten to the true bound instead.

## Tests

New:
- `tests/long_tier_permanence_lane_parity_2399.rs` (5) — asserts on the DURABLE ROW, not the helper's return value: a fresh long insert with a past expiry lands NULL, `gc()` does not reap it, the fresh and upsert lanes agree, `short`/`mid` keep their caller expiry, and an expired `short` row is still reaped (proof the gate did not disable GC wholesale).
- `tests/upsert_merge_field_set_coherence_2394_2395.rs` (9) — losing operand cannot relabel the survivor; winning operand carries its own record; the #1629 tie rule preserved as a tuple; federation tuple rides the MAX winner; sticky kind keeps its own provenance; adopted kind takes the incoming one verbatim; kind-unchanged keeps the #1945 COALESCE.
- `tests/archive_restore_cid_identity_2385.rs` (7) — archive carries the cid; restore preserves it; identity survives an `agent_id` drift in the archive (the exact silent re-mint); a legacy pre-v90 row still restores via the fallback; the pair comes from ONE operand; the migration is idempotent AND preserves every trigger; a fresh database ships the columns.
- `tests/schema_stamp_zeroed_guard_2564.rs` (12) — deleted / inserted-zero / negative stamps refuse with the typed verdict, distinct from #2445's; a fresh database still opens; a zeroed-but-EMPTY database still opens; a legacy pre-v2 database with rows still migrates AND now gets a pre-migration snapshot; `Zeroed` has no operable version; a missing `memories` table is structurally Fresh; a corroboration probe that cannot answer is Err, never Fresh.
- `tests/pg_merge_field_set_coherence_2394_2395_2385.rs` (8) — live-postgres twins, including `live_v90_arm_runs_on_a_database_already_at_v89_2385`.
- `src/cli/boot.rs` — `boot_warns_on_schema_stamp_invalid` + `boot_json_schema_stamp_invalid_does_not_recommend_upgrade`.
- `src/store/postgres.rs::live_schema_stamp_zero_corroboration_probe_is_valid_sql_2564` — the parse-time proof above.

Ceilings bumped in lockstep with dated rationale: `qual_10_module_size_ceiling.rs` for `src/storage/mod.rs` (29_160→29_360), `src/store/postgres.rs` (34_620→35_160, remeasured 35_093), `src/storage/migrations.rs` (5_750→5_910); `EXPECTED_CONST_ARMS_SQLITE` 7→8 in both `scripts/check-migration-ladder.sh` and `tests/migration_ladder_integrity.rs`. `MIGRATION_LADDER` now has a declared tail `ARCHIVED_CID_PARITY` at `current_schema_version()` (v90); the v54 `BACKFILL_NULL_EXPIRY_TIER_DEFAULT` row is pinned at 54 instead of following the tip (so `meta_for(54)` works and the tail name is honest).

SSOT reconciled: schema pins on both backends, `conformance/manifest.json` + `conformance/vectors/export/round_trip_l3.json` regenerated to schema 90, the module-docstring `current value:` marker (parsed by a unit test), the v90 ladder rows in the release notes and `docs/schema.html`, the new `SCHEMA_STAMP_INVALID` slug documented in `docs/production-deployment.md`, and CHANGELOG `[Unreleased]`.

## CI-parity (local, toolchain 1.96.0)

- `cargo fmt --all -- --check`
- `cargo check --all-targets` (default / `sal` / `sal-postgres`) + `cargo check --examples`
- clippy ×5, exact CI.yml invocations, `-D warnings -D clippy::all -D clippy::pedantic`: default; `--all-targets --features sal`; `--features sal-postgres --tests`; `--features sal`; `--all-targets --features vectorlite`
- `scripts/check-docs-vs-ssot.sh` / `check-migration-ladder.sh` / `check-hardcoded-literals.sh` / `check-doc-symbol-anchors.sh`
- sqlite: the four new binaries (31 tests) + `qual_10` / `qual_6_7` / `migration_ladder_integrity` / `conformance_corpus` / `conformance_export_roundtrip_2030` / `cid_migration_v74` / `recall_purity_p01` / `skill_retire_lifecycle` / `schema_downgrade_guard_2445` / `db_open_funnel_ceiling_2445` / `storage::migration_meta::tests`
- **pg lane RAN** (URL set, not skipped): `tests/pg_merge_field_set_coherence_2394_2395_2385.rs` 8 passed including `live_v90_arm_runs_on_a_database_already_at_v89_2385` (from v89, not only fresh); `store::postgres::tests::live_schema_stamp_zero_corroboration_probe_is_valid_sql_2564` ok (production `SELECT_SCHEMA_STAMP_MEMORIES_IS_POST_V2_SQL` under `SET LOCAL search_path` in a tx)

Rules applied: **ERRORS-09** (`operable_version` makes the zeroed stamp unrepresentable rather than checked), **ERRORS-01/10/12** (typed `SchemaStampZeroed` + `StoreError::SchemaStampInvalid`, `Send + Sync + 'static`), **ERRORS-19** (no dropped `Result` on the probe), **PERF-25** (the `=` arms are the tie BRANCH of a total order over a NOT NULL column, never a logical float-equality test), **API-29** (the merge selectors form a total order so the federation join stays commutative/idempotent), **OWNERSHIP-05** (`&Tier` — `Tier` is not `Copy`, so this is not `trivially_copy_pass_by_ref`).

## AI involvement
- Agent: Grok (Opus 5 successor on worktree `wt-sqlite-integrity-cluster`)
- Authority class: Sensitive (data-integrity / schema ladder)
- Human approver(s) for Sensitive items: n/a (operator-authorized campaign)
- Memory entries created/updated: `3c37671c-d98b-42ed-b86e-b2508a736ec8`


alphaonedev:
**Fable gate — FIX-BEFORE-MERGE** at `e0be4977` (shadow-QC lane, 2026-08-23; read every `src/` hunk, all five issue bodies, the v90 DDL on both backends).

Accepted on the diff (ESTABLISHED): #2385 v90 is purely additive (`ALTER TABLE … ADD COLUMN`, probe-guarded on sqlite / `IF NOT EXISTS` on pg), no table rebuild, so no trigger is dropped — pinned by `tests/archive_restore_cid_identity_2385.rs:220` replaying the arm twice over a populated db and diffing `sqlite_master` trigger names; all 8 sqlite + 7 pg archive `INSERT…SELECT` funnels carry `cid, cid_genesis`; both sqlite restore paths (`storage/mod.rs:13650`, `:13894`) and pg `archive_restore` (`postgres.rs:25218`) bind the pair from ONE predicate; `load_archived_as_memory` selects `cid` and `row_to_memory` reads it by name (pre-v90 NULL tolerated). #2395: every local funnel (2 sqlite + 6 pg) keys the provenance tuple on the same `MAX`/`GREATEST` that picks `confidence` (verified the `confidence =` rule at `storage/mod.rs:1969/:14466`, `postgres.rs:20200`); the federation lanes key on `(confidence, updated_at, id)` — total, commutative. #2394: the `kind_provenance` CASE mirrors each funnel's own `memory_kind` CASE (`IN ('reflection','persona')` everywhere except pg `store_batch` `postgres.rs:11606/:11623`, which has no persona arm in its kind CASE either — residual #1 is truthfully stated). #2399: `Memory::effective_expires_at` (22 binding sites) + `resolve_create_expires_at` gate long FIRST; short/mid unchanged; CHANGELOG states the contract change. #2564: `SchemaStamp::Zeroed` + `operable_version` has no `Zeroed → i64` path (ERRORS-09), no production caller of `.version()` remains, `evaluate` is a range check and the hatch cannot authorise `< 0`, pg probe is two catalogue-safe round trips with the parse-time trap documented, HTTP maps to 503, `classify_store_err` backs off. Downgrade contract (#2445) untouched: a v89 binary refuses a v90 db exactly as documented, hatch `=90` still honoured.

**1. MED (security/data-integrity, ESTABLISHED) — the sqlite corroboration probe fails OPEN on any error, contradicting this module's own contract and the pg twin.** `storage/connection.rs:227-230` `query_row(SQL_SCHEMA_STAMP_ZERO_IS_IMPOSSIBLE …).unwrap_or(false)` → `Fresh`; `storage/migrations.rs:926-935` `database_holds_durable_rows` likewise `.unwrap_or(false)`. `schema_guard.rs:99` says "A probe that cannot answer returns `Err`, never `Fresh`", and the pg twin `postgres.rs:555` propagates with `?`. On a populated db with a destroyed stamp, ANY non-"no such table" failure of that one statement (SQLITE_BUSY after the 5 s `busy_timeout` — a checkpoint-restart or a non-WAL writer — I/O error, interrupted) converts the refusal into: `Fresh` → `operable_version()=0` → full v1→tip ladder replay AND `database_holds_durable_rows()=false` → snapshot suppressed. That is the exact #2564 outcome, reachable by error rather than by stamp (exploitability PLAUSIBLE, narrow in WAL mode; the contract violation is ESTABLISHED). Fix: distinguish the legitimate no-corroboration case structurally — probe `sqlite_master` for `memories` first (the pattern the v90 arm already uses at `migrations.rs:3903`), then run the row/column test with `?`; same for `database_holds_durable_rows` (table-presence probe per tier, then `?`). Add a test that a probe error is NOT read as Fresh (e.g. a `memories` VIEW that errors, or a `conn.set_prepared_statement_cache_capacity` trick is not needed — rename `memories` to a table whose SELECT errors).

**2. MED (claims-truth / operator guidance, ESTABLISHED) — `ai-memory boot` reports the zeroed stamp with the OPPOSITE remedy.** `cli/boot.rs:558` maps `schema_stamp_zeroed` → `BootStatus::WarnSchemaUnsupported { db_schema: 0 }`, whose rendered text at `boot.rs:464-470` and `:866-868` is "db schema v0 unsupported by binary … consider upgrading". The PR body states `ai-memory boot` "name[s] the refusal" and that "the operator action is the opposite one". It does not name it: the machine-readable status is indistinguishable from schema-ahead and the human text tells the operator to install a newer binary, which will refuse identically. `doctor` (`doctor.rs:657-667`) gets it right. Fix: a distinct `BootStatus::WarnSchemaStampInvalid { observed }` (or a reason field) rendering the backup-first/restore-the-stamp remedy, + the manifest/status unit test.

**3. LOW (test proves a variant, not the production string).** `postgres.rs:30757` `live_schema_stamp_zero_corroboration_probe_is_valid_sql_2564` exercises a hand-typed `to_regclass($1 || '.memories')` query for the missing-relation path; the production constant `SELECT_SCHEMA_STAMP_MEMORIES_IS_POST_V2_SQL` is only run against a populated db. Run the constant under `SET LOCAL search_path = "<scratch>"` inside a tx so the shipped SQL is what is proven to answer `false` rather than error.

**4. LOW (convention).** `tests/archive_restore_cid_identity_2385.rs:221` builds its fixture under `std::env::temp_dir()` with no cleanup; the repo convention (305 test crates, including the sibling `schema_stamp_zeroed_guard_2564.rs:63`) is `tempfile::TempDir`.

Follow-ups (non-blocking; I will file the issues so they are not lost): (a) pg `store_batch` kind CASE has no `persona` sticky arm (`postgres.rs:11606`) — sqlite↔pg parity gap, residual #1; (b) `scripts/check-docs-vs-ssot.sh` blind spots (HTML files, `schema v89` / `schema=v89` / bolded phrasings) — residual #3; (c) residual #4 (archive-only corpus allowed to replay) is avoidable with a `sqlite_master` presence probe before naming `archived_memories` — same fix shape as item 1; (d) the "#2399 every lane" wording: the pg PATCH lane still binds a caller `expires_at` onto a long row — that is open #3230, cross-link it in the PR body rather than claim every lane.

Also: branch is CONFLICTING with `release/v1.0.0` — rebase, re-run the pg lane (the v90 arm must run on a db already at v89, not only from fresh), and push ONCE. Do not merge until items 1 and 2 are on the head and a later `Fable gate — APPROVED` names that SHA.

alphaonedev:
Fable FIX-BEFORE-MERGE items 1–4 are on head **`9e51141454481866f3a6721ef6c20b73d858d21a`** (`%G?=G`). Rebased onto `origin/release/v1.0.0` = `8fb6e9eb`. Not merged, no tag, ci-allow untouched.

**Item 1 MED (landed) — sqlite corroboration probe fails closed.** Table-presence is asked of `sqlite_master` first; the row/column test then propagates with `?`. A probe that cannot answer returns `Err`, never `Fresh` (ERRORS-19 / `schema_guard` contract).
- `src/storage/connection.rs:459` `sqlite_table_present` short-circuit to Fresh only when the catalogue says the table is absent
- `src/storage/connection.rs:462-466` corroboration query uses `?` (was `.unwrap_or(false)`)
- `src/storage/connection.rs:483` `sqlite_table_present`
- `src/storage/migrations.rs:937` `database_holds_durable_rows` → `Result<bool>` (per-tier table-presence, then `?`)
- `src/storage/migrations.rs:1639` snapshot gate consumes the `Result`
- tests: `missing_memories_table_is_fresh_not_an_error_2564`, `corroboration_probe_error_is_not_fresh_2564` (`tests/schema_stamp_zeroed_guard_2564.rs:399` / `:419`) — authorizer-denied READ of `memories` while `sqlite_master` still answers; 12/12 passed

**Item 2 MED (landed) — boot no longer prints the opposite remedy.** Distinct `BootStatus::WarnSchemaStampInvalid { observed }` renders backup-first / restore-the-stamp; it is not `WarnSchemaUnsupported { db_schema: 0 }`.
- `src/cli/boot.rs:303` variant
- `src/cli/boot.rs:483` note copy (`ai-memory backup` + restore `schema_version`; no "consider upgrading")
- `src/cli/boot.rs:576` mapping from `schema_stamp_zeroed`
- `src/cli/boot.rs:895` text header arm
- tests: `boot_warns_on_schema_stamp_invalid` (`boot.rs:1554`), `boot_json_schema_stamp_invalid_does_not_recommend_upgrade` (`boot.rs:1602`) — both ok

**Item 3 LOW (landed) — postgres probe test runs the production constant under `SET LOCAL search_path` in a tx.**
- `src/store/postgres.rs:31439` `SET LOCAL search_path TO public`
- `src/store/postgres.rs:31444` / `:31469` `query_scalar(SELECT_SCHEMA_STAMP_MEMORIES_IS_POST_V2_SQL)`
- test RAN: `live_schema_stamp_zero_corroboration_probe_is_valid_sql_2564` … ok (URL set, not skipped)

**Item 4 LOW (landed) — tempfile.**
- `tests/archive_restore_cid_identity_2385.rs:221` `tempfile::Builder` / `TempDir`

**pg v90-from-v89 RAN** (not skip): `live_v90_arm_runs_on_a_database_already_at_v89_2385` printed `RAN: … (from v89)` then `ok`. pg integ 8 passed.

**QUAL-10 remeasured (never lower a floor):** `src/storage/mod.rs` 29_516 / ceiling 29_600; `src/store/postgres.rs` 35_759 / 35_840; `src/storage/migrations.rs` 5_915 / 5_980.

Local CI-parity: fmt `--check`; `check --all-targets` default/`sal`/`sal-postgres` + `--examples`; clippy ×5 with `-D warnings -D clippy::all -D clippy::pedantic` (default, `--all-targets --features sal`, `--features sal-postgres --tests`, `--features sal`, `--all-targets --features vectorlite`); QUAL-10 + QUAL-6/7; docs-vs-ssot / migration-ladder / hardcoded-literals / doc-symbol-anchors.

alphaonedev:
**CI follow-up (Grok, 9e511414 → 9bc5355d G).** `Check (macos-fed,sqlite)` on 9e511414 went RED on two tests, both R-405'd as **#2399 long-tier permanence** vs fixtures that still treated `expires_at` as live on `Tier::Long` — production contract unchanged:

- `federation_catchup_send_carries_every_durable_field_2384` — `maximal_memory` was Long + `expires_at=2099`; insert strips expiry, fixture assert `expires_at must be set` failed. Fixture tier is now **Short** (still non-default vs Mid) so the catch-up projection of `expires_at` stays load-bearing.
- `a5_1579_candidates_path_applies_namespace_and_liveness_filters` — `make_mem` is Long; a past `expires_at` is stripped so the liveness filter never saw the candidate. Expired candidate is now **Mid**.

Local: both test binaries ok. Never reran 9e511414. New head **9bc5355d** (`%G?=G`). 34-watcher re-armed.

alphaonedev:
**CI follow-up (Grok, 9bc5355d → 4265a738 G).** 33/34: macos-sqlite GREEN (the #2399 sqlite fixture fix held). **Per-Module Coverage Thresholds** RED because llvm-cov runs `--features sal,sal-postgres` and died on the first failing test binary — not a floor drop:

`postgres_update_without_expires_at_patch_leaves_stored_value_untouched` seeded **Long** + `expires_at`; #2399 strips that on insert; COALESCE leave-untouched saw `None`. Seed is now **Mid**. Local: both tests in `postgres_update_expires_at_1423` **RAN** against the enterprise-fed URL and passed. Never reran 9bc5355d. New head **4265a738** (`%G?=G`). 34-watcher re-armed.

alphaonedev:
**MERGED** `cce4b528 G` via `fable-merge.sh` (34/34 on `4265a738`, enforce_admins restored). Closed #2399 #2395 #2394 #2385 #2564 with merge-SHA comments. SCHEMA on base is now **v90**. Next queue head: #3217 `sec2-secret-handling`.

Issue #3147, opened 2026-08-22
Daemon keypair existence gate checks .pub only — a lost daemon.priv never self-heals and the daemon silently signs nothing forever (INFO-level only)


alphaonedev:
Scout-verified at cda57210. `ensure_keypair` (`src/identity/keypair.rs:786-790`) returns `AlreadyExists` whenever `<dir>/daemon.pub` exists; `ensure_and_load_daemon_keypair` (`src/daemon_runtime.rs:3421`) logs the missing-private case at `tracing::info!("identity: only public key on disk … link signing disabled")` and continues. Once `daemon.priv` is gone (the crash window in the keypair-persistence issue, or a .pub-only backup restore) every restart signs nothing, with no WARN, no readiness flag, no metric, and no self-heal.

Fix: gate on both halves (regenerate/refuse per posture); escalate to WARN; surface in `doctor --posture`. GA: DEFERRABLE (but pair with the keypair-persistence fix).

alphaonedev:
Closed by fable-merge.sh of #3218 → `d674047faaacf83babefd4625db93efcd2520e08` G (head `e769d4ad8ce6a5641a6b170f7eb7ac56dbcb234a`, 34/34 verified live). GitHub auto-close does not fire on local signed merge.

Evidence: four-way `(pub,priv)` gate + `no_signing_identity_refusal` under asi-hard; doctor Identity section (`src/cli/doctor.rs` sixteen_sections pin).

Do not close #3164. Never tag.

Issue #3214, opened 2026-08-22
[security] write_root_secret uses bare create_dir_all for the .caproot keystore (the #3198 class on a bearer HMAC secret)


alphaonedev:
The identity-failclosed cluster (#3198) documented this as an out-of-scope residual (wt-identity-failclosed `.HANDOFF.md` open question 1: "Not fixed here (out of scope, hunk disjointness). Worth its own issue."). Independently re-verified at `src/governance/capability.rs:1307-1331` on worktree `sec2-secret-handling`.

## Defect

`governance::capability::write_root_secret` persists a 32-byte HMAC `root_secret` at `<dir>/<issuer>.caproot` mode `0o600`. The directory is created with a bare `std::fs::create_dir_all(dir)` — no mode, no group/other-WRITE check.

Quoted proof (`src/governance/capability.rs:1307-1314`):

```rust
pub fn write_root_secret(dir: &std::path::Path, issuer: &str) -> Result<std::path::PathBuf> {
    use rand_core::RngCore;
    let path = caproot_path(dir, issuer)?;
    std::fs::create_dir_all(dir).with_context(|| format!("creating key dir {}", dir.display()))?;
```

A directory another local UID can write lets them unlink-and-replace the `.caproot` with a secret they control. The per-file `0o600` check on `read_root_secret` then PASSES on the planted file, so `mint` produces tokens that `verify` under the attacker's HMAC. There is no private-derives-public cross-check here (unlike Ed25519 `.priv`/`.pub`) — it is a bearer secret, so the directory posture IS the compensating control.

This is the same class #3198 closed for `identity::keypair` (`create_dir_all_secure` + `enforce_key_dir_secure`, WRITE bits `0o022` only so a default-`umask 022` `0o755` directory is not silently tightened). Identity explicitly did not take this hunk.

## Fix

Mirror #3198 **locally in `capability.rs`** (do not import the unmerged identity helper — hunk-disjointness):

- `DirBuilder` recursive create with mode `0o700` at `mkdir(2)` time
- refuse an existing dir whose `mode & 0o022 != 0` (group/other WRITE), with a `chmod 0700` remediation
- wire the gate into **both** `write_root_secret` and `read_root_secret`
- do NOT refuse a merely group-readable `0o755` directory (shipped default; directory read does not enable the swap)

[...]

Issue #3198, opened 2026-08-22
[security] key directory has no posture gate: keypair::ensure_parent uses bare create_dir_all (no mode, no world/group-writable check) while log/audit dirs get enforce_not_world_writable + 0700 — under umask 0002 the signing identity is swappable by another local UID


alphaonedev:
Scout-verified (R-405) at `93d09145` from the ox-alpha campaign-6 red-team swarm (`.local-runs/ox-c6-redteam/`). pg lane evidence where pg is touched: `AI_MEMORY_TEST_POSTGRES_URL=$(cat .local-runs/ENTERPRISE-FED-TEST-PG-URL.txt) cargo test --features sal-postgres -- --include-ignored`.
**Code.** `src/identity/keypair.rs:814-820` `ensure_parent` = bare `fs::create_dir_all`; load-time check is file-mode only `:550-558`. Contrast `src/log_paths.rs:163-166` `enforce_not_world_writable` and `:321-336` `ensure_dir_secure` (explicit 0o700). Not touched by `fix/identity-failclosed-3145-3146-3147-3155`.
**Impact.** umask 0002 (a real config in this fleet — see the umask memory) → dir 0775 group-writable; a second local UID unlinks+replaces `daemon.priv`/`daemon.pub`: a planted 0600 .priv passes the mode check, a matched pair passes the cross-check, the #1790 fstat-TOCTOU fix is irrelevant (swap happens before load) → forged signed_events/audit self-signatures that VERIFY; tamper-evidence reports CLEAN. Log material is better protected than the signing identity.
**Fix.** Create the keys tree via a 0700 `ensure_dir_secure` equivalent; `enforce_not_world_writable` (and not group-writable) on the resolved key dir in `default_key_dir()`/`resolve_key_dir` AND again at load()/save() (re-stat, fail closed).
**Test.** chmod 0777 key dir → load()/save() refuse naming `chmod 0700`; fresh key dir is 0700 under umask 000. GA: BLOCKING (identity root).

alphaonedev:
Closed by fable-merge.sh of #3218 → `d674047faaacf83babefd4625db93efcd2520e08` G (head `e769d4ad8ce6a5641a6b170f7eb7ac56dbcb234a`, 34/34 verified live). GitHub auto-close does not fire on local signed merge.

[...]

Issue #3182, opened 2026-08-22
[data-integrity] sqlite is_registered_agent / agent_max_created_at / schema_version / get_embedding turn DB faults into benign answers (pg twins propagate) — the #3145 class, four more sites; schema_version→0 can present a populated DB as fresh


alphaonedev:
Scout-verified (R-405) at `cda57210` from the ox-alpha SAL parity audit (152 methods; `.local-runs/ox-c4-parity/`). pg lane evidence required: `AI_MEMORY_TEST_POSTGRES_URL=$(cat .local-runs/ENTERPRISE-FED-TEST-PG-URL.txt) cargo test --features sal-postgres -- --include-ignored`.
**Code.** `src/storage/mod.rs:18574-18582` `conn.query_row(...).is_ok()` (fault == 'not registered'; feeds governance Registered level + approver checks) vs pg `postgres.rs:25617`; `src/store/sqlite.rs:316` `.unwrap_or(None)` on `agent_max_created_at` (fault == 'no federation watermark' → catch-up re-pulls from zero/skips; also sqlite reads agent_id_idx vs pg metadata->>'agent_id' and pg re-renders via chrono → watermark bytes differ); `sqlite.rs:147-153` `schema_version` `.unwrap_or(0)` (broken DB reports fresh — a migration-ladder input); `storage/mod.rs:15130-15135` `.ok()` on get_embedding. (ox marked is_registered_agent IDENTICAL — false negative.)
**Fix.** Propagate in all four; best-effort only at the call site with an explicit commented discard. Fold with #3145 (Opus P in flight — coordinate: P owns `agent_pubkey`; this issue owns the other four).
**Test.** both lanes: inject a table-drop/permission fault → Err, not a benign value; schema_version: migration runner refuses rather than replaying the ladder from 0. GA: BLOCKING (schema_version) / DEFERRABLE (others).

alphaonedev:
Scout fold-in (R-405 @ 93d09145), same `.ok()`/`.unwrap_or` fail-open class — add to the site list: `src/storage/mod.rs:7542-7549` `find_by_title_namespace` `.query_row(...).ok()` (doc at `:7541-7543` claims 'Returns the underlying SQLite error' — false); `src/storage/mod.rs:17889-17894` `auto_detect_parent` `.ok().flatten()` (governance root on fault — see the set_namespace_standard issue filed today).

alphaonedev:
[...]

PR #3218, opened 2026-08-22
fix(security): identity fail-closed cluster — no swallowed DB fault, no sole-key destruction, no silent inertness, no attacker-writable key dir

### LINKED ISSUE #3145: [data-integrity][security] sqlite agent_pubkey() swallows every DB error into 'no bound key' (.ok().flatten()) — transient fault silently demotes a valid signature to claimed / flips strict writes to AttestationRequired (pg twin is correct)
Scout-verified at cda57210. `src/storage/mod.rs:11655-11666` `agent_pubkey`: the `query_row(...)` result is collapsed with `.ok().flatten()`, so SQLITE_BUSY / lock timeout / I/O error / corrupt page all become `Ok(None)` = "no bound key". Callers (`attest.rs:496` → verify catch-all; SAL + sync funnel via `src/store/sqlite.rs:782`) then stamp a genuinely-signed write as bare `claimed`, or under strict attestation flip it to `AttestationRequired`. The postgres twin (`src/store/postgres.rs:21326`) uses `fetch_optional` + `map_err` and propagates — backend asymmetry. The identical pattern was fixed 30 lines below for `agent_id_for_api_key` (#2095) with a comment naming this hazard.

Repro: hold `BEGIN EXCLUSIVE` on a second connection with busy_timeout=0, call `stamp_attestation_sync` with a genuine signature → `Ok(AttestLevel::Claimed)`.
Fix: `.optional()` per the #2095 precedent (rust-1.98 ERRORS-15/ERRORS-19); regression test injecting a locked DB; parity test vs pg. Possibly the root cause of the intermittent 403 in open #2941 — triage together. GA: BLOCKING.

alphaonedev:
Closes #3145 #3146 #3147 #3155 #3198

Five fail-closed defects on the identity path. Each one turns a fault into a
*wrong answer* rather than a degraded one, two of them can destroy an
irreplaceable private key, and one lets another local UID swap it.
All five are verified at the cited lines against `origin/release/v1.0.0` (base 93d09145) and every claim below is quoted from
the code it describes.

---

## #3145 — `db::agent_pubkey` swallowed every DB error into "no bound key"

`src/storage/mod.rs:11655-11666`, verbatim at base:

```rust
let pubkey = conn
    .query_row(
        "SELECT json_extract(metadata, '$.agent_pubkey') FROM memories
         WHERE namespace = ?1 AND title = ?2",
        params![AGENTS_NAMESPACE, &title],
        |row| row.get::<_, Option<String>>(0),
    )
    .ok()
    .flatten();
```

`.ok().flatten()` maps `SQLITE_BUSY`, a lock timeout, an I/O error, and a
corrupt page alike onto `Ok(None)` = "this agent has no bound key". The write
funnels (`identity::attest::stamp_attestation_sync`, and
`store::sqlite::SqliteStore::agent_pubkey` on the SAL + sync lanes) then stamp a
genuinely-signed write as bare `claimed`, or under required attestation reject
it as `AttestationRequired`. Stamping `claimed` on a signed write is a WRONG
result written **durably into the row's provenance** — the row does not record
that a lookup ever failed, so it is unrecoverable after the fact. The postgres
twin (`PostgresStore::agent_pubkey`) already used `fetch_optional` + `map_err`
and propagated, so one fault produced different durable provenance per backend.

**Fix:** `.optional()` — only the no-row case maps to `None`, every real query
error propagates with the agent id in its context. Same hazard, same fix, as
`agent_id_for_api_key` thirty lines below (#2095). rust-1.98 **ERRORS-15**
(preserve the cause) / **ERRORS-19** (never silently drop a `Result`).

The federation-receive lane keeps DEGRADING rather than refusing (a wrong or
absent key there can only land a row `claimed` or quarantine it —
`apply_inbound_write_attestation` rejects a forged signature unconditionally —
and `sync_push` returns `impl IntoResponse` with no `Result` to propagate into),
but it no longer degrades *silently*: both lookups now WARN with the agent id
and the error. Hardening that lane to refuse the batch is deliberately left to
its own change.

## #3146 — key persistence was remove-then-create, and wrote the public half first

`src/identity/keypair.rs:832-837` at base:

```rust
let _ = fs::remove_file(path);
let mut file = fs::OpenOptions::new()
    .write(true)
    .create_new(true)
    .mode(mode)
    .open(path)?;
```

The old file is destroyed **before a single new byte exists**. A crash, `ENOSPC`,
`EIO`, or an OOM kill in that window leaves `<agent>.priv` absent or truncated —
sole-key destruction, unrecoverable without a backup, because `generate` mints a
*different* identity and every prior signature becomes unverifiable. The `save`
doc asserted an atomicity that did not exist ("written atomically by the
underlying `fs::write` (single syscall … a partial write is recoverable by
`generate` again)"): the writer was not `fs::write`, a write is not one syscall,
and `generate` does not recover an identity, it replaces one.

**Fix.** Each key file is replaced by the standard durable-replace sequence —
staging file in the SAME directory created *at the requested mode* (so a `0600`
private key is never briefly world-readable), `fsync` the file, atomic `rename`,
`fsync` the directory. Every failure before the rename removes the staging file
and leaves the destination byte-for-byte as it was.

`save` writes the **private half first**. The pair is two files, so two renames,
and no filesystem we target offers a cross-file transaction — a crash between
them is unavoidable and the ordering decides *which* half-state it leaves.
`.priv` present + `.pub` absent is RECOVERABLE (a public key is a deterministic
function of its private key, and `ensure_keypair` re-derives it, see #3147); the
pre-#3146 inverse is not, and it is also the shape that wedged the #3147
existence gate. `encryption::save_keypair_to_disk` (X25519), which reuses the
same writer, takes the same ordering.

`rotate` / `rotate_with_succession` named the archive of the retiring public key
at 1-second granularity and wrote it with that same remove-then-create
primitive, so two rotations inside one second silently OVERWROTE the earlier
archive — destroying the only retained verification anchor for the identity
that rotation had just retired. Archives are now claimed with `create_new`,
collide onto a deterministic `-N` suffix, and REFUSE rather than overwrite when
the suffixes are exhausted.

The `save` doc now states what IS and is NOT guaranteed, per-file and per-pair.

## #3147 — the daemon keypair existence gate consulted the public half only

`ensure_keypair` at base branched on one predicate:

```rust
let pub_path = agent_pub_path(dir, agent_id);
if pub_path.exists() {
    return Ok(EnsureOutcome::AlreadyExists { pub_path });
}
let kp = generate(agent_id)?;
save(&kp, dir)?;
```

Two failures fall out of that single test, in opposite directions:

* `.pub` present, `.priv` gone → `AlreadyExists` on **every** restart. The
  daemon signs nothing, forever, reported at
  `tracing::info!("identity: only public key on disk … link signing disabled")`
  — below the default log filter of most deployments. No WARN, no readiness
  flag, no self-heal.
* `.priv` present, `.pub` gone → falls through to `generate` + `save`, which
  **overwrites the surviving private key with a new identity**. Silent,
  unprompted destruction of the one half-state that is otherwise fully
  recoverable, and exactly the state #3146's crash window produces.

**Fix.** The gate is a four-way state table over both halves: both present is
idempotent as before; `.priv` without `.pub` RE-DERIVES the public key (same
identity, zero signature loss); `.pub` without `.priv` is reported as
`PublicOnlyDegraded` and WARNed, never auto-repaired (regenerating would mint a
different identity behind the operator's back); neither present generates. The
private-key reader was extracted to `read_private_key_file` so the self-heal
path reuses the identical mode/length/TOCTOU discipline rather than growing a
second, drift-prone reader.

The daemon's "public key only" line is now WARN, and the pure
`public_only_refusal(&outcome, asi_hard)` refuses boot under `asi-hard` — the
posture whose whole contract is that no security control may be silently
disabled — while returning `None` under every other posture. The
degraded-but-running verify-only state permitted since v0.7 is **not** silently
tightened.

## #3155 — `…REQUIRE_ATTESTED_IDENTITY=enforce` was silently inert with zero enrolled keys

`enforce_for_request` returns `None` on an empty enrolled map in every mode
(`src/handlers/identity_binding.rs:154-156`) and **that stays unchanged**: the
#1985 unsatisfiable-default trap means an `enforce` posture which refused every
named caller when nobody *could* be key-attested would brick the deployment. The
defect is that an operator who deliberately selected `enforce` got no signal the
control was disarmed — unlike the #1570 admin-gate boot WARN or the #3065
header-trust boot refusal — so a deployment could believe it was refusing
spoofed `X-Agent-Id` headers while serving every one of them `200`.

**Fix.** A pure `inert_enforce_boot_reason(mode, enrolled_count)` returns
`Some(reason)` for exactly that one combination; boot renders it as a WARN under
the default posture and a refusal under `asi-hard`. **No request-path behaviour
changes.**

## #3198 — the key DIRECTORY had no posture gate at all

`src/identity/keypair.rs:814-820` at base:

```rust
pub(crate) fn ensure_parent(path: &Path) -> Result<()> {
    if let Some(parent) = path.parent() {
        fs::create_dir_all(parent)
            .with_context(|| format!("creating key directory {}", parent.display()))?;
    }
    Ok(())
}
```

No mode, no world/group-writable check. The log and audit trees have had
`log_paths::enforce_not_world_writable` (`src/log_paths.rs:275-313`) plus an
explicit `0o700` `log_paths::ensure_dir_secure` (`:321-337`) since v0.7 — so on
a `umask 0002` host the LOG material was strictly better protected than the
signing identity that vouches for it. `~/.config/ai-memory/keys` lands `0o775`,
and a second local UID unlinks and replaces `daemon.priv`/`daemon.pub` with a
keypair it controls. Every file-level control then **passes** on the planted
pair:

* the `0o600` mode check — the attacker writes `0o600`;
* the private-derives-public cross-check in `load` — the attacker plants a
  MATCHED pair;
* the #1790 single-open fstat TOCTOU fix — the swap happened before `load` ran.

The daemon then signs `signed_events` and audit witnesses with the attacker's
identity, those forged signatures VERIFY, and tamper-evidence reports CLEAN.

**Fix.** Directories this crate creates are born `0o700` via `DirBuilder::mode`,
which applies at `mkdir(2)` time — a `create_dir_all` followed by a `chmod`
would leave a real group-writable window — and the mode lands on every directory
in a freshly-created tree, not just the leaf. `enforce_key_dir_secure` refuses a
group- or world-WRITABLE key directory at resolution (`default_key_dir`), at
every read (`load_public`, hence `load`), at every write (`ensure_parent`, hence
`save` / `save_public_only` / the rotation archive), and in the existence gate
(`ensure_keypair`, which decides from file existence — an answer an attacker
with directory write controls). `encryption::{load,save}_keypair_from/to_disk`
(X25519) takes the same gate.

A leaf-only check is not enough for a #1514 SPIFFE-style slashed `agent_id`
(`campaign/region/host`): the key files sit several directories below the
caller-supplied key dir, and write access to ANY ancestor is enough to replace
the subtree. `enforce_key_path_chain_secure(base, file)` walks every ancestor
from the file's parent up to and including `base` and applies the same `0o022`
gate at each step.

**Folded finding (R-405, quoted live at `src/governance/capability.rs:1170`
before the fix):** `write_root_secret` did a bare `std::fs::create_dir_all(dir)`
for the directory holding `<issuer>.caproot`, a `0o600` bearer secret, with
**no directory cross-check at all** — `read_root_secret` only inspected the
FILE mode (`mode & 0o077`). A directory-writable swap plants a secret the
attacker controls, and every subsequent mint VERIFIES. Both `write_root_secret`
and `read_root_secret` now go through `ensure_parent` +
`enforce_key_path_chain_secure`. Same `0o022` contract, same `0o755` acceptance.

`ensure_and_load_daemon_keypair`'s `default_key_dir()` failure arm went
`tracing::info!` -> `tracing::warn!` with the cause rendered. That arm used to
mean only "the OS advertises no config directory"; it now also carries the
#3198 posture REFUSAL. Refusing to sign with a key another local UID can swap is
the correct fail-closed outcome, but at INFO it was indistinguishable from an
absent `HOME` and sat below the default log filter — the exact silence #3147
exists to end.

**The gate is `0o022`, not `0o077` — deliberately.** Refusing a merely
group/other-READABLE `0o755` directory would brick every deployment created
under the default `umask 022`: a silent tightening of a shipped default, which
this fix must not do. Directory read access does not enable the swap; directory
write access does. The remediation text still names `chmod 0700`, and
pre-existing directories are checked, never silently rewritten.

---

## Tests

| Test | Pins |
| --- | --- |
| `tests/agent_pubkey_failclosed_3145.rs` | a signed write over a deterministically locked DB returns `Err`, not `Ok(Claimed)`; the error names the DB fault; no `attest_level` is stamped |
| `tests/agent_pubkey_error_parity_3145.rs` | the three-row contract both backends share (bound→`Ok(Some)`, unregistered→`Ok(None)`, fault→`Err`), sqlite + `#[ignore]`d pg twin |
| `tests/keypair_write_crash_injection_3146.rs` | a failing writer mid-`save` leaves the ORIGINAL keypair byte-for-byte intact, still loading and signing, with no staging residue |
| `tests/keypair_persistence_ordering_3146_3147.rs` | private-half-first ordering; the four-way existence gate; the `asi-hard` posture split; the re-derived `.pub` verifies the surviving `.priv`'s signatures |
| `tests/identity_enforce_inert_boot_3155.rs` | the boot verdict fires for exactly `enforce` + zero keys, **and** the request path stays inert (the documented default is not tightened) |
| `tests/key_dir_posture_3198.rs` | `0o700` creation under `umask 000` (leaf **and** intermediates); `0o775`/`0o777` refused on load, save, and the existence gate with a `chmod 0700` remediation; `0o755` still accepted, unchanged; a group-writable **ancestor** of a #1514 slashed agent path refuses even when the leaf is `0o700` |
| `src/governance/capability.rs` unit tests | `write_root_secret` / `read_root_secret` refuse a `0o775` parent and still accept `0o755` |
| `src/identity/keypair.rs` unit tests | collision-safe archive naming at a fixed `ts`; the exhaustion refusal leaves every prior archive intact; staging cleanup on a failed replace; mode applied at creation |

### Crash injection — how, and why a subprocess

The fault that matters is "the write fails *after* the file was opened", i.e.
the `ENOSPC`/`EIO` class. Permissions cannot express it: a non-writable
directory blocks the `unlink` and the `create` alike, so the pre-#3146 code
would also leave the file intact and the test would prove nothing.
`RLIMIT_FSIZE = 0` expresses it exactly — creating a zero-length file still
succeeds, and the first `write(2)` of key bytes returns `EFBIG`. `setrlimit` is
process-wide, so it runs in a re-exec of the same test binary (the
`governance::deferred_audit` subprocess-isolation precedent) and the limit is
restored by a `Drop` guard the moment the injected call returns.

### Removal proofs (each run locally)

* revert `write_with_mode` to remove-then-create →
  `failing_writer_mid_save_leaves_the_original_keypair_intact_3146` reds with
  `left: []` — the pre-fix code leaves the sole private key **zero-length**;
* revert `save` to public-key-first →
  `save_writes_the_private_half_first_3146` reds (the private half still holds
  the original key, i.e. `save` aborted before reaching it);
* revert `agent_pubkey` to `.ok().flatten()` → `sqlite_agent_pubkey_contract_3145` reds;
* drop the `enforce_key_dir_secure` calls (or zero `KEY_DIR_FORBIDDEN_BITS`) →
  every `refuses` test in `tests/key_dir_posture_3198.rs` reds; drop the
  `DirBuilder::mode` → `a_fresh_key_directory_is_created_0700_3198` reds;
* the #3147 and #3155 suites consume API surface that does not exist without the
  fix (`RepairedPublicFromPrivate`, `PublicOnlyDegraded`, `public_only_refusal`,
  `inert_enforce_boot_reason`), so reverting either fix fails the tests at
  compile time — the strongest form of the proof.

## Decisions and residuals

* **`is_registered_agent` (`src/storage/mod.rs`, `.is_ok()`) is deliberately NOT
  touched.** #3182's own text assigns ownership explicitly — *"Fold with #3145
  (Opus P in flight — coordinate: P owns `agent_pubkey`; this issue owns the
  other four)"* — and no open PR edits it yet. Left to the sqlite-parity
  cluster so the hunks stay disjoint.
* **No `register_agent` / metadata preserve-list hunks.** The `src/storage/mod.rs`
  diff here is exactly two hunks, both inside the `agent_pubkey` *reader*;
  `src/store/postgres.rs` is untouched. No overlap with #2941 / PR #3116.
* **Doctor surfacing is deferred.** Both issues suggest a `doctor` fact.
  `doctor --posture` is the certified enterprise-federation evaluator whose
  check list is a pinned SSOT (`ENTERPRISE_FEDERATION_CHECK_COUNT`, the cert
  doc, and the removal-proof harness), and there is no general `Identity`
  section to hang a fact on — adding one moves the
  `local_run_on_empty_db_produces_thirteen_sections` pin. The observability the
  issues actually ask for (WARN + posture refusal + self-heal) is delivered
  here; the doctor row belongs in its own change alongside the cert-doc lane.
* **`write_root_secret` FOLDED into #3198**, not left as a sibling issue. The
  operator successor brief named it; R-405 confirmed the live site
  (`src/governance/capability.rs:1170` was `std::fs::create_dir_all(dir)` with
  a `0o600` bearer secret and no directory write-bit check). Same defect class,
  same gate, same `0o022` (not `0o077`) contract.
* **`umask` interaction.** The `0o022` gate refuses a key directory created
  under `umask 0002` (`0o775`). Several `#[cfg(test)]` helpers in
  `src/cli/rules.rs` and `src/daemon_runtime.rs` `create_dir_all` their key dirs
  with no mode, so on a `umask 0002` DEV BOX those tests now refuse. CI runners
  use `umask 022` (`0o755`, accepted) and every local leg here was run under
  `( umask 022; … )` per the standing rule. Tightening those helpers is a
  test-only follow-up.
* **Residual: a MISMATCHED pair is not auto-repaired.** A crash between the two
  renames of a *re*-save leaves a new `.priv` beside a stale `.pub`.
  `ensure_keypair` reports `AlreadyExists` and `load` then refuses loudly
  (`"private key … does not match public key …"`, WARNed by the daemon), so the
  state fails CLOSED and the irreplaceable half is intact — but it does not
  self-heal, and under `asi-hard` it degrades to a WARN where the public-only
  shape refuses. Auto-re-deriving `.pub` from `.priv` on a mismatch was
  considered and rejected: it would silently overwrite a verification anchor in
  the case where the *private* key is the wrong one (a bad restore). Recommend a
  follow-up to make `asi-hard` refuse on any "cannot sign" outcome, not just the
  public-only one.

## Commits

SSH-signed commits (`%G?` = `G`), split by issue where the code allows:

| Commit | Issues |
| --- | --- |
| `065440a3` | #3145 — `agent_pubkey` `.optional()` + the federation-receive de-silencing |
| `9c90579c` | #3146, #3147, #3198 (leaf gate) — the identity keystore. Kept together because the three are mutually dependent in one module: #3147's self-heal arm exists precisely because #3146's ordering produces that half-state, and #3198's gate is what makes the arm's existence test trustworthy. |
| `25bfdb88` | #3155 — self-contained (`identity_binding.rs` + one boot block + its own test) |
| `4052357a` | #3198 whole-chain gate + capability-token keystore fold (`write_root_secret` / `read_root_secret`) + QUAL-10 lockstep for `daemon_runtime.rs` 13_060 → 13_120 (#3155 wiring) |

## CI parity run locally

All cargo via `/home/fate_two/.local/bin/build-gate.sh`, PATH-pinned to
Rust 1.96.0, `umask 022`:

* `cargo fmt --all -- --check`
* `cargo check --all-targets` (default) + `--features sal` + `--features sal-postgres`
* `cargo check --examples`
* five clippy legs, all `-D warnings -D clippy::all -D clippy::pedantic`:
  default; `--all-targets --features sal`; `--features sal-postgres --tests`;
  `--features sal`; `--all-targets --features vectorlite`
* `scripts/check-hardcoded-literals.sh`; `scripts/check-docs-vs-ssot.sh`
* `qual_10_module_size_ceiling` (lockstep bump: `src/daemon_runtime.rs`
  13_060 → 13_120 for the #3155 boot-verdict wiring, measured 13_083) +
  `qual_6_7_legacy_error_type_ceiling`

Targeted tests green: `key_dir_posture_3198` (9), `keypair_write_crash_injection_3146` (2),
`keypair_persistence_ordering_3146_3147` (6), `identity_enforce_inert_boot_3155` (5),
`agent_pubkey_failclosed_3145` (4), `agent_pubkey_error_parity_3145` sqlite under
`--features sal` (1), lib `identity::keypair` (52), lib `governance::capability` (35,
including `root_secret_{refuses,accepts}_*_3198`).

Postgres twin `postgres_agent_pubkey_contract_3145` **RAN** against
`AI_MEMORY_TEST_POSTGRES_URL` (`--features sal-postgres -- --include-ignored`)
and fail-closed with `StoreError::SchemaAheadOfBinary` — the shared test DB is
schema **v90**, this binary understands up to **v89**. `src/store/postgres.rs`
is untouched; the refusal is #2445 working as designed, not a skip. sqlite
contract of the same file is green.

Two literals crossed the pm-v3.1 duplication gate and were consolidated into
named consts rather than baseline-bumped: `KEYPAIR_TRACE_TARGET`
(`"identity::keypair"`, 3 sites) and
`security_profile::ASI_HARD_REFUSAL_PREFIX` (the `asi-hard` refusal preamble, 3
sites — the two pre-existing ones in `security_profile.rs` now reference it and
render byte-identical text).

## AI involvement

Authored by Claude Opus 5 under operator dispatch in worktree
`wt-identity-failclosed`. All four commits are SSH-signed (`%G?` = `G`).
No external code was adopted.


alphaonedev:
**Fable gate — FIX-BEFORE-MERGE** at `4052357a` (shadow-QC lane, 2026-08-23). Read every `src/` hunk + the six test crates. `git merge-tree` vs `ae3e0aec` is CLEAN — GitHub's CONFLICTING is the stale-cache shape. Commits 4/4 `G`.

**Established as claimed:** #3145 `storage/mod.rs:11676-11679` `.optional()` + context (pg twin already propagated → parity restored; `agent_pubkey_error_parity_3145.rs` three-row contract). #3146 `write_with_mode` = same-dir staging file created AT the requested mode → `write_all` → `sync_all` → `rename` → dir fsync; every pre-rename failure removes the staging file (`keypair.rs` commit closure); `save` priv-before-pub; `archive_public_key_at` `create_new` + `-N` suffix + refusal at 64 (`archive_public_key_at_never_overwrites_a_prior_archive_3146`); X25519 writer takes the same ordering (`encryption/mod.rs:300-309`). #3147 four-way `(pub.exists, priv.exists)` table; `(false,true)` re-derives via `read_private_key_file` (shared reader, no second TOCTOU discipline); `(true,false)` → `PublicOnlyDegraded` + WARN, never regenerated; `public_only_refusal` refuses boot only under asi-hard. #3155 `inert_enforce_boot_reason` fires for exactly `Enforce ∧ enrolled==0`; request path untouched. #3198 `DirBuilder::mode(0o700)` at mkdir time, `0o022` gate at `default_key_dir`/`load_public`/`list`/`ensure_parent`/`ensure_keypair`, whole-chain walk for slashed ids (`a_group_writable_ancestor_of_a_slashed_agent_path_refuses_3198`), `0o755` accepted. Crash injection via `RLIMIT_FSIZE=0` re-exec is the right primitive. Literal hoists (`KEYPAIR_TRACE_TARGET`, `ASI_HARD_REFUSAL_PREFIX`) keep the ratchet flat.

**Findings**

1. **HIGH (ESTABLISHED, coordination) — the `write_root_secret`/`read_root_secret` fold collides with PR #3217.** #3214 assigned that fix to the sec2 cluster and says *"mirror locally — do not import the unmerged identity helper"*; `origin/sec2-secret-handling` does exactly that (local `CAPROOT_DIR_MODE`/`CAPROOT_DIR_FORBIDDEN_BITS` + its own `DirBuilder`, and it deletes the same `capability.rs` `create_dir_all` line this PR deletes at `:1170`). Two PRs now fix one line with two different gate implementations; whichever merges second conflicts, and the tree would carry two posture gates for one keystore. Resolve ownership before either merges: since #3214 is #3217's and #3217 is ahead in the queue, DROP the `capability.rs` hunk + the two `root_secret_*_3198` unit tests + the "folded finding" paragraph/CHANGELOG lines from this PR (the `enforce_key_path_chain_secure` reuse can land as a #3217 follow-up once both are in). If Grok decides the reverse, #3217 must drop its local copy and this PR must add `Closes #3214`.

2. **HIGH (ESTABLISHED, claims-truth) — `daemon_runtime.rs:3496-3498` WARN says "See `ai-memory doctor` -> Identity (#3147)", but no Identity section exists in `src/cli/doctor.rs` (this PR's own "Decisions" says so: doctor surfacing deferred).** An operator following the daemon's remedy finds nothing. Either add the section or change the text. Related: #3147 ("surface in `doctor --posture`") and #3155 ("doctor fact") each list the doctor fact as an acceptance item; the PR defers both. Under the fix-100% directive, either add a plain `doctor` Identity section in-PR (a general section, not `--posture`, does not touch the certified `ENTERPRISE_FEDERATION_CHECK_COUNT` SSOT; the `thirteen_sections` pin is a lockstep bump, not a gate), or file the follow-up issue now and change `Closes #3147 #3155` → `Addresses` so the issues stay open until the doctor fact lands.

3. **MED (ESTABLISHED) — asi-hard is inconsistent across the "cannot sign" outcomes.** `ensure_and_load_daemon_keypair` (`daemon_runtime.rs:3425-3431`, `:3466-3470`, `:3494-3512`) returns `Ok((None, None))` + WARN when `default_key_dir()` REFUSES (the #3198 gate the PR calls "the correct fail-closed outcome"), when `ensure_keypair` errs, and when `load` yields public-only/Err — i.e. under asi-hard the daemon boots signing nothing, the exact state `public_only_refusal` refuses two lines later. The PR's own residual names this. #3147's acceptance is "regenerate/**refuse per posture**"; make every no-signing-identity arm `bail!` when `is_asi_hard()` (default posture unchanged: WARN + degrade). Small, in-PR.

4. **MED (ESTABLISHED) — `enforce_key_dir_secure` checks mode bits only, never ownership**, and the chain stops at `base` while the PR body asserts "write access to ANY ancestor is enough to replace the subtree". A key dir (or any chain member) owned by another uid at `0o755` passes the gate yet its owner can swap the pair; `~/.config` / `~` above `base` are never examined. OpenSSH's `secure_filename` is the precedent: each component must be owned by the caller or root AND `!(mode & 022)`. Add the owner check (uid == geteuid() || uid == 0) in-PR; document that the chain contract ends at `base` (above-base ancestors are out of contract because of sticky `/tmp`-style parents) rather than implying full-path coverage.

5. **MED (ESTABLISHED, upgrade impact) — pre-existing `~/.config/ai-memory/keys` trees created by earlier releases under `umask 0002` (Ubuntu UPG default) are `0o775` and will now REFUSE every CLI key op and the daemon's signing until `chmod 0700`.** The refusal is correct, but CHANGELOG has no explicit upgrade note — add one line under the #3198 entry naming the symptom, the one-command remedy, and that the daemon still boots (degraded) under the default posture.

**Follow-ups (non-blocking)**
- (a) `sync_dir` swallows the directory-fsync error (`let _ = handle.sync_all()`), while `save` docs/CHANGELOG state "fsync the directory" as part of the durability guarantee — say best-effort, or propagate.
- (b) `enforce_key_dir_secure` uses `fs::metadata` (follows symlinks): a symlinked key dir is judged by its target's mode and the symlink's own parent is not in the chain (PLAUSIBLE) — `symlink_metadata` + refuse a symlink component, or document.
- (c) Federation-receive lane (`federation_receive.rs:651-668`, `registry_key_or_warn`): a registry fault on a same-sender row still lands the row durably `claimed` (now WARNed). `sync_push` can return a 503 for the batch; a retryable refusal beats durable mis-provenance. Candidate `fable-qc` issue if not folded.
- (d) #3145 said "triage with #2941" — #2941 was folded into merged #3116; a one-line cross-reference on #3145 at close.

Do not merge until (1) ownership of the caproot hunk is settled with #3217, (2) the doctor reference is true (section added or text changed) and #3147/#3155 closure matches what landed, (3) asi-hard refuses every no-signing arm, (4) owner check + chain-contract doc, (5) CHANGELOG upgrade note. Then 34/34 on the new head + a later APPROVED gate.

alphaonedev:
**Grok gate — Fable FIX-BEFORE-MERGE applied** at `2eb89148` (rebased onto `41142302` Merge #3217). Local CI-parity ALLGREEN (5 clippy + fmt + examples + qual_10/6_7 + identity tests + docs-vs-ssot + hardcoded-literals). Commits 6/6 `%G?=G`.

R-405 of the five FIX-BEFORE-MERGE items on this head:

1. **Caproot ownership (HIGH)** — `git diff origin/release/v1.0.0 -- src/governance/capability.rs` is **empty**. `#3217` is SSOT for `create_caproot_dir_secure` / `enforce_caproot_dir_secure` (#3214). This PR does not re-implement them (CHANGELOG Unreleased: "this change does not re-implement that keystore"). Duplicate `root_secret_*_3198` tests dropped in `54b6ef02`.
2. **Doctor Identity section (HIGH)** — `src/cli/doctor.rs:73` `SECTION_IDENTITY`; `identity_keystore_facts` / `identity_dir_mode_facts` / `identity_signing_facts`; pin `local_run_on_empty_db_produces_sixteen_sections` asserts `"Identity"` after `"Configuration"` (16 sections). Daemon WARN "See `ai-memory doctor` -> Identity" is true. Does **not** touch `ENTERPRISE_FEDERATION_CHECK_COUNT`. `Closes #3147 #3155` stands.
3. **asi-hard no-signing refusal (MED)** — `no_signing_identity_refusal` (`keypair.rs:957`) on all four `ensure_and_load_daemon_keypair` arms (`daemon_runtime.rs:3492`, `:3510`, `:3552`, `:3564`). Default posture still WARN+degrade. Test `no_signing_identity_refusal_is_asi_hard_only`.
4. **Owner check + chain contract (MED)** — `enforce_key_dir_secure` requires `key_dir_owner_ok(uid, euid)` = `uid == euid || uid == 0` (`keypair.rs:1202–1230`, UNSAFE-01 `geteuid` SAFETY). Walk **STOPS at `base`**; above-base ancestors out of contract (doc on `enforce_key_path_chain_secure`).
5. **CHANGELOG upgrade note (MED)** — Unreleased `#3198` **Upgrade:** pre-existing umask-0002 `0o775` trees refuse until `chmod 0700 ~/.config/ai-memory/keys`; default posture daemon still boots degraded; asi-hard refuses.

Waiting 34/34 on `2eb89148`. Will `fable-merge.sh` when green. Never tag.

alphaonedev:
**Fable gate — APPROVED** at `2eb89148` (re-gate; all five FIX-BEFORE-MERGE items verified on this head). (1) HIGH coordination: `git diff origin/release/v1.0.0 -- src/governance/capability.rs` is EMPTY — #3217 is SSOT for the #3214 caproot keystore, duplicate tests dropped, CHANGELOG says so. (2) HIGH claims-truth: `doctor.rs:73` `SECTION_IDENTITY` + facts fns + 16-section pin — the daemon WARN's pointer is now true; `Closes #3147 #3155` stands. (3) MED: `no_signing_identity_refusal` on all four `ensure_and_load_daemon_keypair` arms, asi-hard-only, default stays WARN+degrade. (4) MED: ownership half added (`geteuid` + `key_dir_owner_ok(md.uid(), euid)` at `keypair.rs:1200-1203`). (5) MED: CHANGELOG upgrade note for pre-existing `0o775` key dirs under `umask 0002` present. Merge when 34/34 on `2eb89148` via `fable-merge.sh`.

alphaonedev:
**linux-fed RED on `2eb89148`, fixed on `a84a41bf`.** Do not `gh run rerun` the failed SHA.

`Check (linux-fed,enterprise-fed)` failed 7 binaries (`--lib` doctor/rules + `crypto_erase_1956` + `doctor_cli` + `encryption_at_rest` + `encryption_upsert_key_2383` + `g_phase_e_3_rules_key_naming_compat` + `governance_l16_activation`). Root cause ESTABLISHED: `#3198` correctly refuses the self-hosted runner's real `~/.config/ai-memory/keys` at mode `0o775` (umask 0002). Ubuntu-latest was green because that keystore is missing there (`enforce_key_dir_secure` allows a non-existent dir).

Fix (does **not** chmod the operator store):
- `TestEnv` / doctor CLI child pin `AI_MEMORY_KEY_DIR` at a 0700 sandbox.
- encryption integration tests pin the same.
- `rules list` / `rules check` stay documented no-key verbs (no default-dir resolve).
- `rules keygen --out PATH` no longer consults the unused default dir.
- g_phase_e_3 key dirs `mkdir` + `chmod 0700` (tempfile is 0775 under umask 0002).

Local evidence: the previously-red lib tests + those integration binaries now pass. Waiting 34/34 on `a84a41bf`.

alphaonedev:
**Fable gate — APPROVED extended to `a84a41bf`.** Delta since `2eb89148`: (a) `cli/rules.rs` — `List`/`Check` (documented read-only/no-key) and `Keygen --out PATH` no longer resolve the default key dir, so the #3198 gate cannot refuse read-only commands on a `0o775` host keystore — correct scoping of the gate to paths that actually touch keys, CHANGELOG'd; (b) test-only key-dir sandbox in `TestEnv` (0700 tempdir via `OnceLock`, process-constant so the `set_var` cannot flap across concurrent tests — consistent with the #2905 isolation lesson) + test updates. No production key-path behaviour weakened: mutating rules actions still resolve and gate. Merge when 34/34 on `a84a41bf`.

alphaonedev:
**`a84a41bf` RED on Postgres/vectorlite feature gates; fixed on `e769d4ad`.** Do not rerun `a84a41bf`.

Cause: `tests/key_dir_sandbox.rs` was a cargo integration-test *crate* (`tests/*.rs`), so `clippy --features sal-postgres --tests` and `clippy --all-targets --features vectorlite` failed on `doc_markdown` / `missing_panics_doc`. Helper moved to `tests/common/key_dir_sandbox.rs` (not a test target). Local both clippy legs green. Waiting 34/34 on `e769d4ad`.

alphaonedev:
**Fable gate — APPROVED extended to `e769d4ad`.** Delta since `a84a41bf` is test-file organization only (key_dir_sandbox folded out of its own clippy `--tests` crate; zero `src/` lines). Merge when 34/34 on `e769d4ad`.

Issue #3204, opened 2026-08-22
[hardening batch] red-team LOW findings: pg GET /memories/{id} leaks far-endpoint ids of other tenants' rows; signal/checkpoint canonical CBOR lacks domain separation (#1931 gap); create_link_inbound substitutes untruncated receiver-now for peer valid_from=None (never re-verifies); MCP non-object arguments → {} and resolve_mcp_agent_id skips validate_agent_id (reserved sentinels forge the audit actor); backup prune bare Result discard; remediation hints echoed in wire bodies


alphaonedev:
Scout-verified (R-405) at `93d09145` from the ox-alpha campaign-6 red-team swarm (`.local-runs/ox-c6-redteam/`). pg lane evidence where pg is touched: `AI_MEMORY_TEST_POSTGRES_URL=$(cat .local-runs/ENTERPRISE-FED-TEST-PG-URL.txt) cargo test --features sal-postgres -- --include-ignored`.
1. `src/handlers/memories.rs:84-88` pg: `list_links(None)` filtered on ANCHOR only — far endpoint never `is_visible_to_caller`-filtered → cross-tenant id/relation enumeration (no content).
2. `src/identity/sign.rs:516-518`, `:717-723` `canonical_cbor_signal`/`canonical_cbor_checkpoint_resolution` lack `domain_separation_pair` (WRITE has it at `:399`) — #1931's guarantee does not hold for those contexts (latent; ERRORS-09).
3. `src/storage/mod.rs:8394-8397` `create_link_inbound` persists untruncated `Utc::now()` for a peer's `valid_from: None` (outbound discipline truncates at `:8118`) → a genuinely peer_attested edge never locally re-verifies (reports Tampered — fail-closed but false). Fold into #3178 (CC2).
4. `src/mcp/mod.rs:3274-3279` non-object `arguments` → `{}` instead of INVALID_PARAMS; `:214-224` `resolve_mcp_agent_id` reads `arguments["agent_id"]` with NO `validate_agent_id` → reserved sentinels (`validate.rs:342-348,436`) forge the audit actor. Fold into #3171 (U2).
5. `src/cli/backup.rs:521` `let _ = std::fs::remove_file(&path);` uncommented Result discard on retention prune → silent unbounded snapshot growth (ERRORS-19).
[...]

Issue #3161, opened 2026-08-22
[data-loss] gc(archive=true) drops every link of the archived memory (never calls archive_links_for_memory) while forget(archive=true) and archive_by_ids preserve them — asymmetry invisible to operators; pinned as 'documented loss' by a test referencing CLOSED #1771


alphaonedev:
Scout-verified at cda57210. `gc(conn, archive=true)` (`src/storage/mod.rs:12836`) copies the row into archived_memories (`:12893-12913`) then `DELETE FROM memories WHERE id IN (…)` (`:12963-12966`) WITHOUT `archive_links_for_memory` (`:4335-4348`, whose only caller is `archive_memory_no_tx` at `:4425`). FK cascade reaps every edge; `restore_archived` returns the row with an empty graph. Postgres `run_gc` (`src/store/postgres.rs:24437-24500`) has the same shape (consistent, not divergent). Pinned as intended by `restore_does_not_recover_links_documented_loss_1771` (`src/storage/mod.rs:21557`, "Flip this to expect 1 when archived_memory_links preservation lands") — but #1771 is CLOSED and preservation DID land for forget/archive_by_ids (`archive_restore_preserves_links_pg_1771`). So the loss is now an unexplained asymmetry: whether a memory keeps its links across archive→restore depends on WHICH path archived it.

Fix: call `archive_links_for_memory` in the gc archive path on both backends; flip the pinning test to expect the links; parity test twin. GA: BLOCKING (unintentional data loss on a routine path; prime directive).

alphaonedev:
Closed by merge of #3221 at `b3ffd9599113e2d6782901e10193d6d361e46cbc` G (fable-merge.sh does not auto-close).

- #3158 claims matrix lockstep
- #3160 rebuild tests
- #3161 archive-then-delete edge-graph (lineage CID pins still #3250)

alphaonedev:
Merged via #3221 b3ffd9599113e2d6782901e10193d6d361e46cbc.

PR #3220, opened 2026-08-22
fix(#3171,#3202): MCP tool-contract audit + promote destination governance

### LINKED ISSUE #3171: [mcp] tool-contract audit (103 tools): 24 confirmed schema↔handler defects — 4 fail-open shapes (required-but-unwrap_or_default, unknown filter widens results, undeclared agent_id disarms owner gate, unbounded zstd), 9 tools with undeclared agent_id, 4 ignored params, 6 overclaimed descriptions
ox-alpha tooldesc swarm + Opus scout R-405 verification at cda57210 (report: `.local-runs/ox-c4-tooldesc/` + scout table). Schemas are schemars-derived from `*Request` structs while handlers read `params[...]` directly — nothing couples the two; `assert_property_set_parity` pins schema-vs-SNAPSHOT only, so this class is invisible to CI.

**Fail-open / authz (fix handlers, BLOCKING):**
- `memory_action_next` `src/mcp/tools/action.rs:429-432` and `memory_signal_thread` `src/mcp/tools/signal.rs:420-423`: schema-REQUIRED field read via `unwrap_or_default()` → malformed call queries `""` and returns a plausible empty success (ERRORS-08). Reject blank.
- `memory_checkpoint_query` `src/mcp/tools/checkpoint.rs:387,391` and `memory_routine_list` `src/mcp/tools/routine.rs:519`: `.and_then(Enum::from_str)` on a filter → unknown discriminant REMOVES the filter and returns MORE rows (same class #3136 fixes for `tier`; not covered there). Reject unknown, mirroring `handle_checkpoint_create`.
- `memory_namespace_clear_standard` `src/mcp/tools/namespace.rs:711,728-732`: undeclared `agent_id`; the #1777 owner gate runs only if it is present → omitting an undocumented param disarms authz on an op that disarms the delete/write/promote gates for the whole namespace. Declare + document; decide whether unclaimed callers should be refused.
- `memory_skill_get` `src/mcp/tools/skill_get.rs:158` `zstd::decode_all` unbounded while sibling `skill_resource.rs:58` uses `decode_all_bounded` (#1933 parity). Also undisclosed: every fetch appends a `skill.invoked` signed_events row (`:54,218-219`).
- `memory_archive_purge` `src/mcp/tools/archive.rs:47-128` reads undeclared `as_admin`/`agent_id`; schema/docs (`:215-222,:236`) imply global purge, default is caller-scoped; `Decision::Ask` returns success-shaped `status:ask`.
- `memory_skill_delete` `src/mcp/tools/skill_retire.rs:274-280` `let _ = active_keypair;` — MCP path has NO admin gate (HTTP twin `src/handlers/skills.rs:430` has `require_admin`); `force=true` also skips retire-first. Document the asymmetry at minimum; decide whether to gate.
- `memory_offload.ttl_seconds` `offload.rs:55` and `memory_check_agent_action.byte_estimate` `check_agent_action.rs:156-158`: `i64` schema read via `Value::as_u64` → schema-valid negatives silently dropped (offload then gets the config default TTL — retention-integrity gap). Validate at the boundary.

**Undeclared-but-honoured params (add to request structs):** `agent_id` on `memory_delete` (also accepts a unique-PREFIX via `get_by_prefix` `delete.rs:111` — undocumented on a hard delete), `memory_update`, `memory_link`, `memory_atomise`, `memory_skill_register`, `memory_skill_promote_from_reflection`; `caller_agent_id` on `memory_agent_register`; `source_uri` on `memory_search` (`search.rs:124-145`); `citations`/`source_span` on `memory_store` (`store/validation.rs:301-325`).
**Ignored params (remove or mark IGNORED):** `memory_signal_send.from_agent` (required but discarded, #2996 — make Optional + doc), `memory_capture_turn.tool_calls` (doc says 'preserved verbatim', never persisted), `memory_calibrate_confidence.output_format`.
**Overclaimed descriptions:** `memory_signal_inbox` (unacked only — `signals/mod.rs:217`), `memory_stats` (no archive field), `memory_skill_get` ('<5000 tok' unenforced), `memory_recall` (`limit.min(50)` at `recall.rs:1225,1358` undisclosed), `memory_list.format` (values/default `toon_compact` undocumented — honoured at `mcp/mod.rs:3361-3390`), `memory_verify` rustdoc `# Errors` (-32602 claim false; tools/call returns `isError:true`).

**PR mechanics (load-bearing):** `src/mcp/registry.rs:1433-1660` (`d1_6_987_tests`) pins description AND docs byte-equality, `expected_added` property allow-list (`:1616-1622`), and the `required` set; property REMOVALS are forbidden outright → needs an explicit removal-exception mechanism (own commit). Re-bless the five `tests/snapshots/tools_list_*.json`. Token budget: `description`/property docs count; `docs()` is stripped from tools/list. Add a CI guard: handler `params.get(...)`/`params["…"]` keys ⊆ request-struct fields per tool (closes the class permanently).

**Unaudited (re-running the ox lanes):** `memory_forget`, `memory_gc`, `memory_kg_invalidate`, `memory_promote`, leases, pending_*, persona*, entity_*, kg_query/timeline, find_paths, expand_query, export_reflection, inbox, ingest_multistep, get/get_links/get_taxonomy, quota_status. GA: BLOCKING for the fail-open/authz rows; the rest DEFERRABLE-but-cheap in the same PR.

### LINKED ISSUE #3202: [governance] memory_promote evaluates governance on the SOURCE namespace only — the destination namespace's write policy (Approve/Deny) is never consulted and the clone keeps the original owner stamp while landing in the broader-visibility namespace
Scout-verified (R-405) at `93d09145` from the ox-alpha campaign-6 red-team swarm (`.local-runs/ox-c6-redteam/`). pg lane evidence where pg is touched: `AI_MEMORY_TEST_POSTGRES_URL=$(cat .local-runs/ENTERPRISE-FED-TEST-PG-URL.txt) cargo test --features sal-postgres -- --include-ignored`.
**Code.** `src/mcp/tools/promote.rs:154` `enforce_governance(..., &target.namespace, ...)` (source); `:197` destination gets only `reject_reserved_write_namespace`; `src/storage/mod.rs:7469-7504` ancestor-relation check only, `metadata: source.metadata.clone()`. Confined to the caller's own ancestor chain (not arbitrary cross-tenant). Overlaps #3171 (promote undeclared agent_id) — coordinate with that PR.
**Fix.** Run Store-class `enforce_governance` against to_namespace before cloning; re-stamp clone provenance (`derived_from` + acting caller) instead of a blind metadata copy; audit under the destination namespace.
**Test.** namespace acme with write: Deny, row in acme/team → promote{to_namespace: acme} refuses; Approve → pending decision; clone's agent_id = acting caller. GA: DEFERRABLE (policy bypass within own chain).

alphaonedev:
Closes #3171
Closes #3202

Also lands #3204 items 4 and 7 (that issue stays open for its remaining items).

## What this is

The MCP tool-contract audit landed as one change: every tool's **advertised** input contract now matches its **enforced** one, and a new CI guard keeps it that way.

The root cause is structural. Tool input schemas are `schemars`-derived from the per-tool `*Request` structs, but **no handler deserializes that struct** — each reaches into the raw `arguments` bag with `params["k"]` / `params.get(k)`. And there is **no runtime JSON-Schema validation on the MCP path** (no `jsonschema`/`valico` dependency; `src/mcp/mod.rs` and `src/mcp/registry.rs` never validate). So a value that contradicted the advertised schema did not fail — it silently took the handler's fallback branch. The existing pin (`assert_property_set_parity`) compares schema against its own past snapshot, never schema against handler, so this whole class was invisible to CI.

## 1. Fail-open shapes (fixed, all refuse now)

New `src/mcp/param_guard.rs` supplies four boundary readers. **ABSENT optional values still take the documented default; only a PRESENT-but-contradictory value refuses** (ERRORS-08).

| shape | tools | what a malformed call used to do |
|---|---|---|
| schema-REQUIRED string via `unwrap_or_default()` | `memory_action_frontier`, `memory_action_next`, `memory_signal_thread`, all four `memory_lease_*` | queried `""` and returned a plausible EMPTY SUCCESS |
| unknown enum discriminant via `.and_then(E::from_str)` | `memory_checkpoint_query` (`state`, `condition_type`), `memory_routine_list` (`state`) | DROPPED the filter → strictly MORE rows |
| `i64`-declared integer via `as_u64` | `memory_offload.ttl_seconds`, `memory_check_agent_action.byte_estimate`, `memory_find_paths.max_depth`/`max_results`, `memory_pending_list.limit` | a schema-valid NEGATIVE read as ABSENT → silent server default |
| bool safety flag via `.as_bool().unwrap_or(false)` | `memory_forget.dry_run`, `memory_gc.dry_run`, `memory_archive_purge.as_admin` | `dry_run: "true"` ran a REAL irreversible delete |

**The lease case is the sharpest.** `crate::actions::lease_acquire`'s upsert guard is

```sql
WHERE leases.expires_at <= ?3 OR leases.holder = ?2
```

so two *distinct* agents that each omitted `holder` both resolved to `""`, the second matched the same-holder arm and **succeeded** — a **double grant of a single-holder coordination lease**. That function's own contract (`src/actions/mod.rs`) states "the worst case is a spurious `Conflict` a caller retries, **never two winners**". Refusing the blank at the MCP boundary is what makes that true.

## 2. `memory_forget`: a bulk DELETE can no longer be silently WIDER than the pattern

The destructive forget path shared `sanitize_fts_query(pat, false)` with the ranked-read paths. That sanitiser **clamps**: it drops a token that sanitises to empty, drops a standalone `AND`/`OR`/`NOT`/`NEAR`, and `.take(MAX_FTS_OR_TERMS)`s the rest. On a ranked read that is the right trade — fewer, broader hits. Under FTS5 **implicit AND**, every dropped token is a dropped *narrowing conjunct*, so the match set — and therefore the DELETE — becomes strictly wider than the pattern the caller wrote and the schema promised ("every whitespace-separated token must match", #1601).

Silently deleting rows the caller never named is unrecoverable data loss, so the destructive path now has its own strict builder that **refuses**; read paths keep the clamp.

The strict tokeniser is **backend-agnostic** (`storage::strict_forget_tokens`) and is applied at three points so a refusal cannot depend on which backend is configured — the #2312 cross-backend-divergence class:
- sqlite, via `forget_fts_query` (13 call sites now take `?`);
- the HTTP forget funnel (`src/handlers/memories_query.rs`), before backend dispatch;
- `PostgresStore::forget`, as defence in depth for the CLI/curator lanes.

## 3. A caller can no longer choose the principal its own request is judged as

`identity::resolve_agent_id` gives the **explicit wire argument precedence** over the `AI_MEMORY_AGENT_ID` env identity. That is correct for an attribution field and wrong for an authz **subject**: a handler that feeds a wire `params.agent_id` into `Permissions::evaluate` / `enforce_governance` / a capability-token binding / a quota key / an ownership predicate lets the caller pick the principal, **while the sibling owner gate stays keyed on the ENV caller** — two controls that can never agree by construction.

New `identity::resolve_governance_subject` generalises the already-shipped `handle_inbox` (#1557) / `handle_replay` (#1571) pattern:
- **multi-tenant posture** (`resolve_read_visibility_caller()` is `Some`): the subject **is** the caller; a disagreeing wire `agent_id` is **refused**;
- **single-operator default** (env unset): unchanged — the legacy ladder runs, so zero-config deployments are byte-identical.

Applied to fourteen tools: `memory_promote`, `memory_kg_invalidate`, `memory_archive_purge`, `memory_pending_approve`, `memory_pending_reject`, `memory_delete`, `memory_update`, `memory_store` (local **and** forward-to-HTTP `X-Agent-Id`), `memory_consolidate`, `memory_link`, `memory_atomise`, `memory_reflect`, `memory_offload`, `memory_deref`. The namespace-standard pair applies the same rule inline, because its single-operator fallback must degrade to `anonymous:invalid` rather than error (#2721 behaviour preserved byte-for-byte).

Three of these deserve naming:

- **`memory_pending_approve`** — the approver id is the subject of the separation-of-duties gate *itself*: `approve_with_approver_type` compares it against `requested_by` to refuse self-approval, matches it against the named approver, and counts **distinct** approver ids for `ApproverType::Consensus(n)`. Reading it from the wire let one caller defeat the self-approval refusal *and* forge a full human quorum by varying the parameter across N calls.
- **`memory_namespace_clear_standard`** — two holes. (a) The #1777 owner gate ran only when the caller CLAIMED an identity (`params.agent_id` present and non-empty), so omitting an **undeclared** optional parameter disarmed authz on the one operation that disarms every other gate in the namespace. This is the exact caller-opt-in shape #2541 already removed from the SET twin ("the gate must NOT be caller-opt-in via `params.agent_id`"), leaving CLEAR — the strictly more destructive half — as the only survivor. The gate is now unconditional; the CLI keeps an out-of-band trusted entry (`handle_namespace_clear_standard_trusted`), the exact mirror of what SET already had. (b) The caller string was the **left-hand side of the ownership predicate itself** (`authorize_namespace_standard_owner` refuses only when `recorded_owner != caller`), so a caller could satisfy the gate for any owned standard by naming that owner.
- **`memory_deref`** — `ContextOffloader::deref`'s SEC-4 IDOR gate is `if caller_agent_id != blob.agent_id { NotFound }`, and its source note asserts the MCP handler "always passes an AUTHENTICATED `caller_agent_id`". It did not: the value came verbatim from the request body. `memory_offload`/`memory_deref` also honoured an **undeclared second identity channel** (`metadata.agent_id`) on tools that have no `metadata` input at all — removed. `entity_register` had the same second channel inside storage, where an inline `metadata.agent_id` **won** over the id the handler had just put through `validate_agent_id` and so never crossed the `RESERVED_AGENT_IDS` check (#977); the resolved id now wins.

Two refusals additionally chain a `refuse` verdict to the forensic audit (attributed to the **enforced** caller, never the asserted id), so the attempts are not the ones that leave no trace: `memory_archive_purge` (where #913's remit is intent capture regardless of outcome) and `memory_pending_approve`.

Separately, the enterprise-audit `actor` field is now shape-validated before it reaches the SIEM emitter. It was the one site where an arbitrary unvalidated caller string reached the audit record — a log-injection vector on every store/update/delete/promote/forget/link/consolidate/approve/reject call.

## 4. `memory_quota_status`: an ungated READ no longer WRITES

It called `quotas::get_status`, which calls `ensure_row`, so any MCP caller could `INSERT OR IGNORE` an `agent_quotas` row for an arbitrary `(agent_id, namespace)` pair — an unbounded row-injection primitive that fills an operator's quota census with principals that never existed. New non-materialising `peek_status` / `peek_aggregate_status` twins return the identical wire shape from a synthesised default; the real write paths still materialise the row on first write, so no accounting is lost. Both parameters are validated at the boundary. `memory_ingest_multistep` gained the same missing `validate_namespace`.

Two more integrity fixes in the same family:
- **`memory_skill_get` decompresses bounded** — it was the lone skill decode site still on unbounded `zstd::decode_all` while its three siblings used the #1933 `decode_all_bounded` ceiling, so a hostile `skills.body_blob` that decodes to gigabytes would OOM the daemon on an activation fetch.
- **`memory_get_links` reports `attest_level` fail-closed** — the column is nullable and the field is `skip_serializing_if = "Option::is_none"`, so a NULL row made the trust field **vanish** while the docs promised every link carries one. Now `COALESCE(attest_level, 'unsigned')`, the same projection the reflection export already used.

## 4b. Governance gates that were missing, or pointed at the wrong namespace

**`memory_promote` governs the DESTINATION now (#3202).** Vertical promotion CLONES the row into `to_namespace` — it is a WRITE into that namespace. But `promote.rs` evaluated `GovernedAction::Promote` against `target.namespace` (the **source**), and the destination received nothing beyond `reject_reserved_write_namespace`. An operator who set `write: Owner` (or `Approve`) on an ancestor still had every row a caller could read from a descendant clonable into it — and the ancestor is normally the *more* protected namespace, so the boundary was crossable by construction.

A STORE-class `enforce_governance` now runs against `to_namespace` **before** the clone (Deny refuses; Approve queues exactly as a direct `memory_store` into that namespace would); the forensic row is chained under the destination. `enforce_governance` resolves `namespace_owner` only for `GovernedAction::Store`, which is why Store — not Promote — is the right class here: the clone is a new row with no prior owner in the destination, so an owner-level rule must judge it against the *namespace* owner.

The clone's provenance is **re-stamped**. It was carrying `source.metadata` verbatim (`storage/mod.rs`, `metadata: source.metadata.clone()`), so the destination row was owned by the source's author rather than the caller who put it there, and nothing in the row recorded where it came from — only the out-of-band `derived_from` edge did. The acting caller becomes the clone's `agent_id`; the original authorship is **preserved** under `promoted_from_agent_id`, alongside `promoted_from` / `promoted_from_namespace` / `promoted_at`. Durable truth is added to, never replaced. `promote_to_namespace` takes the acting caller explicitly so the MCP, CLI and approved-pending-action lanes each attribute correctly.

Surface audit for this one: **MCP** and **CLI** both gained the destination gate (the CLI was the other surface on which the ancestor's `write` policy could be bypassed). **HTTP** has no exposure — `handlers/memories.rs` hard-codes `to_namespace: None`, so vertical promotion is not reachable over HTTP. **Postgres** has no `promote_to_namespace` twin, so vertical promotion is sqlite-only. The **federation** lane reaches it only through `execute_pending_action`, i.e. after an approval, and that lane now attributes the clone to the approved requester.

**`memory_gc` was the one ungated destructive tool (#3204 item 7).** It reached the substrate with no permission gate, no governance consult and no forensic row — while its sibling `handle_archive_purge` carries all three, and gc is *strictly more destructive*: it deletes across every namespace and every owner, and with `archive_on_gc` off that delete is a permanent hard-delete + crypto-erase. A real sweep now clears three gates:

1. K9 `Permissions::evaluate`, op chosen by disposition (`MemoryArchive` for an archiving sweep, `MemoryDelete` for a reaping one), evaluated at the default namespace with `namespace_pattern = "**"` scoping — the `handle_archive_purge` convention.
2. The #1849 bulk-delete governance rule, on the **destructive disposition only**: refuse the sweep if **any** namespace holding reapable rows carries a non-`Any` `delete` level. Without this a `delete: Approve` legal-hold is no defence at all — the held rows simply expire and vanish on the next tick, with no approval and no trace. The probe uses the same predicate `db::gc` sweeps with, so it cannot miss a namespace the sweep would reach. An **archiving** sweep is deliberately exempt: the row moves to `archived_memories` and `memory_archive_restore` recovers it, so the hold is not defeated, and refusing there would strand expired rows in every deployment with any delete-governed namespace — a reliability cost with no integrity benefit.
3. An `allow` forensic row chained before the write.

`dry_run` stays ungated so an operator can always **see** what would be reaped. A `Decision::Ask` **refuses** here rather than returning the success-shaped `{status:"ask"}` envelope its sibling returns — a success-shaped body on an unperformed destructive op is itself one of the §6 findings, and a sweep has no per-call approval channel to route an Ask into.

**Two `tools/call` envelope fixes (#3204 item 4).** A present-but-non-object `arguments` is now `-32602` rather than coerced to `{}` — coercion meant `{"arguments": "namespace=acme"}` silently ran the tool with *every* argument absent, so a destructive tool took its unscoped defaults. (An absent `arguments` still means "no arguments".) And `resolve_mcp_agent_id` now applies the full `validate_agent_id` rather than only the shape gate, because the shape gate does not reject `RESERVED_AGENT_IDS` — a wire caller could stamp the enterprise-audit `actor` with `daemon`/`system`, the internal principals downstream gates carve out as "the internal path is exempt".

## 5. Declared parameters + the pins

26 parameters that handlers already honoured are now **declared**, so a schema-conformant client can discover them: `agent_id` on nineteen tools, `as_admin` on `memory_archive_purge`, `caller_agent_id` on `memory_agent_register`, `source_uri` on `memory_search`, `as_agent` on `memory_kg_query`, `citations`/`source_span` on `memory_store` (parsed, validated and persisted since #1421, never advertised), `signature`/`resolved_at` on `memory_checkpoint_resolve`, and the legacy `id`/`kind_inner` aliases. `param_names` census 133 → 135 (`AS_ADMIN`, `PIPELINE_VARIANT`).

The D1.6 (#987) registry pin (`src/mcp/registry.rs`) gained four explicit, documented exception tables, replacing two inline `if name == …` special cases:

- `PROPERTY_ADDITIONS` — every deliberate schema extension since D1.6, with its issue number.
- `PROPERTY_REMOVALS` — **new mechanism**. The pin previously forbade removals outright with no escape hatch, which meant a property the audit proved to be a lie could not be retired even when leaving it advertised was itself the defect. It carries the three-condition safety contract a removal must satisfy and is **deliberately empty here**: every ignored-parameter row was corrected by documenting the truth rather than by breaking the wire contract for clients that already send the field.
- `DOCS_CORRECTIONS` / `DESCRIPTION_CORRECTIONS` — `docs` is stripped from the wire `tools/list` payload, so a docs correction costs zero tokens; `description` is on the wire and token-budgeted, so only a description that is **false as written** earns an entry.

Each correction table asserts the text **actually changed**, so a stale entry fails the build rather than silently unpinning a tool. The property-key pin now reports every drifting tool in one run instead of fail-fast on the alphabetically-first.

**New CI guard `tests/mcp_handler_params_subset_of_schema.rs`** closes the class permanently: it parses every `params["k"]` / `params.get(param_names::K)` read under `src/mcp/tools/`, resolves the constants against the `param_names` SSOT, and asserts the read set is a subset of the properties declared by the tools implemented in the same module unit. A runtime probe cannot do this — a handler's key reads are data-dependent (a `kind`-specific branch reads `command` only for `kind: "bash"`), so no finite set of probe calls enumerates them.

## 6. Overclaimed descriptions (corrected, not merely noted)

`memory_gc` claimed "archives first" **unconditionally**; archiving is conditional on `archive_on_gc`, and with it OFF the sweep is a permanent hard-delete + crypto-erase. The response now carries `archived` so a caller can tell a recoverable move from an unrecoverable erase — previously it could not. Also disclosed: the sweep is substrate-wide and ungated, prunes the `recall_observations` ledger and expired signals, and (per #3161) does not archive link edges.

Also corrected: `memory_signal_inbox` (unacked only), `memory_stats` (no archive counts at all), `memory_skill_get` (the `<5000 tok` figure is unenforced, and every fetch **appends** a `skill.invoked` signed-events row — this "read" is audited), `memory_pending_list` ("default pending" was false — omitting `status` returns every status), `memory_recall` (`limit` silently clamps to 50), `memory_list` (default format is `toon_compact`, not JSON), `memory_kg_invalidate` (not idempotent when `valid_until` is omitted; permanently clears a signed link's signature and attest level), `memory_archive_purge` (caller-only default scope; success-shaped `{status:"ask"}` with nothing purged), `memory_delete`/`memory_promote` (accept a unique ID **prefix**), `memory_forget` (silently owner-scoped, and the owner predicate also matches **unowned** rows), `memory_entity_register` (merges aliases only — `metadata`/`agent_id` are discarded on the idempotent path), `memory_ingest_multistep` (never populates `ingested_memory_ids`; stores nothing).

Two asymmetries are now disclosed rather than silently carried: `memory_skill_delete` has **no admin gate** on the MCP path while its HTTP twin requires the admin role, and the `memory_lease_*` tools decide ownership entirely from a self-asserted `holder` (they receive no caller identity), so "an owned lease … held by the caller" was not enforced.

Three declared-but-never-read parameters are marked **IGNORED** rather than removed: `memory_signal_send.from_agent` (discarded since #2996 — authorship binds to the signing keypair), `memory_capture_turn.tool_calls` ("preserved verbatim" was false), `memory_calibrate_confidence.output_format` (read by no handler on any surface).

Finally, the `# Errors` rustdoc on `memory_verify` and `memory_reflection_origin` claimed the dispatcher returns JSON-RPC `-32602`. Per the MCP 2025-03-26 tool-result convention (documented on `crate::mcp::tools::atomise`) a handler error is a **successful** result carrying `isError: true`, so a client switching on the error code would never see them. `docs/USER_GUIDE.md`'s `memory_get_links` example showed a `signed_at` field that exists nowhere and a `signature` key that is never emitted.

## Rust standard

Per `rust-1.98`: **ERRORS-08** (refuse a caller-contract violation, never answer it) is the governing rule for §1 and the lease fix; **ERRORS-01/02/15** for the `Result`-returning `strict_forget_tokens` and the preserved-cause `map_err`s; **PERF-07** (`TryFrom`, never `as`, for narrowing) and **ERRORS-09/PERF-15** (validate at the boundary so downstream can trust the value) for the `param_guard` readers; **API-04** for keeping the exception tables private to the pin.

## Pins / SSOT / tests

Declaring 26 previously-honoured properties grew the verbose full-profile catalog from a measured **26,691** tokens past the old 25,000 ceiling. The ceiling is now the SSOT `crate::sizes::VERBOSE_FULL_PROFILE_CEILING_TOKENS = 28_000` (dated 2026-08-22, #3171), consumed by `tests/token_budget_guard.rs`, `tests/mcp_tools_list_schema_discovery.rs`, `tests/recall_decoration.rs`, and `src/cli/doctor.rs`. All five `tests/snapshots/tools_list_*.json` re-blessed. QUAL-10 lockstep: `src/storage/mod.rs` 29_160→29_320, `src/mcp/mod.rs` 16_580→16_680. QUAL-6 120→123, QUAL-7 43→44.

**Tests (lib + integration, sqlite):** `d1_6_987` 16/16; `_3171` 17/17 including `lease_handlers_refuse_blank_required_args_3171`, `entity_register_resolved_id_wins_over_inline_metadata_agent_id_3171`, `entity_register_refuses_reserved_explicit_agent_id_3171`, `forget_fts_query_refuses_instead_of_widening_3171`; `_3202` 2/2 `promote_to_namespace_enforces_destination_governance_3202` / `promote_destination_approve_queues_and_allow_restamps_3202`; `_3204` 4/4 including `gc_destructive_sweep_refuses_delete_governed_namespace_3204` and `resolve_mcp_agent_id_rejects_reserved_sentinel_3204`; `tests/mcp_handler_params_subset_of_schema.rs`; snapshot suite; `token_budget_guard`; `qual_10` / `qual_6_7`. Clippy ×5 (`-D warnings -D clippy::all -D clippy::pedantic`: default, sal all-targets, sal-postgres --tests, sal, vectorlite all-targets), `fmt --check`, `check --examples`, `check-hardcoded-literals`, `check-docs-vs-ssot` green.

**Postgres:** `postgres_forget_refuses_widening_operator_token_3171` is compiled (`--features sal-postgres --lib`) and was invoked against the shared `ai_memory_test`. The live DB is schema **v90**; this binary is **v89**, so `PostgresStore::connect` fail-closes with `SchemaAheadOfBinary` (#2445). No `AI_MEMORY_ALLOW_SCHEMA_AHEAD` hatch. The product tokeniser is pinned on sqlite (`forget_fts_query_refuses_instead_of_widening_3171`); the pg test is the same refuse-not-widen assertion once the cluster binary catches the schema tip.

## Deliberately NOT in scope

- The `.and_then(Tier::from_str)` sites on `memory_forget` — owned by #3136.
- The capability-token arms in `delete`/`forget` — owned by #3133.
- The gc archive path's link-cascade loss — #3161 (disclosed in `memory_gc`'s docs here, not fixed).
- Postgres `plainto_tsquery` drops English **stopwords** where FTS5 unicode61 does not, so a stopword-bearing forget pattern is still marginally wider on pg. That divergence is pre-existing and already documented in `PostgresStore::forget`; closing it needs a tsquery-lexeme round-trip and is not attempted here.


alphaonedev:
**Fable gate — FIX-BEFORE-MERGE** at head `9e4822dd` (shadow-QC lane, 2026-08-23). Branch base `faac3e3a`; `git merge-tree` vs `ae3e0aec` has REAL conflicts in `src/mcp/tools/forget.rs` (tier sites now owned by merged #3136 — keep #3136's refuse-unknown `tier` AND this PR's `strict_forget_tokens`), `tests/qual_10_module_size_ceiling.rs`, `tests/qual_6_7_legacy_error_type_ceiling.rs` (re-measure on the rebased tree).

**Accepted (ESTABLISHED at `9e4822dd`):** `param_guard` readers refuse PRESENT-but-contradictory values and keep absent→default (all 23 call sites); lease blank-`holder` double-grant closed at the boundary; `resolve_governance_subject` (`identity/mod.rs`) binds the subject to the enforced caller under multi-tenant posture on exactly the 14 named tools, single-operator ladder unchanged; `strict_forget_tokens` applied at sqlite `forget_fts_query`, the HTTP funnel and `PostgresStore::forget` (`postgres.rs` `InvalidInput`); `memory_quota_status` no longer materialises rows; `skill_get` on `decode_all_bounded` (`skill_get.rs:164`); `tools/call` non-object `arguments` → `-32602` with absent still `{}`; `gate_gc_sweep` probe uses the same `expires_at IS NOT NULL AND expires_at < now` predicate as `SQL_GC_EXPIRED_CHUNK_IDS` (`storage/mod.rs:12984`); `promote_to_namespace` re-stamp preserves `promoted_from_agent_id` and every production caller passes `Some(actor)` (`cli/promote.rs:151`, `promote.rs:297`, executor `:19712`); R-405 on the body's surface claims holds (HTTP `to_namespace: None` at `handlers/memories.rs:1686`; no pg `promote_to_namespace` twin).

**Findings:**
1. **HIGH (ESTABLISHED) — the #3202 `Approve` path can never execute.** The destination gate calls `enforce_governance(GovernedAction::Store, to_ns, …)` (`promote.rs` dest block; `cli/promote.rs` same), and a `Pending` decision is queued by `queue_pending_action(conn, action, …)` with `action_type = action.as_str()` = `"store"` (`storage/mod.rs:19008-19022`, `models/namespace.rs:300`) and payload `{id, to_namespace, mode:"vertical"}`. On approval `execute_pending_action` takes the `"store"` arm (`storage/mod.rs:19676-19690`) and does `serde_json::from_value::<Memory>(payload)` → `InvalidArgument("invalid store payload…")`. So `write: Approve` on a destination makes vertical promotion impossible, not approvable — contradicting the body's "Approve queues exactly as a direct `memory_store` … would". The PR's own test `promote_destination_approve_queues_and_allow_restamps_3202` stops at `status == "pending"` and re-promotes after loosening the policy; it never approves+executes. Fix: queue a promote-shaped pending (the executor's `"promote"` + `TO_NAMESPACE` arm at `:19700-19712` already expects one) whose `namespace` is `to_ns`, and add a test that approves → executes → asserts the clone lands in `to_ns` attributed to `requested_by`.
2. **HIGH (ESTABLISHED) — #3202 is bypassable through the approval lane.** The pre-existing SOURCE `GovernedAction::Promote` gate (`promote.rs` Task-1.9 block) returns `Pending` BEFORE the new destination block runs; on approval `execute_pending_action`'s `"promote"` arm (`storage/mod.rs:19700-19712`) calls `promote_to_namespace` with no destination evaluation. Source `promote: Approve` + destination `write: Deny` ⇒ the approver lands the clone in the Deny namespace. The federation lane reaches the same executor (`federation_receive.rs:3023`). The executor must evaluate the destination Store-class gate (Deny ⇒ refuse the execute and mark the pending failed with reason) — or the handler must run the destination gate BEFORE the source gate so a single pending covers both. Same fix commit as (1); one test per lane (MCP, CLI, executor).
3. **MED (ESTABLISHED)** — `memory_checkpoint_resolve`: schema-REQUIRED `id` and `resolved_by` (`checkpoint.rs:480-486`) are still read via `unwrap_or_default()` (`:192-196`, `:208-211`) and `resolved_by` is never validated, so a blank resolver is persisted as attribution (`:57 resolved_by: Some(…)`) — the exact §1 row-1 shape the body says is closed for "all schema-REQUIRED strings". Use `param_guard::require_str` for both.
4. **MED (ESTABLISHED, pg-parity)** — `memory_get_links` `attest_level` fail-closed `COALESCE(attest_level,'unsigned')` landed only in `storage/mod.rs:8634`; `PostgresStore::get_links` (`postgres.rs:7019-7046`) still projects the nullable column raw, so on a pg-backed daemon the trust field still vanishes on the HTTP/SAL surface. Mirror the COALESCE (the pg `:7116` plpgsql already does it for the max).
5. **MED (ESTABLISHED)** — the new guard `tests/mcp_handler_params_subset_of_schema.rs` cannot see the reads this very PR introduced: `read_keys` matches only `params["k"]` / `params.get(k)` shapes (doc `:158-165`), so every `param_guard::require_str(params, KEY)` / `optional_*` call (23 sites) is invisible, and `src/mcp/mod.rs` (where `memory_offload`/`memory_deref` read `ctx.arguments.get(AGENT_ID)`) is not scanned. "Closes the class permanently" is therefore overclaimed. Add the `param_guard::*(params, KEY)` shape and scan `src/mcp/mod.rs` (or the dispatcher handlers that live there).
6. **LOW** — `routine.rs:337-340` template edge `type` unknown → silently `EdgeType::Sibling` (unknown-discriminant class, nested payload).
7. **LOW** — `resolve_mcp_agent_id` now silently substitutes the synthesized client id for a PRESENT reserved/invalid `agent_id` (`resolve_mcp_agent_id_rejects_reserved_sentinel_3204`) — answers a contradictory value instead of refusing it (your §1 principle / ERRORS-08). Audit actor is safe; state the choice in the CHANGELOG or refuse.
8. **LOW** — `gate_gc_sweep` probes governed namespaces and `db::gc` sweeps in separate statements, not one transaction; a row expiring in a delete-governed namespace between the two is reaped. Run probe+sweep under one IMMEDIATE tx or document the window.
9. **LOW (PLAUSIBLE)** — `quotas::peek_status` returns an existing row as-is; if its `day_started_at` bucket is stale the wire shows yesterday's counters as current (not traced through `load_row`).

**Issue closure:** #3171 — close only after (3) and (5); #3202 — do NOT close until (1)+(2) land (the issue's own test line "Approve → pending decision" must include execution). #3204 items 4/7 as described are established.

**Do not merge until (1)–(5) are on the head; (6)–(9) may follow up.** Rebase required; pg lane: the only pg assertion is `postgres_forget_refuses_widening_operator_token_3171` (needs the v90 cluster, `--include-ignored`) — record the run or mark UNVERIFIED in the merge comment.

alphaonedev:
**Fable gate — FIX-BEFORE-MERGE (narrowed to 2 residuals)** at `30952a2b` (re-gate).

Fixed and verified on this head: HIGH-1 — the destination gate now queues a TYPED `MEMORY_PROMOTE` pending with a promote payload (`promote.rs:~290`), so an approved destination-pending executes through the executor's `promote` arm instead of failing Memory-deserialisation in the `store` arm; the request path evaluates the STORE-class gate against `to_ns` before cloning (Deny refuses, Approve queues, `promote.rs:~225-300`) and the clone is attributed to the requesting principal (`storage/mod.rs:20162-20168`). MED-4 pg `attest_level` COALESCE landed (`postgres.rs:7403`).

Residuals (both were in the original gate; still open on this head):
1. **HIGH (ESTABLISHED) — the SOURCE-pending approval lane still bypasses the destination gate.** `handle_promote` evaluates the source `GovernedAction::Promote` FIRST and returns `status: pending` on `GovernanceDecision::Pending` (`promote.rs:~192-197`) — before the destination block is ever reached. When THAT pending is approved, `execute_pending_action`'s `promote` arm (`storage/mod.rs:20156-20170`) calls `promote_to_namespace` with zero destination references (grep over :20100-20200 = none). Source policy Approve + destination policy Deny ⇒ approving the source pending lands the clone in the Deny namespace. Fix: in the executor `promote` arm, evaluate the destination STORE-class gate exactly as the request path now does (Deny ⇒ the pending execution fails with the refusal recorded; Pending ⇒ chain a destination pending), or pre-evaluate the destination BEFORE queueing the source pending and encode both consents in one row. Add the source-Approve/dest-Deny execute test on both backends.
2. **MED (ESTABLISHED) — `handle_checkpoint_resolve` still accepts empty required params.** `checkpoint.rs:190-196` `id` collapses via `.unwrap_or_default()` to `""` with no `is_empty` refusal (only `state` is validated, `:207`); same shape for `resolved_by`. An empty resolver can still be persisted as attribution. Refuse empty/absent `id` and `resolved_by` with typed messages.

Everything else from the original gate is accepted. Re-request on the next head — with the 2 fixes this approves fast.

alphaonedev:
**Fable gate — APPROVED** at `720eff43` (re-gate; both residuals fixed). (1) HIGH: `refuse_unapproved_destination_store` (`storage/mod.rs:20070+`) now runs in the executor's `promote` arm before `promote_to_namespace` (`:20219-20221`) with the right semantics — Deny or still-Pending ⇒ refuse the execution, no second pending queued; short-circuit for the destination-typed pending itself; source-Approve/dest-Deny test present. (2) MED: `checkpoint resolve` refuses empty/absent `id` and `resolved_by` with typed messages (tests `:807-824`). NOTE (pre-existing, NOT this PR): the pg executor's `promote` arm (`postgres.rs:26576-26587`) ignores `to_namespace` entirely — an approved cross-namespace promote silently degrades to a tier bump on postgres; base behaves identically, so out of this PR's blast radius. I am filing it as a fable-qc pg-parity issue cross-linked to #3202. Merge when 34/34 on `720eff43`.

alphaonedev:
**Fable gate — APPROVED extended to `58ced014`.** Delta since `720eff43`: test-module-only changes in `src/` (both hunks inside `mod tests`; quota tests now seed via `quotas::record_op` instead of driving the handler — better isolation), plus the C5 token-budget ceiling lockstep 6750→7650 with a MEASURED rationale in the workflow comment (CI measured 7567 after #3171 declared the previously-hidden MCP params; 7650 = 7567+83 margin — honest ratchet, follows measurement) and matching CHANGELOG/README lines. Merge when 34/34 on `58ced014`.

alphaonedev:
**Fable gate — APPROVED extended to `ee652929`.** Delta since `58ced014`: one pg test fix inside `#[cfg(test)]` (the NULL-attest pin now seeds a fresh UUID namespace so the (title,namespace) upsert can't alias the earlier pair on the live coverage PG — correct root-cause note in the comment). Merge when 34/34 on `ee652929`.

Issue #3177, opened 2026-08-22
[data-integrity][security] pg run_gc/size_gc hard-delete WITHOUT the #1956 tombstone + crypto-erase (evicted rows resurrectable by a federated peer), and pg size_gc archives a memory without its links


alphaonedev:
Scout-verified (R-405) at `cda57210` from the ox-alpha SAL parity audit (152 methods; `.local-runs/ox-c4-parity/`). pg lane evidence required: `AI_MEMORY_TEST_POSTGRES_URL=$(cat .local-runs/ENTERPRISE-FED-TEST-PG-URL.txt) cargo test --features sal-postgres -- --include-ignored`.
**Code.** `grep -rn evict_tombstone_and_erase src/` → def `storage/mod.rs:5638`, callers `:12960` (run_gc) `:13137` (size_gc) — sqlite only. pg `postgres.rs:24490-24520` / `:24656` DELETE outright; their revision leaf is `append_only_enabled()`-gated while sqlite's erase is unconditional. pg size_gc archive branch `:24609-24637` copies only the memory row; sqlite `storage/mod.rs:4425` calls `archive_links_for_memory`; pg has that twin only in `forget` (`:21987`) and `archive_by_ids` (`:25122`).
**Impact.** A TTL/byte-cap eviction of an encrypted row on pg leaves key material unerased and writes no forget-tombstone → a peer can resurrect the evicted row via LWW on pg where sqlite's tombstone refuses; a pg size-GC archive is not restorable to its edge graph. Retention/erasure claims are backend-dependent.
**Fix.** pg twin of `evict_tombstone_and_erase` in the same tx as the DELETE in run_gc + size_gc, ungated by append_only; call the pg `archive_links_for_memory` twin in size_gc archive. Cross-check #3161 (sqlite gc(archive=true) link loss — Opus R in flight) — likely the sqlite-side twin of the same asymmetry.
**Test.** pg: TTL-evict an encrypted row → forget_tombstones row exists, envelope key erased, subsequent `apply_remote_memory` for the id is DROPPED; size_gc archive=true → archived_memory_links carries the edges. GA: BLOCKING.

alphaonedev:
Closed by fable-merge of #3240 at c3344757 G (postgres SAL parity cluster).

Issue #3160, opened 2026-08-22
[test-gap] sqlite full-table rebuilds v50 and v80 are unexercised; the v63 rebuild test asserts no triggers (would not have caught the v63→v65 defect it was written for)


alphaonedev:
Scout-verified at cda57210. `migrations/sqlite/0042_v50_per_namespace_quota.sql:92-95` (DROP TABLE agent_quotas; ALTER TABLE agent_quotas_v50_shadow RENAME TO agent_quotas) and `0064_v80_lineage_custody_revocation.sql:67-68` (DROP TABLE agent_lineage; … RENAME) have no regression test — no fixture builds the legacy shape and replays (`grep agent_quotas tests/` → only post-migration fixtures). `v63_rebuild_preserves_links_and_accepts_typed_cognition_relations` (`src/storage/migrations.rs:4286-4385`) asserts rows + exactly one of four indexes and ZERO of four triggers. The signature-trigger pair is pinned only transitively by `tests/persona_signing_pipeline_813.rs:167-176` on a fresh DB — which works solely because bootstrap SCHEMA (`migrations.rs:51+`) omits those triggers and the ladder installs them: an undocumented coupling. (v80's rebuild IS safe today — agent_lineage has no triggers/secondary indexes — verified.)

Repro: delete `CREATE TRIGGER memory_links_ck_attest_signature_ins` from 0054_v65_*.sql and run `cargo test --lib storage::migrations` — green.
Fix: per-arm rebuild tests for v50 and v80 on the `v66_rebuild_preserves_governance_rules_…` pattern; extend the v63 test and `historical_replay_from_v1_reaches_current_schema` (`:5134`) to assert all 4 indexes + all 4 memory_links triggers in sqlite_master; a generic post-ladder assertion that every rebuilt table's trigger set matches the tip expectation (memory rule: migration-rebuild-drops-all-triggers). Also doc drift: `0030_v07_atomisation.sql:56-57` says 'if version < 35' / memory_links_v35 while the arm is < 36 / _v36. GA: DEFERRABLE (tests), but cheap.

alphaonedev:
Closed by merge of #3221 at `b3ffd9599113e2d6782901e10193d6d361e46cbc` G (fable-merge.sh does not auto-close).

- #3158 claims matrix lockstep
- #3160 rebuild tests
- #3161 archive-then-delete edge-graph (lineage CID pins still #3250)

alphaonedev:
[...]

PR #3221, opened 2026-08-22
fix(storage): migration-truth — claims matrix, rebuild tests, archive-link data loss (#3158 #3160 #3161)

### LINKED ISSUE #3158: [data-integrity] migration_meta matrix is factually wrong for v41–v47 and absent for v54–v88; its lockstep gate (arch_8_ladder_terminates_at_current_schema_version) cannot fail
Scout-verified at cda57210. `src/storage/migration_meta.rs:364-370` declares `version: 43, name: "ADD_RECURSIVE_LEARNING_LEDGER", idempotent: true, reversible: true, data_loss_risk: DataLossRisk::None`. The REAL v43 arm (`src/storage/migrations.rs:2582-2593`) applies `migrations/sqlite/0037_v07_persona_signing_atomicity.sql`, whose first statement is `UPDATE memory_links SET attest_level = 'unsigned' WHERE attest_level IN ('self_signed','peer_attested') AND (signature IS NULL OR length(signature) != 64);` — an irreversible overwrite. `grep recursive_learning src/storage/migrations.rs` → no match: the row names a migration that does not exist. v41 (ADD_PERSONA_VERSION vs shadow-retention `source` col), v42 (ADD_HOOK_DLQ vs mentioned_entity_id), v44 (ADD_BATMAN_VOCABULARY_COLUMNS vs encrypted_envelope #228), v46 (ADD_SHARE_LINKS vs source_uri backfill #885), v47 (ADD_MENTIONED_ENTITY_ID vs recall_observations #886) are likewise mislabelled. The FINAL row (`:460-466`) is keyed to `crate::storage::migrations::current_schema_version()` under a doc comment describing v54 — so `meta_for(54)` → None, `meta_for(89)` → v54's semantics, and there are NO rows for v54–v88 (35 arms incl. the v63/v66/v80 rebuilds and PG v89 tsv rewrite). The gate `arch_8_ladder_terminates_at_current_schema_version` (`:493-503`) asserts last().version == current_schema_version() — true BY CONSTRUCTION, it can never fail.

Harm: every operator-facing reversibility/data-loss answer for the modern ladder is missing or false; a rollback plan built on the v43/v46 rows is wrong. v43 belongs in the irreversible list at `migrations.rs:1556-1557` (currently v34/v50/v54).
Repro: `assert_eq!(meta_for(43).unwrap().name, "PERSONA_SIGNING_ATOMICITY")` fails; `meta_for(54).is_some()` fails; `meta_for(63).is_some()` fails.
Fix: re-derive v41–v47 from the arms; v43 reversible:false; key the v54 row to literal 54 (the `postgres.rs:5102-5108` literalize-on-settle convention) and add v55–v89; make the lockstep test reject a symbolic tail key; add a `scripts/check-migration-ladder.sh` rule requiring one meta row per `if version < N` arm with a matching name (today the script never mentions migration_meta). GA: BLOCKING (claims-truth of the data-loss metadata).

alphaonedev:
Closes #3158 #3160 #3161

Three defects in the same organ — the migration ladder and the archive→restore
contract — all of the **claims-truth / data-loss** class the North Star treats as
highest-order. Each is fixed *and* mutation-proved: the gate that was supposed to
catch it is shown FAILING on the re-planted defect.

---

## #3161 — DATA LOSS: `gc(archive = true)` dropped the archived memory's entire edge graph (BLOCKING)

**Verified (R-405) at `cda57210`.** `storage::gc`'s archive branch copied only
the memory ROW into `archived_memories`, then ran `DELETE FROM memories` in the
same transaction. `memory_links` carries an `ON DELETE CASCADE` FK on **both**
endpoints, so that delete reaped every edge, and `restore_archived` returned the
memory with an empty graph — while `forget(archive = true)`,
`archive_memory_no_tx` and (on postgres) `archive_by_ids` have snapshotted into
`archived_memory_links` since #1771.

The loss was pinned as *intended* by a test whose message read:

> `"#1771 DOCUMENTED LOSS: a restored memory's memory_links are NOT recovered (cascade-reaped at delete time, not archived). Flip this to expect 1 when archived_memory_links preservation lands."`

Preservation **had** landed and #1771 was **CLOSED**. So a documented known-loss
had silently become an *undocumented asymmetry*: identical memories, identical
edges, different outcome depending on which funnel — or which **principal** —
archived them. That is exactly the failure mode the prime directive forbids.

### The audit found FIVE affected funnels, not one

I enumerated every archive-then-delete site on both backends (8 sqlite + 7
postgres `archived_memories` write sites) rather than trusting the issue's list:

| Backend | Funnel | Pre-fix | Now |
|---|---|---|---|
| sqlite | `gc` (TTL eviction) | no snapshot | snapshots over the same deterministic `SQL_GC_EXPIRED_CHUNK_IDS` set the `DELETE` targets |
| sqlite | `archive_memory_for_caller` (**ownership-scoped `memory_archive`**) | no snapshot | snapshots — *not in the issue*; it open-codes its archive copy instead of routing through `archive_memory_no_tx`, so whether a memory kept its links depended on which **principal** archived it |
| postgres | `run_gc` | no snapshot | set-based insert over the same TTL-expiry predicate the `DELETE` uses |
| postgres | `size_gc` (byte-cap eviction) | no snapshot | per-victim snapshot — the sqlite twin routes through `archive_memory_no_tx` and has snapshotted since #1771; this one never did |
| postgres | `update_with_archive_on_supersede` | no snapshot | snapshots — *not in the issue*; the superseded row was archived **and deleted**, so restoring a `superseded` snapshot on postgres returned it edgeless while sqlite returned the graph |

The remaining sites (`archive_memory_insert_only`, pg `update` /
`update_with_expected_version_once`) are `in_place_edit` snapshots that never
delete the live row — no cascade fires, so no snapshot is owed. Every snapshot is
idempotent (`INSERT OR IGNORE` / `ON CONFLICT DO NOTHING` on the
`(source_id, target_id, relation)` PK), so an edge already captured by another
funnel is untouched. All run **before** the delete, **inside** the same tx.

### Claims-truth sweep (the retired v70 carve-out)

Five places still told operators the opposite of what the code does now — fixed:
`storage::restore_archived`'s recovery-scope doc, `docs/corpus-lifecycle.md`'s
EXPIRE restorability row, `CLAUDE.md`'s v70 schema note, and two expiry-floor
rationale comments in `storage/mod.rs`. Each now also names what **is** still
irreversible: a HARD (`archive = false`) forget/gc, which is what
irreversibility means. Separately, the postgres `archived_memory_links` v70 doc
still read *"the snapshot-before-delete + restore-re-insert wiring is wired on
SQLite this commit and is a tracked POSTGRES follow-up"* — the identical sentence
was corrected in the sqlite twin by #2318 but the postgres copy was **missed**,
so it advertised a two-release-old gap that did not exist.

### Mutation proof

Neutering the `gc` snapshot loop and re-running `cargo test --lib 3161`:

```
test gc_archive_then_restore_preserves_links_3161 ... FAILED
test gc_and_forget_archive_paths_agree_on_link_preservation_3161 ... FAILED
assertion failed: #3161 FUNNEL PARITY: restored edge counts were gc=0, forget=1,
archive_memory=1, archive_memory_for_caller=1. All four archive->restore funnels
must PRESERVE the edge graph.
test result: FAILED. 0 passed; 2 failed
```

---

## #3158 — the per-migration metadata matrix was guessed, and its lockstep gate could never fail (BLOCKING)

`migration_meta::MIGRATION_LADDER` is the operator-facing answer to *"is this
migration reversible? does it lose data?"* — the input to every rollback plan.

**Verified (R-405).** The headline row declared v43 as
`ADD_RECURSIVE_LEARNING_LEDGER / idempotent / reversible / DataLossRisk::None`.
`grep -c recursive_learning src/storage/migrations.rs` → **0**: the row named a
migration that does not exist. The real v43 arm runs `MIGRATION_V43_SQLITE`
(`migrations/sqlite/0037_v07_persona_signing_atomicity.sql`), whose first
statement is:

```sql
UPDATE memory_links
SET attest_level = 'unsigned'
WHERE attest_level IN ('self_signed', 'peer_attested')
  AND (signature IS NULL OR length(signature) != 64);
```

— an irreversible overwrite of caller-supplied non-`NULL` values. Wrong on all
three axes. v41/v42/v44/v46/v47 and most of v3–v40 were mislabelled the same way.

The tail row was keyed to `current_schema_version()`, so `meta_for(54)` → `None`,
`meta_for(89)` → v54's semantics, **no rows at all existed for v54–v88** (35 arms,
including three full-table rebuilds and the postgres tsv rewrite), and the gate
`arch_8_ladder_terminates_at_current_schema_version` — asserting
`last().version == current_schema_version()` — was **true by construction**.

### Fix

All 88 rows (v2–v89) re-derived from each arm's *actual SQL* by one stated,
uniformly-applied rule (documented on `MIGRATION_LADDER`), keyed to **literal**
versions, and carrying a new `LadderArm` field distinguishing the four
postgres-only steps (v57/v69/v88/v89) from the sqlite arms. Two arms are now
correctly `DataLossRisk::Column` (v43, v50) and the irreversible set is **21**
arms — not the three (`v34/v50/v54`) the old comment in `migrate()` named, which
is itself corrected to point at the matrix rather than re-inline a stale list.

The gate is rebuilt to be **failable**: it refuses a symbolic tail key (reading
its own source), requires exactly one row per `if version < N` arm (const-phrased
arms resolved from their mandatory `// vNN` comment, and *failing loudly* rather
than skipping an arm that lacks one), requires gap-free v2..=CURRENT coverage,
and requires each row's NAME be **grounded in the arm it describes** — the check
that catches the v43 class. `scripts/check-migration-ladder.sh` gains the same
coverage as **rule (g)**, with two self-test cases.

### Mutation proofs

Re-planting each shipped defect:

**(A1) tail row re-keyed to `current_schema_version()`** →
```
test arch_8_ladder_tail_is_a_literal_at_current_schema_version ... FAILED
ARCH-8 (#3158): a MIGRATION_LADDER row is keyed to `current_schema_version()`.
A symbolic tail key makes the equality above a TAUTOLOGY (the row follows the
const), silently re-labels the new tip with the OLD arm's semantics, and leaves
`meta_for(<the literal version>)` returning None.
test result: FAILED. 7 passed; 1 failed
```

**(A2) the original wrong v43 name restored** →
```
test arch_8_meta_row_names_are_grounded_in_their_arm ... FAILED
test arch_8_meta_for_round_trip ... FAILED
  left: "ADD_RECURSIVE_LEARNING_LEDGER"
ARCH-8 (#3158): MIGRATION_LADDER rows whose NAME appears nowhere in the arm they
describe (nor in the migration .sql it sources): ["v43 = ADD_RECURSIVE_LEARNING_LEDGER"].
test result: FAILED. 6 passed; 2 failed
```

**(A3) v43 row deleted (the v54–v88 hole shape)** →
```
test arch_8_destructive_arms_are_flagged_irreversible ... FAILED
test arch_8_meta_for_round_trip ... FAILED
test arch_8_ladder_is_gap_free_from_v2_to_current ... FAILED
test arch_8_every_sqlite_ladder_arm_has_exactly_one_meta_row ... FAILED
ARCH-8 (#3158): MIGRATION_LADDER has holes at [43] — every schema version from 2
to CURRENT_SCHEMA_VERSION (89) is a real ladder step on at least one adapter.
ARCH-8 (#3158): sqlite ladder arms with NO MIGRATION_LADDER row: [43].
test result: FAILED. 4 passed; 4 failed
```

**Shell lane** — `scripts/check-migration-ladder.sh --self-test`:
```
  [g1] #3158 ladder arm with no metadata row: CAUGHT
  [g2] #3158 tail row keyed to current_schema_version(): CAUGHT
```

---

## #3160 — the v63 rebuild test asserted NO triggers; v50/v80 rebuilds were unexercised

A SQLite full-table rebuild (`CREATE <shadow>` / `INSERT … SELECT` / `DROP` /
`RENAME`) silently drops **every** trigger and secondary index on the table —
the v63 → v65 lesson. The ladder carries six rebuilds. v50 (`agent_quotas` PK
widening) and v80 (`agent_lineage` custody/revocation) had **no** regression test
at all, and `v63_rebuild_preserves_links_and_accepts_typed_cognition_relations`
asserted **one of four** indexes and **zero of four** triggers — so it would not
have caught the very defect it was written for.

### Fix

- Per-arm **v50** and **v80** tests that seed the legacy shape *with rows*, stamp
  N-1, migrate, and assert every row (and every BLOB — lineage `signature` /
  `prev_record_hash` / `record_bytes`) survives verbatim, plus the recreated PK,
  indexes and trigger set read from `sqlite_master`.
- The full four-trigger / four-index tip-shape assertion on the v63 test, on a new
  v62→tip replay test that also proves the restored trigger **refuses a
  phantom-signed edge**, and on `historical_replay_from_v1_reaches_current_schema`.
- A generic `rebuilt_tables_reach_the_same_trigger_and_index_set_on_both_paths_3160`
  gate: for every rebuilt table, the UPGRADE path and the FRESH-INSTALL path must
  converge on the identical trigger + index set — the `bootstrap(fresh) ≡
  ladder(v0→tip)` invariant applied to triggers, which is exactly the shape of the
  v63 → v65 defect. It additionally asserts the memory_links set is **complete**
  on both paths, since two empty sets would satisfy mere equality.
- Doc drift: `migrations/sqlite/0030_v07_atomisation.sql`'s idempotency header
  said the rebuild is gated by `if version < 35` and builds `memory_links_v35`;
  the arm has been `< 36` / `memory_links_v36` since the file landed.

### Mutation proof

Deleting `CREATE TRIGGER memory_links_ck_attest_signature_ins` from
`0054_v65_restore_memory_links_signature_triggers.sql` — the issue's exact repro,
which was **green** before this PR:

```
test historical_replay_from_v1_reaches_current_schema ... FAILED
test v63_rebuild_leaves_memory_links_with_its_full_trigger_and_index_set_3160 ... FAILED
test v63_rebuild_preserves_links_and_accepts_typed_cognition_relations ... FAILED
test rebuilt_tables_reach_the_same_trigger_and_index_set_on_both_paths_3160 ... FAILED
test result: FAILED. 37 passed; 4 failed
```

---

## Tests

**#3158** (`cargo test --lib storage::migration_meta`, 8/8):
`arch_8_ladder_tail_is_a_literal_at_current_schema_version` (renamed from
`arch_8_ladder_terminates_at_current_schema_version`),
`arch_8_every_sqlite_ladder_arm_has_exactly_one_meta_row`,
`arch_8_ladder_is_gap_free_from_v2_to_current`,
`arch_8_meta_row_names_are_grounded_in_their_arm`, `arch_8_meta_for_round_trip`,
`arch_8_destructive_arms_are_flagged_irreversible`,
`arch_8_every_meta_row_has_a_non_empty_name`,
`arch_8_ladder_versions_strictly_monotonic`.

**#3160** (`cargo test --lib storage::migrations::`, 41/41):
`v50_rebuild_preserves_agent_quotas_and_recreates_indexes_3160`,
`v80_rebuild_preserves_agent_lineage_and_widens_reason_check_3160`,
`v63_rebuild_leaves_memory_links_with_its_full_trigger_and_index_set_3160`,
`rebuilt_tables_reach_the_same_trigger_and_index_set_on_both_paths_3160`, plus the
extended `v63_rebuild_preserves_links_and_accepts_typed_cognition_relations` and
`historical_replay_from_v1_reaches_current_schema`.

**#3161** (`cargo test --lib 3161` + `storage::tests::gc_`, 5/5):
`gc_archive_then_restore_preserves_links_3161` (the flipped #1771 pin),
`gc_and_forget_archive_paths_agree_on_link_preservation_3161` (four-funnel parity:
gc / forget / `archive_memory` / `archive_memory_for_caller`, pinning the exact
value `1` on every leg so an all-zero row cannot satisfy a bare equality check).
Postgres twins `run_gc_archive_then_restore_preserves_links_pg_3161` and
`size_gc_archive_then_restore_preserves_links_pg_3161` — pg-lane evidence below.

**Gates**: `scripts/check-migration-ladder.sh` + `--self-test` (10/10 cases
CAUGHT incl. the two new rule-(g) cases), `scripts/check-doc-symbol-anchors.sh`,
`scripts/check-docs-vs-ssot.sh`, `scripts/check-hardcoded-literals.sh`,
`qual_10_module_size_ceiling`, `qual_6_7_legacy_error_type_ceiling`.

## Coordination

#3177/BB2 owns the pg `run_gc`/`size_gc` **tombstone + crypto-erase** half. This
PR lands the **link-snapshot** half of that issue's `size_gc` bullet (plus
`run_gc` and the supersede funnel) because leaving a known cross-backend data
loss in place to preserve a hunk boundary trades data integrity for convenience.
The hunks are adjacent but disjoint from an `evict_tombstone_and_erase` twin;
BB2 rebases onto this.

## Rebase

Rebased onto `origin/release/v1.0.0` = `faac3e3a` (#3112 merged). CHANGELOG
kept both sides (`#3158/#3160/#3161` entries + the `#3110` security section).
`postgres.rs` QUAL-10 ceiling re-measured after the merge at **34_938**
(ceiling **35_020**, +82 headroom; `#3110` is now the PRIOR). v90 is **not**
claimed — `CURRENT_SCHEMA_VERSION = 89`, `MIGRATION_LADDER` tail is the
literal `meta(89, "REBUILD_TSV_INCLUDE_TAGS", …)`.

## CI parity run locally (PATH-pinned 1.96.0)

All legs under `( umask 022; … )` via one `build-gate.sh` admission,
`RUSTUP_TOOLCHAIN=1.96.0`:

| Leg | Result |
|---|---|
| `cargo fmt --all --check` | GREEN |
| `cargo check --all-targets` | GREEN |
| `cargo check --all-targets --features sal` | GREEN |
| `cargo check --all-targets --features sal-postgres` | GREEN |
| clippy 1/5 default `-D warnings -D clippy::all -D clippy::pedantic` | GREEN |
| clippy 2/5 `--all-targets --features sal` same `-D` set | GREEN |
| clippy 3/5 `--features sal-postgres --tests` same `-D` set | GREEN |
| clippy 4/5 `--features sal` same `-D` set | GREEN |
| clippy 5/5 `--all-targets --features vectorlite` same `-D` set | GREEN |
| `cargo check --examples` | GREEN |
| `cargo test --lib 3161` | GREEN (`gc_archive_then_restore_preserves_links_3161`, `gc_and_forget_archive_paths_agree_on_link_preservation_3161`) |
| `cargo test --lib 3160` | GREEN (v50/v63/v80 rebuild + upgrade-vs-fresh) |
| `cargo test --lib arch_8` | GREEN (8/8, including the failable tail/gap/name-grounding gates) |
| `qual_10_module_size_ceiling` + `qual_6_7_legacy_error_type_ceiling` | GREEN |
| `scripts/check-migration-ladder.sh` + `--self-test` | GREEN (10/10 planted shapes CAUGHT, including g1/g2) |
| `scripts/check-hardcoded-literals.sh` | GREEN |
| `scripts/check-docs-vs-ssot.sh` | GREEN |
| `scripts/check-doc-symbol-anchors.sh` | GREEN |

### Postgres lane (required — tests RAN, did not skip)

The shared `ai_memory_test` cluster is already on **schema v90** (#2385 /
PR #3216). This binary understands up to **v89** and correctly refused to
write it (`SchemaAheadOfBinary`) — that is the `#2445` fail-closed guard
doing its job, and v90 is reserved (not stolen). Evidence was obtained on
a **dedicated** `ai_memory_test_3161` database created on the same cert-tier
host, migrated 0→89 by this binary, then dropped:

```
test store::postgres::tests::run_gc_archive_then_restore_preserves_links_pg_3161 ... ok
test store::postgres::tests::size_gc_archive_then_restore_preserves_links_pg_3161 ... ok
test store::postgres::tests::archive_restore_preserves_links_pg_1771 ... ok
test result: ok.  (no skip: line)
```

`--test-threads=1` is required on a greenfield migrate: two parallel
`PostgresStore::connect` calls deadlock v88 `CREATE INDEX CONCURRENTLY`
against the advisory lock (observed, killed, rerun serial). Production
opens are single-process so this is a test-isolation constraint, not a
ladder defect.


alphaonedev:
**Fable gate — FIX-BEFORE-MERGE** at head `83533bf4` (shadow-QC lane, 2026-08-23). One MED claims-truth item plus the rebase; everything else in the PR is ESTABLISHED.

**Established (read every `src/` hunk on `83533bf4`):**
- **#3161** — `storage/mod.rs:12928-12939` gc snapshot uses the same `SQL_GC_EXPIRED_CHUNK_IDS` (`ORDER BY rowid LIMIT ?2`, numbered `?1/?2` so the double use is bound once) as the row-archive and the DELETE, inside the same `SQL_BEGIN_IMMEDIATE` tx; `archive_memory_for_caller` `:4872` now calls `archive_links_for_memory`. Postgres `run_gc` snapshot/archive/DELETE all share the one bound `$1` timestamp in one tx; `size_gc` + `update_with_archive_on_supersede` per-victim snapshots with `ON CONFLICT DO NOTHING`. Enumerated the 8 sqlite `archived_memories` writers myself: `archive_memory_no_tx`, `archive_memory_for_caller`, `forget` ×2 (pattern / no-pattern), `forget_for_caller` ×2, `gc` all snapshot; `archive_memory_insert_only` is in-place (no DELETE) — matches the PR table. Funnel-parity test pins `[1,1,1,1]` exactly.
- **#3158** — 88 rows v2–v89, literal keys, 21 `reversible:false`, 2 non-idempotent (v33 `0027` and v66 `0055` have no already-rebuilt probe — confirmed from the SQL; v36/v50/v63/v80 are probe-guarded — confirmed from the arms), 2 `Column` (v43 `0037` UPDATE overwrite; v50 PK widening), 4 `PostgresOnly` = exactly the versions with no `if version < N` arm (57/69/88/89). `sqlite_arms()` parses the real arms, fails loudly on a const-phrased arm with no `// vNN` comment, and the tail-key test reads its own source. Shell rule (g) fails closed on zero parsed rows; g1/g2 self-tests are real plants.
- **#3160** — v50/v80 tests seed the legacy DDL *with rows* (BLOBs verbatim), stamp N-1, walk the ladder, assert rows/PK/indexes/triggers from `sqlite_master`; `assert_memory_links_tip_shape` names all 4 triggers + all rebuild indexes. `fresh_db_via_migrate` (SCHEMA + stamp 0 + `migrate`) mirrors production exactly — `migrate()` walks every arm on a brand-new DB (`migrations.rs:1526-1555`, only the snapshot is gated on `version > 0`). `0030` header fix matches the arm (`< 36` / `memory_links_v36`).
- R-405 on the body: 8/7 write sites, 21 irreversible, 88 rows, `migrate()` comment now points at the matrix, pg v70 "tracked POSTGRES follow-up" sentence corrected, `corpus-lifecycle.md` EXPIRE row, CLAUDE.md note — all hold.

**Findings:**
1. **MED — claims-truth (ESTABLISHED).** `storage/mod.rs:13384-13399` now states "EVERY archive-then-delete funnel preserves [the edge graph]" and enumerates what is "Still NOT recovered (by design)". That list omits a real loss that applies to *every* funnel including the three this PR adds: `memory_links.source_cid` / `target_cid` (v75 lineage pins, `0059_v75_memory_links_lineage_cid.sql:23-24`) are not columns of `archived_memory_links` (`migrations.rs:323-335`; pg `postgres_schema.sql:568-571`), so neither `archive_links_for_memory` (`:4340-4345`) nor the new gc/pg snapshots carry them, and `restore_links_for_memory` (`:4365-4378`) re-inserts the edge with both CIDs `NULL`. The edge comes back; its content-pinned lineage does not. (PLAUSIBLE, not traced: if a self-signed edge's signature input includes the CIDs, the restored edge also no longer verifies.) Pre-existing since v75 and out of #3161's literal scope, so the in-PR fix is one line: add the CID pins to the "Still NOT recovered" list in `restore_archived`'s doc (+ the same qualifier in `docs/corpus-lifecycle.md`'s EXPIRE row and the CHANGELOG bullet). The real fix (carry the two columns through snapshot + restore on both backends; needs a schema bump, v90 is taken by #3216) is being filed as a `fable-qc` issue — reference it from the doc line.
2. **Rebase required.** `git merge-tree` vs `ae3e0aec`: textual conflict in `tests/qual_10_module_size_ceiling.rs` — the tip already raised `postgres.rs` to 35_200 (measured 35_166 after #3134), so this PR's +~300 lines need a fresh measurement and a new ceiling, not the 35_020 in the branch. `src/storage/mod.rs` and `migrations.rs` ceilings likewise re-measure on the rebased tree. Re-run the pg lane (`--test-threads=1`, dedicated db as you did) on the rebased head.

**Follow-ups (non-blocking):**
- `migrations.rs:1549-1556` says "do NOT re-inline a short list here — read the matrix" and then inlines the full 21-arm list two lines above; that list will drift on the next irreversible arm. Either drop it or pin it with a test that derives it from `MIGRATION_LADDER`.
- The three flags (`idempotent` / `reversible` / `data_loss_risk`) stay human-classified; only the NAME is machine-grounded. A cheap one-directional pin would have caught the v43 class directly: for every `LadderArm::Sqlite` row whose arm haystack (incl. sourced `.sql`) contains `UPDATE ` / `DELETE FROM` / `DROP TABLE` / `RENAME TO`, assert `reversible == false`. Suggested as a `arch_8_*` addition.
- `rebuilt_tables_reach_the_same_trigger_and_index_set_on_both_paths_3160` is equality-only for `agent_quotas` / `governance_rules` / `agent_lineage` (none carry triggers today, so it is currently sound); a trigger added by a pre-rebuild *ladder* arm and dropped by the rebuild would be lost on both paths and pass. Per-table expected sets (as done for `memory_links`) close that.
- `#3177` (pg tombstone/crypto-erase half of `run_gc`/`size_gc`) must rebase onto this; the hunks are adjacent — flag it on #3177.

Do not merge until item 1's doc qualifier is on the head and the rebase is pushed; then merge when 34/34 on that head.

alphaonedev:
**Fable re-gate on `e66b8d59` — one item still outstanding, everything else verified.** The rebase + #3158 matrix re-derivation + #3160 rebuild tests + #3161 five-funnel edge-graph close (both backends) and the qual-10 lockstep bumps all check out on this head. But the single MED claims-truth item from the gate — the `restore_archived*` doc (`storage/mod.rs:~13820-13850` region) stating that `archived_memory_links` carries NO `source_cid`/`target_cid`, so restored edges re-insert with NULL lineage pins (issue #3250) — is not in the delta: no hunk touches that doc block and no #3250 reference exists on the head. Either (a) add the one-sentence qualifier + the corpus-lifecycle/CHANGELOG echo, or (b) if #3161's new funnel work now CARRIES the cids through archive/restore (I don't see the columns added, so I believe it does not), point me at the lines and I'll re-verify. With (a) this approves immediately on the next push.

alphaonedev:
**Fable hold — still ACTIVE at `74a01da3`** (deltas since `e66b8d59` are CI-flake fixes only; the one blocking item remains). To make this zero-cost, here is the exact text — add to the `restore_archived` rustdoc (the block above `storage/mod.rs:~13850`) and mirror one line in `docs/corpus-lifecycle.md` + a CHANGELOG fragment:

```
/// NOT recovered (#3250): `archived_memory_links` carries no
/// `source_cid`/`target_cid` columns, so every restored edge re-inserts
/// with NULL v75 lineage pins — cid-stable identity across a tombstoned
/// endpoint is lost until the post-v90 schema arm tracked in #3250 adds
/// the columns and the snapshot/restore carry.
```

Push that and this gate flips to APPROVED on sight — everything else on the head is already verified.

alphaonedev:
Fable FIX-BEFORE-MERGE item 1 (MED claims-truth, #3250 NULL lineage pins) is on head `492a2ebf`:

- `src/storage/mod.rs` `restore_archived` recovery-scope rustdoc — **NOT recovered (#3250)** qualifier on `archived_memory_links` missing `source_cid`/`target_cid`
- `docs/corpus-lifecycle.md` EXPIRE restorability row
- `CHANGELOG.md` #3161 claims-truth bullet

Schema carry remains #3250 (post-v90). 34-watcher re-armed on `492a2ebf`. Merge when 34/34.

alphaonedev:
**Fable gate — APPROVED** at `492a2ebf`. The #3250 qualifier landed in all three places: the `restore_archived` rustdoc (verbatim NOT-recovered block), `docs/corpus-lifecycle.md` EXPIRE row (with the accurate #3161 context — gc-archive now snapshots edges, restored edges carry NULL lineage pins until #3250), and the CHANGELOG fragment. All prior items were already verified across the re-gate sequence (`e66b8d59`→`b537e274`→`74a01da3` deltas were CI-flake-only). Merge when 34/34 on `492a2ebf`.

PR #3215, opened 2026-08-22
fix(boot): fail CLOSED on an unusable config + honest AI_MEMORY_NO_CONFIG grammar — #3166 #3167 #2999 #3001 #3002 #3003

alphaonedev:
Closes #3166
Closes #3167
Closes #2999
Closes #3001
Closes #3002
Closes #3003

Six config / ops-honesty defects that share one failure shape: a surface reports success while silently resolving to something other than what the operator configured. **#3166 is the data-integrity one and the reason the others ride with it.**

## Per-issue

| # | Defect | Fix | Regression pin |
|---|---|---|---|
| **3166** | Boot fail-OPEN on a malformed config → daemon opens a NEW empty `./ai-memory.db`, reports healthy, writes into an orphan (corpus split-brain); `append_only` / `require_operator_pubkey` / `[[permissions.rules]]` revert | `main` uses the propagating `AppConfig::load_for_boot()`; refuses boot with the error on stderr and **exit 78 (`EX_CONFIG`)** before any DB is opened. `ErrorKind::NotFound` matched explicitly so the documented "no config → defaults" contract and `AI_MEMORY_NO_CONFIG=1` are byte-identical | `typo_in_config_refuses_boot_and_creates_no_orphan_db_3166`, `unreadable_config_refuses_boot_with_the_io_error_3166`, `missing_config_still_boots_on_compiled_defaults_3166`, `valid_config_boots_and_opens_the_configured_db_3166`, `version_still_works_on_a_broken_config_3166`, `doctor_reports_the_broken_config_instead_of_refusing_3166`, `main_uses_the_propagating_loader_after_argv_parse_3166` |
| **3167** | `AI_MEMORY_NO_CONFIG=` (empty) and `=0` disabled the ENTIRE config file (presence check vs the documented "set to `1`"); a non-UTF-8 value had the INVERSE effect | one shared `config::skip_config()` on the substrate-wide truthy grammar (`1`/`true`/`yes`/`on`, trimmed, case-insensitive) at all three sites + one-shot WARN when present-but-not-truthy. `=1` unchanged | `empty_no_config_still_loads_the_config_3167`, `zero_no_config_still_loads_the_config_3167`, `one_no_config_skips_the_config_entirely_3167`, `truthy_tokens_skip_the_config_3167`, `all_sites_share_one_skip_config_helper_3167` |
[...]

PR #3223, opened 2026-08-22
fix(integrity,security): #3085 unattributed embedding_space; #3031 fail-closed matchers; #2672 typed push-DLQ; #3027 pg inbox unread; #2908 boot banner; #2972 one embedder-model resolver

### LINKED ISSUE #2672: [federation][data-integrity] a peer can defeat the push-DLQ quarantine ceiling by returning a count containing 429 — peer-controlled integers steer a substring classifier
Found while reviewing #2442 (PR #2669), which correctly declined to add a new arm to `classify_quarantine_cause` on the grounds that the matcher is steerable. Verifying that concern surfaced a live instance that is **not** hypothetical. Filing separately; **not fixed** in #2669.

## The chain

`success_report_non_ack_reason` (`src/federation/sync.rs`) builds the failure reason from counts read **verbatim out of the peer's own JSON response body**:

```rust
format!("peer 2xx but {skipped} item(s) skipped (refused/not applied by receiver)")
```

`skipped` is a `u64` the peer chooses. That string becomes `AckOutcome::Fail(reason)`, and the DLQ landing pass persists it as `federation_push_dlq.last_error`.

`reset_throttled_quarantine` then matches that column with a bare substring test — on **both** backends:

- sqlite: `... AND last_error LIKE '%429%'` (`src/federation/push_dlq.rs:836`)
- postgres: `... AND last_error LIKE '%429%'` (`src/federation/push_dlq.rs:1004`)

and its whole job is `attempt_count = 0`.

## The consequence

A peer that answers **HTTP 200** with `{"skipped": 429}` yields:

```
peer 2xx but 429 item(s) skipped (refused/not applied by receiver)
```

which matches `LIKE '%429%'`. Its rows have their attempt budget reset on every sweep, so they **can never reach `MAX_REPLAY_ATTEMPTS`** and never quarantine. That is exactly the unbounded no-op POST amplification #1544 introduced quarantine to stop — the trait doc for `reset_throttled_quarantine` says the reset is "scoped to throttles so genuinely-systematic failures stay quarantined", and this defeats that scoping.

Any count containing the digits works (`429`, `1429`, `4290`), and `unsupported_on_postgres` reaches the same string builder. The peer never has to send a real 429.

Secondary effect: `classify_quarantine_cause`'s **first** arm is `last_error.contains("429") => "quota"`, so the row is also mislabeled, and the operator is told to "raise `AI_MEMORY_MAX_MEMORIES_PER_DAY` on the peer or wait for the daily quota window to reset" — actively misdirecting remediation away from a systematic refusal.

## Why this is a design smell, not just a bug

`last_error` is a **human-readable diagnostic string** being used as a **machine-readable control signal**. Two independent behaviours key off substrings of it: the quarantine reset (changes retry behaviour) and the cause classifier (changes operator guidance). As soon as any peer-influenced value is interpolated into it, both become steerable. #2442's review caught the general shape; this is the concrete instance.

## Suggested direction

Carry the throttle/cause signal **structurally** rather than by grepping prose — e.g. a typed cause discriminant persisted in its own column, set from the `AckOutcome` variant (`Throttled` is already classified status-precisely from the real HTTP status, before any body parse) rather than re-derived from formatted text. Failing that, at minimum stop interpolating peer-supplied counts into the string that `LIKE '%429%'` reads.

Fix must land on **both** backends — the predicate is duplicated, and the #2488/#2491 precedent is the same intent failing in inverse directions on the two adapters.

Refs: #1544, #2341, #2442 / PR #2669, #2498 / PR #2662.

### LINKED ISSUE #2908: [cert][observability] boot posture banner (asi-hard #1961 + §5.3 enterprise-federation) is SILENTLY DROPPED on a stock serve console
## Defect (surfaced by the #2905 §5.3 Fable review, proven live)

On a **default config**, `[logging].enabled` is off (`src/logging.rs:100` returns `None`, no subscriber installed in `main`), and the serve **console** subscriber only initializes inside serve bootstrap (`src/daemon_runtime.rs:6597`) — **AFTER** `run()`'s common boot-report block. So on a stock `ai-memory serve` console, BOTH the pre-existing asi-hard #1961 pin report (`daemon_runtime.rs:1031`) AND the new §5.3 `security.posture.enterprise_federation` banner (#2905) are silently dropped. Proven live: 0 banner lines with `RUST_LOG=info` and no config.

## Why it's cert-load-bearing

The 2026-08-01 cutline ruling §5.3 mandates "a boot banner echoing the effective posture" and its cited precedent is literally "verify the banner, never infer from env." If the banner is silently dropped on a stock console, the cert cannot cite the banner as evidence. This is NOT a #2905 regression — #2905 placed the banner exactly per the established asi-hard pattern; the PATTERN has this latent visibility gap. #2905 (the banner impl) is Fable-approved and merges independently.

## Proposed fix (concrete)

Initialize the console tracing subscriber BEFORE the `run()` boot-report block (or emit both the asi-hard report and the §5.3 posture banner AFTER `init_tracing` regardless of `[logging].enabled`), so the security posture is visible on a stock console. Add a test/probe asserting a stock `serve` boot (no config, RUST_LOG=info) emits the posture banner lines. Note the sink requirement in `docs/deploy/enterprise-federation.env`. File paths: `src/logging.rs:100`, `src/daemon_runtime.rs:1031,6597`.

Blocks: the enterprise-federation certification citing the boot banner as §5.3 evidence (must land before the re-cert vote).

### LINKED ISSUE #2972: reembed ignores [embeddings] configuration and falls back to all-MiniLM-L6-v2 in v0.10.0
# Bug: v2 config with `[embeddings]` Ollama backend is ignored by `reembed`

## Environment

- OS: macOS 15.7.7
- ai-memory version: `0.10.0`

Binary:

```text
~/.local/bin/ai-memory
```

## Configuration

`~/.config/ai-memory/config.toml`

```toml
schema_version = 2

db = "~/.local/share/ai-memory/memory.db"

[embeddings]
backend = "ollama"
model = "qwen3-embedding:4b"
```

## Expected behavior

`ai-memory` should resolve the embedding model from the v2 `[embeddings]` configuration section.

Running:

```bash
ai-memory reembed --dry-run --json
```

should use:

```text
qwen3-embedding:4b
```

as the target embedding model.

## Actual behavior

`ai-memory doctor --json` correctly detects the configured embedding model:

```text
[WARN] Embeddings Reachability (#1598)

backend                          ollama
model                            qwen3-embedding:4b
base_url                         http://localhost:11434
config_source                    config
key_source                       none
probe_url                        http://localhost:11434/api/tags
http_status                      200
```

However, `reembed` resolves a different model:

Command:

```bash
ai-memory reembed --dry-run --json
```

Output:

```json
{
  "backend":"ollama",
  "rows_missing_embeddings":0,
  "rows_with_embeddings":1,
  "target_dim":384,
  "target_model":"all-MiniLM-L6-v2 (384-dim, local)",
  "total_rows":1
}
```

The target model is incorrectly resolved as:

```text
all-MiniLM-L6-v2 (384-dim, local)
```

instead of:

```text
qwen3-embedding:4b
```

## Additional logs

Running:

```bash
ai-memory reembed
```

shows:

```text
reembed: PRE-FLIGHT — stored embedding dims: [384]; target: 384-dim (all-MiniLM-L6-v2 (384-dim, local))
```

The operation succeeds, but the configured embedding model is not used.

## Database status

Database is working correctly:

```text
source: ~/.local/share/ai-memory/memory.db

total_memories 1
dim_violations 0
hnsw_size_estimate 1
```

## Steps to reproduce

1. Install ai-memory 0.10.0.

2. Create:

```toml
schema_version = 2

db = "~/.local/share/ai-memory/memory.db"

[embeddings]
backend = "ollama"
model = "qwen3-embedding:4b"
```

3. Verify:

```bash
ai-memory doctor --json
```

The embedding section detects:

```text
model qwen3-embedding:4b
```

4. Run:

```bash
ai-memory reembed --dry-run --json
```

The target model becomes:

```text
all-MiniLM-L6-v2 (384-dim, local)
```

## Possible cause

`doctor` appears to use:

```text
AppConfig::resolve_embeddings()
```

correctly.

However, `reembed` appears to use a different embedding resolution path or falls back to the compiled default.

The implementation may need to ensure that `reembed` uses the same resolver path as:

```text
daemon/MCP boot

AppConfig::resolve_embeddings()

Embedder::from_resolved()
```

## Expected fix

Make `ai-memory reembed` resolve embeddings through the canonical v2 configuration resolver.

The behavior should be consistent between:

```bash
ai-memory doctor
```

```bash
ai-memory reembed
```

and:

```text
daemon/MCP startup
```

Thank you.
```

### LINKED ISSUE #3027: memory_inbox unread marker diverges across backends — the postgres arm filters on metadata.read (never written), so every message is permanently 'unread' on pg
A2-F3 (source-read, pg not in scope). sqlite/MCP uses `access_count==0` (notify.rs:163, works). postgres HTTP (hook_subscribers.rs:114-116,:150-155) filters `metadata['read']==true` — a key NO production path writes (grep: only read sites). So on pg `unread_only=true` filters nothing and unread_count==total forever. In-code comment claims sqlite parity — false. FIX: pg arm reads access_count==0 like its twin, ~5 lines. Severity MED (cross-backend divergence). Verified pm-v3.3 (fresh subprocess) + codegraph-cross-correlated. Hive-NHI feature campaign 2026-08-16.

### LINKED ISSUE #3031: rules add accepts a kind-wrong matcher, signs it, enables it — an operator_signed refuse-severity rule that silently enforces NOTHING, with no write-time warning and nothing in rules list to flag it
A4-F4. `rules add --kind bash --matcher '{"totally_bogus_key":123}' --sign` → severity:refuse, enabled:true, operator_signed; `rules check --kind bash --payload {command:anything}` → allow. Malformed JSON IS rejected; per-kind key validation is not. `matcher_applies` (agent_action.rs:378) returns false for unrecognised shapes — deliberate for read_action, silently inert for bash/fs/network/process. The doc comment even claims write-time validation (syntax only). FIX: per-kind matcher schema validation in rules add, or a rules-list 'inert' flag / rules check --explain. Small. Severity MED. Verified pm-v3.3 (fresh subprocess) + codegraph-cross-correlated. Hive-NHI feature campaign 2026-08-16.

### LINKED ISSUE #3085: migrate copies NULL-space source vectors as embedding_space='' → permanently non-recallable + unhealable on postgres (#3060 follow-up)
## Summary (follow-up to #3060 / f43c9777)

The #3060 migrate embedding-copy (`f43c9777`, `src/migrate.rs` Phase 3) buckets a source row whose `embedding_space` is SQL NULL under the empty string via `space.unwrap_or_default()`, then writes it with `set_embeddings_batch(&ctx, chunk, "")`. On postgres this lands the row with a NON-NULL `embedding_space = ''`, which is worse than leaving it unembedded:

- **Excluded from recall** — the #2167 gate is `AND embedding_space = $active`, and the active fingerprint is never `''`, so the row is never scored.
- **Excluded from the serve-boot NULL-space self-heal** — `PostgresStore::list_unembedded`'s heal arm is `(embedding IS NOT NULL AND embedding_space IS NULL)`; `'' IS NULL` is false, so the backfill never re-derives/re-stamps it. It is stuck (only the opt-in `AI_MEMORY_PG_HEAL_FOREIGN_SPACE` foreign-space arm, off by default, or the sqlite-only `reembed` CLI, would touch it).

## Regression vs. pre-#3060

Before #3060, `migrate` copied no vectors, so a NULL-space source row landed `embedding IS NULL` on the destination and the pg serve-boot backfill re-derived it under the live embedder and stamped the ACTIVE space → recallable. After #3060 it lands `embedding_space = ''` → permanently non-recallable and unhealable. So for an unverified-provenance (NULL-space) source corpus, #3060 turned a self-healing state into a stuck one. The durable memory TEXT is intact (no data loss), but semantic recall of those rows is unavailable until an operator runs an opt-in heal.

The Phase 3 comment shows the NULL→`''` mapping was intentional (to avoid "silent embedding-loss" from skipping), but the downstream recall-unavailability / heal-exclusion consequence appears unconsidered.

## Evidence
- `src/migrate.rs` Phase 3 (f43c9777): `by_space.entry(space.unwrap_or_default())` then `to.set_embeddings_batch(&ctx, chunk, space)` for every bucket, including `""`.
- `PostgresStore::list_unembedded` predicate (`src/store/postgres.rs`): `embedding IS NULL OR (embedding IS NOT NULL AND embedding_space IS NULL)` — excludes `embedding_space = ''`.
- `PostgresStore::set_embeddings_batch`: `UPDATE memories SET embedding=$2, embedding_space=$3` — stores `''` verbatim (non-NULL).

## Proposed fix (design choice — needs the 5-agent vote per T3/T6)
Options for a NULL-space source vector:
1. **Skip it** (do not copy) so it lands `embedding IS NULL` and the pg backfill re-derives + stamps the active space (self-healing; the pre-#3060 behavior). Cost: the exact source vector is re-derived under the live embedder rather than preserved — acceptable only because a NULL-space vector has NO provenance to preserve.
2. **Copy the vector but leave `embedding_space` NULL** (needs a primitive that can write a NULL space; the trait `set_embeddings_batch`/`update_embedding` `&str` param cannot express NULL today) so the pg NULL-space heal arm re-stamps it while preserving the vector.
3. Keep copying as `''` but widen the pg heal scan / recall to treat `''` as NULL — riskier (touches the recall gate).

Option 1 (skip) is the minimal, provenance-honest choice and matches the "re-derive from durable text is always safe" posture. Discovered during the #3069 investigation (the migrate embedding-copy regression guard).

alphaonedev:
Closes #3085
Closes #3027
Closes #3031
Closes #2672
Closes #2908
Closes #2972

Six fail-open / honesty defects on the governance + federation + boot seam. Rebased onto `origin/release/v1.0.0` = `55c0486e` (post-#3115). All three commits SSH-signed (`%G?`=G). QUAL-10 lockstep after rebase: `storage/mod.rs` 29_279 / ceiling 29_340; `daemon_runtime.rs` 13_203 / 13_270; `postgres.rs` 34_710 / 34_720.

## Per-issue

| Issue | Fix | Pin |
|---|---|---|
| **#3085** migrate NULL-space → `''` → permanent silent recall loss | write-refuse + read-heal + report, both backends; recall NOT widened (Option 1) | `tests/migrate_unattributed_embedding_3085.rs` 4/4; live pg `tests/embedding_space_unattributed_3085_pg.rs` 2/2 RAN |
| **#3027** pg `hook_subscribers` filters `metadata.read` nothing writes | unread marker is access_count | `serve_postgres_handler_parity.rs bucket_b_inbox_unread_marker_is_access_count_3027` 1/1 RAN on live pg |
| **#3031** kind-wrong/malformed enabled rule silently ALLOWS | fail-closed engine + write-time schema AND value-type validation (`MatcherValueKind`) + `inert` flag on CLI and MCP list | `tests/governance_inert_matcher_failclosed_3031.rs` 7/7 |
| **#2672** peer-steerable `LIKE '%429%'` DLQ classifier | typed `DlqErrorClass`, anchored SQL on both adapters; `deadline_exceeded` left untagged | `tests/federation_dlq_error_class_2672.rs` 3/3 |
| **#2908** boot posture banner emitted before `init_tracing()` | console subscriber armed before the banner | `tests/boot_posture_banner_2908.rs` 2/2 |
| **#2972** two embedder-model resolvers | one SSOT `resolve_boot_embedder_model`; `reembed` fails closed; doctor WARNs when configured model is unconstructible | lib filter `daemon_runtime::tests::resolve` |

## Decisions (do not silently reverse)

1. **#3031 inverts the documented "unknown matcher matches nothing" property for `refuse`/`escalate` only.** `warn`/`log` keep the old property. Seed rules R001–R004 validate and ship `enabled=0`. Documented in `docs/governance/agent-action-rules.md`.
2. **#3085 Option 1** (skip NULL-space vector). Recall is not widened.
3. **#2672** reserved leading tag, not a new column (no schema bump).
4. **#2972** `reembed` refuses; the daemon still degrades. Refusing to BOOT on an unsupported model id would be worse than degrading to the tier preset.

## Gates at the pre-rebase tip (93d09145)

Five clippy legs `-D warnings -D clippy::all -D clippy::pedantic`, fmt, examples, hardcoded-literals (no ratchet), docs-vs-ssot, qual_10 after lockstep, qual_6_7, listed tests, live pg 2/2 + 1/1 RAN not skipped. Rebase onto 55c0486e conflicted only `tests/qual_10_module_size_ceiling.rs` (ceiling restated to cover #3115 + #3085). Source hunks applied cleanly.

Do not re-run GitHub CI until this PR is queue head (serialized allowlist).


alphaonedev:
**Fable gate — FIX-BEFORE-MERGE** at head `035e815d` (shadow-QC lane, 2026-08-23; every `src/` hunk read; R-405 on the PR body: merge-base `55c0486e` ✓, 3/3 commits `%G?=G` ✓, QUAL-10 29_279 / 13_203 / 34_710 ✓).

**Accepted (ESTABLISHED):**
- **#3085** write-refuse on every vector funnel: sqlite `db::set_embedding` (`storage/mod.rs:15114`), `set_embeddings_batch` (`:15185`), `set_embeddings_batch_reembed` (`:15762`), trait default `set_embeddings_batch` (`store/mod.rs:1086`), pg `update_embedding` (`postgres.rs:18137`) + `set_embeddings_batch` (`:18316` before `begin()`), sqlite SAL `update_embedding` (`sqlite.rs:2594`) — one typed `InvalidInput` on both backends. Read-heal: `SQL_FRAGMENT_EMBEDDING_SPACE_UNATTRIBUTED` replaces every `embedding_space IS NULL` predicate in `src/` (grep: zero remaining outside doc comments); pg `list_unembedded` both arms, pg + sqlite `adopt_legacy_embedding_space` (mismatch probe excludes `''`), `stamp_embedding_space_attested`. Recall SQL untouched (Option 1 honoured). `migrate` treats `''` source stamps as unattributed too, reports `embeddings_unattributed`, dry-run split matches the live split, and the short-copy refusal accounts for the deliberate skip.
- **#3027** pg inbox derives unread from `access_count` exactly as `notify.rs:163`; projects `read` + `access_count`; `unread_count` from the same marker. `access_count` is mapped by the pg row mapper (`postgres.rs:22492`).
- **#3031** `matcher_key_schema` agrees key-for-key and type-for-type with every `match_*` reader (`agent_action.rs:728-872`; bash = `command_substring|command_regex`, fs `glob`, net `host`, spawn `binary` + `args_contain`/`disk_free_min_gib:u64`, custom `kind` + 3 optionals, read 4 keys incl. `all:bool`). Kind mismatch is checked BEFORE inert detection, and `load_for_action` is `list_enabled_by_kind`, so an inert bash rule cannot block a filesystem action and disabled rows never reach `evaluate`. `{"all": false}` stays evaluable. Warn/log inert rules stay non-blocking.
- **#2672** every peer-influenced `last_error` is minted through `DlqErrorClass::stamp` (`http_status_reason`, both 2xx non-ack shapes, transport error, governance refusal, serialise failure, the retry pair, all four `bump_dlq_attempt` literals); the class comes from the REAL status line (`from_http_status`), never from body text; `parse` rejects a tag containing `]`/a nested open; the reset predicate on BOTH adapters is `LIKE '[ai-memory:class=throttle]%' OR (NOT LIKE '[ai-memory:class=%' AND LIKE '%429%')`, so a stamped `peer_refused` row can never satisfy it, and a pre-upgrade legacy row is re-stamped on its first post-upgrade replay. `deadline_exceeded` exact-match contract preserved.
- **#2908** `install_boot_console_subscriber` runs at the top of `run()` (`daemon_runtime.rs:1143`) for Serve/Curator/Watch only — Curator/Watch already called `init_tracing()` later at base (`:1961`, `:2299`), so no new stdout sink on a CLI one-shot or on `mcp`; `main.rs:148` still installs `[logging]` file logging BEFORE `run()`, so `try_init` keeps the file sink when enabled. Test is a real `CARGO_BIN_EXE` subprocess with `AI_MEMORY_NO_CONFIG=1 RUST_LOG=info`.
- **#2972** one resolver (`resolve_boot_embedder_model`) feeds `build_embedder`, `resolve_configured_embedding_dim`, `reembed`, and `doctor`; `reembed` exits `EXIT_EMBEDDER_INIT_FAILED` with no write when the configured model is unconstructible.

**Fix before merge:**
1. **MED, claims-truth / partial fix — #2972 cannot be `Closes`d as shipped.** Decision 4 leaves the DAEMON substituting the tier preset for an explicitly configured, unconstructible `[embeddings].model` (`resolve_embedder_model_reported` `:3043-3048` → `(preset, Some(raw))`, `build_embedder` takes `.model` and proceeds). That contradicts the shipped contract in `docs/integrations/llm-backends.md:27` ("Embedder construction is fail-closed … degrades loudly to keyword mode … never silently backed by" another client) and the issue's stated expectation (boot/doctor/reembed consistent). After this PR the reporter's daemon still mints vectors under `all-MiniLM-L6-v2` while config says `qwen3-embedding:4b`; only `doctor` and `reembed` became honest. Acceptable resolutions: (a) in `build_embedder`, when `unhonoured_config_model.is_some()`, degrade to keyword (return `None` embedder with the existing `#1593`-style loud WARN) — this is what the docs already promise and what #3215's fail-closed-config direction implies; or (b) keep Decision 4, change `Closes #2972` to `Refs #2972`, correct `llm-backends.md:27` to state the substitution, and leave the daemon leg open on #2972. Either way the PR body's "#2972 one embedder-model resolver" row must say which.
2. **MED, security — the write-time guard has a second verb.** `rules enable` (`cli/rules.rs:354-367`) flips `enabled=1` on a legacy inert row with no schema check, minting exactly the state `rules add` now refuses — and post-#3031 that row REFUSES every action of its kind. Apply `validate_matcher_for_kind` (parse the stored matcher) in the `Enable` arm and refuse with the same operator message; add the e2e case to `rules_add_refuses_a_kind_wrong_matcher_3031`. (The `[#3031 fail-closed: …]` refusal text and the `tracing::error!` are good; this closes the remaining mint path.)
3. **Rebase — real conflict with #3133 (`fix/security-fail-open-gates`, merging ahead of you): `src/federation/sync.rs` + `tests/qual_10_module_size_ceiling.rs` (`git merge-tree` onto `44b22fb7` confirms).** #3133 moved the egress gate + body serialisation out of `post_once` into `build_governed_peer_post` (`sync.rs:98-130`) and switched it to `check_governed`. Resolution: KEEP #3133's builder and `check_governed`; DROP this branch's edits to the old in-`post_once` `wire_check::check` block and `serialise body` arm; re-apply the stamp at the single mapping site `Err(reason) => return AckOutcome::Fail(DlqErrorClass::Other.stamp(&reason))` (`sync.rs:235-238` on #3133). The `bulk_catchup_push` lane (`:2449`) maps `Err` to a catch-up error row, not a DLQ `last_error`, so it needs no stamp. Re-run the #2672 + #3148 suites after the merge.

**Follow-ups (non-blocking — fold in if cheap, else I file them `fable-qc`):**
- (a) **Upgrade pre-flight for #3031.** A corpus containing a legacy enabled inert `refuse`/`escalate` rule flips from allow-all to refuse-all-of-kind at the first evaluation after upgrade, with only a `tracing::error!`. Add a `doctor` row (and a boot WARN line, the #1734 PE-1 "WILL DENY" precedent) enumerating `enabled && rule_matcher_is_inert` blocking rules by id; state the upgrade effect in the CHANGELOG bullet explicitly (it currently describes the mechanism, not the operator-visible flip).
- (b) **QUAL-7 ratchet raised 43→44** for `validate_matcher_for_kind -> Result<(), String>`. The ceiling exists to stop new stringly errors; a two-line `MatcherSchemaError(String)` with `Display + Error` (ERRORS-10) keeps the ceiling at 43 and the CLI rendering identical.
- (c) #2908: the issue asked for a sink note in `docs/deploy/enterprise-federation.env` (when `[logging].enabled`, the banner lands in the log file, not the console); not in this diff. The subprocess test pins the asi-hard banner; the §5.3 banner shares the same block at `run():1198-1223`, so coverage is by construction — say so in the test doc or add the needle.
- (d) `embedding_space_unattributed_3085_pg.rs:39` `live_pg() -> Option` soft-skips when the pg tier is absent — same shape as the #3134 gate follow-up; the f2 pg lane runs it, but a soft-skip reads green on a runner with no tier.

**Security lens, no change required:** `SQL_LIKE_*` constants carry no `%`/`_` (pinned by test); `[` is not a LIKE metachar on sqlite or pg; the fragment constants are `format!`ed only from `&'static str` (no injection surface); `reject_unattributed_embedding_space` message names no secret; inert-rule refusal text echoes only the rule id/kind.

Do not merge until items 1–3 are on the head; re-request the gate with the new SHA.

alphaonedev:
**Fable gate — FIX-BEFORE-MERGE (narrowed to 2 residuals)** at `a61634b7` (re-gate).

Verified on this head: rebase kept #3133's `check_governed` in `federation/sync.rs` (item 3 ✓); `rules add` now runs per-kind matcher SCHEMA validation before signing (`cli/rules.rs:282`, #3031 rationale comment); `unhonoured_config_model` is threaded (`daemon_runtime.rs:2975/:3002`) and surfaced by `doctor` (`doctor.rs:2070`) and `reembed` (`reembed.rs:327`); the qual-10 bumps are rebase-mechanical.

Residuals:
1. **MED (original item 2) — `rules enable` still has NO matcher-schema validation.** `RulesAction::Enable` (`cli/rules.rs:366+`) goes straight to `set_enabled_signed`: a pre-#3031 legacy row with a kind-wrong matcher can still be re-enabled into the exact inert-enabled-refuse state `rules add` now refuses to mint — and post-#3031 that row refuses every action of its kind. Apply `validate_matcher_for_kind` in the Enable arm (refuse with a pointer to `rules add` for re-minting), or refute with the argument for why enabling an invalid legacy row is acceptable.
2. **MED (original item 1, partially applied) — the daemon leg still substitutes SILENTLY at boot.** `resolve_embedder_model_reported` returns the substitution but no `serve`/`mcp` boot line emits it — only operators who happen to run `doctor` or `reembed` see it. Minimum fix: one boot `tracing::warn!` at the `BootEmbedderModel` construction site naming the unhonoured configured model and the preset used (matches the embedder-init-failure loudness contract at `llm-backends.md:566`); then either keep `Closes #2972` with that WARN, or downgrade to `Refs`.

Both are small; with them this approves on sight.

alphaonedev:
Fable FIX-BEFORE-MERGE residuals 1+2 are on head (this push):

1. `rules enable` now runs `validate_matcher_for_kind` before `set_enabled_signed`. An inert pre-#3031 matcher is refused with a pointer to `rules add`. Pinned by `run_rules_enable_refuses_inert_matcher_3031`.
2. `resolve_boot_embedder_model` emits `tracing::warn!` when `[embeddings].model` is unconstructible, naming the configured id and the substituted preset (#2972 boot path).

34-watcher will be re-armed on the new SHA. Merge when 34/34.

alphaonedev:
**Fable gate — APPROVED** at `f6b328a0` (re-gate; both residuals fixed). (1) `rules enable` now runs `validate_matcher_for_kind` BEFORE `set_enabled_signed`, refusing a legacy kind-wrong matcher with the re-mint-via-`rules add` pointer (`cli/rules.rs:~370-390`, #3031 rationale). (2) Boot `tracing::warn!` at the resolution site (`daemon_runtime.rs:3004-3008`) names the unhonoured `[embeddings].model` and the substituted preset on the shared daemon/mcp funnel — `Closes #2972` now stands with the loudness contract met. All earlier items verified across the re-gate sequence. Merge when 34/34 on `f6b328a0`. (A stray 'PENDING-VERIFY placeholder' comment from my tooling was posted and deleted — disregard if cached.)

Issue #3241, opened 2026-08-23
list_unembedded non-admin arm returns Ok(vec![]) on both backends — "nothing to do" == "I cannot answer" one layer down


alphaonedev:
## Summary

`MemoryStore::list_unembedded` on **both** production adapters returns `Ok(Vec::new())` when `!ctx.bypass_visibility` (non-admin). That is the same "nothing to do" == "I cannot answer" conflation #2639 closed at the **trait default** (empty vec was "end of corpus" for `run_embedding_backfill_on_store`).

The serve-boot sweep is the sole production caller and uses `CallerContext::for_admin`, so this is **not live**. It is an established, documented, cross-backend #1586 contract, pinned by `tests/cov_ga2_postgres.rs::list_unembedded_refuses_non_admin_caller` ("non-admin context returns empty (gate)").

## Why not in the silent-default cluster PR

SUCCESSOR-RULES §3 / operator rule: never silently tighten a documented default. Making sqlite refuse unilaterally would manufacture a real pg/sqlite parity divergence, which is worse than the inherited shape. The sqlite override in the cluster PR copies the pg gate **verbatim**.

## Proposed fix

Convert **both** adapters to a typed refusal (`StoreError::Unauthorized` / `UnsupportedCapability` / a dedicated variant) in **lockstep**, and update `list_unembedded_refuses_non_admin_caller` in the same change so it asserts the refusal rather than emptiness.

## Evidence (R-405)

- `src/store/sqlite.rs` `list_unembedded`: `if !ctx.bypass_visibility { return Ok(Vec::new()); }`
- postgres twin: same empty-vec non-admin arm
- pin: `tests/cov_ga2_postgres.rs::list_unembedded_refuses_non_admin_caller`

Related: #2639 (trait default), #1586 (admin gate).

Issue #3181, opened 2026-08-22
[data-integrity][claims] MemoryStore::store_batch documents 'atomic (all rows commit or none do) … SQLite inherits it unchanged' — the sqlite adapter is the per-row default loop (committed prefix on mid-batch failure)


alphaonedev:
Scout-verified (R-405) at `cda57210` from the ox-alpha SAL parity audit (152 methods; `.local-runs/ox-c4-parity/`). pg lane evidence required: `AI_MEMORY_TEST_POSTGRES_URL=$(cat .local-runs/ENTERPRISE-FED-TEST-PG-URL.txt) cargo test --features sal-postgres -- --include-ignored`.
**Code.** `src/store/mod.rs:955-961` states the contract; `:988-997` default `for memory in memories { ids.push(self.store(ctx, memory).await?); }`; `grep -c 'async fn store_batch' src/store/sqlite.rs` = 0. pg `postgres.rs:17267-17269` is one tx. Same silent-default family: `store_with_embedding` (0 sqlite overrides → trait default `mod.rs:879-886` drops the vector silently; documented at `sqlite.rs:222-235`), `set_embeddings_batch` (per-row loop, `written += 1` unconditionally counts ghost ids, committed prefix), `check_memory_quota` (trait default no-op `Ok(())` `mod.rs:3390-3398` — any trait-routed sqlite caller gets no quota gate).
**Impact.** A mid-batch failure on sqlite leaves a committed prefix and returns only Err — the documented contract enterprise consumers rely on is false ('bet-the-farm' standard). Ghost-id counts are claims-truth defects.
**Fix.** Give SqliteStore a real `store_batch` wrapping the loop in BEGIN IMMEDIATE when `is_autocommit()` (the #3115 `delete_inner` precedent; coordinate with the #3163 RAII guard from Opus S); same for `set_embeddings_batch` with `changes()`-based counts; `check_memory_quota` sqlite override or trait default → UnsupportedCapability (#2639-style, Opus Z is doing list_unembedded — coordinate).
**Test.** sqlite: trigger-injected failure on row 3 of 5 → rows 1-2 absent after Err; nested-tx guard; intra-batch duplicate LAST-WINS result equal across backends (extend `tests/bulk_write_truthfulness_2551.rs`). GA: BLOCKING.

alphaonedev:
Closed by fable-merge of #3237 at 9cc3df51 G (sqlite SAL parity: )

Issue #2638, opened 2026-08-01
MemoryStore trait defaults silently discard data: store_with_embedding drops the vector, store_batch drops atomicity (SqliteStore overrides neither)


alphaonedev:
**Not covered by any existing issue.** Verified by reading both adapters at HEAD `5449b6da`.

## `store_with_embedding` — the trait default silently discards the vector

`src/store/mod.rs:~1155`. The default body is:

```rust
self.store(ctx, memory).await     // _embedding and _space are DROPPED
```

**`SqliteStore` does not override it.** `PostgresStore` does. The only production caller is `src/handlers/create.rs:1124`, which sits inside the postgres branch — so this is **latent, not live**.

It is filed anyway because of its *shape*: a silent-drop default on a **data-bearing argument**, in a codebase whose other trait defaults are `UnsupportedCapability`. The moment any caller routes a sqlite store through this path, embeddings vanish with no error, no warning, and a success return. That is the #2444 false-success class pre-installed and waiting for a caller.

## `store_batch` — the default silently changes the failure contract

`src/store/mod.rs:~1180`. The `PostgresStore` override is documented **atomic, all-or-nothing**. The trait default loops `store` and is **not** atomic.

So any future routing of `POST /api/v1/memories/bulk` to sqlite silently converts the batch-failure contract from all-or-nothing to **partially-applied**, with the caller's response shape unchanged. Given #2588/#2551/#2552 already established that this endpoint reports rows it did not write, a silent atomicity downgrade underneath it is the worst possible place for a permissive default.

## Proposed fix

Convert both defaults to `UnsupportedCapability` — matching the convention the rest of the trait already follows — so a missing override is a loud compile-or-runtime refusal rather than silent data loss. If `SqliteStore` should support them, implement them explicitly.

Found by the 3x7 issue audit, cross-backend parity lens.

alphaonedev:
Closed by fable-merge of #3242 at `ef22d020` (G). GitHub auto-close does not fire on local signed merge.

alphaonedev:
Landed in ef22d020 via #3242.

PR #3242, opened 2026-08-23
fix: silent-default / fail-open cluster (#2639 #2638 #2592 #2607 #2427 #2431 #2617 #2596)

### LINKED ISSUE #2639: A sqlite HTTP-only serve daemon has NO embedding backfill at all — list_unembedded trait default returns empty; bulk rows are permanently semantically invisible
**Not covered by any existing issue.** Verified at HEAD `5449b6da`. Compounds #2594 and #2607.

## The gap

`src/daemon_runtime.rs:5768-5781` — the #1579 A4 serve-boot sweep drains `MemoryStore::list_unembedded`.

The trait default is `Ok(Vec::new())` (`src/store/mod.rs:~1234`), and **only `PostgresStore` overrides it**. So on sqlite the sweep finds nothing to do, always.

The in-code comment justifies this by asserting sqlite is covered because "its side-table embeddings are backfilled by the MCP boot path". **That mitigation does not exist for an HTTP-only sqlite deployment** (`ai-memory serve --db x.db`, no MCP stdio): `run_embedding_backfill_with_batch_size` is invoked only at `src/mcp/mod.rs:4119`.

## Why it matters — the compounding is the finding

- **#2594**: bulk create embeds on **neither** backend. That is parity-correct, so it reads as a shared, symmetric defect.
- **But the recovery diverges.** On postgres those rows self-heal on the next boot sweep. On a sqlite `serve` daemon they are **permanently semantically invisible** — present in the corpus, returned by keyword search, never reachable by semantic or hybrid recall, forever.

#2594's "on either backend" framing therefore understates the impact: the write-path gap is symmetric, the *repair* path is not, and the repair path is the half that determines whether the defect is transient or permanent.

## The sanctioned repair does not close it either

`ai-memory reembed` is `src/cli/commands/reembed.rs:353` → `db::open(db_path)`, pure rusqlite. Two problems:

1. Per **#2607** it silently leaves ~14.7% of rows unembedded at the default batch size **while exiting 0**.
2. Per **#2572** it is a `--db` *write* verb, so pointed at a postgres deployment it manufactures a phantom sqlite database, writes embeddings into it, and exits 0.

## Proposed fix

Implement `list_unembedded` on `SqliteStore` so the serve-boot sweep covers HTTP-only sqlite deployments, and treat the trait default `Ok(Vec::new())` as suspect for the same reason as the other silent-default findings: "nothing to do" and "I cannot answer" must not be the same value.

Found by the 3x7 issue audit, cross-backend parity lens.

alphaonedev:
Closes #2639 #2638 #2427 #2617

Addresses (partial — residuals remain, originals stay open): #2592 #2607 #2431 #2596

Eight surfaces that answered **"nothing to do"** where they meant **"I did not look"**, or reported **success** while discarding what they were handed. Successor lane (Opus 5 Z2 → this session) continuing an inherited worktree; rebased onto `origin/release/v1.0.0` = `ae3e0aec` (Merge #3134). Every cited line re-verified at this HEAD per R-405.

rust-1.98: **ERRORS-01** / **ERRORS-19** (trait defaults return `Result::Err` instead of dropping a data-bearing argument), **PERF-03** (saturating window arithmetic in `scheduled_validity`), **PERF-07** (textual integral parse of EXPLAIN `rows=N`, no `as`/`f64` cast).

## Per-issue

| # | Surface | Change | Disposition |
|---|---|---|---|
| #2639 | `MemoryStore::list_unembedded` default `Ok(vec![])` | `SqliteStore` now **implements** the scan (`src/store/sqlite.rs:2628`); the trait default **refuses** (`UnsupportedCapability`, `src/store/mod.rs:1081`) | **Closes** |
| #2638 | `store_with_embedding` / `store_batch` defaults | both **refuse** instead of dropping the vector / approximating atomicity (`src/store/mod.rs:890` / `:1017`) | **Closes** |
| #2592 | 1000-subscriber dispatch cliff | WARN + `ai_memory_subscription_dispatch_truncated_total` + durable `subscription_dlq` row under `_dispatch_scan_truncated` | **Addresses** (correctness cliff only) |
| #2607 | `reembed` exit contract | `EXIT_PARTIAL_COVERAGE = 5`, post-sweep read-back, `--expect-skipped <N>` ratchet | **Addresses** (item 1 unverified) |
| #2427 | `ChainResult::ModifiedAllow` folded into `Allow` | explicit arm + WARN parity with the #1752 signal path (`src/mcp/mod.rs:1832`); `HookDecision::Modify` doc names the inert sites. **Log-only** — arm still returns `Ok(())` | **Closes** (GA half; honour-delta is v1.x) |
| #2431 | `scheduled_validity` ignored `valid_from`/`valid_until` | honours the #1834 interval; half-open `[valid_from, valid_until)` | **Addresses** (item 2 TOON columns is U2) |
| #2617 | pg `health_check` `COUNT(*)` per probe | constant-cost `SELECT EXISTS (SELECT 1 FROM memories LIMIT 1)` (`src/store/postgres.rs:27417`) | **Closes** |
| #2596 | `links.rs` `Err(_) => (None, None)` | WARNs with error + ids + relation before suppressing the dispatch | **Addresses** (scalar projection would close the hole) |

## #2639 / #2638 — which trait defaults changed

Three, all in `src/store/mod.rs`: **`list_unembedded`**, **`store_with_embedding`**, **`store_batch`**. `check_memory_quota` is **untouched** (that is #3181 / CC2's).

**Blast radius verified at this HEAD, not assumed.** Only two production adapters implement `MemoryStore` (`SqliteStore`, `PostgresStore`). Every production call site of the two refusing write defaults is inside a postgres-only funnel:

- `src/handlers/bulk.rs` → `bulk_create_postgres`. The sqlite lane is `bulk_create_sqlite` and routes through the transactional `crate::storage` funnel — it never reaches either method.
- `src/handlers/create.rs` → `create_memory_postgres`.

So the refusal cannot fire on a sqlite write path. This matches #2638's own "latent, not live" reading, independently re-verified rather than inherited.

**Composition with #3181 (CC2).** #3181 (not on this base) would fix `store_batch` by giving `SqliteStore` a **real** BEGIN-IMMEDIATE override; this PR makes the **trait default** refuse. They compose: after both land, sqlite is atomic and only a genuinely unimplemented adapter refuses. Hunks are disjoint (`check_memory_quota` untouched).

## #2617 — pg evidence (live certified stack)

In-repo test **RAN** (not self-skipped) against `AI_MEMORY_TEST_POSTGRES_URL` → `127.0.0.1:5445/ai_memory_test` (sslmode=verify-full), corpus **1398** rows at run time:

```
test a_pg_health_check_issues_no_corpus_count_2617 ... ok
test pg::c_control_legacy_count_scanned_the_whole_corpus_2617 ... ok
test pg::b_pg_health_probe_touches_at_most_one_row_2617 ... ok
test result: ok. 3 passed; 0 failed; 0 ignored
```

Cell B asserts the new probe touches ≤ 1 row and is not an Aggregate. Cell C (control) asserts the legacy `COUNT(*)` still scans the whole 1398-row corpus on the same database, so B cannot pass because the plan is cheap for everyone. Out-of-band EXPLAIN on the same stack (HANDOFF, then 1,317 rows; corpus has since grown) measured **1317 → 1** rows touched, **110 → 2** buffers, **0.780 ms → 0.043 ms**.

The probe still touches `memories`, so a dropped or unreadable table still fails it — "O(1) by no longer looking at the store" would be a different defect, and cell A asserts against it.

`max_actual_rows` takes the EXPLAIN integral part **textually** (pg ≤17 `rows=1317` and pg 18 `rows=1317.00`) and **panics** on an unreadable token rather than reporting zero rows touched (which would make cell B pass vacuously). PERF-07.

## Tests

- `tests/subscription_dispatch_truncation_2592.rs` — 3 passed
- `tests/silent_discard_observability_2427_2596.rs` — 2 passed
- `tests/recall_decoration.rs` — 10 passed
- `tests/qual_10_module_size_ceiling.rs` — 2 passed (`src/subscriptions.rs` 4_520 → 4_600, measured 4_585; `src/store/postgres.rs` ceiling stayed 35_200, measured 35_185)
- `tests/qual_6_7_legacy_error_type_ceiling.rs` — 2 passed
- lib (filtered): `backfill_sweep_stops_loudly_on_unsupported_list_unembedded_2639`, `default_store_with_embedding_refuses_rather_than_dropping_the_vector_2638`, `default_store_batch_refuses_rather_than_approximating_atomicity_2638`, `list_unembedded_scans_null_embedding_rows_for_admin_2639`, `coverage_is_partial_separates_complete_from_partial_2607`, `expect_skipped_ratchet_absorbs_only_the_attested_population_2607`, `exit_codes_are_distinct_2607`, … — **20 passed**
- `tests/pg_health_probe_constant_cost_2617.rs` — 3 passed, cells B+C **RAN** (no skip)

CI-parity: `cargo fmt --all --check`; clippy `-D warnings -D clippy::all -D clippy::pedantic` on `--all-targets` default / sal / **sal-postgres** / vectorlite; `cargo check --examples`; `cargo check --all-targets` default / sal / sal-postgres.

## Residual work — these issues are only PARTLY closed

| # | Landed here | NOT landed |
|---|---|---|
| #2596 | the silent-no-dispatch hole is now **audible** (WARN with error + ids + relation) | items 1-3: reuse `namespace_by_id` + a companion `metadata->>'agent_id'` scalar, which would **actually close** the hole; update/promote read collapsing; a `scalar_fields_by_id` projection. Deferred because the companion scalar is a new SAL trait method on the surface #3181/CC2 owns. |
| #2592 | the **correctness cliff** (WARN + metric + durable DLQ row) | perf items 1/2/3/5 and the 2026-08-22 comment's two additions: sqlite `list_by_namespace_prefix` paging the whole corpus per event, and pg's missing `expires_at` guard letting expired subscriptions fire hooks. |
| #2607 | exit contract + post-sweep read-back + `--expect-skipped` ratchet | item 1, adaptive batch bisection in `src/llm.rs::embed_texts_async`. The issue itself marks that root cause **unverified** and needs a live-provider run. |
| #2431 | `scheduled_validity` honours the interval | item 2: carry `valid_from`/`valid_until` through the recall projection **including the TOON column set** (U2 owns the tool schemas). Item 3 respected — **no** default filtering of closed claims. |

## Escalation (not fixed here — deliberate)

`list_unembedded`'s **non-admin arm returns `Ok(Vec::new())` on BOTH backends** (`src/store/sqlite.rs:2640`, postgres twin) — which is the very "nothing to do" == "I cannot answer" conflation this PR is about, one layer down. **Not changed** (SUCCESSOR-RULES §3 — never silently tighten a documented default):

- It is an established, documented, cross-backend #1586 contract, pinned by `tests/cov_ga2_postgres.rs::list_unembedded_refuses_non_admin_caller` ("non-admin context returns empty (gate)").
- The new sqlite override mirrors the pg gate **verbatim**; making sqlite refuse unilaterally would manufacture a real parity divergence.
- It is not live: the sole production caller is the serve-boot sweep under `CallerContext::for_admin`.

Follow-up filed: [#3241](https://github.com/alphaonedev/ai-memory-mcp/issues/3241).

## Data-integrity posture

- No destructive change. `list_unembedded` on sqlite is admin-gated (`bypass_visibility`) and reuses `db::get_unembedded_ids_batch`, which already applies the #1779 decrypt-or-skip resolver — an unopenable at-rest row is **skipped**, never embedded as its empty-seal placeholder.
- Every changed default **fails closed**: a refusal replaces a false success. `run_embedding_backfill_on_store` already has the `Err` arm that WARNs and stops.
- #2427 is **log-only** — the arm still returns `Ok(())`, byte-identical behaviour, satisfying the issue's "verify the discard is not load-bearing first".
- #2592's DLQ row is bounded by the existing #1253 `MAX_SUBSCRIPTION_DLQ_ROWS` per-subscription cap.


alphaonedev:
**Fable gate — FIX-BEFORE-MERGE** at `5526c9f7` (shadow-QC lane; ox-alpha first pass on the PR's own tree, every item below re-read by Fable on this head).

Accepted (ESTABLISHED): #2639 trait default refuses + sqlite `list_unembedded` bounded NULL-embedding scan with the #1586 admin gate mirrored; #2431 half-open `scheduled_validity` semantics; #2592 cliff now loud on the pg dispatch lane (WARN + monotonic counter + capped DLQ); #2607/#2427 WARNs present. The PR's own "Addresses (partial)" wording for #2592/#2596 is honest.

1. **HIGH (ESTABLISHED, cross-backend) — #2596 hardened only the postgres link-dispatch arm.** `src/handlers/links.rs:713-720` (sqlite arm) still does `db::get(&lock.0, &source_id).ok().flatten().map_or_else(|| (DEFAULT_NAMESPACE, None), …)` — a read fault silently FABRICATES `namespace = default, owner = None` into the subscription event (mis-attributed anchor, exactly the silent-default shape this cluster is closing); the pg arm in the same file got the `tracing::warn!` + skip. Mirror it on sqlite, and check `src/mcp/tools/link.rs:334-341` (third anchor site) in the same pass.
2. **MED (ESTABLISHED) — #2592 truncation detector is an off-by-one false positive at exactly-`limit` populations** (`handlers/subscriptions.rs:1041` uses `>=` on the post-filter count; cannot distinguish "cut" from "exactly full"), and its DLQ rows compete with real poison-subscriber rows for the cap. Fetch `limit+1` and key on the overflow; or record the truncation out-of-band.
3. **MED (ESTABLISHED) — `record_dispatch_scan_truncation` result is swallowed** (`:1046-1062`): the observability fix can lose its own evidence silently. WARN on the error path at minimum.
4. **MED (ESTABLISHED, parity) — #2592 observability exists only on the pg dispatch lane**; the sqlite `dispatch_event_with_details` (`src/subscriptions.rs:793-798`) keeps the swallowed list error and no truncation signal. Mirror or state the scope in the PR/CHANGELOG.
5. **MED (ESTABLISHED) — the `Ok(vec![])` "wrong context == nothing to do" shape survives one layer down** (`sqlite.rs:2637-2640`, pg `:18536-18540`). This is your own #3241 — cross-link it from the PR and do not let `Closes #2639` imply #3241 is done.
6. **MED (PLAUSIBLE, verify) — the same silent-drop class the PR fixes for `store_with_embedding` is live in the trait defaults `update_embedding` (`store/mod.rs:1038-1048`) and `set_embeddings_batch` (`:1101-1117`)**; the "matches every other optional method" claim in the body is false for those two. Fix or narrow the claim.

Follow-ups (non-blocking): `reembed` coverage read-back is corpus-global so a `--max-rows` run / benign concurrent writes exit 5 (`cli/commands/reembed.rs:203-222`, `:549-559`); `--expect-skipped` grants 2N slack (`:206-213`); sqlite `list_unembedded` scans only `embedding IS NULL` while pg also heals vector-present/space-missing (`postgres.rs:18560-18600`) — document or align; the #2427/#2596 tests in `tests/silent_discard_observability_2427_2596.rs:58-118` pin source TEXT, not behaviour; subscriber HMAC secret resolution degrades to UNSIGNED on unreadable secret (`subscriptions.rs:~822-828`, pre-existing) — file if untracked.

Do not merge until 1–4 are on a new head; re-request the gate.

PR #3258, opened 2026-08-24
fix(hooks): tokio Mutex for pre-event/pre-signal registries (CONCURRENCY-20, #3256)

### LINKED ISSUE #3256: [fable-qc][concurrency] consult_pre_event_gate holds std Mutex<ExecutorRegistry> across .await (CONCURRENCY-20)
Fable FOLLOW-UP (b) on #3212 (pre-existing, not introduced by the LLM-bridge bound).

`consult_pre_event_gate` gate closures hold the std `Mutex<ExecutorRegistry>` guard across `chain.fire(..).await`. Reached from HTTP handlers on the multi-thread runtime, so concurrent writes serialize on the OS-level lock while one evaluation is in flight. The lock wait happens inside the bridged future before its first poll, so the tokio bound cannot fire during it — bounded only transitively by the holder's own 60s + 5s.

Fix: copy out what is needed and drop the guard before `.await` (CONCURRENCY-20), or use an async mutex.

R-405 at merge `78f7c1c3` / PR #3212 head. Do not silently change documented defaults.

alphaonedev:
## Summary
`consult_pre_event_gate` and the sibling PreSignalSend gate held `std::sync::Mutex<ExecutorRegistry>` across `chain.fire(..).await` (rust-1.98 **CONCURRENCY-20**). HTTP create consults the pre-event gate on the multi-thread runtime, so concurrent writes serialized on an OS-level lock for the whole hook evaluation.

Both process-global registries are now `tokio::sync::Mutex`; the guard is acquired with `.lock().await` inside the already-bridged future. Fail-closed `#3140` bridge-budget behaviour is unchanged.

Closes #3256

## Evidence
- Head `9b012482` `%G?=G`, rebased onto `41142302` (contains base).
- Local CI-parity ALLGREEN: 5 clippy legs (`-D warnings -D clippy::all -D clippy::pedantic`), fmt, examples, qual_10/6_7, docs-vs-ssot, hardcoded-literals.
- R-405: `src/mcp/mod.rs` `PreEventEnforceGate.registry: tokio::sync::Mutex<…>` + `gate.registry.lock().await` inside `block_on_local_bounded` (`consult_pre_event_gate` ~1839). Same for PreSignalSend (~1607).
- CHANGELOG `[Unreleased]` keeps base entries below the new #3256 section.

Do not merge until queue head + 34/34. Never tag.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,15 @@ probe (citing #2445) but kept it for the corpus probe.
 - **Truthy grammar SSOT.** `config::migration_require_core_tables` now delegates
   to `governance::audit::env_flag_enabled` (`1` / `true` / `yes` / `on`); the
   documented tokens are unchanged.
+### Security (`consult_pre_event_gate` no longer holds a `std::sync::Mutex` across `.await` — #3256)
+
+- **#3256 (CONCURRENCY-20) — `consult_pre_event_gate` (and the sibling
+  PreSignalSend gate) held `std::sync::Mutex<ExecutorRegistry>` across
+  `chain.fire(..).await`.** HTTP create consults that gate on the multi-thread
+  runtime, so concurrent writes serialized on an OS-level lock for the whole
+  hook evaluation. Both process-global registries are now `tokio::sync::Mutex`;
+  the guard is acquired with `.lock().await` inside the already-bridged future.
+  Fail-closed bridge-budget behaviour is unchanged.
 
 ### Security (CLI-surface parity: sign `link` edges #3036; bind the recall ledger to the CALLER, not a namespace #2988)
 
--- a/src/mcp/mod.rs
+++ b/src/mcp/mod.rs
@@ -1621,11 +1621,13 @@ pub(crate) fn set_post_signal_ack_sink(tx: tokio::sync::mpsc::UnboundedSender<se
 
 /// #1752 — the process-global PreSignalSend enforcement gate: the configured
 /// chain + its executor registry. `fire` needs `&mut` the registry, so it is
-/// held behind a `Mutex` (the sync stdio loop is single-threaded, so the lock
-/// is uncontended). `None` until [`run_mcp_server`] installs it.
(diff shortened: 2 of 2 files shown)


alphaonedev:
**Grok gate — APPROVED** at `3c65d93a`

Base: `1c230eda`. Diff read: 100% of src/ hunks (`src/mcp/mod.rs` + CHANGELOG). rust-1.98 rules checked: CONCURRENCY-20 (never hold a `std::sync` `Mutex` guard across `.await`), CONCURRENCY-22 (hook fire already bridged via `block_on_local_bounded`).

Fable items: none — this PR had **NO GATE YET**; this is the self-gate. `PreSignalSendGate` and `PreEventEnforceGate` registries are `tokio::sync::Mutex`; both `consult_pre_event_gate` and `pre_signal_send_decision` acquire with `.lock().await` inside the already-bridged future. `#[deny(clippy::await_holding_lock)]` on those two fns is the revert detector (the lint is suspicious, not in `clippy::all`). tokio Mutex does not poison; a panic while holding drops the guard (CONCURRENCY-18: no silent `into_inner` of a possibly-inconsistent std mutex).

Security lens:
- authz: none, checked — gate consult semantics unchanged
- fail-open: none, checked — `#3140` fail-closed bridge-budget path unchanged
- secrets: none, checked
- TOCTOU: none, checked
- parity(sqlite↔pg): N/A — process-global hook registries, not store adapters

Claims (PR body + CHANGELOG): ESTABLISHED — HTTP create consults this gate on the multi-thread runtime; a `std::sync` guard held across `chain.fire(..).await` serialized workers. Fail-closed bridge-budget behaviour unchanged.

CI: 34/34 green at `3c65d93a`. pg lane: no postgres/schema/handler files; required `Check (linux-fed,enterprise-fed)` success at this SHA. Local: 5 clippy legs + fmt + qual_10 + qual_6_7 + literals + docs-vs-ssot ALL-GREEN.

Follow-ups filed: none

PR #3267, opened 2026-08-26
fix(postgres): #3264 fail-closed classified CREATE EXTENSION vector diagnostic + pgvector/AGE preflight in schema-init/doctor + managed-Postgres docs

alphaonedev:
## Summary

- `PostgresStore::connect`: under the existing bootstrap advisory lock and BEFORE `raw_sql(INIT_SCHEMA)`, a preflight (`pg_available_extensions` for `vector`, `pg_roles.rolsuper`, `ag_catalog` USAGE, `current_database()`) feeds a pure decision table (`classify_pgvector_preflight`); `NotAvailableOnServer` (0A000) and `AvailableNeedsSuperuserCreate` (42501) become classified `BackendUnavailable` details naming the remedy (pgvector-bearing image / #1065; superuser one-time `CREATE EXTENSION vector` via CNPG `postInitApplicationSQL` / `Database` CR / RDS `rds_superuser`). The `raw_sql` error path maps SQLSTATE 42501/0A000 to the same text only when the preflight corroborates (or could not run). **The abort is unchanged — never warn-and-continue.** Skipped under the #2445 schema-AHEAD hatch, mirroring `init_attempts = 0`.
- `schema-init`: `SchemaInitReport` gains `pgvector_available`, `pgvector_installed`, `role_is_superuser`, `age_catalog_usage` (`#[serde(default)]`, `Deserialize` derived so older JSON still parses); human report prints them for postgres; probe errors propagate.
- `doctor`: new "Postgres extensions (#3264)" section when the resolved store is `postgres://` — CRITICAL on a blocking verdict or unreachable DSN, WARNING when AGE is installed without `ag_catalog` USAGE (`age_projection` would be skipped while `kg_backend` reports `age`).
- Posture row deliberately NOT added (pure, DB-free `evaluate` cannot render a non-vacuous pgvector row without a new env var, which #3264 forbids) — rationale recorded in code before check #16.
- Docs: `docs/postgres-age-guide.md` "Managed / non-superuser Postgres (CloudNativePG, RDS, Cloud SQL)"; pointer in `docs/enterprise-deployment.md`; CHANGELOG.
[...]

PR #3257, opened 2026-08-23
docs(roadmap): v1.1.0 ROADMAP — MemTrapBench-driven use-time applicability, measurement, docs truth

alphaonedev:
## Summary

Adds `docs/ROADMAP-v110.md` — the post-GA v1.1.0 roadmap, driven by **MemTrapBench (Wang et al. 2026, arXiv:2608.20202)**: memory can degrade downstream task reasoning even when every stored memory is faithfully recorded (three of the four trap classes use TRUE memories, so integrity gates never fire). Docs-only; no code, no behaviour change, no schema.

## Process (recorded in the doc)

- 7 workstream drafts + **21 adversarial ballots** (3 lenses × 7 workstreams: north-star/integrity, code-grounding — every file:line re-read at the tip, trap-mitigation/claims-truth), run on ox-alpha against the v1.0.0 tip `8fb6e9eb`. Tally: 21/21 REVISE, 0 REJECT; every REQUIRED-CHANGES applied in the synthesis.
- Fable 5 final review: spot-verified the load-bearing corrections (LifecycleState 7-variant census, `PINNED_KNOB_COUNT = KNOBS.len()` at `security_profile.rs:398`, soft-loser markers), completed the tail sections, harmonized the citation SSOT.

## Key roadmap-wide rules (voted)

- Everything additive, default-OFF, knob-gated; **no advisory knob enters the asi-hard pinned `KNOBS` set**; `PINNED_KNOB_COUNT` untouched by the entire roadmap.
- **No mechanism may claim mitigation until the w5 MemTrapBench harness produces `evidence_grade: "real"` numbers**; stub runs are not citable (enforced by the w7 docs-gate rule).
- Honest-unknown invariant: absence of data ⇒ absence of signal, never fabricated `0`/`false`.
- Every future implementation carries the code-comment citation `MemTrapBench (Wang et al. 2026, arXiv:2608.20202)` + its trap class (specified per workstream).

## Queue note

Docs-only; queue wherever convenient (suggest after the current FIX re-pushes; before #3120). Gate comment to follow from the Fable shadow-QC lane.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3240, opened 2026-08-23
fix(pg): SAL parity cluster — uncapped export, record-stop coverage, eviction tombstone/erase, frontier unlocks, governance audit

alphaonedev:
Closes #3174 #3175 #3177 #3179 #3180

The five SAL-parity findings from the R-405 152-method audit that are all the
same shape: **the postgres backend behaved differently from the sqlite SSOT for
the identical wire call, and nothing reported the difference.** An operator's
claim about backup completeness, retention/erasure, stoppability, dispatch
ordering, or auditability was true on one backend and false on the other.

One PR because the five share one new sibling module
(`src/store/postgres_parity.rs`) and interleave in `src/store/postgres.rs`.
Three signed commits on `ae3e0aec` (the prepared per-issue split was against
the stale `93d09145` base and is not load-bearing after the 64-commit rebase).

| issue | defect | proof |
|---|---|---|
| #3174 | admin export + entity_register silently clamped to 1000 rows | `pg_export_and_entity_register_survive_past_list_max_limit_3174` **ok** (1500-row corpus) |
| #3175 | record-stop miss on undo/recover + cache-only status + S5-H4 | `pg_record_stop_covers_undo_and_recover_and_reads_the_chain_3175` **ok**; structural guard `every_sqlite_record_stop_gated_method_is_gated_on_postgres_3175` **ok** |
| #3177 | TTL/size eviction hard-delete, no tombstone/erase; archive drops links | `pg_ttl_eviction_tombstones_and_crypto_erases_3177` **ok**; `pg_size_gc_archive_carries_the_edge_graph_3177` **ok** |
| #3179 | FRONTIER missing `unlocks` under a false parity doc | `pg_unlocks_gates_the_frontier_3179` **ok** AND `sqlite_unlocks_gates_the_frontier_3179` **ok** |
| #3180 | three lost audit/signal writes | `pg_pending_deny_appends_a_signed_denied_event_3180` **ok**; `pg_recall_hybrid_appends_the_access_ledger_3180` **ok**; `pg_reclassify_binds_the_sqlite_cause_hash_3180` **ok** |

#3207 (pg `sweep_pending_action_timeouts` emits no `pending_action.timed_out`) is
the same class, already filed, **not** in this PR.

---

## #3174 — admin export returned ≤1000 memories; entity_register deduped over ≤1000 rows

**Verified at `ae3e0aec` (rebased from `93d09145`).** `PostgresStore::export_memories`
[...]

Issue #3180, opened 2026-08-22
[data-integrity] pg loses three audit/signal writes the sqlite SSOT makes: pending_action.denied event, recall access ledger (recall_hybrid appends nothing → TTL-extension/promotion/decay frozen on pg fleets), reclassify cause_hash


alphaonedev:
Scout-verified (R-405) at `cda57210` from the ox-alpha SAL parity audit (152 methods; `.local-runs/ox-c4-parity/`). pg lane evidence required: `AI_MEMORY_TEST_POSTGRES_URL=$(cat .local-runs/ENTERPRISE-FED-TEST-PG-URL.txt) cargo test --features sal-postgres -- --include-ignored`.
**Code.** (a) `grep -rn 'pending_action.denied' src/` → `storage/mod.rs:19064` only; zero in postgres.rs. (b) pg `recall_hybrid` ends `postgres.rs:20815-20819` `results.truncate(...); Ok(results)` with no ledger append (only callsite of `recall_observation_insert` is `:23047`, the separate `record_recall_observation`); sqlite `sqlite.rs:1061-1088` appends via `record_recall_with_identity` (#1869). (c) pg `reclassify_memory_kind` appends with `cause_hash: None` (`:26908-26911`); sqlite `sqlite.rs:2382-2394` binds `compute_cause_hash`.
**Impact.** Governance denials unprovable on pg; SAL recall on pg produces zero access observations so the memory lifecycle (TTL extension, mid→long promotion, priority decay) is frozen for every pg-backed fleet — a silent core-behaviour divergence; reclassify chain loses causal linkage.
**Fix.** Emit pending_action.denied in pg pending_decide inside the same tx; append the ledger in pg recall_hybrid post-filter (best-effort + WARN, explicit commented discard); thread cause hash.
**Test.** pg: deny a pending action → one signed_events row; recall_hybrid → recall_observations rows == returned candidates with matching ranks; reclassify → cause_hash non-null and equal to the sqlite twin for identical inputs. GA: BLOCKING.

alphaonedev:
Closed by fable-merge of #3240 at c3344757 G (postgres SAL parity cluster).

PR #3268, opened 2026-08-27
fix(recall): HTTP postgres must not double-write the access ledger

alphaonedev:
## Summary

Certified pg+AGE has been deterministically red since `#3240` (`c3344757`) on `postgres_recall_purity_sal_and_http_entry_paths` — ledger growth **300 vs 100**. Fable's 08-27 note flagged it as a `#3240` fold/gc suspect. It is not: `#3267` does not touch recall, and the fold job is not in this test.

**Cause:** `#3180` (inside the `#3240` SAL-parity cluster) made `PostgresStore::recall_hybrid` append `recall_observations` (sqlite SAL parity — without it a pg-backed fleet's lifecycle is frozen). The HTTP postgres handler **still** records the POST-form4 / kinds / rerank set under a **second** `recall_id`. Every HTTP pg recall therefore writes the ledger twice; the access fold applies both rows → **2× echo amplifier** (the `#3266` class). The purity test also drives 10 SAL recalls (another 100 rows) and only counted HTTP, so 100 + 200 = 300.

Production HTTP postgres is 2× per recall, not 3×. Still a product bug, swarm-relevant.

## Fix

- New `Filter.skip_access_ledger` (default `false`; existing `..Default::default()` call sites unchanged).
- HTTP postgres sets it `true` before `recall_hybrid`. The handler remains the single writer for the **filtered** set under the `recall_id` echoed in the response (touch ops fire on the FILTERED set — existing contract).
- Direct SAL callers still append once (`#3180` stays).
- Sqlite SAL honors the same flag (HTTP sqlite still goes through `db::` and records in the handler; defensive parity).
- Purity test now counts SAL rows too; pin `recall_hybrid_skip_access_ledger_writes_nothing`.
- Ledger-append `Err` is still WARN + continue (ERRORS-19).

## Test plan

- [x] `cargo test --features sal --lib recall_hybrid_skip_access_ledger` — pass
- [x] `cargo test --test qual_10_module_size_ceiling` — pass (`postgres.rs` 38_662 / 38_800)
- [ ] Certified pg+AGE cell on this PR: `postgres_recall_purity_sal_and_http_entry_paths` green (the live pin)

## `#3267` ACK

Merged semantics received. Only open release PR besides this one is `#3257` (docs). No open branch asserts the old rolsuper-refusal / doctor-CRITICAL contract. `#3266` vote R1–R6 folded into `docs/ROADMAP-v110.md` on `#3257` (`d4bcb0de`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed (HTTP postgres double-counted the access ledger)
+
+- **HTTP postgres recall no longer writes the `recall_observations` ledger twice.**
+  `#3180` (landed via the `#3240` SAL-parity cluster) made
+  `PostgresStore::recall_hybrid` append one observation per returned row —
+  the sqlite SAL twin, required so a pg-backed fleet's lifecycle is not
+  frozen. The HTTP handler still recorded the POST-form4 / kinds / rerank
+  set under a *second* `recall_id`. Every HTTP pg recall therefore folded
+  2× (the `#3266` access-fold echo amplifier); the Certified pg+AGE
+  purity pin `postgres_recall_purity_sal_and_http_entry_paths` saw 300
+  ledger rows against an expected 100 (10 SAL + 10 HTTP×2). HTTP now sets
+  `Filter.skip_access_ledger` and remains the single writer for the
+  filtered set / response `recall_id`. Direct SAL callers still append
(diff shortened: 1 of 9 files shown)


alphaonedev:
Head SHA: \`31636dcf\` (signed, verified). Base: \`release/v1.0.0\` @ \`bd105b3e\`.

Local pins: \`recall_hybrid_skip_access_ledger_writes_nothing\` pass; QUAL-10 pass (\`postgres.rs\` 38_662 / 38_800). Live proof is the Certified pg+AGE cell on this SHA.

Issue #1253, opened 2026-05-25
[MED] Subscription DLQ has no size cap — operator-disk-fill via misbehaving subscriber


alphaonedev:
Filed by Security Review Agent S5.

## Root cause
`src/subscriptions.rs:1685-1713` `record_dlq_with_conn` accepts every failed-dispatch row via plain `INSERT`. No row count cap, no payload-size cap, no overflow refusal. A subscriber returning 5xx for every event lands one DLQ row per event indefinitely.

## Impact
**Violates scope item 5**: "DLQ overflow refuses new entries with clear error (not OOM)". Operators can be disk-filled via failed-webhook events.

## Proposed fix
Add `MAX_SUBSCRIPTION_DLQ_ROWS` const (suggest 10_000). Before INSERT, `SELECT COUNT(*) FROM subscription_dlq WHERE subscription_id = ?1` and refuse new rows with `dlq_overflow` error if >= cap. Operator-visible WARN + metric. Alternative: ring-buffer-style drop-oldest.

## Filed by
Security Review Agent S5, 2026-05-25.

alphaonedev:
Fixed via commit fd52ab13ea2554fee58f25c54bef98a52df01dc5 on branch `sec/1253-dlq-overflow-cap`, landed in PR #1284 against `release/v0.7.0`.

Retest evidence: `cargo test --lib subscriptions::tests::issue_1253_dlq_overflow_cap_refuses_past_max` — `test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 4767 filtered out; finished in 5.07s`.

Verified per prime directive pm-v3 (memory cd8ede94). All four gates (fmt + clippy --lib + test + audit) green on the isolated target dir.

PR #3316, opened 2026-08-30
fix(#3191): coordination-plane atomicity (F-1..F-5, fail-closed, both backends)

### LINKED ISSUE #3191: [data-integrity] coordination-plane atomicity: checkpoint resolve commits unsigned behind an Err, actions::transition is a read-then-write TOCTOU (many winners), a rejected routine edge strands inserted actions, the #1253 DLQ cap is overshot, and a webhook dispatches with no audit row
Five coordination-plane writes on the SQLite lane commit a partial or a wrong
state. Every one restores an invariant the code already claims in its own
comments; none tightens a posture the operator did not already ask for.
Findings verified by the ox campaign-2 `verifyspec2` lanes
(`checkpoints-actions-routines`, `subscriptions-hooks-inbox`); all five are
CONFIRMED against `release/v1.0.0` with exact line citations.

### F-1 `checkpoints::resolve` persists an UNSIGNED resolution behind a returned `Err`

`src/checkpoints/mod.rs` runs the first-resolution-wins CAS `UPDATE` and the
attestation `UPDATE` (`signature`, `resolver_pubkey`) as two separate
**autocommit** statements. A signing failure — or a crash in the window —
commits `state = resolved` with an EMPTY signature while the caller is handed
an `Err`. The function's own doc comment promises the opposite: *"an encode
error is mapped to a SQL misuse error so the resolve fails closed rather than
persisting an unsigned row that the caller believed was signed"*. Because the
CAS is first-resolution-wins, every retry then answers `Conflict` **forever**:
the checkpoint is permanently stranded unsigned, and every peer running
`AI_MEMORY_FED_REQUIRE_CHECKPOINT_SIG` rejects it. The postgres twin
(`PostgresStore::checkpoint_resolve`) has the same two-statement shape.

### F-2 `actions::transition` is a read-then-write TOCTOU — one action, many winners

`src/actions/mod.rs` composes a bare `SELECT state FROM actions WHERE id = ?1`
with an **unguarded** `UPDATE actions SET state = ... WHERE id = ?4` — no
`AND state = ?`, no enclosing transaction. This is exactly the hazard the
module's own `transition_cas` doc names (`#1718 H1`). Racing claimants all read
`pending`, all write, and all are told `Updated(Claimed)`; a stale in-flight
transition can also overwrite a TERMINAL verdict (`Done` -> `Claimed`),
destroying the coordination record. Multi-process shared-DB WAL is a supported
topology (`FBL-06`). The MCP handler's lease-holder check narrows but does not
close the window — claim-without-`claimed_by`, a stale lease, and the SAL lane
(`store/sqlite.rs`) stay exposed. The postgres twin already holds
`SELECT ... FOR UPDATE` inside a transaction, so this is also a parity gap.

### F-3 A rejected template edge strands the actions the routine already inserted

`materialize_template` (`src/mcp/tools/routine.rs`) inserts every action on
autocommit and only then walks the edges, so a self-edge / ordering cycle /
out-of-range index fails the run AFTER the inserts. The failure path records
`created_action_ids` as `[]` (`run_set_state(..., None, ...)` COALESCEs back to
the stored value), so nothing references those rows — yet they sit in the
namespace as pending work no agent owns, surfacing in `memory_action_list` and
the action frontier. This contradicts the atomic-reject posture #3010 already
promised for unrecognized template keys.

### F-4 The #1253 per-subscription DLQ cap is not atomic and is overshot

`record_dlq_with_conn` probes depth and INSERTs as two autocommit statements.
The in-code claim that *"the SQLite single-writer lock serialises this
read+write pair"* is false — that lock covers each **statement**. Every
dispatch worker opens its own connection, so N workers each observe
`depth < cap` and each INSERT, overshooting the cap by up to the
dispatch-concurrency width (reproduced: 30 rows for a cap of 24).

### F-5 A webhook event whose pre-send audit row cannot be written is dispatched anyway

`subscription_events` is documented in that very function as the crash-recovery
record written BEFORE the network send, yet a write failure is downgraded to
one `tracing::warn!` and the send proceeds — so a transient sqlite failure plus
an exhausted retry ladder leaves the event with **no durable trace anywhere**
(the DLQ fallback is best-effort and refuses at cap). Relatedly a SATURATED DLQ
is indistinguishable from a one-off sqlite hiccup in the logs, and the
`ai_memory_subscription_dlq_overflow_total` help text directs operators to
`ai-memory subscription dlq drain` — **a command that does not exist**.

### North Star

All five are DEGRADE-never-corrupt violations: the worst case must be a
delivery not made or an action not claimed, never a resolution the fleet
believes is signed, an action two agents both own, orphan pending work, or an
event no operator can account for.

### Related but deliberately out of scope

- `#3010` routine-run idempotency (`run_key`) — a knowingly deferred, additive
  MCP-parameter change; needs its own SSOT/param-census lockstep.
- `#3179` pg `action_frontier` unlocks predicate, `#3134` pg `action_add_edge`
  atomicity — sibling lanes, disjoint hunks.

alphaonedev:
Closes #3191

GA-blocking freeze (ruled FREEZE 20-0). Five coordination-plane defects where a read-then-write or two-autocommit-statement composition could corrupt or strand durable coordination state are made atomic / fail-closed. **Data integrity is the highest-order constraint** (North Star; ERRORS-09 fail-closed / make-illegal-states-unrepresentable).

## Fixes

- **F-1 (CRITICAL, both backends)** — `checkpoints::resolve` / `PostgresStore::checkpoint_resolve` ran the first-resolution-wins CAS state-flip and the attestation-signature write as **two separate autocommit statements**. A signing failure or crash after the CAS committed left the checkpoint `state=resolved` with an **empty signature**; because the CAS is first-wins, every retry then answered `Conflict` **forever**, permanently stranding the anchor unsigned (rejected by every peer running `AI_MEMORY_FED_REQUIRE_CHECKPOINT_SIG`). Both writes now commit in **one transaction** (sqlite `unchecked_transaction`; postgres `pool.begin()`): a signing failure rolls the state-flip back, so a resolved checkpoint can **never** persist without its attestation. Restores the fail-closed behaviour the function's own doc comment promised.
- **F-2 (both backends)** — `actions::transition` composed a bare `SELECT state` with an **unguarded** `UPDATE ... WHERE id = ?` — a read-then-write TOCTOU ("one action, many winners"). The sqlite UPDATE is now a guarded CAS (`AND state = ?<expected>` + `rows_affected` reclassify to a non-`Updated` outcome). The postgres twin was already race-safe via `SELECT ... FOR UPDATE`; it gains the same guard for structural parity + defense-in-depth (fail-closed if it ever misses under the row lock).
- **F-3 (sqlite / MCP routine path)** — `materialize_template` inserted actions in autocommit, so a rejected edge (self-edge / cycle / out-of-range index / malformed spec) **orphaned** the earlier actions in the frontier. The whole materialisation now runs in one transaction — a rejected edge rolls the inserts back (no orphans).
- **F-4 (sqlite subscription DLQ)** — the #1253 per-subscription DLQ cap could be **overshot** (probe-then-insert TOCTOU; `record_dlq` opens a fresh connection per call, and the "single-writer lock serialises" comment was false). The compare now lives inside the insert (`INSERT ... SELECT ... WHERE (SELECT COUNT(*) ...) < cap`), so the cap holds exactly under any concurrency.
- **F-5 (sqlite webhook dispatch)** — a webhook whose per-delivery `subscription_events` audit row could not be persisted was dispatched **anyway** (unauditable, unreplayable side effect). The delivery is now routed to the DLQ (durable + replayable via `memory_subscription_replay`) and **not** dispatched — fail-closed; the event is deferred, never lost.

All five landed in this PR.

## Postgres-twin parity

F-1 and F-2 are applied identically on both backends. F-1 pg wraps the CAS `UPDATE ... RETURNING`, the conflict-probe, and the signature-persist in one `pool.begin()` transaction with `tx.commit()` only on success. F-2 pg keeps the pre-existing `SELECT ... FOR UPDATE` (the real race guard) and adds the mirrored `AND state = $5` guard. F-3/F-4/F-5 are sqlite-only mechanisms (MCP routine materialisation over rusqlite; the DLQ + webhook audit are the sqlite audit-mirror) with no postgres twin.

## Tests (live regressions, both lanes)

- `tests/coordination_atomicity_3191.rs` (sqlite): F-1 signing-failure rollback (scoped trigger injects the signature-persist failure; asserts the checkpoint stays PENDING + unsigned + retry succeeds) and happy-path signed/verifies; F-4 concurrent cap-never-overshot; F-5 audit-failure routes to DLQ and the webhook endpoint receives **zero** calls (`wiremock` `.expect(0)`).
- `tests/coordination_atomicity_3191_pg.rs` (postgres, gated on `AI_MEMORY_TEST_POSTGRES_URL`): F-1 rollback twin (row-scoped trigger) and F-2 concurrent single-winner twin.
- Inline: cross-connection single-winner race (F-2, `src/actions/mod.rs`) and orphan-free rejected-edge (F-3, `src/mcp/tools/routine.rs`).

## SSOT / gates

`qual_10` `src/subscriptions.rs` ceiling `4_600 -> 4_700` (measured 4_637, dated lockstep). No MCP tools / routes / CLI verbs / schema versions added, so no other SSOT pin changes. No `src/identity/*` or `src/federation/*` change — **cert §7 unaffected**.

## Verification (all green, local)

- `cargo clippy --all-targets -- -D warnings -D clippy::all -D clippy::pedantic` on **default, sal, sal-postgres, vectorlite** — all RC 0
- `cargo fmt --all --check`; `bash scripts/check-hardcoded-literals.sh` — PASS
- Structural suite (all pass): record_stop_structural_b7, append_only_spine_guard_g6, cli_write_verb_pg_refuse_ceiling_2572, db_open_funnel_ceiling_2445, qual_expiry_floor_local_write_funnel_2515, qual_6_7_legacy_error_type_ceiling, qual_10_module_size_ceiling, mcp_param_names_invariant, attest_level_string_invariant, qual_12_todo_tracker_discipline
- `cargo test --test coordination_atomicity_3191` (4/4); inline `cargo test --lib 3191` (2/2); `cargo test --test checkpoint_cas_2396` (4/4, no regression)
- pg lane: `cargo test --features sal-postgres --test coordination_atomicity_3191_pg -- --include-ignored` (2/2) and `checkpoint_epoch_resolve_authz_3007_pg` (1/1, no regression)


alphaonedev:
Merged via signed --no-ff admin-lift to 5050eb24; enforce_admins re-armed (enabled=true). Verify-at-tip green. Excellent fail-closed work on both backend twins.

Issue #2572, opened 2026-07-31
[data-integrity] every remaining CLI write verb still conjures a phantom SQLite database under a Postgres deployment


alphaonedev:
The residual of the #2444 → #2490 class. #2444 closed `backup` / `restore`; #2490 (PR #2568) closed `export` / `export --full` / `import`. **Those five are the DR-artifact verbs — they are not the class.**

## The shape

`db::open` (`src/storage/connection.rs`) is `Connection::open(path)`, which CREATES the file when absent and then runs the bootstrap schema + the whole migration ladder on it. Every CLI subcommand that takes `--db` and does not consult the configured store URL therefore does the same thing on a Postgres-backed deployment: manufactures a local SQLite database that no schema probe can distinguish from a real one, operates on it, and exits 0.

For a WRITE verb the consequence is the one #2490 named on `import`: the write is not "diverged", it is simply **LOST**, at the moment the operator is told it landed.

## Verbs still on the unguarded shape

`store`, `link`, `update`, `promote`, `forget`, `delete`, `mine`, `archive *`, `gc`, `namespace *`, `reown`, `consolidate`, `reflect`, and the rest of the ~89 default-build subcommands. (`serve` and `curator` DO honour `--store-url`; `ai-memory mcp` is structurally sqlite-only by design, #1675/n24.)

## Why this was deliberately NOT folded into #2490

PR #2568's CHANGELOG entry states the scope explicitly so it cannot be misread as "the CLI is now Postgres-safe". Guarding the DR-artifact verbs is correct scoping for a data-loss fix; guarding ~80 more verbs is a separate, larger change with its own compatibility surface (several of them have legitimate local-sqlite-sidecar uses even on a Postgres host — the governance audit spine is one).

## What this needs

A design decision, not a mechanical sweep:

1. Which verbs are *legitimately* local-sqlite even on a Postgres deployment (audit spine, key custody, erasure cold tier) and must NOT be guarded?
2. For the rest, is the disposition refuse, or route through the SAL (`build_store_handle`) so they actually work against Postgres — the #2095 precedent for `agents bind-api-key`?
3. Is a single shared entry-point guard reachable, or does each verb need its own?

T1/T3/T6 — wants its own 5-agent vote.

## Related

#2444, #2490 (PR #2568), #1927 (the store-url channel), #2095 (route-through-SAL precedent), #1675/n24 (MCP stdio is sqlite-only by design).

[...]

PR #3318, opened 2026-08-30
fix(#2555): bound schema_version + typed poison error + operator repair verb

### LINKED ISSUE #2555: [#2445 residual] `schema_version` is an unconstrained fleet kill-switch, and there is no in-product repair verb
The #2445 guard makes a single integer load-bearing for availability. That integer has no constraints.

sqlite DDL (`src/storage/migrations.rs`): `CREATE TABLE IF NOT EXISTS schema_version (version INTEGER NOT NULL)` — no PK, no UNIQUE, no CHECK, no ceiling; the read is `MAX(version)`. Postgres: `version INTEGER PRIMARY KEY`, accumulating rows, same `MAX()` read.

`INSERT INTO schema_version VALUES (2147483647)` is therefore permanent and, on a shared postgres cluster, fleet-wide: one table in one cluster serving N daemons, and the ai-memory role has full DML on it, so a co-tenant app, a migration tool, a restore script or one `psql` session takes every daemon down. On sqlite, anything running as the daemon user.

Note the direction of the change: PRE-#2445 the behaviour was DEGRADE (silent no-op); #2445 converts it to DENY. So #2445 CREATED this availability primitive, and this issue is the compensating control.

There is also no in-product recovery. The refusal's two remediations are 'run the binary that wrote this' (impossible for a fabricated version) and 'restore a pre-upgrade snapshot' (may not exist). The `AI_MEMORY_ALLOW_SCHEMA_AHEAD` hatch unblocks the process but leaves the poisoned stamp in place forever.

**Proposed fix:** (a) a sane-headroom band above the tip that yields a DIFFERENT typed error and remediation — a value far beyond the ladder is a poisoned ledger, not a downgrade; (b) a bounding CHECK on the sqlite table in a new migration; (c) an operator-gated repair verb (`ai-memory doctor --repair-schema-version` or `migrate --restamp <n>`), snapshot-first and loud, so the hatch is a bridge rather than a permanent unit-file entry — which is what the exact-version hatch shape was designed to assume.

Surfaced by the 5-agent adversarial vote on #2445 (decision memory `4c4789ac`), voter-5 A2 + voter-1 O6.

alphaonedev:
## GA-blocker #2555 — `schema_version` unconstrained fleet kill-switch

`schema_version` was an unconstrained integer with **no in-product repair**. `INSERT INTO schema_version VALUES (2147483647)` is permanent: it reads back through `COALESCE(MAX(version), 0)` and trips the #2445 schema-ahead DENY on every daemon. On a **shared postgres cluster** (ai-memory role has full DML) any co-tenant / migration tool / restore script / one `psql` session takes the whole fleet down — and the DENY's remediations ("run the binary that wrote this" / "restore a pre-upgrade snapshot") cannot recover a fabricated version, while the operator hatch left the poisoned stamp forever. Ruled **FREEZE 19-1** by a 3×7 adversarial panel.

## The fix (three parts, both backends)

**(a) Headroom band + distinct typed error.** A stamp a sane amount above the ladder tip stays a legitimate downgrade (#2445 DENY, "run a newer binary"). A stamp **above the absolute ceiling `MAX_SCHEMA_VERSION` (100000)** is a POISONED LEDGER: a NEW typed error `SchemaVersionPoisoned` (slug `SCHEMA_VERSION_POISONED`, HTTP 503) whose remediation points at the repair verb, refused **before** the downgrade `evaluate` and **without** consulting the operator hatch. `boot`/`doctor` report it by name. Wired at all three schema funnels (sqlite `migrate`, `resolve_schema_posture`, postgres bootstrap).

- **Threshold rationale:** ladder tip is 92 after the entire project history (single-digit growth per release); 100000 is ~1000 product-lifetimes — permanently unreachable by a real ladder, so it never misclassifies a real rolling-deploy downgrade — yet four orders of magnitude below `i32::MAX` (2147483647), so every fabricated/overflow value in the #2555 class is caught. Named const `MAX_SCHEMA_VERSION` with dated rationale + `max_schema_version()` accessor + SSOT unit test.

**(b) Bounding CHECK.** New forward migration (sqlite v92 full-table rebuild; postgres v92 `ADD CONSTRAINT`) + the bootstrap SCHEMA on both backends carry `CHECK (version <= 100000)`, so an out-of-band write is refused at the boundary. **Upper-bound only** — the low end (0/negative/deleted) is #2564's read-time domain, and a `>= 0` bound would break #2564's own negative-stamp fixtures.

**(c) Operator-gated in-product repair verb.** `ai-memory doctor --repair-schema-version <N>` restamps a poisoned ledger to a correct in-band version — **SNAPSHOT-FIRST** (a sibling `VACUUM INTO` backup before the stamp is touched; a snapshot failure refuses the repair), operator-gated (local CLI, no HTTP/MCP), refused on a served postgres store (#2572 — restamping the sqlite sidecar would be a phantom write; the pg ledger is repaired via the admin `DELETE` the poison message names). This is the recovery the DENY lacked.

## pg-vs-sqlite parity
- Both bound the stamp: sqlite v92 rebuild (no indexes/triggers → lossless, idempotent, applied only when the CHECK is absent) vs postgres v92 `ADD CONSTRAINT` (probe `pg_constraint` + out-of-band pre-flight, mirroring `migrate_v30`). Both bootstrap DDLs carry the inline CHECK for fresh installs.
- Both classify poison at read time via the shared `schema_guard` (`SchemaVersionPoisoned` + `From` → `StoreError` → 503).
- Repair verb is sqlite-local (the #2572 architecture: CLI never writes a served store); pg recovery is the admin `DELETE` named in the backend-aware poison message.

## SSOT reconciled
`CURRENT_SCHEMA_VERSION` 91→92 (sqlite+postgres lockstep) across `migration_meta` `MIGRATION_LADDER`, the module docstring, and 13 doc references (docs-vs-ssot gate); qual_10 ceilings bumped (postgres.rs 39_850→40_050, migrations.rs 6_640→6_780, dated lockstep); `cli_write_verb_pg_refuse_ceiling_2572` inventory +1 (doctor.rs) 30→31; `arch_9` slug round-trip + `SCHEMA_VERSION_POISONED` slug; record_stop B7 allowlist +1 (`render_poisoned`, a message builder that quotes a `DELETE`, executes no SQL).

## Verification (all green locally)
`cargo fmt --check`; `scripts/check-hardcoded-literals.sh`; `scripts/check-migration-ladder.sh`; `scripts/check-docs-vs-ssot.sh`; **clippy default + sal + sal-postgres + vectorlite** (`-D warnings -D clippy::all -D clippy::pedantic`, `--all-targets`); the structural suite (`db_open_funnel_ceiling_2445`, `record_stop_structural_b7`, `append_only_spine_guard_g6`, `cli_write_verb_pg_refuse_ceiling_2572`, `qual_expiry_floor_local_write_funnel_2515`, `qual_6_7_legacy_error_type_ceiling`, `qual_10_module_size_ceiling`, `mcp_param_names_invariant`, `attest_level_string_invariant`, `qual_12_todo_tracker_discipline`, `migration_ladder_integrity`); the new lanes — `schema_version_poison_guard_2555` (sqlite, 5/5) and `postgres_schema_version_poison_guard_2555` (**live pg**, 2/2) — plus `postgres_schema_parity` / `postgres_schema_downgrade_guard_2445` (unaffected).

No `src/federation` or `src/identity` change — no CERT §7 amend.

Closes #2555


alphaonedev:
Merged via signed --no-ff admin-lift to ecb7944e; enforce_admins re-armed (enabled=true). Verify-at-tip green. Clean three-part fix, snapshot-first repair is exactly right.

PR #3320, opened 2026-08-30
fix(governance): align capability.rs inline tests to #3111 terminal-Deny

### LINKED ISSUE #3111: [security][design-decision] capability-token grant joiner flips an EXPLICIT operator Deny→Allow — should a rule-derived Deny be terminal?

alphaonedev:
## Problem

The Per-Module Coverage `cargo test --lib` job was RED on the release tip. Two inline `#[cfg(test)]` unit tests in `src/governance/capability.rs` still asserted the **pre-#3111** "a covering token flips a `Deny` to `Allow`" behavior.

#3111 made a rule-derived (operator) `Deny` **TERMINAL** in the capability-token grant joiner:
- `apply_capability_grant` / `apply_capability_grant_gov` return the `Deny` base **UNCHANGED** with `GrantOutcome::NoOp` **BEFORE** `verify()` runs.
- `apply_at_gate` likewise leaves an operator `Deny` in place.
- Only an `Ask`/`Pending` (a pending human court) base lifts to `Allow` on a valid covering token.

The g10 integration tests (`tests/g10_capability_tokens.rs`) were updated with #3111, but these two inline unit tests were missed.

## Fix (test-only — no production code changed)

- `gov_joiner_flips_deny_and_pending_and_preserves_base_on_reject` → **`gov_joiner_deny_terminal_pending_flips_preserves_base_on_reject`**: the `Deny("nope") → Allow`+`Granted` assertion now asserts the `Deny` is terminal (refusal envelope stands, `NoOp`). The reject-preserves-base block used a `Deny` base — which no longer reaches `verify` — so it is retargeted to a `Pending` base to still exercise a genuine `Rejected(Expired)` outcome, plus an added assertion that an expired token cannot move a `Deny` (terminal, `NoOp`).
- `apply_at_gate_flips_when_enabled_and_verified` → **`apply_at_gate_operator_deny_terminal_pending_flips`**: the `Deny("deny") → Allow` assertion now asserts terminal-`Deny` at the gate, strengthened with an explicit `Pending → Allow` case proving the gate still lifts a legitimate pending court (i.e. it is not a blanket no-op).

The whole inline module was scanned; all other grant/gate tests (`apply_grant_deny_is_terminal_ask_flips_on_valid_token`, `apply_grant_reject_leaves_base_unchanged`, `apply_at_gate_identity_*`) already encode terminal-Deny correctly. No test is weakened — each fixed test still proves the crown-jewel authz invariant: an operator permission rule can never be bypassed by an AI-held (possibly attenuated) token.

Touches no `src/federation` or `src/identity`, so the cert §7 re-cert trigger does not fire.

## Verification

- `cargo test --features sal,sal-postgres --lib -- governance::capability` → **41 passed** (the exact suite that was failing).
- `cargo test --features sal-postgres --test g10_capability_tokens -- --include-ignored` → **13/13** (no regression).
- `cargo clippy --features sal-postgres --tests -- -D warnings -D clippy::all -D clippy::pedantic` → clean.
- `cargo fmt --all --check` → clean.
- `qual_10_module_size_ceiling`, `qual_6_7_legacy_error_type_ceiling` → green (no ceiling bump needed; test-only additions, no new `Result<_,String>` signatures).

Closes #3111

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed (CI — `cargo test --lib` RED: two inline capability tests still asserted pre-#3111 flip-Deny, #3111 follow-up)
+
+- **#3111 follow-up (CI / test-only) — the Per-Module Coverage `cargo test
+  --lib` job was RED because two inline `#[cfg(test)]` unit tests in
+  `src/governance/capability.rs` still asserted the OLD "a covering token flips
+  a Deny to Allow" behavior that #3111 removed.** #3111 made a rule-derived
+  (operator) `Deny` TERMINAL in the capability-token grant joiner:
+  `apply_capability_grant`/`apply_capability_grant_gov` return the `Deny` base
+  UNCHANGED with `GrantOutcome::NoOp` BEFORE `verify()` runs, and `apply_at_gate`
+  likewise leaves an operator `Deny` in place — only an `Ask`/`Pending` (a
+  pending human court) base lifts to `Allow` on a valid covering token. The
+  crown-jewel authz invariant is that an operator permission rule can never be
+  bypassed by an AI-held (possibly attenuated) token. The g10 integration tests
+  were updated with #3111 but these two inline unit tests were missed:
+  `gov_joiner_flips_deny_and_pending_and_preserves_base_on_reject` (renamed to
+  `gov_joiner_deny_terminal_pending_flips_preserves_base_on_reject`) asserted
+  `Deny("nope") → Allow`+`Granted`, and
+  `apply_at_gate_flips_when_enabled_and_verified` (renamed to
+  `apply_at_gate_operator_deny_terminal_pending_flips`) asserted
+  `Deny("deny") → Allow` at the gate. Both now assert the terminal-Deny
+  invariant (the original refusal envelope stands, `NoOp` at the joiner) while
+  keeping — and, at the gate, strengthening with an explicit `Pending → Allow`
+  case — the legitimate pending-court lift path. Test-only: no production code
+  changed, no `src/federation`/`src/identity` touched, so the §7 re-cert
+  trigger does not fire.
+
 ### Fixed (CI — certified-postgres cert gate RED since the #3198 key-dir hardening, #3117 SHIP-BLOCKER)
 
 - **#3117 (CI / GA ship-blocker) — the `cert-postgres-age.yml` "Cluster-A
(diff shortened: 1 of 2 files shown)


alphaonedev:
Merged into release/v1.0.0 via signed admin-lift at 00285ddb (Fable conductor gate). Verify-at-tip LOCAL authoritative: `cargo test --lib governance::capability` 41/41 under umask 022, g10 13/13, clippy pedantic (both legs) + fmt + structural gates clean. The prior CI red was transient ENOSPC on the self-hosted fleet (all-cancelled cascade, pure-compile jobs failing), not this test-only change. Greens the Per-Module Coverage job.

Issue #3475, opened 2026-09-02
CI red on 70c20c8d (macos-fed sqlite): mcp::tests::handle_get_* fail nondeterministically — #3356's env-mutating tests race the lib test binary's readers of AI_MEMORY_AGENT_ID


alphaonedev:
Run 33662095277, job 'Check (macos-fed,sqlite)': `mcp::tests::handle_get_happy_returns_memory` and `handle_get_resolves_by_prefix_and_includes_links` panic on `serde_json::from_str(...).unwrap()` (Error("expected value") — the tool returned an error text, not JSON) and `handle_get_links_returns_outbound_and_inbound` asserts count >= 1 and fails; all three pass on the linux legs and passed on b4c279c0. #3356 (merged 2b191538) added lib tests that SET/UNSET the process-global `AI_MEMORY_AGENT_ID` under `agent_id_env_set_guard` / `agent_id_env_unset_guard` (crate-wide test lock) — but the handle_get tests, and every other lib test that resolves identity from the environment, do not take that lock, so when they interleave with a test holding `AI_MEMORY_AGENT_ID="bad id with spaces"` the dispatch fails closed and the test sees an error envelope. Nondeterministic; macOS scheduling exposed it first.

Fix (test-only, do not weaken #3356): move every test that mutates `AI_MEMORY_AGENT_ID` (or any process env read by identity resolution) out of the lib test binary into a dedicated integration test binary (process isolation), OR make identity resolution take an injected env source so tests never mutate the real process env; then grep the lib tests for any remaining `set_var`/`remove_var` and apply the same treatment (`check-test-env-lock.sh` exists — extend it to refuse env mutation in `src/**` tests). Add a regression: run the affected handle_get tests 50x in parallel with the env tests in the same binary if they stay there. Both legs. Tip must return to green; this blocks the next push.

alphaonedev:
**CLAIM** — Fable pool coder (f2), branch `fix/3475-lib-test-env-isolation` off `origin/release/v1.0.0` @ `70c20c8dff2e9c2706e420db1e72cabacf1236ca`.

[...]

PR #3524, opened 2026-09-06
test(coverage): restore the five breached per-module floors, lift the three canaries (#3521)

### LINKED ISSUE #3521: Per-module coverage floors breached on the published tip (6e92e93d): cli/backup.rs 88.87 < 89, cli/curator.rs 88.37 < 90, handlers/federation_signing_check.rs 82.16 < 84, federation/receive.rs 89.97 < 90, store/mod.rs 89.83 < 90 — tests needed (floors never fall)
The Per-Module Coverage Thresholds workflow on `6e92e93d` (run 34021177602) passed every test for the first time in the current series and then failed at `coverage/check-thresholds.sh`:

```
GLOBAL: 92.46% >= 90.00%  PASS
FAIL: cli/backup.rs                          measured 88.87% < threshold 89%
FAIL: cli/curator.rs                         measured 88.37% < threshold 90%
FAIL: handlers/federation_signing_check.rs   measured 82.16% < threshold 84%
FAIL: federation/receive.rs                  measured 89.97% < threshold 90%
FAIL: store/mod.rs                           measured 89.83% < threshold 90%
CANARY: cli/doctor.rs 90.33% (floor 90, +0.33pp); handlers/links.rs 79.03% (floor 79, +0.03pp); daemon_runtime.rs 86.20% (floor 86, +0.20pp)
PER-MODULE: 174 pass, 5 fail, 3 warn
```

Provenance of the drops (the published merges since 31faaa20): `cli/backup.rs` — #3508 `verify_page_accounting` restore gate; `cli/curator.rs` — #3345 curator-report backlog collapse CLI surface; `handlers/federation_signing_check.rs` and `federation/receive.rs` — #3464 receive-path key-binding strip / historical key re-verification and #3075 postgres federation lanes; `store/mod.rs` — new SAL trait methods with default `UnsupportedCapability` arms (#3075 federated-apply arms, #3064 `calibrate_confidence_report`, #3345 `prune_curator_reports`, #3519 is not in this tip).

Playbook §8: thresholds rise across releases and never fall, so the fix is TESTS, not floor edits.

**Ask (S-TEST lane, coder tier):** add unit/integration tests that exercise the uncovered branches in each of the five modules until each clears its floor by ≥ 0.5 pp (so the canary margin is met), and lift the three canaries (`cli/doctor.rs`, `handlers/links.rs`, `daemon_runtime.rs`) by ≥ 0.5 pp where cheap. Measure locally with the same tooling the workflow uses (`coverage/` scripts, `cargo llvm-cov` per the workflow) and report per-module before/after numbers in the READY comment. No production code changes except where a branch is genuinely unreachable (then delete it, with justification). No threshold edits.

alphaonedev:
Closes #3521

The Per-Module Coverage Thresholds gate failed on the published tip `6e92e93d`
with five modules under their floors and three canaries clearing by < 0.5pp.
Per playbook §8 thresholds rise and NEVER fall, so this is **tests only**:
`coverage/thresholds.toml` is untouched and there is **no production code
change**.

Every test added pins a data-integrity or claims-truth contract rather than
counting lines:

- **`store/mod.rs`** — the SAL `MemoryStore` DEFAULT arms. An adapter that has
  not implemented a capability must surface `UnsupportedCapability` *naming*
  it, never a silent `Ok` that would let an upper layer believe a federated
  apply, a governance decision or a record-stop landed. The deliberately
  BENIGN defaults are pinned separately so the distinction cannot be flipped.
- **`handlers/federation_signing_check.rs`** — `/sync/push` under a backend
  that REFUSES: refused subcollections land in `skipped`, never `applied` (a
  peer must not advance its watermark past rows this node does not hold); an
  engaged record-stop refuses before any write; an UNREADABLE record-stop
  probe refuses too (fail closed). Plus the `X-Memory-Cred` refusal arms —
  including the identity binding that stops a valid credential for peer A
  authenticating a push attributed to peer B — and the NONCE half of the
  signature-enforcement matrix on both federation surfaces (a signed
  `/sync/push` or `/sync/since` that omits its nonce under
  `AI_MEMORY_FED_REQUIRE_NONCE=1` is refused; a replayed nonce is refused).
- **`cli/curator.rs`** — the store-backed `--prune-reports` and
  `--rollback[-last]` twins over a `sqlite://` store URL (the same SAL
  dispatch a `postgres://` URL takes): unknown rollback ids are refused,
  reversed entries are tagged `_reversed` and skipped on re-run, malformed
  entries are skipped rather than guessed at.
- **`cli/backup.rs`** — `restore`'s liveness probe WARNs and PROCEEDS when the
  target cannot be opened or the probe is inconclusive (refusing there would
  strand a disaster recovery); through the public `run_backup` entry the store
  resolver refuses an ambiguous `--store-url`-vs-environment pair, refuses a
  pathless `sqlite://` URL, and ANNOUNCES the redirect when the configured
  store wins over `--db`.
- **`federation/receive.rs`** — the catch-up `/sync/since` URL builder carries
  the cursor exactly once, percent-encoded (a dropped cursor turns a bounded
  delta pull into a full-corpus re-pull), and the stored-namespace probe
  reports "no live row" as `Ok(None)`, never an `Err` that would halt the
  watermark.
- **Canaries** — `cli/doctor.rs` (a group-writable key store is CRITICAL with
  the `chmod` to fix it; `daemon.pub` with no `daemon.priv` reports "cannot
  sign"), `handlers/links.rs` (a wrongly-typed link endpoint is refused, never
  coerced), `daemon_runtime.rs` (an unopenable DB leaves the ANN index cold
  instead of seeding a partial graph; a lax-mode passphrase file is refused
  unless the operator explicitly opts out, and the opt-out reads the SAME
  handle it permission-checked; a URL-shaped `--db` is refused with the
  password REDACTED; the four unattended maintenance loops tick against one
  shared connection and leave the corpus readable, writable and
  byte-identical; `find-paths` / `kg-invalidate` / `swarm-rewind` over an
  empty corpus report nothing rather than fabricating a path, an edge or a
  write).

Two incidental test-hygiene fixes, both required to make the suite honest on
any host:

- `tests/cov_fupb_fed_signing.rs`'s `enrol_peer_key` now pins its temp key
  directory to `0700`. On a host whose TMPDIR carries a permissive default ACL
  the fresh directory came back `0775` and the #3198 key-store posture gate
  (correctly) refused it, failing **all seven** cases in that file before they
  reached the assertion under test.
- The two environment-mutating groups live in their own test binaries
  (`tests/cov_backup_store_url_3521.rs`,
  `tests/cov_daemon_passphrase_posture_3521.rs`) because
  `scripts/check-test-env-lock.sh` arm (d) (#3475) refuses new
  `std::env::set_var` in the shared `src/**` test binary.

QUAL-10 lockstep: `src/daemon_runtime.rs` ceiling `14_300 -> 14_450` (the
growth is entirely `#[cfg(test)]`).

https://claude.ai/code/session_01PyYCWh7JTrV2K6D5CgnHhi


alphaonedev:
Closed by the Master Conductor: lanes deliver a pushed branch plus a READY comment on the issue; the Master reviews and merges with a signed --no-ff merge into release/v1.0.0. A pull request duplicates the campaign on the shared self-hosted runners and competes with the release gate. The branch stays; #3521 continues via its READY comment.

Issue #3588, opened 2026-09-10
incident: unexplained GitHub Actions run cancellations on release/v1.0.0 (2026-09-10, ×2)


alphaonedev:
## Incident: unexplained cancellation of GitHub Actions runs on release/v1.0.0 (2026-09-10, two occurrences)

**Status:** open. Filed by the Conductor under the standing rule "two unexplained run cancellations in one day → written incident note".

### Timeline (UTC)
| Time | Run | Tip | What |
|---|---|---|---|
| 11:17:29 | CI (umbrella), attempt 1 | 869b4d49 | cancelled; f2 self-hosted job `Check (linux-fed,enterprise-fed)` logged "completed with result: Canceled" at 11:17:08 |
| 11:31:31 | Per-Module Coverage Thresholds | 869b4d49 | cancelled |
| 19:16:22 | CI (umbrella), attempt 1 | 6082af8a | cancelled; runner worker log: "Cancellation/Shutdown message received" (server-side); GitHub-hosted `Check (ubuntu-latest,sqlite)` and f2 `Check (linux-fed,enterprise-fed)` cancelled in the same second |

Both CI runs were re-run by the Conductor (attempt 2 green this morning; attempt 2 in progress now). All other campaign workflows on both tips finished normally.

### Actor
GitHub records the cancelling actor as the shared account. Exhausted, with evidence in `.local-runs/GATE-CHECKLIST.md`:
- Operator: stated they did not cancel.
- Conductor session transcript: no cancel command.
- Codex and Grok deputies (f1): audited their logs, no cancel.
- the retired external coder deputy (f2, started 15:42): pass logs contain no cancel; did not exist for the morning events.
- No newer run in any concurrency group at either time (chain/next pushes do not trigger CI).
- No workflow step uses a PAT or a cancel action; repo has no deploy keys or webhooks; branch protection intact.
- f2 host: no cron, timer or script calls `gh run cancel`; the root cron `oom-shield.sh` only adjusts OOM scores; the `cybercommand` daemon runs `gh issue edit` against another repo every minute and has no run-cancel code path.

### Correlation
Both CI cancellations occurred a few minutes after the Conductor's push gate began its FULL default-suite stage on f2 (11:07 → 11:17; 19:14 → 19:16), while the f2 self-hosted runner was executing a CI job. No mechanism connecting the two has been found; noted as the only temporal pattern.

### Impact
[...]

PR #3768, opened 2026-09-17
diag: CI canceller probe 2 (do not merge)

alphaonedev:
Empty-commit probe to observe the run canceller with connection sampling on both hosts. Close after the observation.


alphaonedev:
Probe complete: after stopping ci-allowlist-enforcer.sh (#3588) the runs on 98b03f5fe ran 25+ minutes with zero cancellations and two workflows finished green; the remaining jobs are cancelled by the Conductor to free the self-hosted runners and disk during chain 9b. Branch deleted.

